PR #10, opened 2018-11-14
Lower Norfair & Maridia Inner

DIProgan:
Lower Norfair & Maridia Inner


miketrethewey:
Implemented differently.
Rooms: Lower Norfair: 9f42ccf
Connections: Lower Norfair: 3281de7
Rooms: Inner Maridia: bceb264 & 1f30337
Connections: Inner Maridia: 5824222

PR #12, opened 2018-11-17
Add Pink Brinstar rooms and connections, as well as related techs and helpers

fdtaillefer:


fdtaillefer:
Note: This also relies on the presence of a note property on a link in the schema, and on the quickCrumbleEscape tech, which are on the green brinstar branch.

miketrethewey:
Implemented differently.

PR #31, opened 2018-12-10
Add two ice clip strats in green brin

fdtaillefer:


fdtaillefer:
May as well add Zeela clip, will re-open

fdtaillefer:
Done.

PR #47, opened 2018-12-15
New Spark structure and Wrecked Ship statue req

DIProgan:
Croc and Crateria with LeaveSparked removed. Also added Phantoon requirement needed to activate wrecked ship statue.


fdtaillefer:
Going to merge this PR so it can be split into smaller commits on another branch.

PR #99, opened 2019-01-27
Norfair East heatdata

DIProgan:
New PR with same data because of weird conflicts. Also one commit per room whee.


region/norfair/east.json
```diff
@@ -1869,25 +1903,101 @@
...
+                    "ScrewAttack",
+                    {"cost": [
+                      {"heatFrames": 780}
+                    ]}
+                  ]},
+                  {"and": [
+                    "PowerBomb",
+                    {"cost": [
+                      {"heatFrames": 850},
+                      [
```
fdtaillefer: It makes sense to have an array for ammo, but it's "ammo" itself that has to be an array, rather than be in an array.

region/norfair/east.json
```diff
@@ -2197,7 +2339,15 @@
...
             {
               "id": 5,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 160}
```
fdtaillefer: Is that a typo? You mostly have multiples of 50.

region/norfair/east.json
```diff
@@ -2821,7 +2817,24 @@
...
               "id": 2,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"or": [
+                  {"and": [
+                    "Missile",
```
fdtaillefer: Is there anything special about missiles that strong beam combinations don't do?

region/norfair/east.json
```diff
@@ -2915,12 +2911,24 @@
...
+                {"or": [
+                  {"and": [
+                    "SpaceJump",
+                    "canSwim",
+                    {"cost": [
+                      {"heatFrames": 450}
+                    ]}
+                  ]},
+                  {"cost": [
+                    {"heatFrames": 600},
```
fdtaillefer: I think this should have `canSuitlessLavaDive` as well?

region/norfair/east.json
```diff
@@ -2930,36 +2938,102 @@
...
+                            ]}
                           ]}
                         ]}
                       ]}
                     ]}
                   ]},
                   {"and": [
                     "canSuitlessLavaDive",
-                    "canSuitlessLavaWalljump",
+                    "Morph",
```
fdtaillefer: Feels kinda hard to explain that you absolutely need Morph to do a lava dive, no?

region/norfair/east.json
```diff
@@ -3097,7 +3100,16 @@
...
               "id": 1,
-              "requires": [ "Morph" ],
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                "Morph",
+                {"cost": [
+                  {"heatFrames": 150}
```
fdtaillefer: This must be a typo? Was it meant to be 1500?

region/norfair/east.json
```diff
@@ -3161,9 +3157,30 @@
...
-              "note": "Some lava damage is required without Morph. Is it worth writing down?"
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"or": [
+                  "Morph",
+                  "canSwim",
+                  "canSuitlessLavaDive"
```
fdtaillefer: I didn't put `canSuitlessLavaDive` in for this room because it's pretty shallow and I didn't think you needed to get into the lava too much. You feel this is enough to require the tech?

DIProgan: Can't say I have a strong opinion about it either way. The amount of lava frames needed to traverse without it might be an indication of what to do with it but it was certainly a relevant amount of damage to have in logic.

region/norfair/east.json
DIProgan: 650 so good find.

region/norfair/east.json
DIProgan: Pretty sure I was just being lazy and didn't want to clutter the code with a bunch of beams but I went back and found you don't need the missiles to achieve that time at all so clutter be gone!

region/norfair/east.json
DIProgan: True

PR #101, opened 2019-01-27
Norfair West heatdata

DIProgan:
Steam over

### DIFF
--- a/region/norfair/west.json
+++ b/region/norfair/west.json
@@ -581,10 +581,6 @@
       "name": "Ice Beam Acid Room",
       "area": "Norfair",
       "subarea": "West",
-      "requires": [
-        "Varia",
-        "FIXME"
-      ],
       "nodes": [
         {
           "id": 1,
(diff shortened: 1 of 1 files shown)


region/norfair/west.json
```diff
@@ -765,17 +803,60 @@
...
+                      {"enemyDamage": {
+                        "enemy": 99,
+                        "hits": 1
+                      }}
+                    ]
+                  ]},
+                  {"cost": [
+                    {"heatFrames": 300},
+                    [
+                      {"ammo": {
```
fdtaillefer: Do you think it's worth also looking at alternatives with wave beam?

region/norfair/west.json
```diff
@@ -785,18 +866,60 @@
...
+                        "enemy": 99,
+                        "hits": 1
+                      }}
+                    ]
+                  ]},
+                  {"cost": [
+                    {"heatFrames": 200},
+                    [
+                      {"ammo": {
+                        "type": "PowerBomb",
```
fdtaillefer: Same question here for wave

region/norfair/west.json
DIProgan: Absolutely. Tomorow :)

DIProgan: Difference is less than 50f both from top and bottom door. Leaving them unchanged.

fdtaillefer: Alrighty then

PR #104, opened 2019-01-29
Lower Norfair West heatdata

DIProgan:
A minor cost change for Red Tower found its way in but it's something that should be in the end product anyway.

### DIFF
--- a/region/lowernorfair/west.json
+++ b/region/lowernorfair/west.json
@@ -6,7 +6,6 @@
       "name": "Acid Statue Room",
       "area": "Lower Norfair",
       "subarea": "West",
-      "requires": [ "Varia" ],
       "nodes": [
         {
           "id": 1,
@@ -248,7 +247,6 @@
       "name": "Golden Torizo's Room",
       "area": "Lower Norfair",
       "subarea": "West",
-      "requires": [ "Varia" ],
       "nodes": [
         {
           "id": 1,
@@ -302,8 +300,50 @@
           "yields": [ "DefeatGoldenTorizo" ],
           "unlock": [
             {"or": [
-              "Super",
-              "Charge"
+              "heatProof",
+              "canHeatRun"
+            ]},
+            {"or": [
+              {"and": [
+                "Super",
+                {"cost": [
+                  {"heatFrames": 1200}
+                ]}
+              ]},
+              {"and": [
+                "Charge",
+                {"or": [
+                  {"and": [
+                    "Ice",
+                    "Wave",
+                    "Spazer",
+                    {"cost": [
+                      {"heatFrames": 4000}
+                    ]}
+                  ]},
+                  {"and": [
+                    "Plasma",
+                    {"cost": [
+                      {"heatFrames": 2000}
+                    ]}
+                  ]},
+                  {"and": [
+                    "Plasma",
+                    "Wave",
+                    {"cost": [
+                      {"heatFrames": 1400}
+                    ]}
+                  ]},
+                  {"and": [
+                    "Ice",
+                    "Wave",
+                    "Plasma",
+                    {"cost": [
+                      {"heatFrames": 1250}
+                    ]}
+                  ]}
+                ]}
+              ]}
             ]}
           ]
         }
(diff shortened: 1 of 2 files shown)


region/lowernorfair/west.json
```diff
@@ -68,13 +67,31 @@
           "to": [
             {
               "id": 3,
-              "requires": null,
-              "unlock": null,
-              "note": "FIXME Some acid damage will be required without certain items"
+              "requires": [
```
fdtaillefer: I intended for the node to just represent being at the ledge. Clearing the PB blocks and interacting with the statue was meant to be part of the node's `unlock` reqs.

I'm a bit disappointed I didn't leave a note on the node to explain that.

region/lowernorfair/west.json
```diff
@@ -94,13 +119,31 @@
           "to": [
             {
               "id": 1,
-              "requires": [ "UsedAcidChozoStatue" ],
-              "unlock": null,
-              "note": "Because going to the statue without unlocking it is pointless, this path will assume the acid has been cleared"
```
fdtaillefer: Any reason the note was removed? The flag it explained is still in the requirements.

region/lowernorfair/west.json
```diff
@@ -110,19 +153,50 @@
...
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
                 "UsedAcidChozoStatue",
-                "canPassBombPassages"
+                "canPassBombPassages",
+                {"cost": [
+                  {"heatFrames": 250},
+                  {"ammo": [
```
fdtaillefer: No passage with just Morph Bombs?

region/lowernorfair/west.json
```diff
@@ -133,14 +207,35 @@
...
+                  "canHeatRun"
+                ]},
                 "UsedAcidChozoStatue",
-                "canPassBombPassages"
+                "canPassBombPassages",
+                {"cost": [
+                  {"heatFrames": 300},
+                  {"ammo": [
+                    {
+                      "type": "PowerBomb",
```
fdtaillefer: No passage with just Morph Bombs?

region/lowernorfair/west.json
```diff
@@ -364,28 +452,62 @@
...
+                          {"heatFrames": 950},
+                          {"shinesparkFrames": 11}
+                        ]}
+                      ]}
                     ]}
                   ]}
                 ]}
               ],
-              "unlock": null,
-              "note": "The Shinespark strat requires two separate sparks"
```
fdtaillefer: Was there something wrong with the note?

region/lowernorfair/west.json
```diff
@@ -841,7 +840,16 @@
           "to": [
             {
               "id": 2,
-              "requires": [ "canOpenGreenDoors" ],
+              "requires": [
```
fdtaillefer: I'm a bit surprised there's no fast ripper hits in there. Sounds unlikely without killing them.

region/lowernorfair/west.json
DIProgan: Done in the middle of the process where things where up in the air. Putting it back.

region/lowernorfair/west.json
DIProgan: Looks like I forgot that one yes. To the bat-emulator-mobile!

region/lowernorfair/west.json
DIProgan: Sometimes I don't know if the note is for me or for the end-user.

region/lowernorfair/west.json
DIProgan: We removed the entire part about killed enemies since you didn't like it, remember? I killed two rippers with supers for node 1 to 2 since supers are already required to open the gate anyway and any ammo concerns are mute since rippers drop the supers en masse. 

For node 2 to 1 however I could get through without killing or get hit by any. While that strat was harder to execute than any of the other plausible ones like using PB's or ScrewAttack they where not too far off each-other in the use of heatFrames, The rippers don't deal much damage anyway if you have to get past this sloppy style. The bulk of the damage for most runners will come from executing the GGG so I simply opted for a flat heatFrame count in this link as well.

region/lowernorfair/west.json
DIProgan: I interpret leave-as-is

region/lowernorfair/west.json
fdtaillefer: In general I put notes as comments to explain things that I think aren't obvious. They're meant for anyone who could be reading the file.

PR #114, opened 2019-02-07
Lower Norfair: East heatdata

DIProgan:
Hot stuff

### DIFF
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -1555,7 +1555,6 @@
       "name": "Plowerhouse Room",
       "area": "Lower Norfair",
       "subarea": "East",
-      "requires": [ "Varia" ],
       "nodes": [
         {
           "id": 1,
@@ -1634,7 +1633,6 @@
       "name": "Lower Norfair Farming Room",
       "area": "Lower Norfair",
       "subarea": "East",
-      "requires": [ "Varia" ],
       "nodes": [
         {
           "id": 1,
@@ -1672,7 +1670,15 @@
           "to": [
             {
               "id": 2,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 400}
+                ]}
+              ],
               "unlock": null
             }
           ]
@@ -1682,7 +1688,15 @@
           "to": [
             {
               "id": 1,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 500}
+                ]}
+              ],
               "unlock": null
             }
           ]
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -2506,12 +2506,30 @@
           "to": [
             {
               "id": 1,
-              "requires": null,
+              "requires": [ 
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                "Morph",
+                {"cost": [
+                  {"heatFrames": 150}
+                ]}
+              ],
               "unlock": null
             },
             {
               "id": 3,
-              "requires": null,
+              "requires": [ 
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                "Morph",
+                {"cost": [
+                  {"heatFrames": 100}
+                ]}
+              ],
               "unlock": null
             }
           ]
@@ -2521,7 +2539,16 @@
           "to": [
             {
               "id": 2,
-              "requires": null,
+              "requires": [ 
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                "Morph",
+                {"cost": [
+                  {"heatFrames": 90}
+                ]}
+              ],
               "unlock": null
             }
           ]
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -117,7 +117,6 @@
       "name": "Fast Pillars Setup Room",
       "area": "Lower Norfair",
       "subarea": "East",
-      "requires": [ "Varia" ],
       "nodes": [
         {
           "id": 1,
@@ -188,12 +187,33 @@
           "to": [
             {
               "id": 4,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 200}
+                ]}
+              ],
               "unlock": null
             },
             {
               "id": 5,
-              "requires": [ "canUsePowerBombs" ],
+              "requires": [                
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                "canUsePowerBombs",
+                {"cost": [
+                  {"heatFrames": 200},
+                  {"ammo": [{
+                    "type": "PowerBomb",
+                    "count": 1
+                  }]}
+                ]}                
+              ],
               "unlock": null
             }
           ]
@@ -203,7 +223,15 @@
           "to": [
             {
               "id": 5,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 50}
+                ]}
+              ],
               "unlock": null
             }
           ]
@@ -213,7 +241,15 @@
           "to": [
             {
               "id": 5,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 50}
+                ]}
+              ],
               "unlock": null
             }
           ]
@@ -223,7 +259,15 @@
           "to": [
             {
               "id": 1,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 150}
+                ]}
+              ],
               "unlock": null
             }
           ]
@@ -233,17 +277,46 @@
           "to": [
             {
               "id": 1,
-              "requires": [ "canUsePowerBombs" ],
+              "requires": [                
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                "canUsePowerBombs",
+                {"cost": [
+                  {"heatFrames": 350},
+                  {"ammo": [{
+                    "type": "PowerBomb",
+                    "count": 1
+                  }]}
+                ]}                
+              ],
               "unlock": null
             },
             {
               "id": 2,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 50}
+                ]}
+              ],
               "unlock": null
             },
             {
               "id": 3,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 50}
+                ]}
+              ],
               "unlock": null
             }
           ]
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -1148,7 +1148,6 @@
       "name": "Wasteland",
       "area": "Lower Norfair",
       "subarea": "East",
-      "requires": [ "Varia" ],
       "nodes": [
         {
           "id": 1,
@@ -1192,7 +1191,41 @@
           "to": [
             {
               "id": 4,
-              "requires": [ "canUsePowerBombs" ],
+              "requires": [ 
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},                
+                "canUsePowerBombs",
+                {"or": [
+                  "canWalljump",
+                  {"cost": [
+                    {"heatFrames": 200}
+                  ]}
+                ]},
+                {"and" :[
+                  {"cost": [
+                    {"heatFrames": 800}
+                  ]},
+                  {"or": [
+                    {"cost": [
+                      {"enemyDamage": [{
+                        "enemy": 79,
+                        "hits": 1
+                      }]}
+                    ]},
+                    "ScrewAttack",
+                    "Super",
+                    {"and": [
+                      "Plasma",
+                      {"or": [
+                        "Ice",
+                        "Wave"
+                      ]}
+                    ]}
+                  ]}
+                ]}
+              ],
               "unlock": null
             }
           ]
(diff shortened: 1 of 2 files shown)


region/lowernorfair/east.json
```diff
@@ -537,39 +552,114 @@
...
+                      {"and": [
+                        "Charge",
+                        "Ice",
+                        "Spazer",
+                        "Wave",
+                        {"cost": [
+                          {"heatFrames": 1150}
+                        ]}
+                      ]},
+                      {"canComeInCharged": {
```
fdtaillefer: I am guessing that this has no associated cost because it's just an indicator that we don't need to kill the pirates, and the spark and travel cost is in the second part of the reqs, where actual travel is addressed.

region/lowernorfair/east.json
```diff
@@ -537,39 +552,114 @@
...
+                        "fromNode": 2,
+                        "framesRemaining": 120
+                      }}
+                    ]}
+                  ]}
+                ]},
+                {"or": [
+                  {"and": [
+                    {"canComeInCharged": {
```
fdtaillefer: I think the shinespark needs to be entirely separate from the "kill pirates vs climb up" split, because otherwise combinations that use a shinespark for only one of the two are in logic even though they don't make sense.

region/lowernorfair/east.json
```diff
@@ -842,9 +854,28 @@
           "to": [
             {
               "id": 1,
-              "requires": [ "canSwim" ],
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"and": [
```
fdtaillefer: This `and` (and the next one) are already within an implicit `and`. Were they intended to be two options in an `or`?

region/lowernorfair/east.json
DIProgan: Yes I had to put it in twice to make that structure work.

region/lowernorfair/east.json
```diff
@@ -842,9 +854,28 @@
...
+                  "HiJump",
+                  "canSuitlessLavaDive",
```
fdtaillefer: I think the lack of a tech in the reqs for the dive was indeed a mistake on my part. Are we comfortable with using the same tech as suitless lava though? A suitless acid dive has better physics than suitless lava but deals a lot more damage.

region/lowernorfair/east.json
DIProgan: Would have to add extra heatProof / canHeatRun into it but it's more clean than two canComeInCharged objects for sure.

region/lowernorfair/east.json
```diff
@@ -960,7 +967,36 @@
...
+                  {"and": [
+                    "heatProof",
+                    {"or": [
+                      "canUseMorphBombs",
+                      {"and": [
+                        "canUsePowerBombs",
+                        {"cost": [
+                          {"ammo": [{
+                            "type": "PowerBomb",
+                            "count": 3
```
fdtaillefer: There are ways around using 2 of those PBs. 1 can be avoided with Screw attack, and 1 can be avoided with either X-ray stand-up + ceiling clip, or Spring Ball.

Considerations like this are why I'm more in favor of leaving PB costs to their own task that focuses on them.

region/lowernorfair/east.json
DIProgan: They're meant to be handling gravity vs suitless but I forgot the "or" and also for some reason the gravity one got no heatframes.

region/lowernorfair/east.json
```diff
@@ -960,7 +967,36 @@
...
+                          }]}
+                        ]}
+                      ]}
+                    ]}
+                  ]},
+                  {"and": [
+                    "canHeatRun",
+                    "canUsePowerBombs",
```
fdtaillefer: Is there a reason morph bombs here are kept out of heatRun logic? It feels like it should only add a few hundred frames.

region/lowernorfair/east.json
fdtaillefer: Makes sense. Should be fine once fixed then.

region/lowernorfair/east.json
```diff
@@ -970,14 +1006,75 @@
...
+                            "type": "PowerBomb",
+                            "count": 2
+                          }]}
+                        ]}
+                      ]}
+                    ]}
+                  ]},
+                  {"and": [
+                    "canHeatRun",
+                    "canUsePowerBombs",
```
fdtaillefer: Same question as 2 to 3

region/lowernorfair/east.json
```diff
@@ -970,14 +1006,75 @@
...
               "unlock": null
             },
             {
               "id": 5,
-              "requires": [ "canKillRedKiHunters" ],
+              "requires": [
+                {"or": [
+                  {"and": [
+                    "heatProof",
+                    "canKillRedKiHunters"
```
fdtaillefer: If taking KiHunter hits instead of killing them is expected for a heat run, surely it should also be a valid strat when heatRuns are disallowed but your are heatproof?

region/lowernorfair/east.json
```diff
@@ -1192,7 +1191,41 @@
...
             {
               "id": 4,
-              "requires": [ "canUsePowerBombs" ],
+              "requires": [ 
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},                
+                "canUsePowerBombs",
+                {"or": [
```
fdtaillefer: I get the feeling this `or` closed too early and the walljump and heatframes should've been in an `and`?

region/lowernorfair/east.json
```diff
@@ -1192,7 +1191,41 @@
...
+                        "enemy": 79,
+                        "hits": 1
+                      }]}
+                    ]},
+                    "ScrewAttack",
+                    "Super",
+                    {"and": [
+                      "Plasma",
+                      {"or": [
+                        "Ice",
```
fdtaillefer: I see Plasma was rejected as an option here. I'm guessing that's on the basis that it takes 6 hits instead of 4 and that's the line where avoiding hits becomes unlikely?

region/lowernorfair/east.json
```diff
@@ -1225,19 +1318,107 @@
...
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                "Morph",
+                {"or": [
+                  "Bombs",
+                  {"and": [
+                    "PowerBomb",
```
fdtaillefer: There seems to be some inconsistency on whether the item itself is required alongside its ammo cost. It was handled differently with Supers in writg. Any opinion on which one it should be?

region/lowernorfair/east.json
```diff
@@ -1491,7 +1498,52 @@
           "to": [
             {
               "id": 1,
-              "requires": [],
+              "requires": [
```
fdtaillefer: Do you think it's ok to put this in for now? It will have to be altered later because this is all the cost of unlocking the left door, not of traveling through the room. I think at the minimum we should have a FIXME note explaining that.

region/lowernorfair/east.json
```diff
@@ -1491,7 +1498,52 @@
...
+                    {"cost": [
+                      {"heatFrames": 2000}
+                    ]}
+                  ]},
+                  {"and": [
+                    "Super",
+                    {"cost": [
+                      {"heatFrames": 700}
+                    ]}
+                  ]}
```
fdtaillefer: Should we add the speed echoes strat in there too? I feel like it should be locked behind a tech of some kind but it is a quick way to kill them.

region/lowernorfair/east.json
DIProgan: It's the same ballpark. Seems ok.

region/lowernorfair/east.json
```diff
@@ -1592,7 +1592,15 @@
           "to": [
             {
               "id": 2,
-              "requires": null,
+              "requires": [
```
fdtaillefer: I frequently see people with Gravity going through the acid, but I don't know if it's just a speed strat. Did you check that one for cost-effectiveness?

region/lowernorfair/east.json
```diff
@@ -1602,7 +1610,19 @@
...
               "id": 1,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                {"cost": [
+                  {"heatFrames": 450},
+                  {"enemyDamage": [{
```
fdtaillefer: I feel like that enemy damage probably shouldn't be mandatory if you're heatproof?

fdtaillefer:
I don't see any changes to Ridley's room. I suggest putting down `heatProof` in his `unlock`.

region/lowernorfair/east.json
DIProgan: I think I'll leave it as is for now to not clutter it too much.

fdtaillefer: Could you put in a FIXME note about it, then?

region/lowernorfair/east.json
DIProgan: From what I gather if you're not fast to get through the morph ball passage the Kihunter above becomes hell to handle.

region/lowernorfair/east.json
```diff
@@ -1953,31 +1977,79 @@
           "to": [
             {
               "id": 1,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "heatProof",
+                  "canHeatRun"
+                ]},
+                "canUsePowerBombs",
```
fdtaillefer: I don't understand the PB cost here

region/lowernorfair/east.json
DIProgan: I'll add it

region/lowernorfair/east.json
DIProgan: Isn't that using Morph Ball though?

region/lowernorfair/east.json
fdtaillefer: Hmm. Yeah he seems to like moving down to the door level. However, he doesn't seem to actually stick to your passage up so if you want an extra second or two it seems quite doable to avoid him with morph. He actually seems to like going back up.

region/lowernorfair/east.json
fdtaillefer: The bottom door requires a PB only once, though. If it's in any way possible to open that door, come out, and farm PBs, then that PB cost can likely be ignored, unlike actual passage.

region/lowernorfair/east.json
DIProgan: There's an additional 200 heatframe cost for not being able to walljump up there. The dev will have to add the different heatframes up into one.

region/lowernorfair/east.json
fdtaillefer: I didn't use Morph Ball when I messed around with evading them. There is a reliable way to avoid the first KiHunter coming out of the save door (Angle up to stop momentum, then jump to bonk on the ceiling. Then run left, he *will* dive down, and you can run and jump above him to the right). The second guy is fairly easy to evade. The third guy usually requires taking a hit, and Morph actually hasn't helped me avoid him much.

region/lowernorfair/east.json
fdtaillefer: Oh! Makes sense.

region/lowernorfair/east.json
DIProgan: Yes at some point many of these rooms just become kind of a drag if you need to stay around and avoid everything.

fdtaillefer: On the one hand I think killing the one Dessgeega at 4 with just Plasma and getting to PBs of shame is reasonable to have in logic, but I can see the argument for just having one hard req for killing Dessgeegas while heatrunning. So this is fine IMO.

region/lowernorfair/east.json
DIProgan: Adding Super ammo in writg was kind of a mistake. I will remove super ammo and add "Super" req. Will need some kind of way to handle these big uses of ammo though.

region/lowernorfair/east.json
```diff
@@ -2205,14 +2217,31 @@
           "to": [
             {
               "id": 2,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "canWalljump",
+                  "SpaceJump"
```
fdtaillefer: There are so many alternatives to walljump besides space jump (Successively climbing on two frozen firefleas, midair springball jump, ibj). I don't mind walljump being there only sporadically as a req for now, but when it's put in it should be fully put in.

fdtaillefer: Ok I'm stupid. The Firefleas won't work for the second portion :P

fdtaillefer: Actually hjb + Speed Booster also works

region/lowernorfair/east.json
DIProgan: I couldn't get it done. Overall it seems pretty useless as it's quicker with 6 supers but in the rare case one don't have supers it could be a thing. Probably gonna need a video though.

region/lowernorfair/east.json
DIProgan: Definitely a speed strat

region/lowernorfair/east.json
fdtaillefer: IIRC there was the idea of using an `enemyKill` in costs, which we discussed during preCathedral, but it wasn't clear how it should work.

region/lowernorfair/east.json
```diff
@@ -2610,7 +2609,41 @@
           "to": [
             {
               "id": 4,
-              "requires": [ "canKillRedKiHunters" ],
+              "requires": [ 
+                {"or": [
+                  {"and": [
+                    "heatProof",
+                    "canKillRedKiHunters"
```
fdtaillefer: Same thought here as Red KiHunters shaft. If we can be expected to heatrun here without the means to kill them, it should probably be expected when heatproof as well.

region/lowernorfair/east.json
```diff
@@ -2641,19 +2692,100 @@
           "to": [
             {
               "id": 1,
-              "requires": [ "canKillRedKiHunters" ],
-              "unlock": null
+              "requires": [ 
+                {"or": [
+                  {"and": [
+                    "heatProof",
+                    "canKillRedKiHunters"
```
fdtaillefer: Same thought as the reverse path

fdtaillefer:
Have you considered the application of `canHitBox` in Red Kihunter Shaft?

region/lowernorfair/east.json
DIProgan: Putting the reqs in unlock as discussed.

region/lowernorfair/east.json
DIProgan: Will fix

region/lowernorfair/east.json
DIProgan: In order to avoid taking damage from the multiviolas there are really no other good, fast and reliable option for the heatruns. However I guess I could stick it into the HeatRun portion.

fdtaillefer: Oh, I see. Do you think tanking the damage should be an option? It doesn't seem like it would be that costly if you had Gravity Suit.

region/lowernorfair/east.json
DIProgan: Adding canUseFrozenEnemies between nodes 4 & 5. Adding HiJ+SpeedBooster for 5 to 2.

fdtaillefer: You should also add hjb + springBallJumpMidAir, and replace Space Jump with canFly.

region/lowernorfair/east.json
DIProgan: Roger, roger

region/lowernorfair/east.json
DIProgan: > Have you considered the application of `canHitBox` in Red Kihunter Shaft?

Going down from the top it only helps about 15 frames. Going up it's too inconsistent with the RNG kihunters placement sometimes lurking straight above etc.

region/lowernorfair/east.json
DIProgan: Alrighty

region/lowernorfair/east.json
DIProgan: You're right if you wait a little while longer he seems to go away. Adding bombs with that in mind.

region/lowernorfair/east.json
DIProgan: Code looks better without the divide between heat and proof anyway ^^

region/lowernorfair/east.json
DIProgan: The first one is easy but there's a couple crappy RNG versions for the other two. One is where both the 2nd and 3rd is bunched up in the same place and blocks your path. Often yields 2 hits. Another one is where the 3rd is almost down at the middle section. 1 hit is the usual but I feel 2 happen too often.

fdtaillefer: I have no issues with 2 hits.

region/lowernorfair/east.json
DIProgan: The turnout is usually 3x Multiviola and 2x Desgeega hits. It's survivable, even suitless with a lot of health. Putting it in.

region/lowernorfair/east.json
DIProgan: Done!

region/lowernorfair/east.json
fdtaillefer: This looks like it will complicate the link's reqs significantly though. What's your opinion on adding a junction at the bottom right to simplify this?

Issue #113, opened 2019-02-07
Runway clear requirement


DIProgan:
For a combined runway to function properly usually only one of the parts can have extra requirements to solve beforehand. However some runways are said to require nothing and still needs to be cleared up. See the two top doors in Red Kihunter Shaft. We need to have some type of indication of this such as Requires: [ "cleanup" ]

fdtaillefer:
I think the simplest way to deal with this is to have two runways with no requirements, and for the rule to be that only the shortest runway in the destination room is usable.

Later on we could consider replacing the `null` req of the larger runway with a `canDestroyShotBlocks` which could become relevant for a beamless mode.

DIProgan:
Considers it closed with #129

Issue #110, opened 2019-02-06
Additional heatframes for canLeaveCharged required


DIProgan:
Many heated rooms with canLeaveCharged objects need additional heatFrames to compensate for the time it takes to perform the charge when not in possession of Varia.

fdtaillefer:
I wonder to what extent this also applies to runways in heated rooms?

DIProgan:
It's a thing for sure.

DIProgan:
#99 #100 #101 #104 #114

PR #119, opened 2019-02-11
Runway changes mostly Norfair East

DIProgan:
While working with creating additional runways to accommodate for issue #110  & #113 I ended up having to add heatdata to every runway in all the heatrooms to function with the case that logic demands you to visit a heatroom for the only purpopse to use that runway suitless. So the frames added represents running from the door to the end of the runway and back again trying to emulate some 2-tap'ish movement on the way back. 

That's why the commits are a little all over the place this time. Workload increased a bit. 

Edit: Rest of the runways from the other regions found their way into this as well.
Edit2: And more...


region/norfair/west.json
```diff
@@ -1542,6 +1546,7 @@
                     {"or": [
                       "Grapple",
                       "SpaceJump",
+                      "canSpringBallJumpMidAir",
```
fdtaillefer: I didn't exclude it by mistake, it's missing about half a tile of height off a crouch jump. What did you do?

fdtaillefer: Though it is doable with hjb and that doesn't seem to be there.

region/lowernorfair/west.json
```diff
@@ -589,7 +589,17 @@
             "runways": [
               {
                 "length": 12,
-                "requires": null
+                "requires": [
+                  {"or": [
+                    "heatProof",
+                    "canHeatRun"
+                  ]},
+                  "reset",
```
fdtaillefer: I think this needs to be thought out a bit. I think it could be an array of all nodes you could use to properly reset. That way, we'll be able to use the same thing at Early Supers for travelling from node 1 to 2.

region/norfair/east.json
```diff
@@ -1580,9 +1735,29 @@
           "unlock": [ "canOpenRedDoors" ],
           "sparking": {
             "runways": [
+              {
+                "length": 7,
```
fdtaillefer: What does this smaller runway represent?

region/lowernorfair/east.json
```diff
@@ -125,9 +141,34 @@
...
+              },
               {
                 "length": 12,
-                "requires": null
+                "requires": [
+                  {"or": [
+                    "heatProof",
+                    "canHeatRun"
+                  ]},
+                  {"or": [
```
fdtaillefer: Would charge plasma be a comparable time?

region/lowernorfair/east.json
```diff
@@ -812,9 +909,30 @@
...
+              },
               {
                 "length": 19,
-                "requires": null
+                "requires": [
+                  {"or": [
+                    "heatProof",
+                    "canHeatRun"
+                  ]},
+                  "Super",
```
fdtaillefer: Why specifically Super? There's no other good way to kill it?

region/lowernorfair/east.json
```diff
@@ -894,9 +1012,29 @@
...
+                    "heatProof",
+                    "canHeatRun"
+                  ]},
+                  {"cost": [
+                    {"heatFrames": 85}
+                  ]}
+                ]
+              },
               {
                 "length": 12,
```
fdtaillefer: This one will require a reset from either of the top doors, right?

region/lowernorfair/east.json
```diff
@@ -929,9 +1075,29 @@
...
+                    "heatProof",
+                    "canHeatRun"
+                  ]},
+                  {"cost": [
+                    {"heatFrames": 75}
+                  ]}
+                ]
+              },
               {
                 "length": 12,
```
fdtaillefer: This one will require a reset from either of the top doors, right?

region/lowernorfair/east.json
```diff
@@ -2321,8 +2584,20 @@
           "sparking": {
             "runways": [
               {
-                "length": 7,
+                "length": 0,
```
fdtaillefer: What does this represent?

region/norfair/east.json
```diff
@@ -1567,7 +1714,15 @@
             "runways": [
               {
                 "length": 31,
-                "requires": null
+                "requires": [
```
fdtaillefer: I think the runways in here might need a reset at the opposite door? The crumble blocks ruin them otherwise.

region/lowernorfair/east.json
```diff
@@ -1656,11 +1656,11 @@
                     "heatProof",
                     "canHeatRun"
                   ]},
-                  "Super",
                   {"cost": [
-                    {"heatFrames": 1200}
+                    {"heatFrames": 400}
```
fdtaillefer: I thought the room was symmetrical, is that not the case? The two frame counts are different right now. Also, only one side has the node,

fdtaillefer: Additionally, should this be a tech?

fdtaillefer: Finally, there should probably a way to use this by killing the enemies as well? We could have it require heatproof at worst.

region/lowernorfair/east.json
fdtaillefer: Also, there should probably a way to use this by killing the pirate by any slower means (except maybe morph bombs)? We could have it require heatproof at worst.

region/lowernorfair/east.json
fdtaillefer: Also, there should probably a way to use this by killing the pirate by any slower means (except maybe morph bombs)? We could have it require heatproof at worst.

region/lowernorfair/west.json
```diff
@@ -87,10 +94,19 @@
...
-                {"cost": [
-                  {"heatFrames": 1200}
+                {"and": [
+                  "SpaceJump",
+                  "canUsePowerBombs",
+                  {"cost": [
+                    {"heatFrames": 400}
+                  ]}
+                ]},
+                {"and": [
```
fdtaillefer: ??

region/lowernorfair/west.json
```diff
@@ -87,10 +94,19 @@
                   "heatProof",
                   "canHeatRun"
                 ]},
-                "SpaceJump",
-                "canUsePowerBombs",
-                {"cost": [
-                  {"heatFrames": 1200}
+                {"and": [
```
fdtaillefer: We're already inside an implicit `and`. Is this where you wanted it?

region/brinstar/red.json
```diff
@@ -324,91 +318,103 @@
...
+                    {"or": [
+                      {"and": [
+                        "HiJump",
+                        "canDash"
+                      ]},
+                      {"and": [
+                        "HiJump",
+                        "canSpringBallJump"
+                      ]},
+                      "canIBJ"
```
fdtaillefer: I think maybe we should give more leniency in spike hits for IBJ. 4 spike hits is with flawless execution, but it's not as easy to setup the ibj as it is to set up the other two strats, and the wavers can get in the way and destroy your attempt, while the reqs don't ask for the PBs you would need to salvage your attempt then.

region/wreckedship/main.json
```diff
@@ -807,6 +807,9 @@
                 {"canShineCharge": {
                   "usedTiles": 31
                 }},
+                {"cost": [
+                  {"shinesparkFrames": 40}
```
fdtaillefer: When we talked about it, I thought you preferred putting the shinespark  costs inside the `canShineCharge` and `canComeInCharged` objects. Did you change your mind?

region/brinstar/green.json
```diff
@@ -1110,11 +1109,10 @@
               "requires": [
                 {"or": [
                   "Morph",
-                  "FIXME"
+                  "canTunnelCrawl"
```
fdtaillefer: I think this needs to be a separate, `canMorphlessTunnelCrawl` tech. Because without morph, you are going to softlock if you mess up.

Also, the same tech is required to make it back from 4 to 1.

Finally, there's a one-way strat to go from 1 to 4 with `canTurnaroundAimCancel`, it just won't get you back out.

region/brinstar/pink.json
```diff
@@ -72,12 +72,28 @@
...
+                      {"cost": [
+                        {"shinesparkFrames": 95}
+                      ]}
+                    ]}
+                  ]}
+                ]}
               ],
               "unlock": null,
-              "note": "It's possible to get this shinespark doable with just no e-tanks with crazy good walljumping, or with HiJump and just good walljumping"
+              "note": "It's possible to get this shinespark doable with just no e-tanks with good walljumping, or with HiJump and walljumping"
```
fdtaillefer: Shinesparks cut off at 29 health so getting out without e-tanks requires making it high enough to cut the shinespark frames to 70. With that in mind, you should take another look at the whole commit.

region/wreckedship/main.json
DIProgan: No but it didn't feel like we had come to a definite agreement about it. Guess we did ^^

region/brinstar/pink.json
DIProgan: True, forgot about that part. Will investigate.

region/norfair/west.json
DIProgan: In my notes the HiJump is included just forgot to put it in the code.

region/lowernorfair/west.json
DIProgan: {"reset": [1,2,3]} ?

region/norfair/east.json
DIProgan: Coming in from the Right door running up to the crumbleblocks and charging back in.

region/lowernorfair/east.json
DIProgan: Adding canKillYellowSpacePirates to the heatProof part. Charge+Plasma became reliable when I closed the gap to the pirates a bit so that they won't shoot their beams. Inserting it with a slight heatframes increase to the bunch.

region/lowernorfair/east.json
DIProgan: If you use the fast Supers you can escape an extra cycle of acid rising onto the runway but I guess the cycle isn't that extremely long so I'm including the slower alternatives of Missiles and Charge+Plasma.

region/lowernorfair/east.json
DIProgan: True

region/lowernorfair/east.json
DIProgan: The issue with downward slopes.

region/norfair/east.json
DIProgan: Indeed adding simpified "reset" for now

region/lowernorfair/east.json
DIProgan: It's a bit tricky with the heatframes here since you need to kill the pirates in one direction and not the other but as far as I can remember this is just an artifact from the time I didn't consider hitboxing to achieve the charge. Making them both 400 for now.

DIProgan: My suggestion is canHitBox here and canPlasmaHitBox at Kihunters.

DIProgan: That's the whole issue with duplicating the reqs for a simple runway. Where to draw the line or should the full thing be at both places?

region/lowernorfair/west.json
DIProgan: It's the walljump way to get back up to the statue for whatever reason one would want that after being there once in the case of some mod removing SpaceJump as a requirement and god knows what else.

region/brinstar/red.json
DIProgan: If you use sideway IBJ I think you can get away with 3 spikehits but yes I truly hate setting up IBJ onto spikes.  Adding 2 hits to the IBJ for leniency.

region/brinstar/green.json
DIProgan: Changing

region/brinstar/pink.json
DIProgan: If you have an energy tank it won't cut out at 70 but use up the amount needed to the ceiling. This unregistered health drain while small might still be needed in future rooms causing a potentially broken seed so I will not include the InsaneWalljump but keep a similar note about it.

region/lowernorfair/west.json
fdtaillefer: something like that. Maybe `resetFromNode`?

region/norfair/east.json
fdtaillefer: Thought it might be that, but wasn't sure because there was no corresponding change on the other end.

region/lowernorfair/east.json
fdtaillefer: Oh... I thought that whole thing had been moved to the door's unlock, turns out it's in the link's unlock. That doesn't work, you have to unlock the door regardless of what side you're coming in from.

Then charging the spark wouldn't care about this.

fdtaillefer: If duplication becomes an issue, we could consider using a helper.

region/brinstar/green.json
fdtaillefer: I guess this means I should probably add Morph to `canTunnelCrawl`'s reqs though, and update the note.

region/brinstar/pink.json
fdtaillefer: Do you know how many frames there are between getting out and hitting the ceiling? I think that energy difference can likely be covered by health drops that the logic won't consider.

region/lowernorfair/east.json
```diff
@@ -157,22 +157,31 @@
...
+                        ]}
+                      ]}
+                    ]}
+                  ]}                  
                 ]
               }
             ]
           }
-        },
+        }
```
fdtaillefer: This probably won't compile

region/lowernorfair/west.json
```diff
@@ -94,18 +94,20 @@
...
-                ]},
-                {"and": [
-                  "UsedAcidChozoStatue",
-                  "canUsePowerBombs",
-                  {"cost": [
-                    {"heatFrames": 450}
+                {"or": [
+                  {"and": [
+                    "SpaceJump",
+                    "canUsePowerBombs",
```
fdtaillefer: Are the power bombs still needed for either of the traversal options, btw?

region/lowernorfair/east.json
```diff
@@ -1664,7 +1664,7 @@
                   ]},
                   {"or": [
                     {"and": [
-                      "canHitBox",
+                      "canPlasmaHitbox",
```
fdtaillefer: This room was supposed to stay as `canHitbox`, right?

region/lowernorfair/west.json
fdtaillefer: It looks like we're going to leave this for a later PR with issue #121, so resolving conversation.

region/lowernorfair/east.json
fdtaillefer: Do you want to move this concern to another issue? The PR is already bloated enough as is

region/lowernorfair/east.json
DIProgan: Well f me

region/lowernorfair/west.json
DIProgan: If you're asking if PB's are needed to dig into the statue after the first time, yes. I wasn't sure how to do about this one but figured it was best to keep it in both at unlock and link.

fdtaillefer: Ok. It wasn't my intent, but I don't suppose it matters much either way.

region/lowernorfair/east.json
fdtaillefer: (Issue #125 created)

PR #120, opened 2019-02-12
Junction refactor: Outer Maridia

fdtaillefer:


region/maridia/outer.json
```diff
@@ -166,79 +166,91 @@
...
+                    "usedTiles": 1
+                  }},
+                  {"adjacentRunway": {
+                    "fromNode": 3,
+                    "usedTiles": 1
+                  }}
+                ]}
+              ],
+              "unlock": null,
+              "note": "This direct link involves using an adjacent room without water physics to initiate a jump that takes Samus through the top door. The runway is to get enough speed to clear the edge of the tube."
```
DIProgan: Devs need to know that they can't use adjacentRunways with "canSwim" reqs for this one somehow.

region/maridia/outer.json
```diff
@@ -550,215 +597,35 @@
...
-              "id": 6,
-              "requires": [ "never" ],
-              "unlock": null
-            },
-            {
-              "id": 7,
-              "requires": [ "Morph" ],
-              "unlock": null
+              "unlock": null,
+              "note": "FIXME There are ways to do this suitless if you can come in charged, especially if you can carry momentum, but how to test this?"
```
DIProgan: I would leave out the one from node 2(right door) completely but the bot one looks good with framesRemaining 120 with no momentum to speak of.

region/maridia/outer.json
```diff
@@ -1183,667 +1022,388 @@
...
-                      {"and": [
-                        "HiJump",
-                        "canSpringBallJump"
-                      ]}
-                    ]}
+                  {"and":[
+                    "canSwim",
+                    "HiJump",
+                    "SpeedBooster",
+                    "canWalljump"
```
DIProgan: This feels like a TrickyWalljump

region/maridia/outer.json
```diff
@@ -2058,6 +1618,26 @@
               "unlock": null
```
DIProgan: Not sure how to comment outside of the changed portions so doing it here instead, probably easiest if you do the change. Around row 1594 there's a Mama Turtles req for canSwim or SuitlessMaridia that tries to combine that with reqs that can't be done suitless such as canFly.

region/maridia/outer.json
```diff
@@ -2108,6 +1688,12 @@
               }
             ]
           }
+        },
```
DIProgan: I noticed the Boyon Gate Hall got canSwim reqs on all three runways even though they shouldn't Probably easiest if you fix those.

region/maridia/outer.json
fdtaillefer: That's a damn good point. I have no idea how this could be done cleanly. Should I just take this out of logic until we figure it out?

region/maridia/outer.json
fdtaillefer: I ran some tests by taking out gravity suit after establishing a full-speed charge in the adjacent room, and it's possible to store the spark and use it. I think we could have door 2 by requiring 180 frames (so basically need to establish the entire charge in the adjacent room). I would be inclined to do that, unless you have objections.

region/maridia/outer.json
fdtaillefer: trickyWallJumps are usually delayed wallJumps or the like. Getting the correct speed to get the correct jump height does appear to be tricky, but you're supposed to get high enough to just do a simple walljump off the edge. What's your though on replacing that SpeedBooster by a canTrickyDashJump?

region/maridia/outer.json
fdtaillefer: I tried the 120 suitless and it seems doable. That means 120 with gravity is way too lenient. After more tests on that, I'll bring it down to 80.

region/maridia/outer.json
fdtaillefer: One thing we could do is straight-up put a `submerged` property on our doors, then I imagine we could also put the `submerged` property into the `adjacentRunway` object, where a default value of `null` would mean "doesn't matter".

Is that too specific?

DIProgan: I think it just have to go. Too much work for seemingly one occurrence.

region/maridia/outer.json
DIProgan: I removed the water property in the adjacent room with smile and while I couldn't do it sitting at the very last second before door transition, I  could get it done sitting before the door and using the momentum of the two door tiles. I would say you're good to go.

region/maridia/outer.json
DIProgan: Sounds good

region/maridia/outer.json
fdtaillefer: I think your original point remains valid. But you're also right, luckily in this case having Gravity does make this strat redundant. I'll add something to the note to explain this.

fdtaillefer:
> 
> 
> Looks pretty dandy. At some point we need to decide what to do with the Morph Passage transition. It has no connection data for the nodes concerned so at the moment it is unreachable by the model.

It doesn't? I didn't realize that. There's nothing special about it except it's a different type than horizontal doors. I always intended to handle that as a regular connection, so I'll add it as such.

fdtaillefer:
Ok I think that covers everything. If you can confirm, this will be ready to be merged.

DIProgan:
How would you ent rando the morph passage though? Seems to me both rooms always need to be together but let's merge!

fdtaillefer:
> How would you ent rando the morph passage though? Seems to me both rooms always need to be together but let's merge!

You wouldn't.
Well, you would, but you're randomizing using a pool of one existing connection so they'll always connect together. I think that's fine.

PR #128, opened 2019-02-18
Junction refactor: Inner Maridia (pink)

fdtaillefer:


region/maridia/inner-pink.json
```diff
@@ -452,7 +305,18 @@
...
+                "canSwim",
+                {"canShineCharge": {
+                  "usedTiles": 31
+                }}
+              ],
+              "unlock": null,
+              "note": "This direct link is just for the shinespark"
+            },
+            {
+              "id": 9,
```
DIProgan: I'd like to see the canWalljump and all the other ways to get up here.. Same at node 2 to node 5.

region/maridia/inner-pink.json
```diff
@@ -985,33 +516,23 @@
...
+                  ]},
+                  {"and": [
+                    "canSuitlessMaridia",
+                    {"or": [
+                      "canMochtroidClip",
+                      "canPuyoClip"
+                    ]}
+                  ]}
+                ]},
                 "canSwim"
```
DIProgan: This looks like a leftover that shouldn't be there?

region/maridia/inner-pink.json
```diff
@@ -1317,86 +830,10 @@
                       {"canShineCharge": {
```
DIProgan: At 826/827 there's "or" & "and" without options.

region/maridia/inner-pink.json
```diff
@@ -1317,86 +830,10 @@
...
-                  {"and": [
-                    "canSuitlessMaridia",
-                    "Morph",
-                    {"or": [
-                      "canSpringBallJumpMidAir",
-                      "HiJump",
-                      "Ice"
+                      {"canComeInCharged": {
+                        "fromNode": 4,
+                        "framesRemaining": 180
```
DIProgan: Tested this out and got it done in ~100 frames so 120 should be a good fit.

region/maridia/inner-pink.json
```diff
@@ -1409,62 +846,36 @@
...
+                    "Morph",
                     {"or": [
                       "canPassBombPassages",
                       "canUseSpringBall",
                       "can2HighWallMidAirMorph",
-                      {"and": [
-                        "canSuitlessMaridia",
-                        "Morph"
-                      ]}
+                      "canSuitlessMaridia"
```
DIProgan: Is this referring to removing ones grav suit to make the jump easier?

region/maridia/inner-pink.json
```diff
@@ -1570,23 +979,6 @@
...
-                      "HiJump"
-                    ]}
-                  ]}
-                ]}
-              ],
-              "unlock": null
-            },
             {
               "id": 4,
               "requires": [
```
DIProgan: canWalljump package plz ;)

region/maridia/inner-pink.json
```diff
@@ -1983,23 +1220,17 @@
...
-                            "Grapple",
-                            "canSandMochtroidClimb"
-                          ]
-                        }
-                      ]
-                    }
-                  ]
-                }
+                {"or": [
+                  "canSwim",
```
DIProgan: More walljumps

region/maridia/inner-pink.json
```diff
@@ -2393,6 +1542,10 @@
                     {"or": [
                       "canGravityJump",
                       "canFly",
+                      {"canComeInCharged": {
+                        "fromNode": 1,
+                        "framesRemaining": 50
+                      }},
```
DIProgan: No directlink on this one?

region/maridia/inner-pink.json
```diff
@@ -2613,13 +1764,11 @@
...
               "requires": [
-                {
-                  "or": [
-                    "HiJump",
-                    "SpaceJump",
-                    "canTrickyWallJump"
-                  ]
-                },
+                {"or": [
+                  "HiJump",
```
DIProgan: I think I'd like to see a canWalljump even with HiJump because of the last two jumps into the door.

region/maridia/inner-pink.json
fdtaillefer: My opinion has always been that getting walljumps everywhere should be its own undertaking. Most of the places I've put walljumps have been because they were already in and I was unwilling to take them out. I don't think this needs to be part of the already massive undertaking that junction refactor was.

region/maridia/inner-pink.json
fdtaillefer: I didn't change the logic here this time around but yeah, that doesn't look right. Good catch.

region/maridia/inner-pink.json
fdtaillefer: Yup. Removed a bunch of stuff there and missed the cleanup.

region/maridia/inner-pink.json
fdtaillefer: Oh yeah, you can shinespark. I was only thinking of achieving blue suit by the end of the runway and not sparking at all. I'll add this as an alternate strat though.

region/maridia/inner-pink.json
fdtaillefer: Yeah, that makes the mid-air morph viable.

region/maridia/inner-pink.json
fdtaillefer: We'll have to discuss this :P

region/maridia/inner-pink.json
fdtaillefer: It was intentional, because the two nodes might just as well overlap. I don't see a reason for it not to work, in fact I added `fromNode` because I intended not to have direct links and instead use the same `canComeInCharged` for several linear links. Then I realized that doesn't work if you're visiting several nodes because you can't really stop the spark at will.

But here, I think it should be fine.

fdtaillefer:
Adressed everything except walljumps. As far as I'm aware, by and large the model does not offer support for walljumps. The places that do are the exception rather than the rule, and I would leave any further work to a full pass as its own undertaking.

PR #129, opened 2019-02-18
UsableComingIn addition to Runways

DIProgan:
Will do Maridia later


region/cleanup.json
```diff
@@ -432,8 +443,12 @@
           "sparking": {
             "runways": [
               {
-                "length": 8,
+                "length": 6,
```
fdtaillefer: I thought Homing Geemer Room was the section of West Ocean we split off from the rest of the room. That has a basic 12-tile runway with a door on each side. What place are you at?

fdtaillefer:
Do you think you can add the property to runways in the schema? I think there's probably a way to explicitly state a default value of `true`.

region/cleanup.json
DIProgan: The room left of that was my belief.

schema/m3-region.schema.json
```diff
@@ -215,6 +215,12 @@
                               "null"
                             ]
                           },
+                          "UsableComingIn": {
```
fdtaillefer: If not sure if this thing is case sensitive. but to be consistent I suggest the name use camel case (so usableComingIn) and that the same capitalization be used in the "$id"

region/crateria/east.json
```diff
@@ -565,7 +565,7 @@
           "sparking": {
             "runways": [
               {
-                "length": 12,
+                "length": 4,
```
fdtaillefer: There seems to be the same thing with water as on the right side. Does it need two runways as well?

schema/m3-region.schema.json
DIProgan: It's not case sensitive but aligning it as requested

PR #131, opened 2019-02-22
Missing shinespark costs

DIProgan:
Almost all the shinespark costs, various canComeInCharged's and Maridia usableComingIn's.


region/crateria/central.json
```diff
@@ -406,7 +411,8 @@
                   "canDestroyBombWalls",
                   {"canComeInCharged": {
                     "fromNode": 1,
-                    "framesRemaining": 0
+                    "framesRemaining": 0,
+                    "shinesparkFrames": 75
```
fdtaillefer: If you can come in with 180 frames, you can also get through without sparking

region/crateria/central.json
```diff
@@ -497,11 +503,13 @@
               "requires": [
                 {"or": [
                   {"canShineCharge": {
-                    "usedTiles": 25
+                    "usedTiles": 25,
+                    "shinesparkFrames": 40
```
fdtaillefer: The quick charge doesn't need a spark

region/norfair/crocomire.json
```diff
@@ -541,8 +545,13 @@
...
             "runways": [
               {
-                "length": 31,
+                "length": 4,
                 "requires": null
+              },
+              {
+                "length": 31,
+                "requires": null,
+                "usableComingIn": false
```
fdtaillefer: It would be usableComingIn with Gravity Suit actually, and some acid frames, but since there's no charge strats here it doesn't really matter.

region/maridia/inner-green.json
```diff
@@ -461,7 +461,7 @@
           "sparking": {
             "runways": [
               {
-                "length": 7,
+                "length": 6,
```
fdtaillefer: ? Looks like 7 to me

region/maridia/inner-green.json
```diff
@@ -655,6 +655,13 @@
                 "length": 0,
                 "requires": null
               }
+            ],
+            "canLeaveCharged": [
+              {
+                "usedTiles": 31,
+                "framesRemaining": 80,
+                "requires": [ "canUsePowerBombs" ]
```
fdtaillefer: I think it'd be better if had this require `ShaktoolDoneDigging`.

region/maridia/inner-green.json
```diff
@@ -669,6 +676,13 @@
...
```
fdtaillefer: v

region/maridia/inner-pink.json
```diff
@@ -1012,7 +1035,8 @@
                   "canSwim",
                   {"canComeInCharged": {
                     "fromNode": 2,
-                    "framesRemaining": 30
+                    "framesRemaining": 30,
+                    "shinesparkFrames": 45
```
fdtaillefer: You don't need to go all the way up here. You can do a diagonal left spark for 20 frames.

region/maridia/outer.json
```diff
@@ -1074,6 +1053,22 @@
...
+                  "canSuitlessMaridia"
+                ]},
+                {"canComeInCharged":{
+                  "fromNode": 1,
+                  "framesRemaining": 0,
+                  "shinesparkFrames": 150
+                }}
+              ],
+              "unlock": null,
+              "note": "Suitless shinespark direct link"
```
fdtaillefer: Looks like the line for that link (and 4 to 1) is missing from the image.
As a side note, Mt. Everest is the biggest reason I intend to change the color of spark-only links.

region/maridia/outer.json
```diff
@@ -1103,7 +1099,8 @@
                 {"and": [
                 "canSwim",
                   {"canShineCharge": {
-                    "usedTiles": 31
+                    "usedTiles": 31,
+                    "shinesparkFrames": 150
```
fdtaillefer: It looks more like 65*2, so 130.

region/maridia/outer.json
```diff
@@ -1128,6 +1125,23 @@
...
+                ]},
+                {"canComeInCharged":{
+                  "fromNode": 2,
+                  "framesRemaining": 0,
+                  "shinesparkFrames": 60
+                }}
+                ]}
+              ],
+              "unlock": null,
+              "note": "One-way link for shinesparking."
```
fdtaillefer: Seems to me like it would be better to make a spark link directly to 1, since you can do that with a diagonal spark and 1 to 11 is free (while 11 to 1 has reqs)

region/maridia/outer.json
```diff
@@ -1147,12 +1161,57 @@
...
+              "requires": [
+                "canSuitlessMaridia",
+                {"canComeInCharged":{
+                  "fromNode": 3,
+                  "framesRemaining": 0,
+                  "shinesparkFrames": 70
+                }}
+              ],
+              "unlock": null,
+              "note": "One-way link for diagonal suitless shinesparking."
```
fdtaillefer: Looks like this is also missing from the image

region/maridia/outer.json
```diff
@@ -1147,12 +1161,57 @@
...
+                ]},
+                {"canComeInCharged":{
+                  "fromNode": 3,
+                  "framesRemaining": 0,
+                  "shinesparkFrames": 65
+                }}
+              ],
+              "unlock": null,
+              "note": "One-way link for Shinesparking."
             }
```
fdtaillefer: I'm going to hate myself for pointing this out, but there's also a way to spark directly up to 4 with Gravity and some walljumping. Seems to require about 120 frames. It's a diagonal spark right, from just a bit to the left of the right hill peak.

fdtaillefer: ....Actually no one cares cause if you can get there you can drop down to 2 and use that. Never mind!

region/maridia/outer.json
```diff
@@ -1297,6 +1356,30 @@
...
           "from": 9,
           "to": [
+            {
+              "id": 1,
+              "requires": [
+                {"canComeInCharged":{
+                  "fromNode": 5,
+                  "framesRemaining": 90,
+                  "shinesparkFrames": 65
+                }}
```
fdtaillefer: These two sparks probably require `canSwim` or `canSuitlessMaridia` like many of the other ones?

tech.json
```diff
@@ -257,7 +257,14 @@
   },
   {
     "name": "canHerdBabyTurtles",
-    "requires": [],
+    "requires": [
+      {"cost": [
```
fdtaillefer: I feel like moving the baby turtles is separate from actually executing the shinespark, so I'd leave that damage at the actual link level.

What do you think?

region/maridia/inner-green.json
DIProgan: Yeah if that glitch can't be fixed that might be the way to go

fdtaillefer: It's regardless of the glitch actually. It's also just because you don't need to use PowerBombs ever again once Shaktool has dug the room out.

region/crateria/central.json
DIProgan: Ah yes, I somehow thought that was already added.

region/norfair/crocomire.json
DIProgan: You can even use it without gravity many times but didn't want to add a lot of acid junk for no use.

fdtaillefer: Yeah I agree with that.

region/maridia/inner-green.json
DIProgan: There's an enemy there and I wanted to make sure that it doesn't interfer if coming in. Also an extra runway for 1 more tile felt pointless.

fdtaillefer: Well, that tile could be the one that makes a different in some cases. I don't think having two runways would be bad.

region/maridia/inner-pink.json
DIProgan: With 30 framesRemaining I imagined it was balanced towards short time rather than saving health but if you say 30 frames is enough for the diagonal I guess that's the one intended.

region/maridia/outer.json
DIProgan: Yep missed a couple of lines but almost not sorry since it the pic becomes even messier

region/maridia/inner-pink.json
fdtaillefer: I tested it before writing the comment and it's even doable in 20 frames. You really don't need to do far off the door because a diagonal shinespark doesn't go at 45 degrees.

region/maridia/outer.json
fdtaillefer: Well... we don't know what room will be in there. Which actually is a new issue I hadn't considered. I guess just putting a few extra safety frames should mostly be enough.

DIProgan: I'm including some frames on the other side as well for safety.

region/maridia/outer.json
DIProgan: Good point. Time to fire up paint again.

region/maridia/outer.json
DIProgan: Phew dodged one!

tech.json
DIProgan: As I hardly know how it's performed I'm just gonna change back.

region/maridia/outer.json
DIProgan: Yus

region/crateria/central.json
DIProgan: Changed (after some coaching)

region/maridia/outer.json
```diff
@@ -1079,6 +1079,23 @@
         {
           "from": 2,
           "to": [
+            {
+              "id": 1,
+              "requires": [
+                {"or": [
+                  "canSwim",
+                  "canSuiitlessMaridia"
```
fdtaillefer: suitless*

DIProgan: suiiiit

Issue #118, opened 2019-02-10
Missing logic: East Ocean


fdtaillefer:
It's possible to break free of the water suitless in East Ocean with a combination of hjb and space jump, but this isn't currently defined in the logic.

fdtaillefer:
Resolved by #135

PR #135, opened 2019-02-23
Issue #118: Can get through East Ocean suitless using hjb and SpaceJump

fdtaillefer:


DIProgan:
As long as we don't need the SpeingBall I think this is good ^^

Issue #105, opened 2019-01-31
Missing logic in Post Croc Jump Room: Breaking the speed blocks from the left


fdtaillefer:
There is a way to break the PCJR speed blocks coming in from the left door.
If the runway is long enough to generate the blue suit state, it's possible to carry that state all the way to the speed blocks (and break them) using Space Jump.

With Gravity Suit, it's also possible to simply charge a spark in the acid and then shinespark through the blocks.

fdtaillefer:
Resolved by #137

PR #137, opened 2019-02-23
Issue #105: Missing logic in Post Croc Jump Room: Breaking the speed blocks from the left

fdtaillefer:
…blocks from the left

### DIFF
--- a/region/norfair/crocomire.json
+++ b/region/norfair/crocomire.json
@@ -815,7 +815,7 @@
                       "shinesparkFrames":35
                     }},
                     {"cost": [
-                      {"acidFrames": 130}
+                      {"acidFrames": 140}
                     ]}
                   ]}
                 ]}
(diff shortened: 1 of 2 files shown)


DIProgan:
Maybe destroying blocks with bluesuit should be its own tech?

region/norfair/crocomire.json
```diff
@@ -791,13 +803,25 @@
...
+                    "scope": "room"
+                  }},
+                  {"and":[
+                    "canSwim",
+                    {"canShineCharge": {
+                      "usedTiles": 31,
+                      "shinesparkFrames":35
+                    }},
+                    {"cost": [
+                      {"acidFrames": 130}
```
DIProgan: Are you sure the acid version can be done using full 31 length charge? Looks like it would need a slower/shorter charge to not bonk something and lose the blue suit.

fdtaillefer: I made sure to do it when getting my measurements

fdtaillefer: Doing it with a bit less strict of a setup, it looks more like acid 140 frames though. Guess I can change that.

DIProgan:
Ok to merge

Issue #80, opened 2019-01-14
Missing logic: Speed ball in Morph Ball Room


fdtaillefer:
With a sufficiently short charge, it's possible to go between the Power Bombs and left door in MorphBall room without canPassBombPassages, using a speedball.

fdtaillefer:
Resolved by #140

PR #140, opened 2019-02-23
Issue #80: Add speedball strats to Morph Ball Room

fdtaillefer:


region/brinstar/blue.json
```diff
@@ -128,8 +139,21 @@
...
+                    {"canShineCharge": {
+                      "usedTiles": 21,
+                      "shinesparkFrames": 0
+                    }},
+                    "canMockball"
+                  ]}
+                ]}
+              ],
+              "unlock": null,
+              "note": "FIXME The Power Bombs in the mockball strat are to clear the obstacle between this node and node 4, to open up room to charge the blue suit. FIXME It's 21 tiles, but with an open end. FIXME Only morph is required if the blocks have already been cleared."
```
DIProgan: Sounds like it's all about a future implementation of obstacles for the sake of saving 1 PB.

region/brinstar/blue.json
```diff
@@ -78,7 +78,18 @@
...
             {
               "id": 5,
-              "requires": [ "canPassBombPassages" ],
+              "requires": [
+                {"or":[
+                  "canPassBombPassages",
+                  {"and":[
+                    {"canShineCharge": {
+                      "usedTiles": 18,
+                      "shinesparkFrames": 0
```
DIProgan: We should probably default shinesparkFrames to 0 in schema.

fdtaillefer: Maybe but I don't know how to put logical elements in the schema so it's not in there.

Also, I'd just rather put 0 down manually because it's not a very common case, so writing it down underlines that yes, this is really what I intended.

region/brinstar/blue.json
fdtaillefer: It's not just about saving 1 PB. You could come in from the left with no bombs at all, grab the first item, and leave. But without tracking the bomb blocks properly, the logic can't represent it.

region/brinstar/blue.json
DIProgan: Do you want to add a commit to this with the 0's typed out? Like you said it's not that awfully many. Either that or just merge it.

fdtaillefer: I'd do that as its own thing. No reason why doing that everywhere should hold up a PR of two strats in one room.

Issue #34, opened 2018-12-11
Standardize requirements for event nodes


fdtaillefer:
Many of the event nodes in the model currently have their requirements in the link to the node. It would make more sense to have the requirements directly on the node, unless getting to an even from that node specifically add requirements (such as Botwoon from the right door).

It would be a good idea to move event requirements to the node's "unlock" property.

fdtaillefer:
Note: Events often lock passage between two nodes. If those two nodes can both go to the event, and the event can lead back to both of those nodes, that wall would be broken (from the perspective of a program navigating the nodes) without unlocking the event. So moving from the event node to any other node should require the flag that was yielded by the event.

Issue #33, opened 2018-12-11
Standardize "yields"


fdtaillefer:
"yields" is often used on links, when it makes a lot more sense to put it on a node (like it currently is on Kraid).

Go through all the places where "yields" is used in links and move it to the node if it makes more sense that way.

PR #147, opened 2019-02-24
#33 and #34: Finish standardizing events and `yields`

fdtaillefer:


region/lowernorfair/west.json
```diff
@@ -418,9 +418,17 @@
                     ]}
                   ]}
                 ]}
+              ]},
+              {"and": [
+                "heatProof",
+                {"or":[
+                  "Charge",
```
DIProgan: I promise you, you don't want to kill GT with only Charge even if you're heatproof.

fdtaillefer:
I think the GT kill can be revisited when we revisit boss requirements at large. I'd be inclined to go with this for now.

PR #148, opened 2019-02-24
Reset rework. Closes #121

### LINKED ISSUE #121: Some room traversals require a room reset
There are some links that may not be traversable without resetting the room by exiting and re-entering, in rooms where items can make it so a plausible path could do just that. In those cases, we need an indication that the player needs to exit and re-enter (so that the logic can tell if there's something problematic about doing that)

- Early Supers, from left to right, requires a reset at the left door
- Speed booster hall, in either direction, more or less requires a reset because of the crumble blocks. Otherwise, the heat frames requirements will be way off.
- Any other places?

DIProgan:
Adds several new runways and links that needs a "resetRoomAtNode" to work properly.

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -69,7 +69,7 @@
             "runways": [
               {
                 "length": 12,
-                "requires": null
+                "requires": [ {"resetRoomAtNode": [1, 2, 3, 4, 5, 6]} ]
               }
             ]
           }
@@ -84,7 +84,7 @@
             "runways": [
               {
                 "length": 12,
-                "requires": null
+                "requires": [ {"resetRoomAtNode": [1, 2, 3, 4, 5, 6]} ]
               }
             ]
           }
@@ -443,7 +443,8 @@
               "requires": [
                 {"or": [
                   "canDash",
-                  "canMockball"
+                  "canMockball",
+                  {"resetRoomAtNode": [1]}
                 ]}
               ],
               "unlock": null
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -443,9 +443,9 @@
               "requires": [
                 {"or": [
                   "canDash",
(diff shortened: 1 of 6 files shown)


region/brinstar/green.json
```diff
@@ -443,7 +443,8 @@
               "requires": [
                 {"or": [
                   "canDash",
-                  "canMockball"
+                  "canMockball",
+                  {"resetRoomAtNode": [1]}
```
fdtaillefer: This needs to be outside the `or`

region/brinstar/red.json
```diff
@@ -765,7 +765,7 @@
             "runways": [
               {
                 "length": 8,
-                "requires": null
+                "requires": [ {"resetRoomAtNode": [1, 2, 4, 5]} ]
```
fdtaillefer: I think 5 is a bit convoluted with the green gate. I'd leave that out.

region/norfair/east.json
```diff
@@ -1742,7 +1742,7 @@
                     "heatProof",
                     "canHeatRun"
                   ]},
-                  "reset",
+                  {"resetRoomAtNode": [1]},
```
fdtaillefer: Needs reset at opposite side.
Also, traversing from 1 to 2 requires a reset at 1.

region/norfair/east.json
```diff
@@ -1778,7 +1778,7 @@
                     "heatProof",
                     "canHeatRun"
                   ]},
-                  "reset",
+                  {"resetRoomAtNode": [2]},
```
fdtaillefer: Needs reset at opposite side.
Also, traversing from 2 to 1 requires a reset at 2.

region/brinstar/green.json
DIProgan: Ops

PR #154, opened 2019-02-26
Draygon grapple jump requires the means to destroy the turret. Fixes #146.

### LINKED ISSUE #146: Missing requirement for Draygon grapple escape
The Draygon grapple escape requires the ability to break the turret. This is not currently represented in the logic.

fdtaillefer:


DIProgan:
Merge it

PR #159, opened 2019-02-28
MidAir Springball jumps using walls

DIProgan:
New tech added and various fixes found in the process


region/maridia/inner-green.json
```diff
@@ -93,10 +93,15 @@
             {
               "id": 3,
               "requires": [
-                {"or":[
+                {"or": [
```
fdtaillefer: I think I chose to leave suitless out of this req because the only way to get to node 4 with neither Gravity nor suitless is to come in through the top door, and it seemed unreasonable to view that as a logical softlock unless you really can't get out.

PR #161, opened 2019-03-02
Update documentation for utility and the change in enemy attacks

fdtaillefer:
Plus a random nodeItem -> node fix


DIProgan:
Looks good to me.

PR #162, opened 2019-03-02
Fix PCJR strat that uses a springwall, and add a few springwalls

fdtaillefer:


region/brinstar/red.json
```diff
@@ -402,6 +402,12 @@
                           "canDash",
                           "canSpringBallJumpMidAir"
                         ]}
+                      ]},
+                      {"and":[
+                        "canSpringBallWalljumpMidAir",
+                        {"cost": [
+                          {"spikeHits": 6}
```
DIProgan: Would be more compressed if it shared spikehits with the IBJ

PR #168, opened 2019-03-07
openEnds in Crateria

DIProgan:


region/crateria/central.json
```diff
@@ -216,6 +222,12 @@
                 }
               ],
               "unlock": null
+            },
+            {
+              "id": 6,
```
fdtaillefer: Much like I mentioned to a similar change in Mike's PR, in most other cases the nodes that yield flags have a free link there (and back), and the requirements to get the flag are in an `unlock`. You think we could do that here?

DIProgan: Sure

PR #169, opened 2019-03-07
openEnds in Tourian with Schema

DIProgan:


region/wreckedship/main.json
```diff
@@ -590,17 +590,16 @@
           "id": 2,
           "name": "Bottom Door",
           "nodeType": "door",
-          "nodeSubType": "grey",
+          "nodeSubType": "blue",
+          "lock": [
```
fdtaillefer: This will probably trip up Mike's conversion help too, and the proposed format has evolved since. Can we take it out for now?

schema/m3-region.schema.json
```diff
@@ -221,6 +221,12 @@
                             "title": "Is it usable coming in?",
                             "default": true
                           },
+                          "openEnd": {
+                            "$id": "#/items/properties/nodes/items/properties/sparking/properties/runways/items/properties/openEnd",
+                            "type": "integer",
+                            "title": "Does the runway have an open end beside it (as opposed to a wall)?",
+                            "default": 1
```
fdtaillefer: I don't think this needs a default value. Same for the other one.

region/tourian/main.json
```diff
@@ -562,7 +563,13 @@
             "runways": [
               {
                 "length": 4,
-                "requires": null
+                "requires": null,
+                "openEnd": 0
+              },
+              {
+                "length": 8,
```
fdtaillefer: I think this one might need a reset to be reliable.

region/wreckedship/main.json
DIProgan: I'm not quite sure how that found its way into the PR. Never meant to be pushed.

schema/m3-region.schema.json
DIProgan: This default saves several hundred lines of code.

fdtaillefer: I'd argue having default values hides several hundred lines of real data, that I assume developers then have to fetch in some other schema file that they probably don't care about and probably don't want to bother with.

PR #172, opened 2019-03-07
openEnds in Maridia

DIProgan:


region/maridia/inner-pink.json
```diff
@@ -15,9 +15,9 @@
           "sparking": {
             "runways": [
               {
-                "length": 2,
+                "length": 13,
```
fdtaillefer: This one would need a reset, right?

DIProgan: Nah it's a respawn block

region/maridia/inner-yellow.json
```diff
@@ -18,14 +18,32 @@
                 "length": 5,
                 "requires": null
               }
+            ],
+            "canLeaveCharged": [
```
fdtaillefer: Gonna have to explain that one, I can't figure it out

region/maridia/inner-pink.json
fdtaillefer: Golly, you're right

Issue #57, opened 2018-12-24
Add a lock property to doors


fdtaillefer:
Some doors are only locked when some conditions are met, and then unlocked when further conditions are met.

The model currently represents those strictly as unlock requirements, where the requirements to unlock need to be met OR the requirements to lock need to NOT be met.

It might be an improvement to have separate requirements to lock and unlock the door so the model can adequately represent the situation. The lock property should be optional, and all doors with no lock property should simply be assumed to always require their unlock property to be fulfilled.

Rooms that have locking requirements that I can think of:
- Bomb Torizo Room. Its door locks only if you have Morph Bombs.
- Attic. Its bottom door locks only if you have defeated Phantoon.
- Bowling Alley: The Mid-left door locks only if you have defeated Phantoon.

Many other doors are also locked during the escape sequence and cannot be unlocked from that point on, we might be able to represent those as well. Bomb Torizo Room might be more tricky since it has two separate sets of locking and unlocking requirements (locks with bomb, unlocks by killing Bomb Torizo, locks again with escape, and unlocks by saving the animals). ~~However, we probably don't need to model that second lock/unlock.~~ We will model both locks. Anyway, it turns out all colored doors that get locked during the escape also need two `lock`s.

While working on this, also add a `BeatSuperMetroid` flag that is attained from an additional node next to the ship, and the `SaveAnimals` flag that can be attained in Bomb Torizo Room during escape.

DIProgan:
```json
"lock": [
            {"and": [
              "Bombs",
              {"not": [
                "DefeatBombTorizo"                  
              ]}
            ]},
            {"and": [
              "ZebesSetAblaze",
              {"not": [
                "SavedTheAnimals"                  
              ]}
            ]}
          ],
```

fdtaillefer:
That doesn't solve the issue that `unlock` is once and for all, I don't think.

fdtaillefer:
The format we settled on is:
```json
  "locks": [
    {
      "lock": [
        "Bombs"
      ],
      "unlock": [
        "DefeatBombTorizo"
      ]
    },
    {
      "lock": [
        "zebesSetAblaze"
      ],
      "unlock": [
        "animalsSaved"
      ]
    }
  ],
```
This way, a node can have several independant lock conditions, each with an associated unlock condition.

fdtaillefer:
This has been finalized with #205.

PR #173, opened 2019-03-09
Locks

miketrethewey:


region/crateria/central.json
```diff
@@ -216,6 +234,12 @@
                 }
               ],
               "unlock": null
+            },
+            {
+              "id": 6,
```
fdtaillefer: My reaction would have been to make the node link always accessible (and to give it a free link back to 5 as well) but to have `ZebesSetAblaze` in as an `unlock`. I don't know if that's better, but it's more consistent with how the rest of the flags are handled.

region/crateria/central.json
```diff
@@ -1122,13 +1196,23 @@
...
-              ]},
-              "DefeatBombTorizo"
-            ]}
+              ],
+              "unlock": [
+                "DefeatBombTorizo"
+              ]
+            },
+            {
+              "lock": [
```
fdtaillefer: Since the door's second lock is here, did you intend to also add an animals node also locked by `ZebesSetAblaze` that yields `AnimalsSaved`?

fdtaillefer:
The things I've mentioned could also just be rolled into #57 while we're also adding the other `ZebesSetAblaze` locks and the ones in Wrecked Ship.

PR #181, opened 2019-03-10
Obstacles: Crocomire

fdtaillefer:


fdtaillefer:
I thought about it and decided against it for now. I guess it would involve merging nodes 2 and 4.
The Mellas will be done after. I'll try and limit the scope of obstacles to obstacles.

PR #182, opened 2019-03-11
East Norfair Obstacles

fdtaillefer:


region/norfair/east.json
```diff
@@ -773,6 +773,13 @@
           "utility": [ "farming" ]
         }
       ],
+      "obstacles": [
```
miketrethewey: Why are `ID` and `Name` capitalized?

fdtaillefer: Cause I messed up :P

region/norfair/east.json
```diff
@@ -179,31 +186,32 @@
                   "heatProof",
                   "canHeatRun"
                 ]},
-                {"or":[
-                  {"and":[
+                {"cost": [
+                  {"heatFrames": 100}
+                ]},
+                {"obstacle":{
+                  "id": "A",
```
DIProgan: I don't see in what situation the obstacle is useful here?

region/norfair/east.json
DIProgan: Is this obstacle for the future case of wanting to use the farming and then go back?

region/norfair/east.json
fdtaillefer: The goal was to put a hint in there that these requirements are centered around killing enemies, but maybe I'm jumping the gun. This can be done later if deemed necessary. I'll take it out.

region/norfair/east.json
fdtaillefer: Yes. Same reason I did it in the Kronic Boost Room, where the Violas are farmable for PBs.

By opposition, I didn't do it in Ice Beam Gate Room to get to the small Dessgeegas (who are also farmable for PBs) because I figured it's only a difference of one PB. But thinking about it again, maybe it was worth doing?

DIProgan: Nah

Issue #38, opened 2018-12-12
Reduce use of helpers


fdtaillefer:
For performance purposes, it appears helpers with only one requirement should be eliminated and replaced with the item itself.

fdtaillefer:
The last remaining helper of this kind was `canSwim`, and it was removed by #241.

PR #183, opened 2019-03-11
Remaining canDash into SpeedBooster

DIProgan:
As part of #38

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -13,10 +13,7 @@
       "Morph"
     ]
   },
-  {
-    "name": "canDash",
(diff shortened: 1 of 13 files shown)


fdtaillefer:
Are there plans to remove `canSwim` as well down the road?

DIProgan:
> Are there plans to remove `canSwim` as well down the road?

I simply noted how the usage of canDash shrunk so took it upon myself to make it uniform. I can see canSwim being replaced the same way but haven't seen the same tendency there yet.

PR #184, opened 2019-03-14
Treasure Hunt: canComeInCharged & canLeaveChargeds

DIProgan:
A bunch of extra ones I found along with a couple errors as a bonus.

### DIFF
--- a/region/brinstar/kraid.json
+++ b/region/brinstar/kraid.json
@@ -270,6 +270,7 @@
               {
                 "usedTiles": 27,
                 "framesRemaining": 120,
+                "requires": [ {"resetRoomAtNode": [1, 3]} ],
                 "openEnd": 2
               }
             ]
@@ -281,12 +282,21 @@
           "name": "Lower Right Door",
           "nodeType": "door",
           "nodeSubType": "blue",
-          "sparking": {
+          "sparking": {            
             "runways": [
               {
                 "length": 8,
                 "requires": null,
                 "openEnd": 0
+              }
+            ],
+            "canLeaveCharged": [
+              {
+                "usedTiles": 21,
+                "framesRemaining": 0,
+                "shinesparkFrames": 20,
+                "requires": [ {"resetRoomAtNode": [1, 3]} ],
+                "openEnd": 1
               }
             ]
           },
--- a/region/brinstar/kraid.json
+++ b/region/brinstar/kraid.json
@@ -320,7 +320,7 @@
                 "usedTiles": 31,
                 "framesRemaining": 0,
                 "shinesparkFrames": 15,
-                "requires": [ {"resetRoomAtNode": [1, 3]} , "Morph" ],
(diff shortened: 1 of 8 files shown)


region/brinstar/kraid.json
```diff
@@ -314,6 +314,15 @@
                 "requires": null,
                 "openEnd": 1
               }
+            ],
+            "canLeaveCharged": [
+              {
+                "usedTiles": 31,
+                "framesRemaining": 0,
+                "shinesparkFrames": 15,
+                "requires": [ {"resetRoomAtNode": [1, 3]} , "Morph" ],
```
fdtaillefer: That must require something to break the bomb block as well, surely? Basically access from 1 to 3. Bit annoying that we're back to duplicating some logic, but at least it's not widespread

region/maridia/inner-yellow.json
```diff
@@ -149,6 +157,15 @@
...
                 "openEnd": 1
               }
+            ],
+            "canLeaveCharged": [
+              {
+                "usedTiles": 22,
+                "framesRemaining": 0,
+                "shinesparkFrames": 30,
+                "openEnd": 2,
+                "requires": [ "canSwim", "Morph" ]
```
fdtaillefer: Why gravity? If you time it right the water won't matter at the morph gap, and I can't find a way that goes deeper in the water and still has time to get it off, even with Gravity.

region/maridia/inner-green.json
```diff
@@ -143,6 +143,16 @@
               "id": 4,
               "requires": [
                 {"or": [
+                  {"canComeInCharged": {
```
fdtaillefer: Pretty cool that it gets you directly to the top ledge. Should these sparks require either gravity or suitlessMaridia?

region/brinstar/kraid.json
DIProgan: Indeed, ops.

region/maridia/inner-yellow.json
DIProgan: Because it's such a very tight maneuver even without taking the water into account. I can see large amounts of time being spent here if you need to time for water as well.

region/maridia/inner-green.json
DIProgan: I opted no since it's such a miniscule amount of time that's spent doing Suitless things.

region/maridia/inner-yellow.json
fdtaillefer: It's roughly a 4-second cycle. You really only have to know when to start. and then it's the same execution.
But yeah, I'm a bit concerned that many of those strats require tight execution and aren't locked behind any kind of tech.

DIProgan: At least it's locked behind managing 22 tile charges.

fdtaillefer: Beats having nothing, but the ability to do a 2-tap isn't closely related to executing that spark.

PR #185, opened 2019-03-16
West Lower Norfair Obstacles

fdtaillefer:


region/lowernorfair/west.json
```diff
@@ -782,11 +806,68 @@
                   "canHeatRun"
                 ]},
                 {"or": [
+                  {"obstacle":{
+                    "id": "A",
+                    "requires": [
+                      "canIBJ",
```
DIProgan: Since going 3 -> 2 breaks obstacle 'A', going back up using what used to be this IBJ obstacle req is a free null req? Am I understanding obstacles wrong?

DIProgan: Since going 3 -> 2 breaks obstacle 'A', going back up using what used to be this IBJ obstacle req is a free null req? Am I understanding obstacles wrong?

region/lowernorfair/west.json
```diff
@@ -782,11 +806,68 @@
...
+                            ]}
+                          ]}    
+                        ]}
+                      ]
+                    }},
+                    {"or": [
+                      "canSpringBallWalljumpMidAir",
+                      {"and": [
+                        "HiJump",
+                        "SpeedBooster"
```
DIProgan: If Obstacle 'B' is broken this needs a reset

region/lowernorfair/west.json
fdtaillefer: No, that's me being stupid. Wow, that would've been bad.

region/lowernorfair/west.json
fdtaillefer: Nah, the setup here is inside the doorway of node 2 so not reliant on obstacle B.

fdtaillefer:
Fixes made

DIProgan:
Checks out

PR #186, opened 2019-03-17
East Lower Norfair Obstacles

fdtaillefer:


region/lowernorfair/east.json
```diff
@@ -1488,16 +1500,28 @@
...
+                        ]},
+                        {"obstacle":{
+                          "id": "B",
+                          "requires": null
+                        }}
+                      ]
+                    }}
                   ]},
                   { "and": [
                     "canXRayClimb",
```
DIProgan: Does the xray avoid enemyDamage?

DIProgan:
Looked at everything and it looks great. Just need to add Mickey Mouse ice clip and Metal Pirate fixes later on then.

PR #191, opened 2019-03-23
Green Brinstar Obstacles

fdtaillefer:


region/brinstar/green.json
```diff
@@ -739,8 +767,30 @@
...
+                            }
+                          ]}
+                        ]}
+                      ]}
+                    ]}
+                  ]
+                }}
+              ],
+              "unlock": null,
+              "note": "A speedball strat can be doable here, depending on adjacent runway, but between the grey door and the number of runway tiles in this room that can't be used for that, it might be better to omit it."
```
DIProgan: Legit hell to perform. I'd rather omit it.

fdtaillefer: Well, with a long runway on the other side it's probably quite trivial. If not for the grey door, there'd be some pretty easy setups for it. That we're not able to represent right now.

DIProgan: I forgot you can go back to kill the sidehopper after breaking the block. 180 on that one

PR #192, opened 2019-03-23
Kraid Obstacles

fdtaillefer:


region/brinstar/kraid.json
```diff
@@ -373,8 +405,34 @@
...
+                          "id": "A",
+                          "requires": null
+                        }}
+                      ]}
+                    ]}
+                  ]
+                }}
+              ],
+              "unlock": null,
+              "note": "FIXME Possible speedball here depending on adjacent room?"
```
DIProgan: Not without already balling I think.

PR #195, opened 2019-03-26
Central Crateria Obstacles

fdtaillefer:


region/crateria/central.json
```diff
@@ -661,7 +661,12 @@
               },
               {
                 "length": 28,
-                "requires": [ "canDestroyBombWalls" ],
+                "requires": [
+                  {"obstacle":{
+                    "id": "A",
```
DIProgan: I can't wrap my brain around what this obstacle allows for. If it's access to the node 6(shaft) Pirates, why don't the other doors with their respective blocks got the same?

fdtaillefer: The obstacle between the main part of the room and the bottom left door is the same obstacle that stops the 28-tile runway from being usable. The requirements got complex but they're always the same regardless so I put them on the obstacle and saved the duplication.

It might come in handy if we decide to add some spark strats in here too, they'll use the same obstacle.

PR #198, opened 2019-03-28
Wrecked Ship Obstacles

fdtaillefer:


region/wreckedship/main.json
```diff
@@ -258,16 +294,34 @@
...
-                    "Wave"
-                  ]}
-                ]}
+                "Morph",
+                {"obstacle": {
+                  "id": "B",
+                  "requires": [
+                    {"or":[
+                      "Bombs",
+                      "Wave",
```
DIProgan: This also opens up with Spazer

fdtaillefer: Let's see Plasma try to do that!

DIProgan:
Good to go

PR #205, opened 2019-04-01
Zebes Set Ablaze locks & friends

DIProgan:
Moves Zebes Awake into Pit Room and adds the escape sequence locks according to new lock format.  Bowling Alley and Attic got oil in their locks too.


region/brinstar/blue.json
```diff
@@ -11,7 +11,7 @@
           "id": 1,
           "name": "Far Left Door",
           "nodeType": "door",
-          "nodeSubType": "grey",
+          "nodeSubType": "blue",
```
fdtaillefer: That brings up a question. Is grey a legit door color? I'm not sure which one this should be.

region/crateria/central.json
```diff
@@ -22,7 +22,11 @@
               }
             ]
           },
-          "locks": null,
+          "locks": [
+            {
+              "lock": [ "ZebesSetAblase" ]
```
fdtaillefer: It's `ZebesSetAblaze`. The error seems repeated throughout.

region/crateria/central.json
```diff
@@ -859,7 +895,7 @@
           },
           "locks": [
             {
-              "unlock": [ "Kill All Enemies" ]
+              "unlock": [ "ZebesAwake", "Kill All Enemies" ]
```
fdtaillefer: `ZebesAwake` goes in this `unlock`'s matching `lock`. Same for the other door.

region/crateria/central.json
```diff
@@ -890,45 +927,67 @@
           "nodeType": "item",
           "nodeSubType": "visible",
           "nodeItem": "Missile",
-          "locks": null
+          "locks": [
+            {
+              "unlock": [ "ZebesAwake" ]
```
fdtaillefer: Interesting. Didn't know.

region/wreckedship/main.json
```diff
@@ -831,17 +831,8 @@
...
-                    "DefeatPhantoon"
-                  ]},
-                  {"and": [
-                    "DefeatPhantoon",
-                    "Kill All Enemies"
-                  ]}
-                ]}
-              ]
+              "lock": [ "DefeatPhantoon" ],
+              "unlock": [ "DefeatPhantoon", "Kill All Enemies" ]
```
fdtaillefer: I think `DefeatPhantoon` is unnecessary here. It's implied that the door has been locked before being unlocked.

region/brinstar/blue.json
DIProgan: Seems the colors do nothing anyway so *shrugs*

region/crateria/central.json
DIProgan: The left door is locked both before ZebesAwake and after when you need to Kill all enemies to open it. The right door only locks after ZebesAwake.

fdtaillefer: You're right. That makes `ZebesAwake` a bit redundant since you can't kill the enemies before they're there, but it doesn't matter.

PR #215, opened 2019-04-10
Metroid Rooms logic

DIProgan:
Each room of Metroids is killable with only 3 Power Bombs but they need a great deal of tricky avoidance so I created canMetroidAvoid either for going in reverse or for killing Metroids with Power Bombs.

Fitting the Missile & Super ammo into this goes together with the rest of the issues surrounding enemies so leaving it alone for now.

### DIFF
--- a/region/tourian/main.json
+++ b/region/tourian/main.json
@@ -140,13 +140,37 @@
               {
                 "usedTiles": 31,
                 "framesRemaining": 120,
-                "openEnd": 2
+                "openEnd": 1
               }
             ]
           },
+          "yields": [ "killedMetroidRoom1" ],
           "locks": [
             {
-              "unlock": [ "Kill All Enemies" ]
+              "unlock": [
+                {"or": [
+                  {"and":[
+                    "Ice",
+                    {"or":[
+                      "Missile",
+                      "Super"
+                    ]}
+                  ]},
+                  {"and":[
+                    "Morph",
+                    "PowerBomb",
+                    "canMetroidAvoid",
+                    {"cost": [
+                      {"ammo": [
+                        {
+                          "type": "PowerBomb",
+                          "count": 3
+                        }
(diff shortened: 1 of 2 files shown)


fdtaillefer:
I think taking damage from the metroids should be a viable alternative to `canMetroidAvoid` without needing Ice. It's pretty much what everyone does.

region/tourian/main.json
```diff
@@ -193,7 +223,14 @@
           "to": [
             {
               "id": 2,
-              "requires": null,
+              "requires": [
+                {"or": [
+                  "SpaceJump",
+                  {"cost": [
+                    {"acidFrames": 30}
```
fdtaillefer: In both directions, there's a way to avoid those frames with just Speed Booster as well,

region/tourian/main.json
```diff
@@ -203,8 +240,19 @@
...
-              "requires": null,
-              "unlock": [ "canKillMetroids" ]
+              "requires": [
+                {"or": [
+                  "SpaceJump",
+                  {"cost": [
+                    {"acidFrames": 60}
+                  ]}
+                ]},
+                {"or": [
```
fdtaillefer: Does this not make the room entirely out of logic with PBs? You can unlock the door but you can't reach it.

fdtaillefer:
I'm not sure what the solution is, but here is a list of things that we want to represent:

- Unlocking the door requires killing the Metroids
- Killing the Metroids can be done with freeze + missile/super, or with PBs either by avoiding Metroids or taking damage.
- Traveling through the room in either direction can be done by freezing or avoiding metroids, or by taking damage. But it's also free if the Metroids have previously been killed.
- We have to make sure the combination of reaching the door + unlocking it by using PBs and tanking the damage doesn't end up requiring more energy than in reality.

PR #216, opened 2019-04-11
Rework `resetRoomAtNode` into `resetRoom`, and overhaul logic at the bottom of Mickey Mouse Room

fdtaillefer:


region/lowernorfair/east.json
```diff
@@ -211,7 +214,10 @@
               {
                 "length": 12,
                 "requires": [
-                  {"resetRoomAtNode": [1, 4]},
+                  {"resetRoom":{
+                    "nodes": [1, 2, 3, 4],
```
DIProgan: I'm guessing all 4 is here in preparation for an eventual clip

region/lowernorfair/east.json
```diff
@@ -2702,51 +2708,136 @@
...
+                    {"enemyDamage": [{
+                      "enemy": "Dessgeega",
+                      "type": "contact",
+                      "hits": 1
+                    }]}
+                  ]}
+                ]},
+                {"and":[
+                  {"obstacle":{
+                    "id": "C",
```
DIProgan: Is this thing with requiring obstacle C & E multiple times in the same req really going to work without triggering the contact damage options?

region/lowernorfair/east.json
```diff
@@ -2702,51 +2708,136 @@
...
                   ]
-                }},
-                {"roomReset":{
-                  "nodes": [1],
-                  "obstaclesToAvoid": ["E"]
                 }}
               ],
               "unlock": null,
-              "note": "The clip is done by getting the far right Multiviola in the bottom hallway to climb all the way up. It requires several trips through the hallway, hence killing the Dessgeegas"
+              "note": "Options are to Speed Booster dash or spark through, tank the damage, kill the Dessgeegas and Multiviolas, or go slower to spare the Multiviolas"
```
DIProgan: The Plasma option is listed as slower without killing Violas while the Super option is slower not killing them. Might make sense I guess. Also there's no ScrewAttack option for not killing violas which might also make sense. All of the sense is just hidden very deep down in this room.

region/lowernorfair/east.json
```diff
@@ -3027,6 +3043,270 @@
...
+                  "canHeatRun"
+                ]},
+                {"or":[
+                  {"and":[
+                    {"obstacle":{
+                      "id": "F",
+                      "requires": ["never"]
+                    }},
+                    {"canShineCharge": {
+                      "usedTiles": 22,
```
DIProgan: Would you like yet another complication? ;) If you contact damage on the node 6 Dessgeegas there's a ~50% chance you canShineCharge with 31 tiles killing the last 2-4 enemies :P

region/lowernorfair/east.json
fdtaillefer: Yep, I guess we could remove it if we conclude the clip can't be done but I don't expect that.

region/lowernorfair/east.json
fdtaillefer: It's not going to work, no. The Dessgeega damage isn't in the right place. An `or` is missing. And there seems to be a pointless `and` in there somewhere too.
It's supposed to be:
- heatproof or canHeatRun
- Take damage from Multiviola if it wasn't previously killed
- Take an additional hit from a Dessgeega, OR kill the Dessgeegas. OR kill both the Dessgeegas and Multiviolas

region/lowernorfair/east.json
fdtaillefer: The Supers thing sounds wrong. Pretty sure the time should be the same. I got tripped up counting damage due to sometimes forgetting about the 50 damage from the Multiviola. I'll have to recount that.

Not sure if you mistyped, but there is a Screw Attack option for not killing the Multiviolas. I left out the option for killing them because you can't really catch up to them before reaching node 7.

Plasma is much slower without killing the Multiviolas because the beams pierce and you have to wait till it despawns.

region/lowernorfair/east.json
fdtaillefer: Let's not and say we did?

fdtaillefer: I guess it's relevant as it's a way to get through the room with just PBs and Speed Booster.
I'll make it a direct link from 6 to 1.

DIProgan:
Yeah I think I'm done reviewing this. It's really hard to get an overview so let's just see how it goes.

Issue #199, opened 2019-03-31
Aqueduct top door logic


fdtaillefer:
There are currently requirements to get into the top door at Aqueduct, from being right below it.
Those requirements should be waived, because looking away for long enough will bring a snail that can be used to get into the door.

PR #219, opened 2019-04-13
Fix for issue #199: Aqueduct top door access

fdtaillefer:


region/maridia/inner-pink.json
```diff
@@ -585,14 +585,9 @@
...
-                {"or":[
-                  "canSwim",
-                  "HiJump",
-                  "canSpringBallJumpMidAir"
-                ]}
-              ],
-              "unlock": null
+              "requires": null,
+              "unlock": null,
+              "note": "A jump assist would be needed here if there were no snails to climb on"
```
DIProgan: Shouldn't this be hidden behind some tech?

fdtaillefer: I wasn't sure. It's too simple to be a tech. The only reason to have one is for Enemizer way down the line. I decided it could be done later when needed, but not 100% sure. How do you see it?

DIProgan:
Good to go

PR #221, opened 2019-04-13
Add two more strats to Boulder Room. Resolves #208.

### LINKED ISSUE #208: Missing logic: Boulder Room
Boulder Room is doable right to left with no equipment at all and without an adjacent runway, with a walljump off the bottom wall while the water is low, then another one off the bridge.

We should probably lock that behind `canTrickyWalljump`?

Edit: With the door closed, it's also possible to SpringWall there.

fdtaillefer:
Strats are a springwall, and a tricky walljump that makes the room doable naked.

### DIFF
--- a/region/brinstar/blue.json
+++ b/region/brinstar/blue.json
@@ -790,10 +790,13 @@
                   {"adjacentRunway": {
                     "fromNode": 2,
                     "usedTiles": 1
-                  }}
+                  }},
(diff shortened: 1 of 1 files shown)


fdtaillefer:
> 
> 
> I failed at both of them but that's just me :(

https://youtu.be/6rFYT6WqDZ0
and
https://youtu.be/lI0e4JmPRXc

Issue #209, opened 2019-04-07
Missing logic: ceiling clips with the X-Ray standup


fdtaillefer:
We have missing strats that involve using the X-Ray scope to force a standup in a 2-tile-high tunnel, then jumping and quickly pressing down then forward, to enable a ceiling clip.

We'll need an accompanying tech. This strat can at least be used on the way up Red Kihunter Shaft to save a PB if no Morph Bombs or Spring Ball, and when doing Hi-Jump E-Tank Room left to right.

PR #222, opened 2019-04-13
Fix for #209. XRay standup clips as well as a new junction in HiJump Etank Room.

fdtaillefer:


region/lowernorfair/east.json
```diff
@@ -1397,7 +1397,11 @@
                           "count": 1
                         }]}
                       ]},
-                      "SpringBall"
+                      "SpringBall",
+                      {"and":[
+                        "canXRayStandUp",
+                        "canCeilingClip"
```
DIProgan: How does this even look and does it really not add any heatframes compared to Spring Ball?

DIProgan:
Also good to go

region/lowernorfair/east.json
DIProgan: For future archeology https://www.youtube.com/watch?v=qHOT9zY0w2Q

PR #224, opened 2019-04-13
Zebetites

DIProgan:
Adds handling of Zebetites so that it's possible to exit the room after each one or use the Zebetite Skip (not both).
List the new Zebetites flags in items.json together with the forgotten flags for the Metroid Rooms.


region/tourian/main.json
```diff
@@ -1418,18 +1461,33 @@
...
-              "requires": null,
-              "unlock": [
-                {"or":[
-                  "Missile",
-                  "Super"
-                ]}
-              ],
-              "note": "Ammo for destroying MB1's glass?"
+              "requires": [ "Missile", "Super" ],
+              "unlock": null,
```
fdtaillefer: This was an unlock because the glass doesn't respawn. So, like with the Zebetites, you could destroy that and go refill. I guess node 8 needs to be an event as well?

tech.json
```diff
@@ -486,5 +486,19 @@
     "name": "canQuickCrumbleEscape",
     "requires": [],
     "note":"The combination of a crumble quick drop, and landing on a lower surface and jumping back over the crumble block before it re-forms."
+  },
+  {
+    "name": "canZebetitesSkip",
+    "requires": [ 
+      "Ice",
```
fdtaillefer: I would change that to `canUseFrozenEnemies`. However, there is also a spark-based Zebetite skip which depends on the runway in the adjacent room. I don't believe we can put the requirements for that in a tech.

IMO the ice zeb skip and speed zeb skip should be two separate techs, and the requirements for the speed one will have to be put directly on the room. We may want to put the requirements for the ice one in the room as well for consistency...?

region/tourian/main.json
DIProgan: I figure we fix all the MB stuff in a later PR anyway

fdtaillefer: This is represented like the Zebetites though. If we make them diverge now, we may get confused later.

PR #225, opened 2019-04-13
Additional strats in Climb to save a PB. Resolves #188.

### LINKED ISSUE #188: Missing logic in climb
The bomb blocks to access the top right door and mid right door are missing some strats:

- A Behemoth spark can break the bomb blocks at the top, and also at the bottom. This requires a way to break the bomb blocks at the bottom left, or to comeInCharged.
- A Moonfall with Screw Attack can break the blocks at the bottom if holding right and clipping through that platform.

fdtaillefer:


region/crateria/central.json
```diff
@@ -965,13 +965,51 @@
...
+                    }},
+                    {"canShineCharge": {
+                      "usedTiles": 28,
+                      "shinesparkFrames": 147,
+                      "openEnd": 0
+                    }}
+                  ]},
+                  {"and":[
+                    "ScrewAttack",
+                    "canMoonfall"
```
DIProgan: I think you forgot to add Moonfall in tech

fdtaillefer:
> 
> 
> Took me a second to realize there was a global obstacle but it's all good.

Yeah I don't like putting global reqs on an obstacle because they vary so often depending on context but it saves a lot of trouble here.

PR #229, opened 2019-04-16
Drops extended

DIProgan:
dropRequires - What extra feat is required to obtain the drop of an enemy such as dipping into lava or taking a spikehit. Many enemies are placed where Samus can't walk but their drops can be reached anyway.

amountOfDrops - What amount of drops an enemy leaves behind

### DIFF
--- a/region/brinstar/red.json
+++ b/region/brinstar/red.json
@@ -387,9 +387,14 @@
         },
         {
           "name": "Yapping Maw",
-          "quantity": 3,
+          "quantity": 2,
           "betweenNodes": [1,2],
-          "dropRequires": "spikehit"
+          "dropRequires": [ "never" ]
+        },
+        {
+          "name": "Yapping Maw",
+          "quantity": 1,
+          "betweenNodes": [1,2]
         },
         {
           "name": "Fireflea",
--- a/region/brinstar/red.json
+++ b/region/brinstar/red.json
@@ -388,7 +388,8 @@
         {
           "name": "Yapping Maw",
           "quantity": 3,
-          "betweenNodes": [1,2]
+          "betweenNodes": [1,2],
+          "dropRequires": "spikehit"
         },
         {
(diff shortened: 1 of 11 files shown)


fdtaillefer:
Is there value to making `dropRequires` different from all the regular `requires` properties? Looks like devs would have to put additional effort to ultimately get less data.

region/norfair/east.json
```diff
@@ -3276,7 +3280,9 @@
         {
           "name": "Yapping Maw",
           "quantity": 3,
-          "betweenNodes": [ 1, 2 ]
+          "betweenNodes": [ 1, 2 ],
+          "dropRequires": "never",
+          "note": "Possible but requires multiple spikehits in most configurations so rarely worth it"
```
fdtaillefer: The drops are obtainable with Grapple, the issue is actually killing them. Not like farming Yapping Maws would ever be viable anyway.

DIProgan:
I'll remake dropRequires into more typical req data as good as I can. Several cases of heatframes going into it.

fdtaillefer:
> 
> 
> I'll remake dropRequires into more typical req data as good as I can. There'll be some heatframes and crap too.

One of the important things missing is all those places where grapple works. I wonder if we could just roll with a rule that forced damage to reach drops is just not supported though. Feels excessively complicated to start juggling that you have to go to this room and trade X health for X Supers or something.

region/brinstar/red.json
```diff
@@ -387,9 +387,14 @@
...
         {
           "name": "Yapping Maw",
-          "quantity": 3,
+          "quantity": 2,
           "betweenNodes": [1,2],
-          "dropRequires": "spikehit"
+          "dropRequires": [ "never" ]
+        },
+        {
+          "name": "Yapping Maw",
```
fdtaillefer: Nice. Had to think about it a few seconds to figure it out.

PR #232, opened 2019-04-20
Special grey door requirements: Morph Ball Room

fdtaillefer:


region/brinstar/blue.json
```diff
@@ -135,38 +131,84 @@
...
-                {"obstacle":{
-                  "id": "A",
-                  "requires": [
-                    {"or":[
-                      "Bombs",
-                      {"and":[
-                        "PowerBomb",
+                {"or":[
+                  {"cost":[
+                    {"cost":[
```
DIProgan: The double cost lines

region/brinstar/blue.json
```diff
@@ -135,38 +131,84 @@
...
+                            "ScrewAttack",
+                            "Plasma"
+                          ]}
+                        ]},
+                        {"and":[
+                          {"cost":[
+                            {"enemyDamage": {
+                              "enemy": "Sidehopper",
+                              "type": "contact",
+                              "hits": 2
```
DIProgan: I think it's perfectly fair to expect getting hit only once when entering the room here and then morph down into the corner below while the PB is going off.

region/brinstar/blue.json
```diff
@@ -135,38 +131,84 @@
...
                         {"cost":[
-                          {"ammo": [
-                            {
-                              "type": "PowerBomb",
-                              "count": 1
-                            }
+                          {"enemyDamage": {
+                            "enemy": "Sidehopper",
+                            "type": "contact",
+                            "hits": 8
```
DIProgan: This one is tough to balance. I'm unsure about it but 8 does seem a bit excessive. https://youtu.be/7l-mDFOryYc

region/brinstar/blue.json
fdtaillefer: Dammit

region/brinstar/blue.json
fdtaillefer: And you can do the ol' morph ball juke with the other guy. Sold.

region/brinstar/blue.json
fdtaillefer: Yeah it's intended as a "not-quite-worst-case-but-things-didn't-go-well" scenario.

fdtaillefer: I don't know to what extent it matters, though. This can literally never be logically required.

fdtaillefer: How about 6 instead?

DIProgan: Sure

DIProgan:
Everything looks dandy

PR #235, opened 2019-04-21
Special grey door requirements: Beta Power Bomb Room

fdtaillefer:


fdtaillefer:
> 
> 
> I think you said there was no need to req Morph when using PowerBomb in enemyKill.

That Morph is its own option in the big `or`. With Morph you can move the Hoppers around without getting hit and kill them as convenient, without taking a hit.

The PBs are defined separately because they are combined with breaking obstacle B.

DIProgan:
Good to go

PR #238, opened 2019-04-22
enemyKill

DIProgan:
Bosses is not included yet.
Changes to Green Pirate Shaft & grey door Baby Kraid Room


region/brinstar/red.json
```diff
@@ -329,7 +333,11 @@
                   "canTrickyWalljump",
                   {"and": [
                     "SpaceJump",
-                    "canKillRippers"
+                    {"enemyKill":{
+                      "enemies": [
+                        [ "Ripper","Ripper","Ripper" ]
```
fdtaillefer: I think there's 4 Rippers here too? They're just split across two nodes in (perhaps overzealous) anticipation for walljump things.

region/crateria/west.json
```diff
@@ -236,7 +236,35 @@
           "to": [
             {
               "id": 7,
-              "requires": null,
+              "requires": [
```
fdtaillefer: Good catch. That's a 60 damage that could logically make or break a short hellrun early in an entrando seed.

region/lowernorfair/east.json
```diff
@@ -222,16 +222,29 @@
...
                     {"and": [
                       "heatProof",
-                      "canKillYellowSpacePirates"
+                      {"enemyKill":{
+                        "enemies": [
+                          [
+                            "Yellow Space Pirate (standing)"
+                          ]
+                        ],
+                        "excludedWeapons": [ "Bombs", "PowerBomb" ]
```
fdtaillefer: This is the runway where PowerBombPeriphery would work. It allows you to kill them with 5 PBs without breaking the floor.

However, it is a situational weapon which won't be evaluated automatically. I think maybe we need another property called `includeWeapons` or something, which we can use to add situational weapons to a specific `enemyKill` without having to define a lite of `explicitWeapons`.

region/lowernorfair/east.json
```diff
@@ -842,7 +855,14 @@
                     {"or": [
                       {"and": [
                         "heatProof",
-                        "canKillYellowSpacePirates"
+                        {"enemyKill":{
+                          "enemies": [
+                            [
+                              "Yellow Space Pirate (wall)", "Yellow Space Pirate (wall)", "Yellow Space Pirate (wall)"
```
fdtaillefer: It's ridiculously hard to catch all 3 of them for 3 double-hit PBs without taking damage. Maybe we should split them into 2 groups?

region/lowernorfair/east.json
```diff
@@ -1464,7 +1513,11 @@
                   {"and": [
                     "heatProof",
                     {"or": [
-                      "canKillRedKiHunters",
+                      {"enemyKill":{
```
fdtaillefer: Did you figure that juking one of them should be pretty doable?

region/lowernorfair/east.json
```diff
@@ -1640,6 +1693,16 @@
           "id": "B",
           "name": "Bomb Blocks of Shame",
           "obstacleType": "inanimate"
+        },
+        {
+          "id": "C",
```
fdtaillefer: This is a complex room. I think we'd better update the image with the new obstacles.

region/lowernorfair/east.json
```diff
@@ -1667,7 +1730,7 @@
                 ]},
                 {"or": [
                   {"and": [
-                    { "cost": [{ "heatFrames": 680 }]},
+                    { "cost": [{ "heatFrames": 580 }]},
```
fdtaillefer: the count was off?

region/lowernorfair/east.json
```diff
@@ -1745,41 +1822,55 @@
...
-                    {"or": [
-                      "Super",
+                  {"obstacle":{
+                    "id": "C",
+                    "requires": [
+                      {"obstacle":{
+                        "id": "D",
+                        "requires": null
+                      }},
                       {"and": [
```
fdtaillefer: Since we're at the first level of the obstacle's `requires`, this `and` is implicit.

But I think it's probably meant to be part of a big `or` that's missing?

region/lowernorfair/east.json
```diff
@@ -1796,10 +1887,42 @@
...
                 ]},
-                {"cost": [
-                  {"heatFrames": 200}
-                ]},
-                "Morph"
+                {"cost": [{"heatFrames": 200}]},
+                "Morph",
+                {"or": [
+                  {"and": [
+                    {"cost": [{"heatFrames": 50}]},
```
fdtaillefer: I'm not sure what the implications of this are. Does it mean you are expected to handle the Dessgeega (or take a hit) when arriving at 4?

If so, I think the link from 2 to 4 should be 2 Dessgeega hits when not killing them.

region/lowernorfair/east.json
```diff
@@ -3095,46 +3269,39 @@
...
-                          ]}
-                        ]}
-                      ]}
-                    ]}
+                    {"enemyKill":{
+                      "enemies": [
+                        [ "Multiviola", "Multiviola" ],
+                        [ "Multiviola", "Multiviola", "Multiviola" ]
+                      ],
+                      "explicitWeapons": [ "ScrewAttack", "Plasma", "Super", "Missile", "PowerBomb" ]
```
fdtaillefer: Why did I include Missiles and Supers in there? It doesn't make much sense. If we're trying to avoid getting hit while killing them, I'd rather have Wave.

region/lowernorfair/east.json
```diff
@@ -4181,6 +4396,12 @@
                           "hits": 3
                         }]}
                       ]},
+                      {"and": [
```
fdtaillefer: I think I remember asking about that and you said the setup was too unreliable. Was it somewhere else? Did you find a good setup?

region/brinstar/kraid.json
```diff
@@ -653,7 +669,21 @@
           "to": [
             {
               "id": 2,
-              "requires": [ "canKillStandingGreenSpacePirates" ],
+              "requires": [
```
fdtaillefer: Speed Booster also works. It's pretty annoying coming from the left, but someone good enough can probably avoid damage and jump over the pirates and kill mini kraid with Power Beam.

region/brinstar/kraid.json
```diff
@@ -630,11 +634,23 @@
           },
           "locks": [
             {
-              "unlock": [ "Kill All Enemies" ]
+              "unlock": [ 
+                {"obstacle":{
+                  "id": "A",
+                  "requires": [ "never" ]
```
fdtaillefer: Why never? Seeing how the reqs are already on the obstacle anyway, we could probably simplify that by putting null instead (as in, no additional requirements beyond those on the obstacle)

region/brinstar/kraid.json
```diff
@@ -603,7 +602,12 @@
...
```
fdtaillefer: See comment on node 2

region/brinstar/red.json
DIProgan: Yeah that ripper at node 1 fooled me. Changing that as well.

region/lowernorfair/east.json
DIProgan: I made an "or" choice between two different enemyKill where one has PowerBombPeriphery as the only explicitWeapon. Maybe that would work? Also I made a video executing the rather ridicilous strat just to get a one-time runway open: https://youtu.be/xGtUHJgZlSk

region/lowernorfair/east.json
DIProgan: I think splitting them into two groups is the best choice

region/lowernorfair/east.json
DIProgan: Yes the lower one is pretty much a non-issue comming from the door..

region/lowernorfair/east.json
DIProgan: Yes, another thing I forgot to do.

region/lowernorfair/east.json
DIProgan: I moved 100 heatFrames into killing the Dessgeega at the new Obstacle C. Not the easiest balancing but hope it will do.

region/lowernorfair/east.json
DIProgan: Indeed I'm missing an "or" even if this Obstacle D won't be part of it. Ops!

region/lowernorfair/east.json
DIProgan: Yes and you're right it should be two hits at 2 to 4.

region/lowernorfair/east.json
DIProgan: Yeah it's not a good time trying to hit them with missiles. Changing.

region/lowernorfair/east.json
DIProgan: That might have been about the Kihunter shaft, I don't quite remember. This thing was already in the reqs but it was wrongly inside the kihunter obstacle.

region/brinstar/kraid.json
DIProgan: That's a funny strat, waiting for the spike. Thanks to this I also filled in missing attacks for "Mini-Kraid" in the enemies json.

region/lowernorfair/east.json
fdtaillefer: I imagine they would not be, indeed. I also imagine an obstacle within an `and` can only be considered destroyed if the `and` is actually used to traverse a link.

region/brinstar/kraid.json
fdtaillefer: There's a bit of short charging in that video though. I'm not sure it works otherwise.

fdtaillefer: Got one after like 20 tries. It's really ridiculously tight. You have to go fast and land super close to the last one.

DIProgan: It's only 7 damage of difference from using a spike anyway

PR #241, opened 2019-04-27
Boss Ammo

### LINKED ISSUE #239: enemyDamage needs to be in one and the same format
enemyDamage sometimes use the extra array to allow for multiple damage sources but sometimes does not. Need to decide which it will be. With the eventual removal of "cost" maybe the array is not that interesting anymore.

DIProgan:
enemyKill inserted into some bosses
Made a clean up in flags among others
Removed the enemyDamage array so closes #239


region/ceres/main.json
```diff
@@ -30,25 +30,33 @@
...
           },
           "locks": null
+        },
+        {
+          "id": 3,
+          "name": "Game Start",
+          "nodeType": "door",
+          "nodeSubType": "elevator",
+          "locks": null,
+          "note": "Game start gets its own node so that defeating Ceres Ridley is required to use the eleavtor at node 1."
```
fdtaillefer: I don't think that's necessary. Node 1 could just be locked behind `DefeatCeresRidley`

region/tourian/main.json
```diff
@@ -1244,7 +1243,7 @@
           },
           "locks": [
             {
-              "unlock": [ "DefeatMotherBrain" ]
+              "unlock": [ "ZebesSetAblaze" ]
```
fdtaillefer: This changes nothing functionally, but for a rando you'd probably want to take away `ZebesSetAblaze`. Seems like that'd work out better if the door were logically locked behing `DefeatMotherBrain`.

region/tourian/main.json
```diff
@@ -1266,34 +1265,36 @@
...
-          "note": "Because there is no official event for finishing MB1, the requirements for getting from broken glass to MB2 will be in the 3-to-4 link rather than an unlock on this node"
+          "note": "The requirements for finishing MB1 is in the 3 to 4 link."
         },
         {
           "id": 4,
           "name": "Mother Brain (phases 2 and 3)",
           "nodeType": "event",
           "nodeSubType": "boss",
-          "yields": [
-            "DefeatMotherBrain",
```
fdtaillefer: Oh, you're taking away the flag entirely? Why?

region/lowernorfair/east.json
```diff
@@ -3251,18 +3027,12 @@
                   "heatProof",
                   "canHeatRun"
                 ]},
-                {"roomReset":{
-                  "nodes": [2],
```
fdtaillefer: This one is touchy, but without those Dessgeegas, you can't really line up the iframes properly to charge blue suit without a short charge. It's probably why I put that in.

region/ceres/main.json
DIProgan: We never really defined what properties is locked behind locks but I guess using an elevator makes sense since other doors use it.

region/tourian/main.json
DIProgan: Why have 2 events that trigger from the same thing? ZebesSetAblaze seem to be what all of the changes look for.

fdtaillefer: I view it as not hurting anything vanilla while simplifying things for the combo rando, which has no guarantee of `ZebesSetAblaze` and has `DefeatMotherBrain` as part of its win conditions.

DIProgan: Will reinsert DefeatMotherbrain after Discord discussion

region/ceres/main.json
fdtaillefer: I did, actually. https://github.com/miketrethewey/sm-json-data/blob/master/region/region-readme.md
See Additional Considerations in the locks section.

fdtaillefer:
All good. I intend to merge this when I can also write the documentation update so it doesn't lag too far behind.

PR #242, opened 2019-04-30
Door memory addresses 

DIProgan:
Doors, elevators and "exits"


region/crateria/central.json
```diff
@@ -111,6 +115,7 @@
           "name": "Ship",
           "nodeType": "door",
           "nodeSubType": "elevator",
+          "nodeAddress": "",
```
fdtaillefer: Was that forgotten or missing? Do we want a FIXME note?

fdtaillefer: Oh, it's the ship. Does it need an address?

DIProgan:
```suggestion
          "note":  "FIXME If there's a proper memory address to the ship/elevator it was in hiding",
```

(I would remove the nodeAddress line completely for now)

PR #245, opened 2019-05-06
expectedDamageFrames

DIProgan:
Attempt at boss health handling with expectedDamageFrames and weapon cooldownFrames (rate of fire)

### DIFF
--- a/enemies/bosses/main.json
+++ b/enemies/bosses/main.json
@@ -22,6 +22,7 @@
         ]
       },
       "hp": 18000,
+      "expectedDamageFrames": 300,
       "amountOfDrops": 0,
       "drops": {
         "no-drop": 0,
@@ -75,6 +76,7 @@
         ]
       },
       "hp": 800,
+      "expectedDamageFrames": 0,
       "amountOfDrops": 16,
       "drops": {
         "no-drop": 0,
@@ -120,6 +122,7 @@
         ]
       },
       "hp": 960,
+      "expectedDamageFrames": 1000,
       "amountOfDrops": 12,
       "drops": {
         "no-drop": 0,
@@ -176,6 +179,7 @@
         ]
       },
       "hp": 1000,
+      "expectedDamageFrames": 600,
       "amountOfDrops": 16,
       "drops": {
         "no-drop": 0,
@@ -220,6 +224,7 @@
         ]
       },
       "hp": 0,
+      "expectedDamageFrames": 1000,
       "amountOfDrops": 11,
       "drops": {
         "no-drop": 0,
@@ -265,6 +270,7 @@
         ]
       },
       "hp": 2500,
+      "expectedDamageFrames": 600,
       "amountOfDrops": 16,
       "drops": {
         "no-drop": 0,
@@ -316,6 +322,7 @@
         ]
       },
       "hp": 3000,
+      "expectedDamageFrames": 600,
       "amountOfDrops": 16,
       "drops": {
         "no-drop": 0,
@@ -357,6 +364,7 @@
         ]
       },
       "hp": 6000,
+      "expectedDamageFrames": 600,
       "amountOfDrops": 16,
       "drops": {
         "no-drop": 0,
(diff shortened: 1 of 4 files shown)


enemies/bosses/main.json
```diff
@@ -22,6 +22,7 @@
         ]
       },
       "hp": 18000,
+      "expectedDamageFrames": 300,
```
fdtaillefer: 18000 hp at 20 hp per hit, 4 shots per second, adds up to a 225-second fight, or 13500 frames. Assuming contact damage, you now logically need 2 etanks to beat Ceres Ridley.

enemies/bosses/main.json
```diff
@@ -420,6 +428,7 @@
         ]
       },
       "hp": 13500,
+      "expectedDamageFrames": 1200,
```
fdtaillefer: Why would this fight take more damage if it lasts longer? You don't move and he can't hit you.

fdtaillefer:
There's really very few bosses that you can reliably hit at the rate of fire of your weapon. The expected time between hits should also vary in some cases based on item loadout. Possibilities for mid-fight health farming aren't taken into account. Is this really the correct approach?

weapons/main.json
```diff
@@ -181,11 +197,239 @@
...
+      "damage": 1200,
+      "cooldownFrames": 400,
+      "useRequires": ["Wave", "PowerBomb", "Charge"],
+      "shotRequires": [
+        {"ammo": {
+          "type": "PowerBomb",
+          "count": 1
+        }}
+      ],
+      "situational": false,
```
fdtaillefer: I think all SBAs should be considered situational.

region/tourian/main.json
```diff
@@ -170,7 +170,11 @@
                       "excludedWeapons": [ "Super", "Missile" ]
                     }},
                     {"or": [
-                      {"metroidFrames": 300},
+                      {"enemyDamage": {
+                        "enemy": "Metroid",
+                        "type": "contact",
```
fdtaillefer: I wonder if that attack would be clearer with a more descriptive name, like drain16Frames or something?

fdtaillefer:
For beam combo categories, charge is already handled.
For the others, I would lean towards `waveShot` / `iceshot` / `spazerShot` / `plasmaShot` categories. We can use those where appropriate in immunities and weaknesses, and for the weird cases like Evirs we can list them weapon by weapon (In the case of Evirs, I think that would mean 8 weapons to cover all the charged/uncharged ice combos that don't include plasma)

enemies/bosses/main.json
DIProgan: I thought I changed it back to 0, doh

enemies/bosses/main.json
DIProgan: My train of thought was something along the line of taking damage from thrown back supers that sometimes occur. Would you like to use 0?

fdtaillefer: I think the trajectory of those Supers may depend on your position, but yeah that's unclear.

weapons/main.json
DIProgan: Basically dooming them to never get into logic considerations but maybe it's better that way for now.

region/tourian/main.json
DIProgan: I thought of "drain" myself but felt that maybe it's better to keep a contact value for all enemies. I'm not exactly strongly opinionated on this.

DIProgan:
> For beam combo categories, charge is already handled.
> For the others, I would lean towards `waveShot` / `iceshot` / `spazerShot` / `plasmaShot` categories. We can use those where appropriate in immunities and weaknesses, and for the weird cases like Evirs we can list them weapon by weapon (In the case of Evirs, I think that would mean 8 weapons to cover all the charged/uncharged ice combos that don't include plasma)

Weaknesses and strengths for combined beams seem to be nuts.
For example weapons that have 0.5x on Kihunter (red):
Ice
Spazer
Ice+Wave

Weapons that don't: 
Wave
Ice+Spazer
the rest

PR #247, opened 2019-05-08
Add the X-Ray Climb in Big Pink

fdtaillefer:


region/brinstar/pink.json
```diff
@@ -758,6 +778,22 @@
...
           ]
+        },
+        {
+          "from": 14,
+          "to": [
+            {
+              "id": 5,
+              "requires": ["canStationarySpinJump"],
+              "unlock": null,
+              "note": "The only expected way for Samus to access 5 is to enter room while doing a stationary spinjump, otherwise she is expected to spawn at 14 and fall to 13."
```
DIProgan: Not going to lie this feels weird. The only thing really required is to press jump at the right time and then steer back but what purpose will this tech fill? It's unlikely anyone will have the reaction time to perform it the first time seeing the room unless the layout allows for a preview.

fdtaillefer: If you can get there once, you should be able to get there twice.
And stationary spinjumps are already required for the X-Ray climb, which is really the only thing you can do from the door.

DIProgan: I somehow thought this was the goal of the x-ray climb, not the start. So with that I understand why it has to exist.

PR #249, opened 2019-05-13
Remove "Kill All Enemy" logical elements, use a new lockType concept instead

fdtaillefer:


region/crateria/central.json
```diff
@@ -1082,7 +1109,9 @@
           },
           "locks": [
             {
-              "unlock": [ "ZebesAwake", "Kill All Enemies" ]
+              "lockType": "killEnemies",
+              "lock": ["ZebesAwake"],
```
DIProgan: This door is locked prior to activating ZebesAwake as well

fdtaillefer: It sure is. What I get for not checking.

DIProgan:
Looks good

PR #257, opened 2019-05-26
Add the possibility of bypassing a lock without deactivating it

fdtaillefer:
And apply it in two specific instances (Bowling Alley Skip, and Butterfly Botwoon Skip).

### DIFF
--- a/region/crateria/east.json
+++ b/region/crateria/east.json
@@ -364,7 +364,7 @@
               "lockType": "permanent",
               "unlock": [ "never" ],
               "bypass": [
-                "canCrystalFlash",
+                "canCrystalFlashForceStandup",
                 "canXRayClimb",
                 "canGrappleClip"
               ]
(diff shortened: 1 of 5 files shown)


region/crateria/east.json
```diff
@@ -362,7 +362,12 @@
           "locks": [
             {
               "lockType": "permanent",
-              "unlock": [ "never" ]
+              "unlock": [ "never" ],
+              "bypass": [
+                "canCrystalFlash",
```
DIProgan: Using crystal flash as a glitch tech feels quite different from just using it for health

DIProgan:
If comfortable with canCrystalFlash merge away

fdtaillefer:
> 
> 
> If comfortable with canCrystalFlash merge away

I can try to understand what it's there for. If I can figure out an adequate tech for it, I'll use that.

PR #264, opened 2019-06-16
Change link requirement into strats: Kraid

fdtaillefer:


region/brinstar/kraid.json
```diff
@@ -88,52 +101,136 @@
...
+                      }}
+                    ]}
+                  ],
+                  "note": [
+                    "Needs 1 Super with Morph.",
+                    "Needs 3 Supers with nothing at all."
+                  ]
+                },
+                {
+                  "name": "Kraid Entrance Wiggle",
```
osse101: You can make it through the 2 tile gap without needing to wiggle.

fdtaillefer: Yeah. My intent is for that to be covered by `canTunnelCrawl`. It's essentially a tunnel crawl where the initial jump can get you through without having to follow it up with a bunch of spinjumps.

osse101: On its face you are just jumping through a hole, certainly don't have to know how to perform the tunnel crawl to make it through, unless you get stuck on top.  What it is is hitbox manipulation.  Jumping over Oums is about the only thing I can think of resembling this.  
... downgrab tech ... yatayata ... you can just spinjump through the hole

fdtaillefer: Yeah, initially I was doing it with the bottom two blocks taken out. It looks easier with the top two blocks, and indeed works witha spinjump then.

So what changes do you think I oughta make? Is it renaming the strat, changing the tech from `canTunnelCrawl` to something else (something like `canTwoTileHitboxSqueeze`, I guess?), or both?

osse101: Right to left can be techless using just the spinjump.
Left to right I think the tech should be `canManipulateHitbox`.  I think the expectation is that you hold down to make it through the gap because that looks to be pretty safe, after practicing the jump a bit (not spinning means bonking the block above or below doesn't get you stuck in the gap, and your hitbox is smaller).  Wiggle stays as an alternate tech.
I'm thinking `Kraid Entrance Super Skip` because this reminds me of gap skips.

fdtaillefer: I think both ways should require the tech, because this isn't something the game teaches you. You shouldn't be expected to know that a precise spinjump can get you through a spot, when pressing up against the hole with a straight jump will just slide against a wall.

I like `canManipulateHitbox`, I'll use that.

PR #265, opened 2019-06-17
Change link requirement into strats: Pink Brinstar

fdtaillefer:


region/brinstar/pink.json
```diff
@@ -679,77 +806,138 @@
...
+                },
+                {
+                  "name": "Spore Spawn Skip",
+                  "notable": true,
+                  "requires": ["Morph"],
+                  "obstacles": [
+                    {
+                      "id": "F",
+                      "requires": [
+                        "canSuperReachAround",
```
osse101: alternatively, crystal flash

PR #268, opened 2019-06-23
Change link requirement into strats: Central Crateria

fdtaillefer:


region/crateria/central.json
```diff
@@ -200,92 +238,139 @@
...
-                    "HiJump"
+                  "name": "Base",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
+                      "canWalljump",
+                      "canFly",
+                      "SpeedBooster",
+                      "canDestroyBombWalls",
+                      "HiJump"
```
osse101: springball jump

region/crateria/central.json
```diff
@@ -2117,29 +2248,99 @@
...
+                    {"canShineCharge": {
+                      "usedTiles": 31,
+                      "shinesparkFrames": 130,
+                      "openEnd": 2
+                    }},
+                    "canUseFrozenEnemies"
+                  ]
+                },
+                {
+                  "name": "Kill the Boyons",
```
osse101: I find this one hard to get anything useful out of without a shortcharge, compared to bluesuit jump

region/crateria/central.json
fdtaillefer: This is why trying to have walljump as a tech is so much work.

region/crateria/central.json
fdtaillefer: Hm... you're right. I feel like I could just make it notable as an alternate version, but that feels pointless.

I'll change it into a 1-tap, that'll make it a meaningful alternative.

PR #269, opened 2019-06-23
Change link requirement into strats: East Crateria

fdtaillefer:


region/crateria/east.json
```diff
@@ -80,34 +108,58 @@
...
+                    {"or":[
+                      "SpaceJump",
+                      "Grapple",
+                      {"canComeInCharged": {
+                        "fromNode": 1,
+                        "framesRemaining": 0,
+                        "shinesparkFrames": 50
+                      }},
+                      {"and": [
+                        "Gravity",
```
osse101: This should be split into an `AquamanRises` series of strats.
gravity jump, springball jump, and ibj are all notable tricks imo

region/crateria/east.json
```diff
@@ -534,13 +666,19 @@
...
-              ],
-              "note": "The shot blocks can be destroyed without using bombs"
+              "strats": [
+                {
+                  "name": "Base",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
+                      "canPassBombPassages",
+                      "canUseSpringBall"
```
osse101: you can bug boost through here too.  Only get 1 shot and you have to be quick.

region/crateria/east.json
```diff
@@ -636,22 +865,40 @@
...
+                  "name": "Base",
+                  "notable": false,
+                  "requires": [
+                    "Super",
+                    {"ammo": {
+                      "type": "Super",
+                      "count": 1
+                    }}
+                  ],
+                  "note": "Super is for a Super Block. It's not an obstacle because it respawns."
```
osse101: This doesn't represent that you only need 1 super to get the item and get back out.  not that its trivial to do

region/crateria/east.json
fdtaillefer: What I've tried to do with notable is that if there's nothing unusual about using that tech vs in other places, I don't make it notable. That is, if I figure being able to execute the tech should allow you to execute the strat without further practice, I won't make it notable, except if the specific trick is quite well-known (e.g. PCJR short charge).

The IBJs I've made notable have usually been longer ones, or ones starting in midair (which I want to make a tech for sometime, and a helper allowing you to substitute SpeingBall + regular IBJ). Gravity Jump and vertical MidAir Springball are likewise usually very standard (but for example, their use is notable in Pants Room). Springwalls are pretty much all notable because many of them are different.

I've not done that for X-Ray climb (I made them all notable) and questioned that for a while, but now I feel like those are cases where it's much better to know exactly where the climb ends which you can't really tell by winging it.

I've thought about making a helper for jump assist things, but I'm not sure it helps clarity because what works or doesn't is pretty variable from one place to the next (sometimes hj works and SpringBall doesn't. Sometimes you need both). Maybe `hasSingleJumpAssist` and `hasTwoJumpAssists` could hold value, but in the end it's just for readability.

A thing I'm noticing is as time went on, I started splitting things into more strats compared to earlier in the refactor.

region/crateria/east.json
fdtaillefer: Thanks, I hate it.
I think I'm going to have to required mockball on this one.

region/crateria/east.json
fdtaillefer: Indeed. I don't know how to fit it into the structure and I don't think the logical gains would be worth the effort, but I'll add a note.

fdtaillefer: Or I guess it could just be a notable strat. That works by accident as long as we don't find a new way to access node 14.

PR #270, opened 2019-06-24
Change link requirement into strats: West Crateria

fdtaillefer:


region/crateria/west.json
```diff
@@ -231,59 +249,128 @@
...
+                    {"enemyKill":{
+                      "enemies": [
+                        [
+                          "Green Space Pirate (standing)",
+                          "Green Space Pirate (standing)",
+                          "Green Space Pirate (standing)"
+                        ],
+                        [
+                          "Green Space Pirate (standing)",
+                          "Green Space Pirate (standing)"
```
osse101: what's with the grouping? ammo drops?

region/crateria/west.json
```diff
@@ -799,24 +959,40 @@
...
+                  ]
+                },
+                {
+                  "name": "Shinespark",
+                  "notable": false,
+                  "requires": [
+                    {"canComeInCharged": {
+                      "fromNode": 1,
+                      "framesRemaining": 0,
+                      "shinesparkFrames": 100
```
osse101: There should be a number of runwaytiles field to this right?  You can make this spark in vanilla

region/crateria/west.json
```diff
@@ -799,24 +959,40 @@
...
-                  "canUseMorphBombs",
-                  {"and": [
+              "strats": [
+                {
+                  "name": "Base",
+                  "notable": false,
+                  "requires": ["ScrewAttack"]
+                },
+                {
+                  "name": "Morph Bomb Gauntlet Entrance (In)",
```
osse101: Going left to right is not what I'd consider "In"

region/crateria/west.json
```diff
@@ -954,45 +1146,81 @@
...
+                },
+                {
+                  "name": "Morph Bomb Gauntlet E-Tank Room (In)",
+                  "notable": true,
+                  "requires": null,
+                  "obstacles": [
+                    {
+                      "id": "A",
+                      "requires": ["canUseMorphBombs"],
+                      "note": "FIXME health?"
```
osse101: door to door is possible damageless in both directions

region/crateria/west.json
```diff
@@ -954,45 +1146,81 @@
...
+                },
+                {
+                  "name": "Shinespark",
+                  "notable": false,
+                  "requires": null,
+                  "obstacles": [
+                    {
+                      "id": "B",
+                      "requires": [
+                        {"canComeInCharged": {
```
osse101: This spark can be charged in room with a stutter 4

region/crateria/west.json
fdtaillefer: See documentation for enemyKill: https://github.com/miketrethewey/sm-json-data/blob/master/logicalRequirements.md

Essentially, the groups are an estimation of what can be caught together by a PB blast.

region/crateria/west.json
fdtaillefer: Oops

region/crateria/west.json
fdtaillefer: It may not be super instinctive, but there's no tile count for `canComeInCharged` because it would be redundant with things we've already entered.

See the documentation for `canComeInCharged` here, most notably the third bullet in additional considerations: https://github.com/miketrethewey/sm-json-data/blob/master/logicalRequirements.md.

In short, `canComeInCharged` can be fulfilled by combining the two runways attached to a door transition.

In this case, we have 6 tiles in the runway in this room and 10 in the adjacent runway. One of those is lost in the transition. This leaves us with a 15-tile runway with one open end, including 2 diagonal up tiles and one diagonal down tile. An algorithm would ideally gather all this data and decide whether that's doable under the current configuration.

I don't know how diagonals impact the charging mechanics, but I know they shorten the distance (and that up and down diagonals seem to shorten it by different amounts). That's why all the data is there in the runways. Maybe the topic has been studied by someone in the SM community? In the end, that's out of scope for this project, but it would be useful when writing the algorithm.

region/crateria/west.json
fdtaillefer: That's covered. See my previous comment.

region/crateria/west.json
fdtaillefer: It is, I think the issue at hand is leniency.
I would think in the end we'll want the damageless strat to be notable, with another one having some acid frames of leniency.

region/crateria/west.json
osse101: Runways are part of the door definitions, got it.  I remembered them being a thing but couldn't remember where they'd be located.

PR #272, opened 2019-07-05
Change link requirement into strats: West Lower Norfair

fdtaillefer:


region/lowernorfair/west.json
```diff
@@ -112,34 +103,40 @@
...
-                  {"and": [
+                    {"heatFrames": 200}
+                  ]
+                },
+                {
+                  "name": "Without Acid",
+                  "notable": false,
+                  "requires": [
+                    "canNavigateHeatRooms",
                     "UsedAcidChozoStatue",
```
osse101: `UsedAcidChozoStatue` implies having spacejump making this the same as the base strat.

region/lowernorfair/west.json
```diff
@@ -163,26 +163,40 @@
...
+                    "SpaceJump",
+                    {"heatFrames": 300}
+                  ]
+                },
+                {
+                  "name": "Without Acid",
+                  "notable": false,
+                  "requires": [
+                    "canNavigateHeatRooms",
+                    "UsedAcidChozoStatue",
```
osse101: same as above

region/lowernorfair/west.json
```diff
@@ -740,112 +844,208 @@
...
+                {
+                  "name": "Space Morph Bombs",
+                  "notable": false,
+                  "requires": ["canNavigateHeatRooms"],
+                  "obstacles": [
+                    {
+                      "id": "A",
+                      "requires": [
+                        "canUseMorphBombs",
+                        "ScrewAttack",
```
osse101: Space jump

region/lowernorfair/west.json
```diff
@@ -1033,13 +1330,24 @@
...
+                {
+                  "name": "Base",
+                  "notable": false,
+                  "requires": [
+                    "canNavigateHeatRooms",
+                    "canOpenGreenDoors",
+                    {"heatFrames": 600},
+                    {"or":[
+                      "ScrewAttack",
+                      {"enemyDamage": {
```
osse101: power bomb or super kills should save health over taking contact hits.  
I'm not sure how helpful ice or dboosts are.
CanComeInCharged for sure, if we're doing that.

region/lowernorfair/west.json
fdtaillefer: I tried not to make that assumption to prevent the logic from breaking if someone were to mod the statue.

region/lowernorfair/west.json
fdtaillefer: Same

region/lowernorfair/west.json
fdtaillefer: Indeed

region/lowernorfair/west.json
fdtaillefer: I think my rationale was that trying to kill them was liable to cost as much in heat frames as the energy of charging through.

I'm reconsidering now because heatproof scenarios are different.

osse101: yeah, with heatproof it makes sense.  I didn't realize their contact damage was so low that it doesn't matter much without heatproof.  Super kills save ~10 health, PB kills lose more health but gain drops.  I did find a slick dboost (left to right) however.

PR #276, opened 2019-07-07
Change link requirement into strats: West Norfair

fdtaillefer:


region/norfair/west.json
```diff
@@ -1507,22 +1776,29 @@
...
-                {"heatFrames": 150}
+              "strats": [
+                {
+                  "name": "Base",
+                  "notable": false,
+                  "requires": [
+                    "canNavigateHeatRooms",
+                    {"heatFrames": 150}
+                  ],
+                  "note": "Involves crumble block jumping. Should we have alternate strats?"
```
osse101: The room in general is a good candidate for canWallJump.  Crumble jumping should be its own tech.  Reaching door1 would then resemble a tricky walljump.  
The intended strat to 3 looks to be either a horizontal spark from the door, or highjump+ice from door2.

fdtaillefer: re: Crumble Jump as a tech. It started out being one, there was a discussion, it was removed, and it's been bugging me since. In fact, there was also `canCrumbleSpinJump`. I don't think we want to do that here though, because it would have to impact other regions as well.

So I created issue #284.

As for walljump, well. I see it as its own project. I wish it weren't anywhere, I just left it in where others have put it because it seemed like a shame to waste the work.

osse101: Reaching the item I'm ok with, but I'd like to see reaching door1 be marked nontrivial

fdtaillefer: I can do that for now. Issue #284 will end up overhauling this room completely, probably with a bunch of alternate strats.

PR #277, opened 2019-07-27
Change link requirement into strats: Inner Green Maridia

fdtaillefer:


region/maridia/inner-green.json
```diff
@@ -132,57 +166,131 @@
...
+                        "fromNode": 2,
+                        "framesRemaining": 30,
+                        "shinesparkFrames": 23
+                      }}
                     ]}
-                  ]},
-                  {"and": [
+                  ],
+                  "note": [
+                    "Not sure whether this should require canNavigateUnderwater?",
```
osse101: It does make use of water physics to reach the top platform 🤷‍♀

region/maridia/inner-green.json
```diff
@@ -878,88 +924,236 @@
...
+                {
+                  "name": "Suitless",
+                  "notable": false,
+                  "requires": ["canSuitlessMaridia"]
+                }
+              ]
+            },
+            {
+              "id": 7,
+              "strats": [
```
osse101: missing gravity+springball

region/maridia/inner-green.json
fdtaillefer: Between that and the fact that positioning the spark can be finicky depending on incoming speed, I'll just require it.

region/maridia/inner-green.json
fdtaillefer: Sure is

PR #278, opened 2019-07-28
Change link requirement into strats: Inner Pink Maridia

fdtaillefer:


region/maridia/inner-pink.json
```diff
@@ -1303,17 +1377,33 @@
...
-                      "HiJump",
-                      "canUseFrozenEnemies"
+                      "HiJump"
                     ]}
-                  ]}
-                ]}
+                  ]
+                },
+                {
+                  "name": "Suitless Frozen Zoa",
```
osse101: This is either a mistake, and there should be a junction between 5 and 4 that this is trying to reach. 
Or this is a cosmic brain freeze the bug 1 tile up inside the maze (charge buffer holding down and release charge when hit), downgrab to land on it, morph and wait for it to unfreeze to dboost up the 2nd tile, repeat, then drop onto the bug to cross the final gap to the item.

osse101: Forgot jump+morph was a thing, leaving the comment because it was a fun find.

region/maridia/inner-pink.json
```diff
@@ -83,38 +83,77 @@
...
+                  "notable": false,
+                  "requires": [
+                    "canSuitlessMaridia",
+                    "canSpringBallJumpMidAir",
+                    "canUseFrozenEnemies"
+                  ],
+                  "note": "Differs from crab climb in that the crabs only serve as one-time stepping stone between platforms."
+                },
+                {
+                  "name": "Frozen Jump Assists",
```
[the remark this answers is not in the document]
fdtaillefer: I was probably going for Dual Jump Assist

region/maridia/inner-pink.json
```diff
@@ -124,25 +163,37 @@
...
-                        "canUseFrozenEnemies"
-                      ]}
+                      "HiJump"
                     ]}
-                  ]}
-                ]}
+                  ]
+                },
+                {
+                  "name": "Frozen Crabs",
```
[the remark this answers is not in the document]
fdtaillefer: No, they are not. It's weird, sometimes it seems to work and sometimes I don't have enough height.
I think I can still add it.
The Watering Hole stationary spinjump strat also works here, if you start on top of a frozen crab and face right.

region/maridia/inner-pink.json
fdtaillefer: That is one crazy strat. I love SM for all the different tools you have for problem solving. I think that's part of what makes entrando so appealing, new strats to come up with.

PR #279, opened 2019-08-03
Change link requirement into strats: Inner Yellow Maridia

fdtaillefer:


region/maridia/inner-yellow.json
```diff
@@ -348,29 +424,53 @@
...
-                  ]}
-                ]}
+                  ]
+                },
+                {
+                  "name": "Naked Watering Hole Escape",
+                  "notable": true,
+                  "requires": [
+                    "canSuitlessMaridia",
+                    "canStationarySpinJump"
```
osse101: `canSunkenDualWallClimb`?

region/maridia/inner-yellow.json
```diff
@@ -1084,30 +1414,58 @@
...
-                      [
-                        "Ripper", "Ripper", "Ripper"
+              "strats": [
+                {
+                  "name": "Frozen Rippers",
+                  "notable": false,
+                  "requires": ["canUseFrozenEnemies"]
+                },
+                {
+                  "name": "Space Jump",
```
osse101: Is this notably easier than tricky walljumps?

region/maridia/inner-yellow.json
fdtaillefer: I'm not sure this is a tech. `canSunkenDualWallClimb` involves going back and forth between two walls two tiles apart. Here the walls are one tile apart and you just need to spam jump while facing in one direction.

If we determine it should be a tech, it would be a different IMO.

region/maridia/inner-yellow.json
fdtaillefer: I'm honestly not sure. To be it's noticeably more difficult. You think this strat should just be removed?

osse101: yeah, I'd remove it.  The walljumps are the same level of 'tricky' as the spacejumps, navigable just like the top of red tower.

PR #281, opened 2019-08-04
Change link requirement into strats: Tourian

fdtaillefer:


region/tourian/main.json
```diff
@@ -238,12 +274,22 @@
...
-                  {"acidFrames": 30}
-                ]}
+              "strats": [
+                {
+                  "name": "Base",
+                  "notable": false,
+                  "requires": ["SpaceJump"]
+                },
+                {
+                  "name": "Speedjump",
```
osse101: Springball bounce.
Frozen rinkas.  I used the leftmost one to setup a good angle on the center-left one to cross. 
Right to left I just used the top right rinka to cross.

osse101: The right door should be blue so there should be metroid avoidance strats like the other rooms right?

fdtaillefer: Hmm... I'm not sure how those strats were even setup, I can't reset the room properly. Maybe it was assumed this room isn't as bad to avoid the Metroids?

fdtaillefer: I'm also not sure how to organize this since the acid pit has a bunch of strats and so would avoiding the metroids. I think I'd rather just leave a FIXME for now and figure it out later.

PR #282, opened 2019-08-05
Change link requirement into strats: Wrecked Ship

fdtaillefer:


region/wreckedship/main.json
```diff
@@ -1632,7 +2049,40 @@
           "to": [
             {
               "id": 2,
-              "requires": null
+              "strats": [
```
osse101: Springball bounce on the first and third platforms

fdtaillefer: Neat

PR #289, opened 2019-08-19
Reusable type definitions in the schema

fdtaillefer:


schema/m3-region.schema.json
```diff
@@ -1,6 +1,487 @@
...
+            "Brinstar Reserve Hole-in-one"
+          ],
+          "pattern": "^(.*)$"
+        },
+        "notable": {
+          "$id": "#/definitions/strat/properties/notable",
+          "type": "boolean"
+        },
+        "requires": {
+          "$ref" : "#/definitions/logicalRequirements",
```
osse101: missing `type` field right?  and the `items` that goes with it.

see obstacles:requires and obstacles:bypass too

fdtaillefer: No, that's the beauty of #ref. It's pulling the definition of `logicalRequirements` that's written higher up. So that complex type can now be referenced wherever without duplicating the definition.

osse101: I see it now

PR #290, opened 2019-08-24
Bring back Crumble Jump techs

fdtaillefer:
Includes some other changes to support the alternate strats in places with crumble jumps.


region/lowernorfair/west.json
```diff
@@ -454,12 +454,54 @@
               "id": 3,
               "strats": [
                 {
-                  "name": "Base",
+                  "name": "GT Crumble Jump to Item",
```
osse101: make-able jump starting from inside the doorway

region/lowernorfair/west.json
```diff
@@ -503,11 +545,41 @@
...
-                  "name": "Base",
-                  "notable": false,
+                  "name": "GT Crumble Jump from Item",
+                  "notable": true,
                   "requires": [
                     "canNavigateHeatRooms",
+                    "canCrumbleJump",
                     {"heatFrames": 150}
+                  ],
+                  "note": "It takes a short hop to avoid hitting the ceiling to do this with only one crumble jump."
```
osse101: a short hop and precise wall jump avoids the crumble blocks.

region/norfair/west.json
```diff
@@ -1810,12 +1911,108 @@
...
+                  "notable": true,
+                  "requires": [
+                    "canNavigateHeatRooms",
+                    "canCrumbleJump",
+                    {"heatFrames": 800}
+                  ],
+                  "note": "Do 9 successive crumble jumps up the platforms."
+                },
+                {
+                  "name": "Crumble Shaft Wall/Crumble Climb",
```
osse101: obsoletes the 9 crumble jump strat

region/lowernorfair/east.json
```diff
@@ -3724,13 +3725,34 @@
...
+                },
                 {
                   "name": "MidAir Spring Ball",
                   "notable": false,
                   "requires": [
                     "canNavigateHeatRooms",
-                    {"heatFrames": 250},
-                    "canSpringBallJumpMidAir"
+                    "canSpringBallJumpMidAir",
+                    "canCrumbleJump",
```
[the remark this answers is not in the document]
fdtaillefer: So is the timing. I could put it in but it would have to be a notable strat IMO. It feels way harder than the springwall to me.

region/lowernorfair/west.json
fdtaillefer: Indeed.

region/lowernorfair/west.json
fdtaillefer: I don't understand this one

fdtaillefer: ...Oh. Walljump off the far end. Wow.

region/norfair/west.json
fdtaillefer: True enough, but seeing how it's a notable strat, I think the other one should still stay in. Each of the 9 crumble jumps are individually easier than the single one here IMO.

fdtaillefer:
> 
> 
> As `canQuickCrumbleEscape` is written, Early supers room missiles needs a `canCrumbleJump`/spacejump/springball.
> MissionImpossible exit to Big Pink might be ok as is.



> 
> 
> As `canQuickCrumbleEscape` is written, Early supers room missiles needs a `canCrumbleJump`/spacejump/springball.
> MissionImpossible exit to Big Pink might be ok as is.

Yeah, The early supers one is different because you pretty much need to jump on a crumble block on the way back.

PR #432, opened 2021-09-25
Add requirements to get to door 1 in Landing Site

kjbranch:
The existing logic from location 7 to 1 only requires h_canDestroyBombWalls, but it should require abilities to get up to those blocks. A shinespark can also be used to get from 5 to 1 bypassing 7 (this uses less energy than shinesparking from 4 to 1, but only breaks the first set of blocks).

### DIFF
--- a/region/crateria/central.json
+++ b/region/crateria/central.json
@@ -473,29 +473,6 @@
           "from": 5,
           "to": [
             {
-              "id": 1,
-              "strats": [
-                {
-                  "name": "Shinespark",
-                  "notable": false,
-                  "requires": [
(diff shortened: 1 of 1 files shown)


fdtaillefer:
I'd like to avoid the extra 5 -> 1 link if possible, to keep pathing as linear as possible.

How would you feel about folding the 5 -> 1 shinespark in the existing 7 -> 1 link? While it's not technically correct, you can always reach 7 with Speed Booster and you can technically drop down from 7 to charge the spark.

kjbranch:
I think it would be fine. I can add a note so that it is still clear.

PR #439, opened 2022-11-28
Replace boss enemyKill strats with explicit strats

kjbranch:
Replace enemyKill strats for Phantoon and Spore Spawn. 

Reasons:
- enemyKill is more complicated to follow, requiring processing a lot of information in order to determine the possible strats (e.g., enemy JSON, weapons JSON, and the modifiers in the strat such as farmableAmmo, explicitWeapons, excludedWeapons).
- Because of this, it will be easier for us to use in the randomizer we are designing as we haven't fully implemented enemyKill. It may also make it easier for a player to see all of the possible strats in logic which can be used to kill the bosses.
- This more explicit structure is already used for most of the other bosses/minibosses (BT, Kraid, Crocomire, GT, Back-Side Botwoon, and the other Spore Spawn strats).

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -2863,11 +2863,10 @@
                   "name": "Supers",
                   "notable": false,
                   "requires": [
-                    {"enemyKill":{
-                      "enemies": [
-                        [ "Spore Spawn" ]
-                      ],
-                      "explicitWeapons": [ "Super" ]
+                    "Super",
+                    {"ammo": {
+                      "type": "Super",
+                      "count": 4
                     }}
                   ],
                   "note": [
--- a/region/wreckedship/main.json
+++ b/region/wreckedship/main.json
@@ -956,33 +956,24 @@
               "lockType": "bossFight",
               "unlockStrats": [
                 {
-                  "name": "Regular Phantoon",
+                  "name": "Charge",
                   "notable": false,
-                  "requires": [
-                    {
-                      "enemyKill": {
-                        "enemies": [
-                          ["Phantoon"]
-                        ],
-                        "excludedWeapons": ["Super"],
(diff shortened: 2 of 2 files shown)


miketrethewey:
Could you add some FIXMEs to come back and implement the enemyKill strats at a later time?
(I would but I don't know when I'll be at my battlestation next.)

blkerby:
To clarify, when we say we haven't fully implemented enemyKill we're referring not to the sm-json-data itself but to our Map Randomizer (we have a TODO for it [here](https://github.com/blkerby/MapRandomizer/blob/dcf154f9dd583ce166162c51aea5c4a36bf93f85/python/rando/sm_json_data.py#L694)). I think there are also boss scenarios missing from sm-json-data (for instance there is only one scenario for Ridley and it requires Varia) but that's a separate issue.

The idea here is to try to reduce the dependence on enemyKill requirements on bosses, in order to make the logic simpler to understand both for humans and for our randomizer implementation. So the hope is that it won't be necessary to switch these back to enemyKill later. I'm hoping that any remaining enemyKill requirements on bosses (e.g. for Ridley or Botwoon) could be implemented the same as for other enemies (i.e., not requiring a boss scenario JSON). If that's possible I think that in general this could make it a lot easier for randomizers and other tools to consume the data. At least in the cases here of Phantoon and Spore Spawn, it seems like simple explicit strats are enough, because the fact that you can farm energy & ammo makes the calculations in enemyKill or boss scenarios irrelevant.

fdtaillefer:
Worth noting that there is a separate concept for boss fights that has had work put into it but not full integration (IIRC).
See boss scenarios here https://github.com/miketrethewey/sm-json-data/blob/master/enemies/bossScenarios-readme.md

PR #442, opened 2022-11-29
Add canDownGrab strat to the Electric Death Room

kjbranch:
This link is not quite free as it requires a down grab if you have no items.

### DIFF
--- a/region/wreckedship/main.json
+++ b/region/wreckedship/main.json
@@ -2289,7 +2289,18 @@
                 {
                   "name": "Base",
                   "notable": false,
-                  "requires": []
+                  "requires": [
+                    {"or": [
+                      "Gravity",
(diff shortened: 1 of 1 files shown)


miketrethewey:
Remind me what the "notable" element is for?

kjbranch:
It's basically a field to indicate if the strat is interesting or challenging


Notable Strats
Some strats are flagged as "notable". this means a few things:

The strat's name must be unique amongst all existing strats in the project.
Just having the required items and the ability to perform the required techs may not be enough to execute a notable strat. It might require:
A harder version of a tech
Knowledge specific to the strat (such as a unique setup)
An algorithm using this project to drive its logic should consider allowing to toggle off (or on) any given notable strat.

PR #443, opened 2022-11-29
Add canWalljump checks throughout Brinstar

kjbranch:
Most changes are simple wall jump additions throughout. Additionally there are a couple other changes:

- Switched to using the helper function "h_canDestroyBombWalls" in Blue Brinstar E-Tank Room to make the code simpler.
- Added a Space Jump strat to get to the top of Red Tower, as the previous Space Jump strat now requires a wall jump.

### DIFF
--- a/region/brinstar/blue.json
+++ b/region/brinstar/blue.json
@@ -1140,33 +1140,38 @@
                     "h_canUsePowerBombs",
                       {"or": [
                         "canWalljump",
-                        "h_canFly",
-                          {"canShineCharge": {
-                            "usedTiles": 33,
-                            "shinesparkFrames": 43,
-                            "openEnd": 2
-                          }}
+                        "h_canFly"
                       ]}
                   ],
                   "obstacles": [
                     {
                       "id": "A",
                       "requires": [
-                        {"or": [
-                          "h_canDestroyBombWalls",
-                          {"canShineCharge": {
-                            "usedTiles": 33,
-                            "shinesparkFrames": 43,
-                            "openEnd": 2
-                          }}
-                        ]}
+                        "h_canDestroyBombWalls"
                       ]
                     }
                   ],
-                  "note": [
-                    "Base strat when entering the room from below.",
-                    "The runway here is 31 tiles before breaking the PB blocks, but becomes longer after"
-                  ]
+                  "note": "Base strat when entering the room from below."
+                },
+                {
+                  "name": "Shinespark",
+                  "notable": false,
+                  "requires": [
+                    "h_canUsePowerBombs"
+                  ],
+                  "obstacles": [
+                    {
+                      "id": "A",
+                      "requires": [
+                        {"canShineCharge": {
+                          "usedTiles": 33,
+                          "shinesparkFrames": 43,
+                          "openEnd": 2
+                        }}
+                      ]
+                    }
+                  ],
+                  "note": "The runway here is 31 tiles before breaking the PB blocks, but becomes longer after"
                 }
               ]
             }
--- a/region/brinstar/blue.json
+++ b/region/brinstar/blue.json
@@ -1157,18 +1157,17 @@
                   "name": "Shinespark",
                   "notable": false,
                   "requires": [
-                    "h_canUsePowerBombs"
+                    "h_canUsePowerBombs",
+                      {"canShineCharge": {
+                        "usedTiles": 33,
+                        "shinesparkFrames": 43,
+                        "openEnd": 2
+                      }}
                   ],
                   "obstacles": [
                     {
                       "id": "A",
-                      "requires": [
-                        {"canShineCharge": {
-                          "usedTiles": 33,
-                          "shinesparkFrames": 43,
-                          "openEnd": 2
-                        }}
-                      ]
+                      "requires": []
                     }
                   ],
                   "note": "The runway here is 31 tiles before breaking the PB blocks, but becomes longer after"
--- a/region/brinstar/blue.json
+++ b/region/brinstar/blue.json
@@ -1114,7 +1114,13 @@
                 {
                   "name": "Go Back Up",
                   "notable": false,
-                  "requires": ["Morph"], 
+                  "requires": [
+                    "Morph",
+                      {"or": [
+                        "canWalljump",
+                        "h_canFly"
+                      ]}
+                  ], 
                   "obstacles": [
                     {
                       "id": "B",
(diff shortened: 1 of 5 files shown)


region/brinstar/blue.json
```diff
@@ -1114,50 +1114,50 @@
...
                   "notable": false,
-                  "requires": ["h_canUsePowerBombs"],
+                  "requires": [
+                    "h_canUsePowerBombs",
+                      {"or": [
+                        "canWalljump",
+                        "h_canFly",
+                          {"canShineCharge": {
+                            "usedTiles": 33,
+                            "shinesparkFrames": 43,
```
fdtaillefer, line 1146: Doesn't this mean that the shinespark approach has to pay twice the energy cost?

It might be preferrable to break this strat in two, so we can have a dedicated shinespark strat that clears the obstacle for free.

region/brinstar/green.json
```diff
@@ -1077,7 +1077,22 @@
...
+                {
+                  "name": "Wall Jump",
+                  "notable": false,
+                  "requires": [ "canWalljump" ]
+                },
+                {
+                  "name": "Crouch Jump Down Grab",
+                  "notable": false,
+                  "requires": [ "canDownGrab" ]
                 }
```
fdtaillefer, line 1096: I guess if we want to be thorough we need a SpringBall Jump strat.

There's also a ~20-frame shinespark that can be done here if you come in charged from node 2 with ~150 frames remaining. It's annoying, you have to double-angle to aim up and shoot the shot blocks, but it's a thing you can do.

region/brinstar/blue.json
kjbranch: Good catch. This should fix it.

region/brinstar/green.json
kjbranch: We haven't typically considered strats with tech that is significantly harder than simple strats with the same item requirements. 
Springball jumps haven't been added in other rooms where a simple wall jump works.

region/brinstar/green.json
```diff
@@ -2424,19 +2442,33 @@
...
                   "name": "Base",
                   "notable": false,
-                  "requires": ["Morph"]
+                  "requires": [
+                    "Morph",
+                      {"or": [
+                        "HiJump",
+                        "h_canFly",
+                        "canWalljump",
+                        "canUseFrozenEnemies"
```
fdtaillefer, line 2451: A SpringBall Jump also works here.

region/brinstar/kraid.json
```diff
@@ -767,7 +767,13 @@
                 {
                   "name": "Base",
                   "notable": false,
-                  "requires": []
+                  "requires": [
+                    {"or": [
+                      "HiJump",
+                      "h_canFly",
+                      "canWalljump"
```
fdtaillefer, line 774: SpringBall Jump works here. So does a CrouchJump+DownGrab, and a 16-frame shinespark if coming in charged from node 2 with ~150 frames remaining.

kjbranch: Added the crouch jump down grab. I don't think the others are necessary.

region/brinstar/pink.json
```diff
@@ -1072,7 +1072,17 @@
...
-                  "requires": []
+                  "requires": [
+                    {"or": [
+                      "Grapple",
+                      "h_canFly",
+                      "canWalljump",
+                        {"and": [
+                          "SpeedBooster",
+                          "HiJump"
+                        ]}
```
fdtaillefer, line 1083: A SpringBall Jump (with HiJump) works here. You can also do a non-HiJump SpringBall Jump if starting from a frozen Sidehopper (doesn't work on the ground, but the apex of a small hop works).

There's also a ~24-frame shinespark you can charge in-room at node 3's runway. Can also be extended by charging in the adjacent room (with a `canComeInCharged from node 3, with 180 frames remaining, and an inRoomPath of [3, 13]).

region/brinstar/pink.json
```diff
@@ -2361,6 +2371,10 @@
                   "name": "Waterway Suitless Spark",
                   "notable": true,
                   "requires": [
+                    {"or": [
```
fdtaillefer, line 2374: It's possible to get up there with a SpringBall Jump (having HiJump) by jumping a tiny bit later than you press pause.

If getting on the ledge is no longer free, we also have to consider a `canComeInCharged` from node 1. Needs roughly 60 frames remaining on it, and it's a ~150-frame shinespark.

region/brinstar/red.json
```diff
@@ -272,7 +272,13 @@
                 {
                   "name": "Base",
                   "notable": false,
-                  "requires": []
+                  "requires": [
+                    {"or": [
+                      "HiJump",
+                      "h_canFly",
+                      "canWalljump"
```
fdtaillefer, line 279: SpringBall jumps work here, or a 68-frame shinespark (dead-easy execution, only needs ~30 frames remaining coming in from node 1).

region/brinstar/red.json
```diff
@@ -385,15 +391,24 @@
...
+                  "requires": [
+                    "canUseFrozenEnemies",
+                      {"or":[
+                        "HiJump",
+                        "canWalljump"
+                      ]}
+                  ]
                 },
                 {
                   "name": "IBJ",
```
fdtaillefer, line 414: I think it's worth adding `canWalljump` to `Red Tower Midair Spring Ball` too.

region/brinstar/red.json
```diff
@@ -530,6 +546,19 @@
...
+                  "notable": false,
+                  "requires": [
+                    "SpaceJump",
+                    {"enemyKill":{
+                      "enemies": [
+                        [ "Ripper", "Ripper", "Ripper", "Ripper" ]
+                      ]
+                    }}
+                  ],
+                  "note": "This strat is for killing all of the Rippers and then using Space Jump to get up without wall jumps."
```
fdtaillefer, line 560: Is there really enough time to do that?

region/brinstar/red.json
```diff
@@ -1176,7 +1205,14 @@
                 {
                   "name": "Base",
                   "notable": false,
-                  "requires": ["h_canPassBombPassages"]
+                  "requires": [
+                    "h_canPassBombPassages",
+                      {"or": [
+                        "HiJump",
+                        "h_canFly",
+                        "canWalljump"
```
fdtaillefer, line 1213: SpringBall Jumps work here, as well as a 33-frame shinespark coming in from node 1 (needs about 60 remaining frames).

fdtaillefer (changes requested):
First review done.

If you think some strats are too niche to add as an alternative to wall jumps, you can leave them out if you prefer. I'll add them sometime.

region/brinstar/red.json
kjbranch: Yes, that's how I always do it. lol

kjbranch:
I think that should cover all of the changes besides spring ball jumps and shinesparks that can be replaced by canWalljump.

region/brinstar/blue.json
```diff
@@ -1140,33 +1140,38 @@
...
+                  "name": "Shinespark",
+                  "notable": false,
+                  "requires": [
+                    "h_canUsePowerBombs"
+                  ],
+                  "obstacles": [
+                    {
+                      "id": "A",
+                      "requires": [
+                        {"canShineCharge": {
```
fdtaillefer, line 1166: I think we want the shinespark to be on the strat itself, and just clear the obstacle for free. As it is now, it would mean you can get back up for free if the obstacle is already cleared.

PR #451, opened 2022-12-19
Remove gray door requirements in Metroid Room 4

kjbranch:
Moved the Metroid Room 4 lock logic from lock strats to link strats by instead representing the metroids as obstacle A. Mirrored the logic from 2->1 to 1->2.

Reasons:
- The previous logic to traverse the room from 1->2 was dependent on the locks; that means any randomizer that uses this logic would require the bottom door to remain gray or else the logic would be incomplete. (In its previous form if the bottom door is not gray then going through the room from 1->2 would be free.)
  - This currently creates an issue in Map Rando: https://github.com/blkerby/MapRandomizer/issues/4
- The new approach can make the logic more generic. With the new representation, the data is in a state where any combination of top or bottom doors could instead be set to gray and dependent on the metroid kill.


fdtaillefer:
Good thinking. It would be ideal to add the new obstacle in the room diagram, but the changes themselves seem fine.

kjbranch:
Added the new obstacle to the room diagram.

PR #452, opened 2022-12-20
Fix Metroid Room 1 gray door by adding an extra node

kjbranch:
Removed the dependence on the gray door strats to make the logic more generic so randomizers can more easily change the gray doors.

The logic in this room cannot simply be moved to the strats because there are multiple ways to kill or avoid the Metroids as well as multiple strats to traverse the room.

In order to remedy both solutions, Node 3 and Obstacle A were added. 
- The player will traverse the room from 1->3->2 or 2->3->1. (There isn't really a reason to go back after getting to Node 3, so this should not be a problem.)
- Getting to node 3 represents killing or avoiding the Metroids. 
- Leaving Node 3 will represent crossing the room. 
- Killing the Metroids or having the metroids already dead (f_KilledMetroidRoom1) will clear obstacle A, which will unlock the gray door and yield the same flag.


fdtaillefer:
This is a tough one. You can't get past all the metroids without getting across the gap, which I imagine is why you did it like this with the middle node having an ambiguous meaning. This solution fully represent the navigation in this room.

For example: If you start at the left and want to unlock the door and exit left, you need to kill the Metroids but also cross the pit both ways.

Here's what I propose: Represent this with using two one-way paths, each of which has a center node that isn't used by the other path. Something like this:

```
  /-----> 3 ----->\
 /                 \
1                   2
 \                 /
  \<----- 4 <-----/
```

That modeling should still allow you to cross once by doing the metroid kill one and passing the gap once, but also properly require crossing the gap both ways while also killing the metroids to unlock the door coming from the left.

kjbranch:
Thanks for the feedback, this should be fixed now.

region/tourian/main.json
```diff
@@ -923,7 +1147,13 @@
                 {
                   "name": "Already Cleared",
                   "notable": false,
-                  "requires": ["f_KilledMetroidRoom4"]
+                  "requires": ["f_KilledMetroidRoom4"],
+                  "obstacles": [
```
fdtaillefer, line 1151: Good catch on those two

PR #454, opened 2022-12-22
Fix Plasma Gray Door

kjbranch:
It is possible to get out of Plasma Room without killing all of the Pink Pirates. Those strats were originally not considered because the gray door required all enemies to be killed.

Add tech: canJumpIntoIBJ


region/maridia/inner-yellow.json
```diff
@@ -2112,6 +2112,92 @@
...
+                        "usedTiles": 19,
+                        "shinesparkFrames": 25,
+                        "openEnd": 1
+                      }}
+                    ]}
+                  ],
+                  "note": "If you can kill them without being hurt, you can take path 2->3->1."
+                },
+                {
+                  "name": "Kill one then IBJ",
```
fdtaillefer, line 2149: We could eliminate some duplication here with a new node representing "one killed". Maybe not mandatory right now but if not I'll at least create an issue.

fdtaillefer: I have no other comments here so I'll just merge this and add the issue.

PR #461, opened 2022-12-24
Add left-facing door X-Ray climbs

fdtaillefer:


fdtaillefer:
Note: This may conflict with #443 so it might be simpler to wait for that one to be merged first.

PR #469, opened 2022-12-28
Fix Gray Doors in Kraid's E-Tank Room, Baby Kraid Room

kjbranch:
- Converted Kraid's E-Tank Room Beetoms to an obstacle. Referenced the obstacle in the gray door unlock strats.
- Added strats to get through Kraid's E-Tank Room, Baby Kraid Room without killing the enemies.


miketrethewey:
This does seem to give naïve algorithms a "requires" path of nekkid for passage. 
Is that a path we want to include or are we expecting the consumer to have handlers for enemy contact and obstacles?

region/brinstar/kraid.json
```diff
@@ -1043,6 +1114,34 @@
...
+                      "type": "spike",
+                      "hits": 1
+                    }}
+                  ],
+                  "note": "Quickly run through to tank the Pirate (or its laser) and one of Mini-Kraid's spikes."
+                },
+                {
+                  "name": "Dodge",
+                  "notable": false,
+                  "requires": [],
```
fdtaillefer, line 1138: Since we didn't have a no-reqs path through this room before, should we leave this as `never` until we figure something out, rather than leaving it free?

region/brinstar/kraid.json
```diff
@@ -1076,6 +1175,34 @@
...
+                      "type": "spike",
+                      "hits": 1
+                    }}
+                  ],
+                  "note": "Wait to hit one of Mini-Kraid's spikes, then run through. Tank one Pirate (or its laser)."
+                },
+                {
+                  "name": "Dodge",
+                  "notable": false,
+                  "requires": [],
```
fdtaillefer, line 1199: Same here

PR #484, opened 2023-01-03
Add Beetom strats in the Beetom Room, Etecoon E Tank Room

kjbranch:
Add dodge, kill, and evade strats in the Green Brinstar Beetom Room and the Etecoon Energy Tank Room.

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -1838,13 +1838,39 @@
                   "requires": [
                     {"enemyKill":{
                       "enemies": [
-                        [
-                          "Beetom", "Beetom", "Beetom", "Beetom"
-                        ]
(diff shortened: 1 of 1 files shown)


region/brinstar/green.json
```diff
@@ -1848,9 +1875,35 @@
               "id": 1,
               "strats": [
                 {
-                  "name": "Base",
+                  "name": "Kill the Beetoms",
                   "notable": false,
-                  "requires": []
+                  "requires": [
+                    {"enemyKill":{
```
fdtaillefer, line 1881: Do you think this needs a separate strat for a Bombs kill? It looks like you need to take a couple hits.

kjbranch: Added some bomb strats.

PR #486, opened 2023-01-05
Fix Pit Room Grey Door and Flag

kjbranch:
The pit room has two states. Note neither state is dependent on the flag f_ZebesAwake:
1. no morph/missiles: grey door on left, blue on right, no item, no enemies.
2. morph + missile capacity: grey door on left and right, item and enemies spawn, when enemies are killed the doors are openable and f_ZebesAwake is set.


region/crateria/central.json
```diff
@@ -1853,17 +1862,21 @@
...
                   "notable": false,
                   "requires": [],
-                  "note": "The enemies can be killed with just Power Beam."
+                  "note": "This door is unlocked until having morph and missiles, and then free to open after with just power beam."
                 }
               ]
             }
-          ]
+          ],
+          "yields": [ "f_ZebesAwake" ]
```
fdtaillefer, line 1879: I think we have a modeling problem here. Currently, the flag from a `yields` is obtained by interacting with a node. By chance, the left door works, but it means the right door is modeled as yielding the flag even if the lock itself isn't active.

fdtaillefer: Resolved by #487

fdtaillefer:
Moving the `yields` to locks will be simple now that it's supported. I'll merge this and then do the move real quick.

PR #498, opened 2023-01-29
Fish Tank room diagram to accompany PR#497

kjbranch:
Removed the one way arrow from 5->4, as PR#497 added strats to go from 4->5.


osse101:
I added another link 7->1 and updated the room diagram as part of my PR.

PR #506, opened 2023-02-08
Fix Crocomire Room and Post Croc Farm Room

kjbranch:
Crocomire's Room:
- Add wall jump requirements to strats
- Reduce the acid frames needed to get the item (it's still somewhat lenient)

Post Crocomire Farming Room
- Add wall jump requirements
- Split the tricky hi jump strat to its own
- Add notes to clarify some of the less obvious strats
- Add damage to the d-boost

### DIFF
--- a/region/norfair/crocomire.json
+++ b/region/norfair/crocomire.json
@@ -219,7 +219,7 @@
                   "requires": [
                     "f_DefeatedCrocomire",
                     "Gravity",
-                    {"acidFrames": 50}
+                    {"acidFrames": 30}
                   ]
                 },
                 {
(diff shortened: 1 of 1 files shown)


osse101:
Everything looks in order.
With walljump explicitly added, the farm room needs shinespark strats added for reaching 4 and technically to redraw the box around 1 indicating 1-5-4 spark.  Plus there is a crouch jump+down grab that would take another tech for starting it quickly similar to crumble jump timing.

kjbranch:
Currently my goal has been to improve the difficulty expectations for the lower difficulty settings. I have not been including shinesparks, spring ball jumps, etc. to replace a simple wall jump. I do see potential value in adding them eventually - they could be used for hacks that want to actually disable the wall jump ability. But I'm not planning on adding them at this time. 

IBJ is included from 5->1 and 5->4 in the Base strats as h_canFly.

osse101:
It's fine to make another issue of it as it is low(no) impact.  But I'm gonna treat walljump tech as any other.  
And I would like to play a hack version with wall jump disabled after thinking about these strats.

PR #507, opened 2023-02-08
Add to PCJR and Grapple Tutorial 2, 3

kjbranch:
- Clean up various strats and add wall jumps throughout.
- Add canMellaIceClip to PCJR

### DIFF
--- a/region/norfair/crocomire.json
+++ b/region/norfair/crocomire.json
@@ -1629,8 +1629,7 @@
                   "requires": [
                     "Morph",
                     "canMellaIceClip",
-                    "canWalljump",
-                    "canCrumbleSpinJump"
+                    "canWalljump"
                   ],
                   "note": [
                     "Position the frozen Mella such that you can ice clip through the crumble block, then wall jump up the long channel and mid air morph to get out."
--- a/region/norfair/crocomire.json
+++ b/region/norfair/crocomire.json
@@ -1513,16 +1513,17 @@
                 {
                   "name": "PCJR Door IBJ",
                   "notable": true,
-                  "requires": ["canIBJ"],
-                  "devNote": "Separate strat because it has to start in midair. FIXME We have a tech for this, adjust the requirements"
+                  "requires": ["canJumpIntoIBJ"]
                 },
                 {
                   "name": "Speedjump",
                   "notable": false,
                   "requires": [
                     "SpeedBooster",
-                    "HiJump"
-                  ]
+                    "HiJump",
+                    "canWalljump"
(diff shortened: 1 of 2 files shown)


region/norfair/crocomire.json
```diff
@@ -1621,6 +1622,19 @@
...
                   ]
+                },
+                {
+                  "name": "Indy Jones Mella Ice Clip",
+                  "notable": true,
+                  "requires": [
+                    "Morph",
+                    "canMellaIceClip",
+                    "canWalljump",
+                    "canCrumbleSpinJump"
```
osse101, line 1633: There are ways around the crumble jump.  Falling on the mella and jumping again, refreezing the mella above the crumble, using xray standup.  I don't think it should be a requirement.

region/norfair/crocomire.json
```diff
@@ -1621,6 +1622,19 @@
                     "Obstacle B is present here because you need access to node 2 to bring a Mella over.",
                     "Then you need to manipulate it to go high enough to be used as a stepping stone once frozen."
                   ]
+                },
+                {
+                  "name": "Indy Jones Mella Ice Clip",
```
osse101, line 1627: Mellas take a vague number of hits to position right in my experience.  There could be ways to dodge better but I don't feel that its avoidable.

tech.json
```diff
@@ -536,6 +536,13 @@
                             "canNonTrivialIceClip"
                           ],
                           "note": "Using a frozen Multiviola to clip through a one-tile-high wall."
+                        },
+                        {
+                          "name": "canMellaIceClip",
+                          "requires": [
+                            "canNonTrivialIceClip"
```
osse101, line 543: canManipulateMellas should be here as it is described.  But you may be able to dumbly position them ok.

region/norfair/crocomire.json
kjbranch: Thanks, I removed this.

tech.json
kjbranch: I don't think canManipulateMellas belongs here. That is a method for manipulating them so their peak height increases over time - used to get to the item without the ice clip or to the left door of this same room. The vertical positioning you need to freeze the Mella at is within the range they move initially. Without it is also more in line with the other ice clip techs.

We found out about this trick through sapphronxd and BuggMann: They couldn't figure out how to manipulate the Mellas, so they instead tried to see if the ice clip was possible.

osse101: That's right, the original clip kinda settles this then.

PR #511, opened 2023-02-09
Add an Obstacle to Wrecked Ship Main Shaft

kjbranch:
Convert the left bomb blocks heading to the item into an obstacle. This, as well as the bomb blocks to the right and shot blocks going down can all be destroyed by a single power bomb.


kjbranch:
Updated the Electric Death Room Diagram: Reduced the size of node 3, as it is no longer free to jump out of the water

PR #516, opened 2023-02-10
Add Crystal Flash Clip: Pink Brinstar PB (Mission Impossible)

kjbranch:


osse101:
Cool find

osse101:
I'm not sure if there's a combination of rooms that makes it possible to be a problem, but this should have a health under 50 check.  Since there is no way to damage down in room you would need to do some reverse farming.

Morph+10/10/15 is late enough to have found enemies, so I'm thinking CF clip leading into this CF.  But even then you can damage down after doing the first CF for the ones I can think of.  
Or if enough starting equipment is given and this is like first room.

PR #523, opened 2023-02-12
Add Obstacles and Strat to Red Tower

kjbranch:
- Add obstacles for the power bomb blocks and the top rippers, as there is an accessible farm which could be used to refill your used ammo.
- Fix room diagram to show these obstacles and to make Node 1 and 9 boundaries more clear to show the need for wall jumps or jump assists.


osse101 (commented):
Obstacles look like a good add.  I think there should be obstacle bypasses added to the ice and walljump strats.  Walljump must perform a couple tricks to get around them that can be better shown with an enemy/obstacle reference.  Walljumps around obstacles, quick walljumps, and a precise walljump at the end.  Hero shot/shooting block and falling doesn't seem important.

kjbranch:
The strats that dont require the obstacle (and subsequently don't clear said obstacle):
- Use Frozen Enemies
- Red Tower Walljump Climb

And those that do require the obstacle:
- Wall Jump with Space Jump
- Space Jump
- Red Tower IBJ

osse101 (dismissed):

PR #524, opened 2023-02-12
Split East Ocean Junction and Add Damage Boost Strats

### LINKED ISSUE #301: New strats to cross East Ocean right to left
See https://www.twitch.tv/videos/483749172

With two dboosts, East Ocean can be crossed right to left with nothing.

The Skultera dboost can be avoided by Space Jump or coming in charged.
The Choot dboost can be avoided by `canBombJumpBreakFree`. It is executed on the highest submerged platform and is a diagonal.

In short: the bottom junction needs to be split into bottom left and bottom right, with the softlock pit as the boundary of the bottom right junction.

kjbranch:
Split the junction to allow movement from right to left with one or two damage boosts based on the items and tech settings. This should fix #301.


region/crateria/east.json
```diff
@@ -1368,17 +1368,23 @@
         },
         {
           "id": 3,
-          "name": "East Ocean Junction (Pit)",
+          "name": "East Ocean Junction (Left side Pit)",
+          "nodeType": "junction",
+          "nodeSubType": "junction"
+        },
+        {
+          "id": 3,
```
osse101, line 1376: 4

osse101 (commented):
There is a strat for 3->1 using `canBombJumpWaterEscape` that I thought was already added.  It would be the only reason to include a shinespark from 2->3.

PR #525, opened 2023-02-12
Add canCrouchGateClip Strats to Early Supers Room

### LINKED ISSUE #302: New strat to clip through early supers gates
See https://www.youtube.com/watch?v=HMobh8J0Qog

I'm thinking `canCrouchGateClip` for the tech.

See also https://clips.twitch.tv/LittleClumsyEggSwiftRage and https://clips.twitch.tv/NiceAverageGrouseRuleFive for other ways to go through with nothing.

kjbranch:
There are two new strats: One with a damage boost and one with moon walk.
Added two new tech:
- canCrouchGateClip
- canMoonwalk

This should fix #302

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -1093,8 +1093,10 @@
                     }
                   ],
                   "note": [
-                    "Run then do a small hop and aim down to barely clip into the first gate.",
-                    "Moonwalk backwards while inside the first gate to get the maximum run distance and repeat for the next gate."
+                    "Run then do a very small spin jump to clip into the left side of the first gate.",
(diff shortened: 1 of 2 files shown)


osse101 (approved):
There is a failure case of falling into 5 after succeeding with the first gate clip but that doesn't change much.

kjbranch:
Yeah, I didn't see a difference between Nodes 3 and 5, except that Node 3 has an item. Although if that item would help you get out of 3, you would grab it first anyways. (1 PB can get you all the way from 5 -> 3 -> 1.)

PR #528, opened 2023-02-13
Add Two Crystal Flash Clip Tech and Their Strats

### LINKED ISSUE #355: Missing: CF clip in Botwoon Hallway
Saw a runner use crystal flash clip in a race to get through the Botwoon Hallway so it seems like something to be included.

kjbranch:
canCrystalFlashClip: Can be used to clip into non-respawning crumble blocks and break them.
- East Pants Room CF Clip (with gravity and bombs, and a suitless version)
- Botwoon CF Clip (both directions, with gravity and bombs, and a suitless version)

canCrystalFlashGrappleClip: Can be used to clip through solid blocks.
- Mickey Mouse Crystal Flash Grapple Clip
- Reverse Acid Statue CF Grapple Clip
- Bowling Reserve Crystal Flash Grapple Clip

This should resolve #355


tech.json
```diff
@@ -783,7 +783,28 @@
...
-              "note": "Using a Crystal Flash to take advantage of the fact it forces Samus to stand up."
+              "note": [
+                "Using a Crystal Flash to take advantage of the fact it forces Samus to stand up.",
+                "This can be used while in a morph tunnel to jump through single tile thick ceilings."
+              ]
+            },
+            {
+              "name": "canCrystalFlashClip",
+              "requires": ["canCrystalFlash"],
+              "note": "Using a Crystal Flash below a non-respawning crumble block to clip into the block and break it.",
```
osse101, line 794: Hold down to break the crumble.

osse101 (commented):
Gravity+bombs is a 2 frame window trick and a must know for competetive rando.
Water physics CF is 1 frame and last resort trick.
Grapple clips should be 1 frame and is only different from water physics by moving faster probably.
But its all expert tech so maybe notable covers it

PR #531, opened 2023-02-14
Add Taco Tank

kjbranch:


osse101:
canStationarySpinJump as well

And a description as well.  But if you are looking at putting it in logic you have to well know how its done 🤣 
Triple frame perfect dashing stationary spinjump into delayed CWJ and precise hitbox manip.

PR #533, opened 2023-02-14
Add Waterway Come In Charged

kjbranch:
An alternate strat without wall jump requirements

### DIFF
--- a/region/brinstar/pink.json
+++ b/region/brinstar/pink.json
@@ -2431,7 +2431,7 @@
                 },
                 {
                   "name": "Shinespark Come in Charged",
-                  "notable": true,
+                  "notable": false,
                   "requires": [
                     {"canComeInCharged": {
                       "fromNode": 1,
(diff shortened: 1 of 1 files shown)


region/brinstar/pink.json
```diff
@@ -2428,6 +2428,18 @@
...
                   ],
                   "note": "Doable without a short charge, since there are 32 tiles (plus one open end) to charge it."
+                },
+                {
+                  "name": "Shinespark Come in Charged",
+                  "notable": false,
+                  "requires": [
+                    {"canComeInCharged": {
+                      "fromNode": 1,
+                      "framesRemaining": 5,
```
osse101, line 2438: Looks like around 45frames to me.  You could pretend 5frames reaches the raised runway to start another spark 🏇

kjbranch: It felt small, so I put in a small number. Didn't realize how far off it was. Now I know how to find the values. I could do it in 27 with a little run speed through the door, so I put 40.

PR #536, opened 2023-02-15
Add canConsecutiveWalljump throughout Zebes

kjbranch:
- Update the description to be anywhere that requires 3+ consecutive wall jumps.
- It is now required for canSunkenDualWallClimb and canInsaneWalljump.

### DIFF
--- a/region/crateria/central.json
+++ b/region/crateria/central.json
@@ -573,7 +573,13 @@
                 {
                   "name": "Gauntlet Walljumps",
                   "notable": true,
-                  "requires": ["canDelayedWalljump"]
+                  "requires": [
+                    "canDelayedWalljump",
+                    {"or": [
+                      "HiJump",
+                      "canConsecutiveWalljump"
+                    ]}
+                  ]
                 }
               ]
             }
@@ -1190,7 +1196,11 @@
                   "notable": true,
                   "requires": [
                     "Morph",
-                    "canPreciseWalljump"
+                    "canPreciseWalljump",
+                    {"or": [
+                      "HiJump",
+                      "canConsecutiveWalljump"
+                    ]}
                   ]
                 },
                 {
--- a/region/maridia/inner-yellow.json
+++ b/region/maridia/inner-yellow.json
@@ -879,8 +879,19 @@
                 {
                   "name": "Base",
                   "notable": false,
-                  "requires": [ "canConsecutiveWalljump" ],
-                  "devNote": "This should require something more to kill the Choot."
+                  "requires": [
+                    "canConsecutiveWalljump",
+                    {"or": [
+                      "canPreciseWalljump",
+                      "ScrewAttack",
+                      "h_canUsePowerBombs",
+                      {"enemyDamage": {
+                        "enemy": "Choot",
+                        "type": "contact",
+                        "hits": 1
+                      }}
+                    ]}
+                  ]
                 }
               ]
             }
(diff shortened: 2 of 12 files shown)


region/maridia/inner-yellow.json
```diff
@@ -879,7 +879,8 @@
                 {
                   "name": "Base",
                   "notable": false,
-                  "requires": []
+                  "requires": [ "canConsecutiveWalljump" ],
+                  "devNote": "This should require something more to kill the Choot."
```
osse101, line 883: I would say `canPreciseWallJump` or `Screw Attack` or `Power Bomb` or `enemy hit`
There is a fair sized safe spot on the ledge, and you can continue walljumping when it floats towards that side.
Wave is too gimicky with the camera manip you have to do.
Firing a strong weapon just before landing would be a precise jump and some kind of quickFire tech

region/maridia/inner-yellow.json
```diff
@@ -1704,6 +1706,10 @@
                   "name": "Kill Rippers",
                   "notable": false,
                   "requires": [
+                    {"or": [
+                      "HiJump",
```
osse101, line 1710: HiJump+Walljump ?

region/norfair/crocomire.json
```diff
@@ -1664,7 +1666,10 @@
                   "requires": [
                     "Morph",
                     "canMellaIceClip",
-                    "canWalljump"
+                    {"or":[
+                      "HiJump",
```
osse101, line 1670: HiJump+walljump ?

region/norfair/crocomire.json
```diff
@@ -1624,7 +1624,8 @@
                   "requires": [
                     "canUseFrozenEnemies",
                     "canManipulateMellas",
-                    "HiJump"
+                    "HiJump",
+                    "canConsecutiveWalljump"
```
osse101, line 1628: This only takes 1 walljump, might even be able to do it with none.

tech.json
```diff
@@ -320,7 +320,10 @@
                   "extensionTechs": [
                     {
                       "name": "canInsaneWalljump",
-                      "requires": [ "canDelayedWalljump" ],
+                      "requires": [
+                        "canDelayedWalljump",
+                        "canConsecutiveWalljump"
```
osse101, line 325: `canInsaneWallJump` is better left describing the jump itself.  Many cases are 2 jumps total, taco tank is 1 walljump.

tech.json
```diff
@@ -340,7 +343,7 @@
               "requires": [
                 "canWalljump"
               ],
-              "note" : "Climbing a wall with several consecutive walljumps.",
+              "note" : "Climbing a wall with three or more consecutive walljumps, where a single failure will require restarting.",
```
osse101, line 346: Failure is a strange descriptor since missing a walljump means you fall a bit and try again a little lower.  Accidently breaking spin could be the failure at this skill level; an enemy hit doesn't seem meant to be included in this meaning.

tech.json
kjbranch: I was planning on changing this to canStaggeredWalljump once that tech is merged in. I figure it would take a couple simple wall  jumps to position Samus. Taco Tank and Screw CWJ  don't need that though ....

region/norfair/crocomire.json
kjbranch: I think it should take 2 wall jumps to get onto the Mella and then one more to get from the Mella to the item.
https://youtu.be/a02RaagZqjo

region/crateria/central.json
```diff
@@ -573,7 +573,13 @@
                 {
                   "name": "Gauntlet Walljumps",
                   "notable": true,
-                  "requires": ["canDelayedWalljump"]
+                  "requires": [
+                    "canDelayedWalljump",
+                    {"or": [
+                      "HiJump",
```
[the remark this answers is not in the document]
kjbranch: Although the Delayed Walljumps are a little tighter, only having to do one is much easier for a beginner. Varia Randomizer classifies this strat as Very Hard with and Hardcore without Hi Jump: https://randommetroidsolver.pythonanywhere.com/presets

region/norfair/crocomire.json
osse101: You linked the no HiJump version.
With hijump its jump up to the bug then jump into walljump.

kjbranch:
I think I've addressed all of your concerns here. 
- I will update some of them to canStaggeredWalljumps once #523 is merged.
- I will add these tech to WRITG type ledges in another pull request
- I will add these tech to Lava Dive strats after #529 is merged

region/norfair/crocomire.json
kjbranch: Well, that's embarrassing.

PR #539, opened 2023-02-16
Add Etecoon E Tank Morphless Tunnel Strats

### LINKED ISSUE #503: New Strat - Eticoon(Etank) Spikeway Tunnel Crawl
The morph tunnel across from the eticoon save room is 2 tiles high seeing as how you can stand inside of spikes.  
From the left, starting the crawl is kinda hard.  You need to land aiming down ~3 pixels inside the leftmost tile.  To far left and you fall out, too far right and a spike pushes you out, not aiming down puts you stuck in crouch.  Next step is spinjump right to take a spike hit and get wedged in the ceiling.  Then start the crawl while taking constant spike damage.  About 450 damage rounded up to 500 should do it.
From the right, getting the floor stuck setup is less precise but same idea.  About the same energy cost.

Also you can TurnAroundCancel through but that costs 700hp with both suits and adds a canBePatient.  Its still useful as a backup to messing up the tunnel crawl.

kjbranch:
- Add notable morphless tunnel crawl strats both directions
- Add a notable turn around aim cancel strat from left to right that requires canBePatient
- Add a node to the right of the tunnel

This should resolve #503


region/brinstar/green.json
```diff
@@ -2246,13 +2253,54 @@
...
+                  "notable": true,
+                  "requires": [
+                    "canMorphlessTunnelCrawl",
+                    "canTrickyJump",
+                    {"thornHits": 31}
+                  ],
+                  "note": [
+                    "Moving left to right is a bit more tricky to get into the tunnel.",
+                    "Land aiming down approximately 3 pixels inside the leftmost tile.",
+                    "If Samus is too far left she will fall out, too far right and a spike pushes you out, not aiming down puts her stuck in crouch.",
```
osse101, line 2276: There's a downback setup I like for both directions, but maybe that would be separate reference material for this strat.  Like Xray helping with the setup.

region/brinstar/green.json
```diff
@@ -2291,27 +2339,24 @@
...
-                      "hits": 15
+                      "hits": 25
                     }}
                   ]
                 },
                 {
                   "name": "Dodge the Beetoms",
                   "notable": true,
                   "requires": [
-                    "Morph",
```
osse101, line 2314: I'm not sure how to dodge them even with morph.

region/brinstar/green.json
```diff
@@ -2291,27 +2339,24 @@
...
                 },
                 {
                   "name": "Tank the Damage",
                   "notable": false,
                   "requires": [
-                    "Morph",
                     {"enemyDamage": {
                       "enemy": "Beetom",
                       "type": "contact",
-                      "hits": 15
```
osse101, line 2306: Maybe needs another strat for spike iframing the beetoms. 3 spike hits to cross, maybe one more for safety.
There's also some fuzziness you can do by shaking off the beetoms and running to a farm point.  But that doesn't seem worthwhile enough to add.

region/brinstar/green.json
osse101: Also I forsee notable strats having a required note, or a uniquetech that itself has a note.

osse101 (commented):
There's a downback setup for each tunnelcrawl that helps a lot, or using xray.  But using them would add to the tech requirements when you can do it without.

region/brinstar/green.json
kjbranch: If you want to describe them both, I can add them to the note

region/brinstar/green.json
kjbranch: That's a good call. I dont think unique tech should be notable, since the tech could perfectly describe the strat in that instance. 

It is fairly important that there is a way to get from 4->6 with nothing except that tech, otherwise another node will need to be added - 6 but with no Beetoms on Samus. This strat is a lot easier with the i frames, and because of the farm, it should be equivalent to just the tech. I will add a description of how to do it.

region/brinstar/green.json
osse101: L->R
Fall off with low speed and start a downback right away.  (Doesn't always work.)

R->L
Stand fully on the left side of the nearby ledge and run off then start and hold a downback to land in position.

PR #554, opened 2023-02-20
Add canStaggeredWalljump

kjbranch:
- Lava Dive
- Maridia Elevator

### DIFF
--- a/region/norfair/east.json
+++ b/region/norfair/east.json
@@ -5179,11 +5179,12 @@
                     "canSuitlessLavaDive",
                     "canStaggeredWalljump",
                     "canBounceBall",
+                    "canUseEnemies",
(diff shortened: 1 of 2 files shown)


region/norfair/east.json
```diff
@@ -5213,7 +5213,7 @@
                     "h_canNavigateHeatRooms",
                     "canSuitlessLavaDive",
                     "canIframeSpikeJump",
-                    "canWalljump",
+                    "canStaggeredWalljump",
```
osse101, line 5216: HJBless dive should have `canFastWalljumpClimb` for the spike climb, and `canStaggeredWallJump` for the right side climb.
`canUseEnemies` should also be included in all of these stand on the Namihe strats too.

PR #561, opened 2023-02-21
Add canPrepareForNextRoom to Enemy Hit on Entry

kjbranch:
- Pink Hopper Room
- Morph Ball Room Hoppers
- Blue Hopper Room
- Mickey Mouse Room

### DIFF
--- a/region/brinstar/pink.json
+++ b/region/brinstar/pink.json
@@ -1770,6 +1770,25 @@
                   "note": "This assumes you don't know you're entering this room, and take a Small Sidehopper hit quickly."
                 },
                 {
+                  "name": "Kill on Entry",
+                  "notable": false,
+                  "requires": [ "canPrepareForNextRoom" ],
+                  "obstacles": [
+                    {
+                      "id": "A",
+                      "requires": [
+                        "ScrewAttack",
+                        {"canComeInCharged": {
+                          "fromNode": 1,
+                          "framesRemaining": 0,
+                          "shinesparkFrames": 24
+                        }}
+                      ]
+                    }
+                  ],
+                  "note": "Although there are other ways to kill them on entry, these are done reliably."
+                },
+                {
                   "name": "Pink Brin Reverse GGG",
                   "notable": true,
                   "requires": [],
--- a/region/brinstar/pink.json
+++ b/region/brinstar/pink.json
@@ -1777,12 +1777,14 @@
                     {
                       "id": "A",
                       "requires": [
-                        "ScrewAttack",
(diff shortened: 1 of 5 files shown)


region/brinstar/pink.json
```diff
@@ -1769,6 +1769,27 @@
...
+                  "requires": [ "canPrepareForNextRoom" ],
+                  "obstacles": [
+                    {
+                      "id": "A",
+                      "requires": [
+                        {"or":[
+                          "ScrewAttack",
+                          {"canComeInCharged": {
+                            "fromNode": 1,
+                            "framesRemaining": 0,
```
osse101, line 1784: Running in while blue would also work

region/lowernorfair/east.json
```diff
@@ -4378,6 +4378,30 @@
...
+                      "requires": [
+                        {"or": [
+                          {"and": [
+                            "Wave",
+                            "Plasma"
+                          ]},
+                          "ScrewAttack",
+                          {"canComeInCharged": {
+                            "fromNode": 1,
+                            "framesRemaining": 0,
```
osse101, line 4397: `canComeInCharged: framesRemaining: 180` as well

region/tourian/main.json
```diff
@@ -1802,12 +1802,21 @@
                 {
                   "name": "Avoid Damage",
                   "notable": false,
-                  "requires": ["ScrewAttack"]
+                  "requires": [
```
osse101, line 1805: Its safe to enter not in screw, as most speedruns do.  Then use screw after

tech.json
```diff
@@ -13,7 +13,10 @@
         {
           "name": "canPrepareForNextRoom",
           "requires": [],
-          "note": "The ability to perform an action as part of a strat that occurs following a room transition."
+          "note": [
+            "The ability to know what room is coming next and perform an action immediately on entry.",
```
osse101, line 17: The action begins before entry.  Its taken as part of entering the room, I'd say.

region/tourian/main.json
kjbranch: From the left door?

osse101: 1->2 is the vertical door entry.
For the 2 door, you may have room to land first then jump.  Jumping straight in seems like you would land on the hopper to me.

kjbranch: Looks like the Tourian nodes are numbered differently than most rooms. Glad you caught that. It might be worth putting them into the same order as other regions someday ...

PR #565, opened 2023-02-22
Add obstaclesCleared requirement

blkerby:
Sometimes we want to be able to represent a requirement for an obstacle to already be cleared. In strats we have been doing this by having an "obstacles" field with `requires: ["never"]`, but this doesn't work in some other places where logical requirements may appear (such as in "farmCycles"). By adding obstaclesCleared as a first-class requirement type, we can use it anywhere.


osse101:
This does appear to improve clarity in intent of situations it describes.

PR #568, opened 2023-02-23
Add tricky strat to Kraid's Beetoms

kjbranch:
100 energy without canTrickyJump, 70 with.

### DIFF
--- a/region/brinstar/kraid.json
+++ b/region/brinstar/kraid.json
@@ -536,7 +536,19 @@
                     {"enemyDamage": {
                       "enemy": "Beetom",
                       "type": "contact",
-                      "hits": 5
+                      "hits": 4
+                    }}
+                  ]
+                },
+                {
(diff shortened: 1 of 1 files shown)


region/brinstar/kraid.json
```diff
@@ -536,7 +536,19 @@
                     {"enemyDamage": {
                       "enemy": "Beetom",
                       "type": "contact",
-                      "hits": 5
+                      "hits": 4
+                    }}
+                  ]
+                },
+                {
+                  "name": "Tricky Evade",
```
osse101, line 544: I imagine there is a particular way to evade and I'd like that as a note.

osse101: Looks like you wait by the door briefly for 7 hit.
Using the speedrun item grab I could get 5 hit but that takes run speed entering

PR #578, opened 2023-02-27
Add Several canCrossRoomWaterJump and overhaul East Cac Alley

kjbranch:
- Add node in East Cactus Alley Room
- Add 15 of the first 17 strats here (skipped both Crab Hole strats for now): https://vimeo.com/showcase/10188956


region/maridia/inner-pink.json
```diff
@@ -183,6 +183,61 @@
...
                   ]
+                },
+                {
+                  "name": "Cross Room Jump with HiJump",
+                  "notable": false,
+                  "requires": [
+                    "HiJump",
+                    "canCrossRoomJumpIntoWater",
+                    "canMomentumConservingTurnaround",
+                    "canDownGrab",
```
osse101, line 194: Does not need a downgrab

region/maridia/inner-pink.json
```diff
@@ -183,6 +183,61 @@
...
+                    {"adjacentRunway": {
+                      "fromNode": 1,
+                      "usedTiles": 13,
+                      "inRoomPath": [ 1, 3 ],
+                      "physics": ["normal"]
+                    }}
+                  ],
+                  "note": "Seems to require at least 13 tiles in the adjacent room."
```
osse101, line 219: Remove "Seems to"

region/maridia/inner-pink.json
```diff
@@ -1732,6 +1787,39 @@
...
+                  "note": "Needs max horizontal speed coming from 44 runway tiles in the adjacent room."
+                },
+                {
+                  "name": "Cross Room Jump with Speedbooster, Spring Ball",
+                  "notable": false,
+                  "requires": [
+                    "Morph",
+                    "canCrossRoomJumpIntoWater",
+                    "SpeedBooster",
+                    "SpringBall",
```
osse101, line 1814: `canSpringBallJump` 
and the tech needs a different name 🐸 Maybe springball bounce
Or to not be a tech

region/maridia/inner-pink.json
```diff
@@ -2562,6 +2665,72 @@
                     "canSuitlessMaridia",
                     "Grapple"
                   ]
+                },
+                {
+                  "name": "Cross Room Jump with IBJ",
+                  "notable": false,
+                  "requires": [
+                    "canIBJ",
```
osse101, line 2673: There's some variance on if the mochtroid stays close enough to do damage, possibly tied to how well you do the jump or turnaround.  2 hits if you do get grabbed and then your bombs kill it.

region/maridia/inner-pink.json
```diff
@@ -2562,6 +2665,72 @@
...
+                  ],
+                  "note": "Only requires a runway of approximately one tile in the adjacent room."
+                },
+                {
+                  "name": "Cross Room Jump with HiJump",
+                  "notable": false,
+                  "requires": [
+                    "HiJump",
+                    "canCrossRoomJumpIntoWater",
+                    "canMomentumConservingTurnaround",
```
osse101, line 2691: This does downgrab.  And you have to be right on the very edge for 1 tile.

region/maridia/inner-pink.json
```diff
@@ -3277,6 +3446,26 @@
...
+              "id": 2,
+              "strats": [
+                {
+                  "name": "Cross Room Jump",
+                  "notable": false,
+                  "requires": [
+                    "canCrossRoomJumpIntoWater",
+                    {"adjacentRunway": {
+                      "fromNode": 1,
+                      "usedTiles": 2.5,
```
osse101, line 3459: 2 tiles with a downgrab.  1.5 with an airball

region/maridia/inner-pink.json
```diff
@@ -3755,33 +3950,109 @@
...
+                      "inRoomPath": [ 2, 3 ],
+                      "physics": ["normal"]
+                    }}
+                  ],
+                  "note": "Seems to require a runway of at least 2 tiles in the adjacent room."
+                },
+                {
+                  "name": "Cross Room Jump with Spring Ball and Bomb Boost",
+                  "notable": true,
+                  "requires": [
```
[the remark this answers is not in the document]
kjbranch: I'm thinking of holding off for a few days on the `springballBombJump` if that's ok. It seems like a common tech, so I don't really just want to include it in one location. I'll make a pass of a lot of places soon.

Is a single horizontal bomb boost enough, or does it require an HBJ? I cant seem to get it, but it looks close if I unmorphed at the right time.

osse101 (commented):
It may be possible to trade half a run tile for `canDownGrab`.  At this skill level `canDownGrab` should be expected.

region/maridia/inner-pink.json
```diff
@@ -3860,6 +4149,69 @@
...
+                {
+                  "name": "Base",
+                  "notable": false,
+                  "requires": [
+                    "Gravity",
+                    "Morph",
+                    {"or":[
+                      "SpaceJump",
+                      "canCarefulJump",
+                      {"spikeHits": 1}
```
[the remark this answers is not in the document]
kjbranch: I've added careful jump here with gravity turned off. I think luring a cac thorn over here would be a bit unintuitive, so I think the careful jump should be enough.

region/maridia/inner-pink.json
kjbranch: I've been wanting to rename that one to springball bounce for a while. It's a tech in that it's used with speed, so you kind of have to know how to use it. But I think `canSpringBallJumpMidAir` would be better named `canSpringBallJump`

PR #579, opened 2023-02-28
Fix bug in Etecoon E Tank Room - Beetom ice clip

kjbranch:


region/brinstar/green.json
```diff
@@ -2336,8 +2336,12 @@
               "strats": [
                 {
                   "name": "Etecoon E-Tank Beetom Clip",
-                  "notable": true,
-                  "requires": ["canBeetomIceClip"]
+                  "notable": false,
+                  "requires": [
```
osse101, line 2340: requires taking damage.  2 hits to make an attempt and 1 hit every other try.
But you can leave the beetom here and farm, so 2 hits total?
And it is possible to get it stuck in the wall at the right height, leave to farm, and refreeze in the right spot to leave with full health.  But that's a bit gimicky.

PR #588, opened 2023-03-02
Cross room jump physics and notes changes

kjbranch:
- Fix new coss room jumps to require air physics in Outer Maridia, since they were added at the same time as the physics changes
- Clarify how precise tricky cross room strats are


osse101:
👍

PR #594, opened 2023-03-06
HiJumpless Hotarubi Special

kjbranch:
https://clips.twitch.tv/ClearSavageWolfDoritosChip-il6VkH1_G9BA_P_0


osse101:
Could you change this to `notable:false`.  And heat frames to 250.
The draygon doesn't get in the way if you happen to do this strat first.

kjbranch:
I'm thinking it should still be notable. It's fairly tricky and obscure.

PR #601, opened 2023-03-07
Wrecked Ship - remove notable strats

kjbranch:


osse101:
👍 
I like the clarification on sponge bath being a tricky dash jump.  We just happened to find a spot that worked at the time.

PR #602, opened 2023-03-07
Fix bomb blocks in Climb

kjbranch:


region/crateria/central.json
```diff
@@ -1730,9 +1701,38 @@
...
+                },
+                {
+                  "name": "Come in Charged",
+                  "notable": false,
+                  "requires": [],
+                  "obstacles": [
+                    {
+                      "id": "A",
+                      "requires": [
+                        {"canComeInCharged": {
```
osse101, line 1716: There should be a `canComeInCharged` with `framesRemaining: 180` and  `shinesparkFrames: 0` as well, on both directions.

PR #603, opened 2023-03-08
Remove some notable strats in Brinstar

kjbranch:
- Also modernize some of the code to use correct helpers and obstaclesCleared

### DIFF
--- a/region/brinstar/blue.json
+++ b/region/brinstar/blue.json
@@ -41,7 +41,7 @@
                   "name": "Base",
                   "notable": false,
                   "requires": [ {"obstaclesCleared": ["C"]} ],
-                  "devNote": "Enabled by destroying obstacle C by travelling between nodes 1 and 6."
+                  "devNote": "Destroy obstacle C by travelling between nodes 1 and 6."
                 }
               ]
             }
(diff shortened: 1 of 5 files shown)


region/brinstar/blue.json
```diff
@@ -1223,16 +1179,13 @@
...
                     "Morph",
-                      {"or": [
-                        "canConsecutiveWalljump",
-                        "h_canFly"
-                      ]}
+                    {"or": [
+                      "canConsecutiveWalljump",
+                      "h_canFly"
+                    ]},
+                    {"obstaclesCleared": ["B"]}
```
osse101, line 1186: This scenario already has "A" cleared so I'm wondering if `obstaclesCleared: A` shows intent better.  In the fantasy world of crumble broken and bomb block not, you could spark up for instance.
And there may be room for an ice clip strat.
GMode would fit the need for `h_canDestroyBombWalls`

Doesn't affect things either way as things are 🤷

osse101 (commented):
Looks good
There are some strats that will need a revisit like boulderroom entry and dechora escape.  I'll look through everything someday.

"redundant requirements" reads like requirements were listed in Obstacle requirements, but they go in links-strats-requirements now.

PR #604, opened 2023-03-09
Remove some notable strats in Norfair

kjbranch:


region/norfair/west.json
```diff
@@ -3164,7 +3152,7 @@
                 },
                 {
                   "name": "Croc Escape HiJump Freeze",
-                  "notable": true,
+                  "notable": false,
```
osse101, line 3155: This one needs a note.  Just a simple lure the Geruta left by jumping towards it from below, then freeze

osse101 (commented):
The xray clip in HiJump etank room should have a tech made for the motion.  Just not quite sure what to call it yet.

`canShotBlockOverload` may see changes for GMode support.  Or it may become an extension tech of `canOverloadPLMs`

kjbranch:
> The xray clip in HiJump etank room should have a tech made for the motion. Just not quite sure what to call it yet.

I am feeling like `canXRayStandUp` and `canCeilingClip` cover it pretty well. Unless I'm missing something?

PR #605, opened 2023-03-09
Remove some notable strats in Tourian

kjbranch:
- Redesign some Metroid Avoid PB Kill strats

### DIFF
--- a/region/tourian/main.json
+++ b/region/tourian/main.json
@@ -250,13 +250,7 @@
                 {
                   "name": "Base",
                   "notable": false,
-                  "requires": [],
-                  "obstacles": [
-                    {
-                      "id": "A",
-                      "requires": ["never"]
-                    }
-                  ]
+                  "requires": [ {"obstaclesCleared": ["A"]} ]
                 }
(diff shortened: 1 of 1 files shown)


osse101 (commented):
`enemyKill` seems like it fits better in `obstacles: requires`.  Except where the weapon is also used to keep rinkas away.

--

These are some strats. 🤣 Gotta try out damageless 3 PBs only.  It has an implied rinka tag to gather the metroids but dodging rinkas and metroids doesn't sound easy.  And I'd rather it swap `excludeWeapons` to `explicitWeapons`.

Baby skip needs more variations added.  Minimalist is just `Morph` for items, Classic skip would include speed I'd think.  Wave/no wave just depends on how long you want to jump rope the baby.  Springball skip and spacejump skip wont be added as they just have a higher requirement count, like the note says.

`f_KilledMetroidRoomX` should technically go on the obstacle, no?  You don't need to shoot the door like for `f_ZebesAwake`.

osse101:
Seems fine

PR #606, opened 2023-03-09
Add missing free strat between Aqueduct items

blkerby:
This matters because an X-ray climb can get you to node 8 (the right item) and with that you should also be able to reach node 7 (the left item).


osse101:
Righto, the room diagram already has the connection too.

PR #612, opened 2023-03-11
Update Tricky Ice Techs

kjbranch:
- Clarify that using a super to knock off a wall crawler to freeze is assumed in `canTrickyUseFrozenEnemies`
- Add Morph to x-ray stand up ice clips


osse101:
👍

PR #613, opened 2023-03-11
Rename and remove some notables

kjbranch:
- Naming is designed to be clear what the strat is with only the strat name - the room should be implied

### DIFF
--- a/region/brinstar/pink.json
+++ b/region/brinstar/pink.json
@@ -1052,7 +1052,7 @@
                   "devNote": "This base strat represents only backtracking after entering from 7."
                 },
                 {
-                  "name": "Big Pink Super Reach Around",
+                  "name": "Big Pink Off Screen Super Block",
                   "notable": true,
                   "requires": ["Morph"],
                   "obstacles": [
--- a/region/brinstar/pink.json
+++ b/region/brinstar/pink.json
@@ -1052,7 +1052,7 @@
                   "devNote": "This base strat represents only backtracking after entering from 7."
                 },
                 {
-                  "name": "Spore Spawn Skip",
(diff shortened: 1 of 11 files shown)


region/norfair/crocomire.json
```diff
@@ -1294,7 +1294,7 @@
               "id": 4,
               "strats": [
                 {
-                  "name": "PCJR Bluespace",
+                  "name": "PCJR Bluespace Speed Blocks from the Left",
```
osse101, line 1297: I think the alt name is `Space Boost`, but I I'd like to see something like `PCJR Full Room Space Jump Through Speed Blocks`.
Its also not 100% clear if this is from the door or the acid from the name as `BlueSpaceJump` does work for both.

region/brinstar/pink.json
```diff
@@ -1066,7 +1066,7 @@
                   ]
                 },
                 {
-                  "name": "CF Spore Spawn Skip",
+                  "name": "Big Pink Crystal Flash Standup",
```
osse101, line 1069: Needs to clarify you are shooting the super block
So `Crystal Flash Spore Spawn Skip` .  Standup is just a part of CFing.

region/brinstar/pink.json
```diff
@@ -1052,7 +1052,7 @@
                   "devNote": "This base strat represents only backtracking after entering from 7."
                 },
                 {
-                  "name": "Spore Spawn Skip",
+                  "name": "Big Pink Super Reach Around",
```
osse101, line 1055: `canSuperReachAround` is a bad name.  `canOffScreenSuperShot` fits better.

PR #616, opened 2023-03-12
Add reusableRoomwideNotable and example

kjbranch:
Add Bubble Mountain Damage Boost example

### DIFF
--- a/region/norfair/east.json
+++ b/region/norfair/east.json
@@ -1974,7 +1974,7 @@
       "reusableRoomwideNotable": [
         {
           "name": "Bubble Mountain Damage Boost",
-          "note": "Damage Boost off a Waver to cross the gap."
+          "note": "Crossing between the topmost doors of Bubble Mountain by damage boosting using a Waver."
         }
       ],
       "links": [
--- a/region/norfair/east.json
+++ b/region/norfair/east.json
@@ -1971,6 +1971,12 @@
           "homeNodes": [ 8 ]
         }
       ],
+      "reusableRoomwideNotable": [
+        {
+          "name": "Bubble Mountain Damage Boost",
+          "note": "Damage Boost off a Waver to cross the gap."
+        }
+      ],
       "links": [
         {
           "from": 1,
@@ -2000,6 +2006,7 @@
                       "type": "contact"
                     }}
                   ],
+                  "reusableRoomwideNotable": "Bubble Mountain Damage Boost",
                   "note": "Damage Boost off a Waver to cross the gap."
                 },
                 {
@@ -2247,6 +2254,7 @@
                       "type": "contact"
                     }}
                   ],
+                  "reusableRoomwideNotable": "Bubble Mountain Damage Boost",
                   "note": "Damage Boost off a Waver to cross the gap."
                 }
               ],
--- a/schema/m3-region.schema.json
+++ b/schema/m3-region.schema.json
@@ -1359,6 +1359,45 @@
                 }
               }
             }
+          },
+          "reusableRoomwideNotable": {
+            "$id": "#/properties/rooms/items/properties/reusableRoomwideNotable",
+            "type": "array",
+            "title": "Very similar notable strats that exist within multiple links within a single room.",
+            "items": {
+              "$id": "#/properties/rooms/items/properties/reusableRoomwideNotable/items",
+              "type": "reusableRoomwideNotable",
+              "title": "Reusable Roomwide Notable Strat",
+              "required": [
+                "name",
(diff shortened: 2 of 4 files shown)


strats.md
```diff
@@ -35,6 +26,20 @@ A `strat` can have the following properties:
     * _bypass:_ Some [logical requirements](logicalRequirements.md) that can be fulfilled to bypass the obstacle, if it isn't already destroyed. Voids both the `requires` property and the requirements tied to the `obstacle`'s definition within the room. Naturally, this does not destroy the obstacle.
     * _additionalObstacles:_ An array containing the ID of additional obstacles that may not need to be destroyed to execute the strat, but that will be destroyed by destroying the containing `obstacle` via this `strat`. Additional obstacles are _not_ destroyed when bypassing the main obstacle.
   * _stratProperties:_ An array of keywords, which can be used as a requirement for strats in the destination node. These properties can be used to describe miscellaneous details (such as being in a spinjump) that will impact whether it's possible to follow-up with another specific strat.
+  * _reusableRoomwideNotable:_ The name of the reusable roomwide notable strat. This must share an identical name with the `reusableRoomwideNotable` and the other strats that are connected to this one. This is only applicable for strats where `notable` is `true`.
+
+## Notable Strats
```
osse101, line 31: This looks like it was ordered following the structure, so
```
name
notable
requires
obstacles
reusableRoomwideNotable
Structure
```

schema/m3-region.schema.json
```diff
@@ -1359,6 +1359,45 @@
...
+                  "title": "Shared Reusable Roomwide Strat Name",
+                  "default": "",
+                  "examples": [
+                    "Bubble Mountain Damage Boost"
+                  ],
+                  "pattern": "^(.*)$"
+                },
+                "note": {
+                  "$id": "#/properties/rooms/items/properties/reusableRoomwideNotable/items/properties/note",
+                  "type": "string",
```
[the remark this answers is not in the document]
osse101: On the one hand `note` will be used just like all other notes so this could just be a `$ref`.
On the other, this has a clearer purpose in describing the connected strat.

region/norfair/east.json
```diff
@@ -1971,6 +1971,12 @@
           "homeNodes": [ 8 ]
         }
       ],
+      "reusableRoomwideNotable": [
+        {
+          "name": "Bubble Mountain Damage Boost",
+          "note": "Damage Boost off a Waver to cross the gap."
```
osse101, line 1977: Raises the question of how identical `note`s are to be handled.  Notable strats shouldn't truely be identical (maybe).  A LtoR setup here would wait a while for a waver to come back to the door to be in a good position.  A RtoL setup either rushes to meet the waver, or waits for it to bounce off the right side wall and come at Samus from behind.

This note would be the user facing description.
And the full strat description will have a note that only appears in the spoiler log.
That being the case it seems fine for notes to be the same.

This description sounds pretty boring though so maybe this dboost shouldn't be notable.  The action required is just to wait for the waver to be overhead and jump into it and damage boost.

osse101 (commented):
I'll be thinking of these as display names.

region/norfair/east.json
kjbranch: The `note` in the `reusableRoomwideNotable` will show up when selecting which notable strats the player is interested in being in logic, so it will be more generic and simplified. The `note`s in the strats themselves can be more complex and describe the nuances of the particular strat. (All notable strat `note`s need to be worked on soon.)

As for this strat in particular. I kept it in here because it has an unforgiving fall below it. Similar to the PCJR Grapple, Early Supers Mockball, etc

osse101: I'm thinking there is some need to describe the situation as we're detached from room nodes.  Wording should be more proper.  What makes this strat notable should be mentioned in one of the notes, but could be on either.

"Crossing between the topmost doors of Bubble Mountain by damage boosting using a Waver."
"Using a Waver damage boost to propel Samus across the top of the Bubble Mountain room."
"Avoid falling into the center of Bubble Mountain by using a damage boost to directly cross the top"

osse101 (commented):
`reusableRoomwideNotable` needs to be added to the strat Schema too.

region/norfair/east.json
kjbranch: I think this is a good suggestion to do eventually. Most of the notable strats notes could use some work, as well as many of the other notes. I've edited the Bubble Mountain note, but I'm thinking this will be a project for a future pass through.

osse101:
👍

PR #618, opened 2023-03-13
A few more changes to notable strats

kjbranch:


region/lowernorfair/east.json
```diff
@@ -1535,6 +1535,12 @@
           "note": "Technically you can get the drops but you likely won't"
         }
       ],
+      "reusableRoomwideNotable": [
+        {
+          "name": "Amphitheatre Reverse Acid Dive",
+          "note": "Dive into the acid and wall jump up the left wall to the doorway."
```
osse101, line 1541: There will be reverse acid dive strats that do not use walljump, like spacejump.

region/norfair/crocomire.json
```diff
@@ -1456,7 +1456,7 @@
                 },
                 {
                   "name": "PCJR Big Jump Shinespark",
-                  "notable": true,
+                  "notable": false,
```
osse101, line 1459: This one is pretty unintuitive

osse101 (commented):
👍

PR #624, opened 2023-03-15
Bypass canIframeSpikeJump

kjbranch:
Add strats that simply run in the spikes, like a player might do if they didn't know how to utilize the i-frames.

### DIFF
--- a/region/crateria/central.json
+++ b/region/crateria/central.json
@@ -3362,7 +3362,8 @@
                     {"spikeHits": 1},
                     {"or": [
                       "canIframeSpikeJump",
-                      {"spikeHits": 1}
+                      {"spikeHits": 1},
+                      "canDamageBoost"
                     ]}
                   ]
                 },
--- a/region/crateria/central.json
+++ b/region/crateria/central.json
@@ -3358,7 +3358,13 @@
                 {
                   "name": "Take Spike Damage",
                   "notable": false,
-                  "requires": [{"spikeHits": 1}]
+                  "requires": [
+                    {"spikeHits": 1},
+                    {"or": [
+                      "canIframeSpikeJump",
+                      {"spikeHits": 1}
+                    ]}
+                  ]
                 },
                 {
                   "name": "Shinespark",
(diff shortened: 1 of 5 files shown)


region/norfair/east.json
```diff
@@ -4645,8 +4650,13 @@
                   "notable": false,
                   "requires": [
                     "h_canNavigateHeatRooms",
+                    "Gravity",
                     {"heatFrames": 375},
-                    {"spikeHits": 2}
+                    {"spikeHits": 2},
+                    {"or": [
+                      "canIframeSpikeJump",
+                      {"spikeHits": 2}
```
osse101, line 4658: or canDamageBoost tbh

region/wreckedship/main.json
```diff
@@ -1638,6 +1638,13 @@
                     {"spikeHits": 3}
                   ]
                 },
+                {
+                  "name": "Bowling Spike Run",
+                  "notable": false,
+                  "requires": [
+                    {"spikeHits": 7}
```
osse101, line 1645: Looks inconsistent to the other strats.  Should be using no run speed jumps right?

osse101: 3 damage boosts would fit in a strat too.

region/norfair/east.json
osse101: A note about turning off speed would be nice too while we're here.  I feel it comes up the most in this room.

region/norfair/east.json
```diff
@@ -4585,8 +4585,13 @@
...
```
osse101, line 4593: or dboost

region/crateria/central.json
```diff
@@ -3358,7 +3358,13 @@
                 {
                   "name": "Take Spike Damage",
                   "notable": false,
-                  "requires": [{"spikeHits": 1}]
+                  "requires": [
+                    {"spikeHits": 1},
+                    {"or": [
+                      "canIframeSpikeJump",
+                      {"spikeHits": 1}
```
osse101, line 3365: left to right can make it with one damage boost.  Right to left is right on the line, maybe too tight.

region/wreckedship/main.json
kjbranch: I was doing it as running to the right without jumping. With poor jumps and no run I get 6 hits

kjbranch:
I think damage boosting on the spikes in all of these cases is quite a lot harder than `canIframeSpikeJump`, and kind of still implicitly relies on it. The precision of the damage boost is pretty important to ensure that it even helps reduce the number of hits, and requires quick jumps afterwards, something I would consider part of `canIframeSpikeJump`

osse101:
I can agree with that

PR #628, opened 2023-03-15
Cf Standup and WS Supers

kjbranch:


region/wreckedship/main.json
```diff
@@ -1995,7 +2009,25 @@
...
+                {
+                  "name": "WS East Supers Robot Clip (Left to Right)",
+                  "notable": true,
+                  "requires": [
+                    "Morph",
+                    "h_canDestroyBombWalls",
+                    "canTrickyJump"
+                  ],
+                  "note": [
+                    "Kago through the Workrobot by jumping, pressing down, then morphing and moving right just before hitting the robot.",
```
osse101, line 2023: Kago tech would make sense to add now.  Its only lacking the logically required situations for it to be made a tech, and here is one.

region/wreckedship/main.json
```diff
@@ -1995,7 +2009,25 @@
...
+                    "The bomb blocks can be broken with a power bomb, morph bomb, or using screw attack.",
+                    "If using morph bombs, be careful not to break the bottom bomb tile, or you will be pushed into it and stuck."
```
osse101, line 2025: I think bomb jumping is the easier way to place the bomb at the right height.  And you unmorph before falling again.
Otherwise its a `can short jumpmorph` tech.

And the note should say a second kago will push you right into the opened morph hole.

region/wreckedship/main.json
```diff
@@ -2014,8 +2046,25 @@
...
+                },
+                {
+                  "name": "WS East Supers Crystal Flash Clip (Right to Left)",
+                  "notable": true,
+                  "requires": [
+                    "canCrystalFlash",
+                    "canCeilingClip"
+                  ],
+                  "note": [
+                    "Crystal flash to force a standup then spin jump up then morph to bypass the dead robot. Use Coverns to damage down if necessary.",
```
osse101, line 2060: I was aim down jumping and just falling out before reading this.  No reason to add unnecessary alternatives though.

osse101:
👍

PR #629, opened 2023-03-17
Partial fix to Mama Turtle's Room

kjbranch:


osse101 (commented):
Interesting that you can't stright jump above the turtle.  Spinjump and spinbreak worked pretty well.  

All manner of jump assist could be added to jump from middle to missile wall.

Hijump is enough to climb the right wall to the item, with walljump and wiggle. Or spacejump.
Springball can climb the wall, but cannot escape the water.

kjbranch:
I think I addressed everything you were wanting, but I'm not sure I understood all of your suggestions. We never could get the midair wiggle strat to work, so I'm not sure if it needs a normal wall jump or precise.

osse101:
You want to watch the water level too.  The wiggle is to position yourself aligned with the wall, just countering the water physics.  It could go either way on walljump vs precise.

osse101:
That jump is outright tricky, having trouble getting it again. 
And a mention somewhere that you can ride the turtle back up again if you miss your ibj or grapple.  But not once you've gotten started on your strat.

PR #633, opened 2023-03-18
Add strat Oasis Cross Room Jump with Screw Attack

blkerby:


blkerby:
Tested this out and just barely got it to work in the Practice Hack from East Sand Hall, and it seems to require using the open end, so 4.5 tiles. 

I can't seem to get it from West Sand Hall; maybe it's different from that side somehow in spite of everything looking symmetric? If so I guess we need to adjust the tiles for that side.

blkerby:
Nevermind, it was just collision oscillator randomly making it not work. Added note about that, and reduced usedTiles from 4.5 to 4.

osse101:
This is a case of `physics: air`.  There's no space to turn off gravity, and you would use a different strat anyway if you had it.

kjbranch:
This was merged into #647

PR #636, opened 2023-03-21
Brinstar Reserve canSpeedKeep

blkerby:
I don't know exactly how this is done (haven't practiced speed keeps) but I remember seeing it on a stream before. @osse101 would you mind checking?

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -1523,6 +1523,10 @@
                       "shinesparkFrames": 0,
                       "openEnd": 0
                     }}
+                  ],
+                  "devNote": [
+                    "FIXME: This is a rough way to put in the strat, but it is not capturing the full information.",
+                    "If the player is skilled enough to do a shorter charge, the adjacent runway could be smaller, if the player can do a more tight jump and mockball, usedTiles could be larger."
                   ]
                 }
               ]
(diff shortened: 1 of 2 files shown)


osse101:
https://youtu.be/TpRe3x84Sts

-canMockBall
-A way to describe entering the room while falling and not morphed.  It looks possible to run into the room with very slow run speed and jump in room too.  I wanted to make a tech for `canLowRunSpeedShortcharge` already for other strats.
-I'd call jumping into the bomb blocks a tricky jump.  Holding onto blue in the air would be part of speedkeep.
-An alternate strat using springball instead of speedkeep, which is a bit easier.

blkerby:
Thanks, I added canMockball, canTrickyJump, a SpringBall option, and changed it to use adjacentRunway + canShineCharge (instead of canComeInCharged) to encode the ability to jump through the door from the previous room while blue with low enough speed. This isn't perfect since other runway lengths could also work (assuming a different level of short-charging ability), but this should handle most cases, and we can always refine this later once we have a proper way to express jumping out of and into a room with blue speed.

I realized that canComeInCharged wouldn't work correctly here since it could be (wrongly) satisfied by combining with the in-room runway or by charging a spark and entering the room without blue.

osse101:
👍

PR #637, opened 2023-03-21
Blue speed jump through Blue Hopper Room

blkerby:
Like in this clip: https://clips.twitch.tv/SarcasticWildPepperoniMoreCowbell-FNJCTQxYvX9TDOw_

### DIFF
--- a/region/tourian/main.json
+++ b/region/tourian/main.json
@@ -1855,6 +1855,18 @@
                     "ScrewAttack"
                   ],
                   "note": "Samus must enter the room in a screw attack spin jump to avoid the damage."
+                },
+                {
+                  "name": "Blue Speed Jump through Hoppers",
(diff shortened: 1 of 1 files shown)


osse101:
That works better than I thought it would.  Its hard to remember where the hoppers are and how they react.  
Running faster than clip looks fine.  Running slower has a chance to hit the ceiling but you maybe you could control the jump to get past that.
`canComeInCharged` shows you know which room is coming up so jumping into the door seems fair.
👍

kjbranch:
The link is broken. Is this good to merge? Is this viable with the vanilla spawn locations and not just Map Rando's?

osse101:
The link doesn't include the trailing `_`
Both enemies will still jump left and be out of the way with vanilla spawn.
Good to merge

PR #639, opened 2023-03-21
Fish Tank Fixes

kjbranch:
Fish Tank Fixes. Looking for specific feedback on:
- Can you do a spring ball midair from 5 to 4 without ice?
- How do you get from 6 to 7 with just a tricky jump?

### DIFF
--- a/region/maridia/outer.json
+++ b/region/maridia/outer.json
@@ -1689,7 +1689,10 @@
                       "canSpringBallJumpMidAir",
                       "canGravityJump",
                       { "enemyDamage": { "enemy": "Pink Space Pirate (standing)", "type": "contact", "hits": 1 } },
-                      "Plasma"
+                      { "enemyKill": { 
+                        "enemies": [ [ "Pink Space Pirate (standing)" ] ],
+                        "explicitWeapons": [ "Plasma" ]
+                      }}
                     ]}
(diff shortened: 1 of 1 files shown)


osse101 (commented):
`Plasma` would be better shown as `enemyKill`.  But next to that is freezing the fish in one shot which can't be shown well other than listing beams.

Springball jumping top left door seems like a tricky jump and I didn't know about it till you asked.

Subpixel jump should be added.  Dunno what to tack on it.  InsaneJump probably.

Good changes overall.

region/maridia/outer.json
```diff
@@ -1685,14 +1679,38 @@
...
+                      "HiJump",
+                      "canWalljump",
+                      "h_canFly",
+                      "canCarefulJump",
+                      "canSpringBallJumpMidAir",
+                      "canGravityJump",
+                      { "enemyDamage": { "enemy": "Pink Space Pirate (standing)", "type": "contact", "hits": 1 } },
+                      { "enemyKill": { 
+                        "enemies": [ [ "Pink Space Pirate (standing)" ] ],
+                        "explicitWeapons": [ "Plasma" ]
```
osse101, line 1694: Screw too

osse101:
👍

PR #643, opened 2023-03-24
Reduce heatFrames in Speed Booster Hall

blkerby:
I noticed in a recent Croakomire video that he was able to get through here with no ETanks: https://clips.twitch.tv/DignifiedAbstruseClamPeteZaroll-mvUHUTWl73LfxhXt

The logic currently has this as requiring 400 heat frames (100 energy), which seems like too large a gap from what is actually doable without much difficulty. In this case it resulted in an ETank being easily obtainable out of logic. We fix this by reducing the heatFrames from 400 to 360.

### DIFF
--- a/region/norfair/east.json
+++ b/region/norfair/east.json
@@ -3200,7 +3200,7 @@
                     }},
                     {"and":[
                       "SpeedBooster",
-                      {"heatFrames": 400}
+                      {"heatFrames": 360}
                     ]}
                   ]
                 }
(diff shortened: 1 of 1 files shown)


osse101:
360 is about exactly right for no arm pumps and no adjacentRunway.  Neither being things modelled properly 🤷 
We've been rounding up numbers, but I'd call it an issue if that causes an extra Etank to be required.  So it would be worthwhile to check around for `heatFrames:400` and other breakpoints and lower them to 99energy, 199energy, etc.

PR #647, opened 2023-03-27
Add Cross Room Jumps and Fix Oasis

kjbranch:
- Add wall jumps to Oasis
- Add cross room jumps with Screw Attack to Oasis
- Add Draygon cross room jump with an air ball

I still need to update and add a new Room Diagram


region/maridia/inner-green.json
```diff
@@ -312,16 +462,37 @@
                 {
                   "name": "Base",
                   "notable": false,
-                  "requires": ["Gravity"]
+                  "requires": [
```
osse101, line 465: Gravity alone is enough

region/maridia/inner-green.json
```diff
@@ -312,16 +462,37 @@
...
+                      {"and": [
+                        "HiJump",
+                        {"or": [
+                          "canDownGrab",
+                          "canSpringBallJumpMidAir",
+                          "canConsecutiveWalljump"
+                        ]}
+                      ]},
+                      {"and": [
+                        "canSpringBallJumpMidAir",
```
osse101, line 490: all three of these are needed

region/maridia/inner-green.json
```diff
@@ -276,9 +320,37 @@
                     ]}
                   ],
                   "note": "The spark takes Samus directly to the top platform."
+                },
+                {
+                  "name": "Oasis Cross Room Jump with Screw Attack and Speed",
```
osse101, line 325: Needs unique name

region/maridia/inner-green.json
```diff
@@ -252,11 +262,45 @@
...
+                        "physics": ["air"]
+                      }},
+                      {"adjacentRunway": {
+                        "fromNode": 2,
+                        "usedTiles": 14,
+                        "physics": ["air"]
+                      }}
+                    ]}
+                  ],
+                  "reusableRoomwideNotable":  "Oasis Cross Room Jump with Screw Attack",
```
osse101, line 289: This should be a separate notable.  Its a lot more involved

region/maridia/inner-green.json
```diff
@@ -252,11 +262,44 @@
...
+                      }},
+                      {"adjacentRunway": {
+                        "fromNode": 2,
+                        "usedTiles": 14,
+                        "physics": ["air"]
+                      }}
+                    ]}
+                  ],
+                  "note": [
+                    "Use Screw Attack to break the bomb block by entering from a non-water room with a spin jump, and make it all the way through the top door.",
```
osse101, line 290: This uses exactly 14 tiles no more, no less, but it does look a tile wide.  More rises too fast to do the trick, less does not have the height to reach the door.
Using angle to break spin works better to break spin, not sure why.
The first momentumConservingTurnaround is done as early as possible, right as you pass the above tile horizontally.  The background may help tell when that happens.
The second momentumConservingTurnaround is as late as possible, just before touching the above platform.
A third momentumConservingTurnaround is likely needed on the door as it is opening.

osse101 (commented):
`screw+speed to the top` can reach the area just below 4 but above 6 without minmaxing run speeds and turnarounds.  If its important to land exactly on 4, and it doesn't seem to be, then `through the door` is about as hard as reaching the 4 platform.

osse101:
👍

PR #667, opened 2023-04-03
Move X-Ray tech to their own tech category

blkerby:
There are already enough of these to deserve their own category, and we're about to add more for R-mode and G-mode.

### DIFF
--- a/tech.json
+++ b/tech.json
@@ -801,6 +801,70 @@
       ]
     },
     {
+      "name": "X-Ray",
+      "description": "Techs related to use of X-Ray Scope",
+      "techs": [
+        {
+          "name": "canXrayWaitForIFrames",
+          "requires": [ "XRayScope" ],
+          "note": [
+            "Use X-Ray to pause time until invulnerability frames are finished",
+            "Combine with Plasma Beam to repeatedly damage enemies with one fired shot, called Microwave",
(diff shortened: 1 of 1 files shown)


tech.json
```diff
@@ -800,6 +800,70 @@
...
+          "name": "canXrayTurnaround",
+          "requires": [ "XRayScope" ],
+          "note": [
+            "Using X-Ray to turn around in place without changing pixel position.",
+            "Alternating X-Ray turnarounds with regular turnarounds enables movement in either direction even while crouched."
+          ]
+        },
+        {
+          "name": "canXRayStandUp",
+          "requires": [ "XRayScope" ],
```
kjbranch, line 826: should this require `canXrayTurnaround`?

blkerby: I think it's probably best to treat them as separate tech. For canXrayTurnaround, being able to turn around while holding X-ray isn't really the meaning of the tech (since that would be trivial); rather it's the knowledge about how it can be used to move you left or right, which isn't related to the "canXRayStandup" tech even though it also involves turning around while holding X-ray.

blkerby: In any case, this PR is just to move the techs to a new category. If we want to modify the tech, probably would make sense to do that in a separate PR.

PR #668, opened 2023-04-04
Add G-mode tech, requirements, and node properties

blkerby:
This includes the following:
- New tech:
  - canEnterRMode
  - canEnterGMode
  - canEnterGModeImmobile
  - canArtificialMorph
- New node properties:
  - leaveWithGModeSetup
  - leaveWithGMode
- New logical requirements:
  - comeInWithRMode
  - comeInWithGMode
- New helper:
  - h_canArtificialMorphMovement
- Example strats in Landing Site, Flyway, and Crateria Tube.

The new node properties and logical requirements follow a pattern similar to "canLeaveCharged" and "canComeInCharged" (except without the "can" to avoid confusion with tech: we may later rename "canLeaveCharged" and "canComeInCharged" to "leaveCharged" and "comeInCharged"). 

The node property "leaveWithGModeSetup" is for describing how to leave a door node while taking enemy damage, in a way that can be used to set up R-mode or G-mode in the next room.

The logical requirements "comeInWithRMode" and "comeInWithGMode" are for describing how to use R-Mode or G-Mode once you have it. 

The node property "leaveWithGMode" is for describing how to carry G-mode through another transition.

One characteristic of this design is that "comeInWithGMode" strats can be fairly complex, e.g. involving going across a room to overload PLMs before moving to another point of interest, bypassing junctions that normally would be involved with such movement. 

@kjbranch and I considered an alternative design, where "G-mode" and "PLMs overloaded" would be treated as bits of state that could be tracked similarly to flags/obstacles. That approach could avoid the need for strats that bypass junctions, but at the heavy expense of needing to go through all the strats in the game and add some kind of "nonGMode" requirement to adjust them if they don't work in G-mode, such as if they depend on destroying shot blocks, bombing bomb blocks, etc. It would be more complicated than this because multiple cases would need to be addressed, based on whether PLMs are overloaded or not and whether we are using artificial morph or not. Such a design could also require increasing the amount of nodes in many rooms since some regions assumed to be freely traversable would no longer be (particularly when we consider the case of artificial morph, which can have very restrictive movement options). Overall we think the approach proposed in this PR should be more manageable, but feedback would be welcome. 

We'll need to consider somewhat similar questions later when we want to model carrying spikesuits & blue suits around the game; in that case, making implicit abilities become explicit (canJumpUp, canDash, etc.) may make sense to help determine which strats can be used while preserving the suits. But those cases are a little different because the effects of spikesuits & blue suits on the game's behavior and movement options are not nearly as complex compared to G-mode. The complexity here is a lot of what motivates us to try to keep G-mode contained to dedicated properties and strats, rather than allowing it to spread out and complicate the other strats in the game.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -328,6 +328,10 @@
       "note": [
         "These are ways to navigate stairs and small platforms for strats assuming G-mode artificial morph.",
         "Having Morph means we can unmorph, jump, and remorph, as using artificial morph is not needed in that case."
+      ],
+      "devNote": [
+        "Morph will not be a usable alternative to get up very constrained ledges, such as in a morph tunnel.",
+        "IBJ is not usable for underwater rooms without Gravity."
       ]
     }
   ]
--- a/helpers.json
+++ b/helpers.json
@@ -308,6 +308,16 @@
           "canRJump"
         ]}
       ]
+    },
+    {
+      "name": "h_canArtificialMorphMovement",
+      "requires": [
+        {"or": [
+          "Bombs",
+          "SpringBall"
+        ]}
+      ],
+      "devNote": "FIXME: Add a canIBJ tech requirement to the Bombs case, after Morph is removed as a requirement from canIBJ."
     }
   ]
 }
--- a/helpers.json
+++ b/helpers.json
@@ -310,14 +310,20 @@
       ]
     },
     {
+      "name": "h_canArtificialMorphIBJ",
+      "requires": [
+        "Bombs"
+      ],
+      "devNote": "FIXME: Add a canIBJ tech requirement after Morph is removed as a requirement."
+    },
+    {
       "name": "h_canArtificialMorphMovement",
       "requires": [
         {"or": [
-          "Bombs",
-          "SpringBall"
+          "SpringBall",
+          "h_canArtificialMorphIBJ"
         ]}
-      ],
-      "devNote": "FIXME: Add a canIBJ tech requirement to the Bombs case, after Morph is removed as a requirement from canIBJ."
+      ]
     }
   ]
 }
--- a/helpers.json
+++ b/helpers.json
@@ -324,6 +324,10 @@
           "h_canArtificialMorphIBJ",
           "Morph"
         ]}
+      ],
+      "note": [
+        "These are ways to navigate stairs and small platforms for strats assuming G-mode artificial morph.",
+        "Having Morph means we can unmorph, jump, and remorph, as using artificial morph is not needed in that case."
       ]
     }
   ]
--- a/helpers.json
+++ b/helpers.json
@@ -321,7 +321,8 @@
       "requires": [
         {"or": [
           "SpringBall",
-          "h_canArtificialMorphIBJ"
+          "h_canArtificialMorphIBJ",
+          "Morph"
         ]}
       ]
     }
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -372,7 +372,7 @@ Please refer to the sections on `leaveWithGModeSetup` in [the Region documentati
 A `comeInWithGMode` object represents the need to either have or obtain G-mode when entering the room. It has the following properties:
 * _fromNodes:_ Indicates from what doors this logical requirement expects Samus to enter the room.
 * _mode:_ Takes one of three possible values, "direct", "indirect", or "any", indicating whether this logical requirement expects Samus to enter in direct G-mode, indirect G-mode, or either. Direct G-mode is the state obtained when G-mode is first entered (i.e., the next room after the G-mode setup is performed), while indirect G-mode is the state after passing a door transition with G-mode (usually back into the room where the G-mode setup was performed).
-* _artificialMorph:_ A boolean indicating whether the logical requirement expects Samus to either obtain or already have an artificially morphed state when coming into the room.
+* _artificialMorph:_ A boolean indicating whether the logical requirement expects Samus to either obtain or already have an artificially morphed state when coming into the room, or to have collected the Morph item.
 
 __Example:__
 ```json
@@ -395,7 +395,7 @@ __Additional considerations__
     * Any additional requirements in the `requires` property of the `leaveWithGModeSetup`.
   * A `leaveWithGMode` object must satisfy the following requirements in order to match:
     * The `mode` in the `comeInWithGMode` object must be "indirect" or "any".
-    * If `artificialMorph` is `true`, then the `leavesWithArtificialMorph` property of the `leaveWithGMode` object must be `true`.
+    * If `artificialMorph` is `true`, then either the `leavesWithArtificialMorph` property of the `leaveWithGMode` object must be `true` or there is an additional requirement that the Morph item be collected.
     * Any additional requirements in the `requires` property of the `leaveWithGMode` object.
 * In the case of direct G-mode, `comeInWithGMode` object implicitly requires an energy drain caused by the reserve trigger and preceding setup:
   * If the tech `canEnterGModeImmobile` is enabled and the `gModeImmobile` on the corresponding door is satisfied, then Samus' regular energy will become whatever reserve energy she had before the transition, truncated to her maximum amount of regular energy (based on the number of ETanks collected).
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -370,15 +370,13 @@ A `comeInWithGMode` object represents the need to either have or obtain G-mode w
 * _fromNodes:_ Indicates from what doors this logical requirement expects Samus to enter the room.
 * _mode:_ Takes one of three possible values, "direct", "indirect", or "any", indicating whether this logical requirement expects Samus to enter in direct G-mode, indirect G-mode, or either. Direct G-mode is the state obtained when G-mode is first entered (i.e., the next room after the G-mode setup is performed), while indirect G-mode is the state after passing a door transition with G-mode (usually back into the room where the G-mode setup was performed).
 * _artificialMorph:_ A boolean indicating whether the logical requirement expects Samus to either obtain or already have an artificially morphed state when coming into the room.
-* _immobile:_ A boolean indicating whether the logical requirement expects Samus to enter in a G-mode immobile state, which means an enemy must be able to hit Samus where she spawns in the destination room in order to regain control.
 
 __Example:__
 ```json
 {"comeInWithGMode": {
   "fromNodes": [1],
   "mode": "any",
-  "artificialMorph": false,
-  "immobile": false
+  "artificialMorph": false
 }}
 ```
 
@@ -386,7 +384,6 @@ __Additional considerations__
 
 * A `comeInWithGMode` object implicitly requires X-Ray Scope and a Reserve Tank.
 * A `comeInWithGMode` object implicitly requires the `canEnterGMode` tech.
-  * If `immobile` is `true` then it also requires the `canEnterGModeImmobile` tech.
   * If `artificialMorph` is `true` then it also requires the `canArtificialMorph` tech.
 * A `comeInWithGMode` requires that one of the indicating nodes in `fromNodes` has a matching `leaveWithGModeSetup` or `leaveWithGMode` object in the corresponding door node of the neighboring room:
   * A `leaveWithGModeSetup` object must satisfy following requirements in order to match:
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -343,6 +343,67 @@ __Additional considerations__
 * A `canShineCharge` object implicitly requires the Speed Booster.
 * A `canShineCharge` object implicitly requires the `canShineCharge` tech if it has more than 0 `shinesparkFrames`.
 
+#### comeInWithRMode object
+A `comeInWithRMode` object represents the need to obtain R-mode when entering the room. It has the following properties:
+* _fromNodes:_ Indicates from what doors this logical requirement expects Samus to enter the room.
+
+__Example:__
+```json
+{"comeInWithRMode": {
+  "fromNodes": [1],
+}}
+```
+
+__Additional considerations__
+
+* A `comeInWithRMode` object implicitly requires X-Ray Scope and a Reserve Tank.
+* A `comeInWithRMode` object implicitly requires the `canEnterRMode` tech.
+* A `comeInWithRMode` requires that one of the indicating nodes in `fromNodes` has a matching `leaveWithDamage`.
+  * The `leaveWithDamage` object must satisfy following requirements in order to match:
+    * Samus must have non-zero reserve energy.
+    * Any additional requirements in the `requires` property of the `leaveWithDamage` object.
+
+Please refer to the sections on `leaveWithDamage` in [the Region documentation](region/region-readme.md) for a more detailed explanation of this object.
+
+#### comeInWithGMode object
+A `comeInWithGMode` object represents the need to either have or obtain G-mode when entering the room. It has the following properties:
+* _fromNodes:_ Indicates from what doors this logical requirement expects Samus to enter the room.
+* _mode:_ Takes one of three possible values, "direct", "indirect", or "any", indicating whether this logical requirement expects Samus to enter in direct G-mode, indirect G-mode, or either.
+* _artificialMorph:_ A boolean indicating whether the logical requirement expects Samus to enter in an artificially morphed state (without necessarily having collected the Morph item).
+* _previouslyOverloadedPLMs:_ A boolean indicating whether the logical requirement expects that PLMs have already been overloaded when entering the room. This is only applicable to indirect G-mode.
+* _immobile:_ A boolean indicating whether the logical requirement expects Samus to enter in a G-mode immobile state, which means an enemy must be available in the right location to hit Samus and return control.
+
+__Example:__
+```json
+{"comeInWithGMode": {
+  "fromNodes": [1],
+  "mode": "any",
+  "artificialMorph": false,
+  "previouslyOverloadedPLMs": false,
+  "immobile": false
+}}
+```
+
+__Additional considerations__
+
+* A `comeInWithGMode` object implicitly requires X-Ray Scope and a Reserve Tank.
+* A `comeInWithGMode` object implicitly requires the `canEnterGMode` tech.
+  * If `immobile` is `true` then it also requires the `canEnterGModeImmobile` tech.
+  * If `artificialMorph` is `true` then it also requires the `canArtificialMorph` tech.
+* A `comeInWithGMode` requires that one of the indicating nodes in `fromNodes` has a matching `leaveWithDamage` or `leaveWithGMode` object in the corresponding door node of the neighboring room:
+  * A `leaveWithDamage` object must satisfy following requirements in order to match:
+    * The `mode` in the `comeInWithGMode` object must be "direct" or "any".
+    * Samus must have non-zero reserve energy.
+    * Any additional requirements in the `requires` property of the `leaveWithDamage`.
+  * A `leaveWithGMode` object must satisfy the following requirements in order to match:
+    * The `mode` in the `comeInWithGMode` object must be "indirect" or "any".
+    * If `artificialMorph` is `true`, then the `leavesWithArtificialMorph` property of the `leaveWithGMode` object must be `true`.
+    * If `previouslyOverloadedPLMs` is `true`, then the `leavesWithOverloadedPLMs` property of the `leaveWithGMode` object must be `true`.
+    * Any additional requirements in the `requires` property of the `leaveWithGMode` object.
+
+Please refer to the sections on `leaveWithDamage` and `leaveWithGMode` in [the Region documentation](region/region-readme.md) for a more detailed explanation of these objects.
+
+
 ### Room Pathing Objects
 This section contains logical elements that are affected by Samus' pathing within a room.
 
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -399,7 +399,7 @@ __Additional considerations__
     * Any additional requirements in the `requires` property of the `leaveWithGMode` object.
 * In the case of direct G-mode, `comeInWithGMode` object implicitly requires an energy drain caused by the reserve trigger and preceding setup:
   * If the tech `canEnterGModeImmobile` is enabled and the `gModeImmobile` on the corresponding door is satisfied, then Samus' regular energy will become whatever reserve energy she had before the transition, truncated to her maximum amount of regular energy (based on the number of ETanks collected).
-  * Otherwise, Samus' regular energy will become 4, or whatever reserve energy she had before the transition if it was less than 4.
+  * Otherwise, Samus' regular energy will become 4, reflecting the need  or whatever reserve energy she had before the transition if it was less than 4.
   * Samus' reserve energy will always become zero.
 
 Please refer to the sections on `leaveWithGModeSetup`, `leaveWithGMode`, and `gModeImmobile` in [the Region documentation](region/region-readme.md) for a more detailed explanation of these objects.
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -398,11 +398,11 @@ __Additional considerations__
     * If `artificialMorph` is `true`, then the `leavesWithArtificialMorph` property of the `leaveWithGMode` object must be `true`.
     * Any additional requirements in the `requires` property of the `leaveWithGMode` object.
 * In the case of direct G-mode, `comeInWithGMode` object implicitly requires an energy drain caused by the reserve trigger and preceding setup:
-  * If the tech `canEnterGModeImmobile` is enabled and the `gModeImmobileRequires` on the corresponding door is satisfied, then Samus' regular energy will become whatever reserve energy she had before the transition, truncated to her maximum amount of regular energy (based on the number of ETanks collected).
+  * If the tech `canEnterGModeImmobile` is enabled and the `gModeImmobile` on the corresponding door is satisfied, then Samus' regular energy will become whatever reserve energy she had before the transition, truncated to her maximum amount of regular energy (based on the number of ETanks collected).
   * Otherwise, Samus' regular energy will become 4, or whatever reserve energy she had before the transition if it was less than 4.
   * Samus' reserve energy will always become zero.
 
-Please refer to the sections on `leaveWithGModeSetup`, `leaveWithGMode`, and `gModeImmobileRequires` in [the Region documentation](region/region-readme.md) for a more detailed explanation of these objects.
+Please refer to the sections on `leaveWithGModeSetup`, `leaveWithGMode`, and `gModeImmobile` in [the Region documentation](region/region-readme.md) for a more detailed explanation of these objects.
 
 
 ### Room Pathing Objects
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -387,7 +387,7 @@ __Additional considerations__
 
 * A `comeInWithGMode` object implicitly requires X-Ray Scope and a Reserve Tank.
 * A `comeInWithGMode` object implicitly requires the `canEnterGMode` tech.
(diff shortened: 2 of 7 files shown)


tech.json
```diff
@@ -861,6 +861,45 @@
...
+                },
+                {
+                  "name": "canArtificialMorph",
+                  "requires": [ "canEnterGMode" ],
+                  "note": [
+                    "Ability to obtain artificial morph with G-mode.",
+                    "Turn around after reserves finish filling but before knockback frames expire (a 4-frame window when entering G-mode with 4 reserve energy).",
+                    "For G-mode immobile, instead turn around while taking knockback from the enemy that returns control to Samus (a 5-frame window)",
+                    "This results in Samus entering a morph ball state without needing to have collected or equipped Morph."
+                  ]
```
kjbranch, line 898: There may need to be new helpers for tech that require morph. Going through all the tech that require morph, it looks like only `canResetFallSpeed` and `canCrystalFlash` may be doable with artificial morph.

There are some helpers that might be worth looking at as well, although I imagine the gmode strats simply wont use these?: `h_canBombThings`, `h_canUsePowerBombs`, `h_canUseMorphBombs`, `h_canUseSpringBall`.

region/crateria/central.json
```diff
@@ -118,6 +118,57 @@
...
+              "leavesWithArtificialMorph": true,
+              "strats": [
+                {
+                  "name": "Artificial Morph Overload Scroll PLMs",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
+                      "Bombs",
+                      "SpringBall"
+                    ]},
```
kjbranch, line 155: or 7 PBs?

region/crateria/central.json
```diff
@@ -153,6 +204,32 @@
...
+            {
+              "leavesWithArtificialMorph": false,
+              "leavesWithOverloadedPLMs": true,
+              "strats": [
+                {
+                  "name": "Overload Scroll PLMs",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
+                      "SpeedBooster",
```
kjbranch, line 218: wouldnt this require one of many strats to break these blocks? shinesparking from the right, blue space jumps from any of the 3 locations, bombs, or PBs, etc? might be able to simplify this by using obstacle A.

region/crateria/central.json
```diff
@@ -177,6 +254,49 @@
...
+          "leaveWithGMode": [
+            {
+              "leavesWithArtificialMorph": false,
+              "leavesWithOverloadedPLMs": true,
+              "strats": [
+                {
+                  "name": "Overload Scroll PLMs",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
```
kjbranch, line 266: similar comment, tho this one looks a bit better.

region/crateria/central.json
```diff
@@ -177,6 +254,49 @@
...
+                    ]}
+                  ],
+                  "note": [
+                    "Use SpeedBooster or Screw Attack to break through the bomb blocks on the Gauntlet ledge.",
+                    "Use the scroll PLMs one tile to the left of those blocks to overload PLMs.",
+                    "Return to the door with either SpaceJump, a SpringBall jump (with HiJump), or shinesparking."
+                  ],
+                  "devNote": [
+                    "There are no energy requirement for sparking through the blocks on the Gauntlet ledge, since we can refill at the Ship before and after.",
+                    "The only spark we consider for the return path is the short-charge on the plateau, since this has minimal energy requirements."
```
kjbranch, line 294: could shinespark from the bottom for a smaller runway.
These strats look incredibly difficult to maintain, since they will need to include all room strats, many of which are currently missing from these examples. I wonder if there is another way to do this, by using some links in the room. Even if we could use these links to break obstacle A, that would go a long way

region/crateria/central.json
```diff
@@ -365,6 +534,23 @@
                       "requires": ["h_canDestroyBombWalls"]
                     }
                   ]
+                },
+                {
+                  "name": "G-Mode through Bomb Blocks",
```
kjbranch, line 539: You may want to put this in a link from 1->4, as it could be used to get to the bottom right door without a wall jump (if that is ever removed from the game.) going from 4->5 is free, but 5->4 could require wall jumps. A similar strat could be added for 2->4

logicalRequirements.md
```diff
@@ -343,6 +343,67 @@ __Additional considerations__
...
+  "fromNodes": [1],
+}}
+```
+
+__Additional considerations__
+
+* A `comeInWithRMode` object implicitly requires X-Ray Scope and a Reserve Tank.
+* A `comeInWithRMode` object implicitly requires the `canEnterRMode` tech.
+* A `comeInWithRMode` requires that one of the indicating nodes in `fromNodes` has a matching `leaveWithDamage`.
+  * The `leaveWithDamage` object must satisfy following requirements in order to match:
```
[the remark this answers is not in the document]
blkerby: Yeah, another thing is that it's necessary to leave while in a position to be able to use X-Ray on the first frame in the next room, which for horizontal transitions means you have to be on the ground. Maybe we shouldn't try to make this as general and instead name it in a way consistent with it being used for setting up R-mode/G-mode. Maybe something like "leaveWithGModeSetup"? It's a little awkward since R-mode and G-mode have the same setup, but maybe that's ok.

logicalRequirements.md
```diff
@@ -343,6 +343,67 @@ __Additional considerations__
...
+    * Samus must have non-zero reserve energy.
+    * Any additional requirements in the `requires` property of the `leaveWithDamage` object.
+
+Please refer to the sections on `leaveWithDamage` in [the Region documentation](region/region-readme.md) for a more detailed explanation of this object.
+
+#### comeInWithGMode object
+A `comeInWithGMode` object represents the need to either have or obtain G-mode when entering the room. It has the following properties:
+* _fromNodes:_ Indicates from what doors this logical requirement expects Samus to enter the room.
+* _mode:_ Takes one of three possible values, "direct", "indirect", or "any", indicating whether this logical requirement expects Samus to enter in direct G-mode, indirect G-mode, or either.
+* _artificialMorph:_ A boolean indicating whether the logical requirement expects Samus to enter in an artificially morphed state (without necessarily having collected the Morph item).
```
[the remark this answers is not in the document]
blkerby: Reworded to make this more clear, thanks!

logicalRequirements.md
```diff
@@ -343,6 +343,67 @@ __Additional considerations__
...
+* _previouslyOverloadedPLMs:_ A boolean indicating whether the logical requirement expects that PLMs have already been overloaded when entering the room. This is only applicable to indirect G-mode.
+* _immobile:_ A boolean indicating whether the logical requirement expects Samus to enter in a G-mode immobile state, which means an enemy must be available in the right location to hit Samus and return control.
```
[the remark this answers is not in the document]
blkerby: Reworded, thanks!

logicalRequirements.md
```diff
@@ -343,6 +343,67 @@ __Additional considerations__
...
+  "mode": "any",
+  "artificialMorph": false,
+  "previouslyOverloadedPLMs": false,
+  "immobile": false
+}}
+```
+
+__Additional considerations__
+
+* A `comeInWithGMode` object implicitly requires X-Ray Scope and a Reserve Tank.
```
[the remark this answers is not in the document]
blkerby: That requirement belongs more in `leaveWithGModeSetup` but it was missing there too; I've added it there now to make it more clear, thanks!

region/crateria/central.json
blkerby: Thanks, Bokyubi checked and found this could be done with 4 PBs. Added it now

region/crateria/central.json
blkerby: Bombs and PBs won't work to break the blocks in G-mode, and blue space jump requires an additional item (SpaceJump), so I don't think we lose anything significant by just considering these two options.

blkerby: I added some more requirements now to properly represent what is needed for the Big Jump short-charge from the plateau, which is a simplification to avoid considering all the other easier ways they could get through with the same items. If someone is enabling G-mode logic, I don't think we lose too much by also requiring basic things of lower difficulty/complexity.

region/crateria/central.json
blkerby: Yeah there's some duplication of room strats here. I'm not convinced that it's too different or worse compared to other duplication that we have in the data, but it is something that could be improved. I think it could help to add a concept of room-local helper, an idea we've been throwing around for a while. Directly referencing links or strats in the room (to copy their requirements) would be another way to reduce duplication but could be more fragile: since it would be hard to see when links may have references to them, changes could have unexpected side effects on the strats that reference them, and if you renumber nodes or rename strats then those references would probably break. With the room-local helper approach, you're explicitly putting some things into a larger scope and giving them meaningful names, making them easier to keep track of and know that changing them will affect more than one strat, and it would be easy to search to find all the references to a given room-local helper.

tech.json
blkerby: I think canIBJ is another example and one which even applies to this room. I added a FIXME for those cases.

region/crateria/central.json
blkerby: Done, thanks!

osse101:
additional_considerations should note that using Rmode and Gmode will set your current energy to be the truncated value of your reserve energy.
The hierarchy of extension techs on Gmode is fuzzy.  I would have thought gmode immobile is the easier version, and it is easier for artificial morph, And the roles aren't exactly the same, immobile would be used only if there is a need to take damage after.
I don't like removing Morph as a requirement from useSpringball or all the various bomb jumps.  `or: [Morph, ArtificialMorph]` would be fine but they are collected items, which means managing artificialMorph state.  I don't have an easier alternative.

blkerby:
Thanks, I added notes to the additional considerations now about the reserve trigger and its effects on energy.

With 11 or less, G-mode immobile is the easier version (14-frame window for the X-ray cancel), which would be an argument for making a separate, lower-tier tech for that scenario. But I have a hard time imagining someone wanting to turn on this niche case without wanting to enable G-mode more broadly, so I'm not sure there would be a purpose to carving off a separate tech for that case. The main application of G-mode immobile is to allow keeping a larger amount of energy in the destination room, and if it is more than 12 energy then this results in a 3-frame window for the X-ray cancel, which I think makes it harder than low-energy G-mode (4 frame window). G-mode immobile can also involve waiting around for a while. So these both point to putting it in a higher tier of tech. For a higher chance of successful execution, it could still be smart for a player to go for an 11 energy G-mode immobile if there's a suitable enemy in the next room and they don't need higher energy, but I think that's more of a player strategy rather than something that needs to be put into the logic.

I think the existing helpers (e.g., h_canBombThings, h_canUsePowerBombs, h_canUseMorphBombs, h_canUseSpringBall) are fine and don't need to be changed. G-mode strats can just avoid using them, instead referring to item and tech requirements directly. The only problem is techs that have the Morph item requirement baked into them, because there's no way around that. I think what we can do is add corresponding helpers for those. So, for example, everywhere that's using "canIBJ" now we could replace with "h_canIBJ", defined as ["canIBJ", "Morph", "Bombs"]. Then we can remove the item dependencies from the tech "canIBJ" so that G-mode strats with artificial morph can instead use requirements like ["canIBJ", "Bombs"].

osse101:
I haven't tried out how it feels but, Immobile is easier because you have more control.  You see your health filling and know when it will stop.  The low health Gmode timer starts when fade in happens and you cant be sure when that is.

It would fit to combine the two into one canGMode.  Just a lot of text and two situations to describe which tends towards separate techs. But the skill for both is still releasing xray before your reserves fill.  
`Basic Setup`'s opposite would be a complex setup.  And that would be freezing a pipe bug so it can rise to screens to reach a door, or supering a geemer to fall on your head as you go in the door.  And that's not what this is talking about.
`Abilitiy to ... with high energy` - Having energy isn't a skill so this means the 3 frame window which we're going to treat as the only way to do this trick.  I don't see 4 vs 3 frames as a meaningful difference.  Removing immobile doesn't take any action, but does require knowing which rooms it is possible to work in.
Artificial Morph done off of low health vs immobile seems like a larger skill differential.  `Frames after realeasing XRay` vs `you can see the enemy coming`.

So I think I'm OK with canGMode extending to canGModeImmobile for immobile being more situational.  However next is how will each be used.  I wouldn't want immobile to be used whenever is possible because it gives more health because then its being treated as the same tech.  To have a need for the additional health for an upcoming strat would put value on the enemy knowledge.  But also most likely you can farm the random enemies as much as you need either way.  Another difference is the artificial morph timing.  Splitting the method of gaining morph would be a reason to keep them separate.

blkerby:
Thanks, I've clarified the tech descriptions now, e.g. removing the confusing reference to "basic setup".

Yeah you're probably right that G-mode immobile is not necessarily harder than regular G-mode, but maybe it makes sense to keep it as an extension of (low-energy) G-mode because of being more situational.

When it comes to which strats are "used" (e.g. G-mode immobile to keep more energy), this seems more like a randomizer thing or player strategy thing, outside the scope of the data here which is just trying to define which strats are possible with given tech/items. With Map Rando right now, the randomizer's intended route always tries to conserve energy when possible because it has no way of knowing if that energy will be needed in a later step or not. For example in Bubble Mountain when going from the top right to the top left, assuming the tech is enabled the spoiler log will always say to go down and do the insane walljump back up instead of doing the dboost. So this would happen with G-mode immobile too; it will always try to use it if it's applicable and allows it keep more energy. It would be possible to have the randomizer go back and do a second pass over the spoiler route to try to replace unnecessarily difficult/complicated strats with easier ones, but this is not trivial and we haven't prioritized it since so far it doesn't seem like people pay much attention to the spoiler routes anyway.

One thing to keep in mind is that you can't farm enemies during G-mode. Their drops won't spawn until you use X-ray to cancel G-mode. You can use the Ship to refill. I haven't checked yet if energy refill stations work or not. We should probably add a logical requirement for doing a refill so that we can express strats that use them like this (e.g., in a `leavesWithGMode` where you enter the room with direct G-mode, do a refill, and then go back out while retaining G-mode).

schema/m3-region.schema.json
```diff
@@ -860,6 +860,89 @@
...
+                  "properties": {
+                    "requires": {
+                      "$id": "#/properties/rooms/items/properties/nodes/items/properties/gModeImmobile/properties/requires",
+                      "$ref": "m3-requirements.schema.json#/definitions/logicalRequirements",
+                      "title": "G-Mode Immobile Requirements",
+                      "description": "Logical requirements in order to allow control to be returned to Samus when coming into this door with G-mode immobile"
+                    },
+                    "note": {
+                      "$ref" : "m3-note.schema.json#/definitions/note",
+                      "$id": "#/properties/rooms/items/properties/nodes/items/properties/locks/items/properties/note"
```
osse101, line 938: We're not in `locks`

osse101:
It looks like gaining GMode is described well enough here.  
For setting up same room indirect, you need to make sure the other room doesn't drop you without a way back..  But with crumbles being solid that's only like Ridleys room, with walljump disabled.  And an artificial morph still keeps you in the door.  Immobile looks to need to lower ridleys health before he'll fly up to the door.

Describing strats was always going to be strange.  Needing to go through a room in ball form the whole way is too different from normal movement so I can agree with keeping GMode strats a little separate.

There is very little, but some, interest in sharing a hit in transition by enemy list with `canResetFallSpeed` and `canIFrameEnemy` in the future.

blkerby:
Good point about setting up indirect. I was going off of what bobbob mentioned about always being able to come back, which I guess is still true in the vanilla game if we assume that you can do artificial morph. But I think we'd like this to be more general so that the rules can make sense for other possible (non-vanilla) rooms and so we can account for any artificial morph tech requirements. I think `spawnAt` is the general way that we have to identify doors that do not have free passage back into them. So I changed the docs now for `leaveWithGMode` to say that it's only implicitly present if the door does not have a `spawnAt`. In such cases we would have to put in `leaveWithGMode` explicitly to express any requirements for getting back into the door. Right now Ridley's Room doesn't use a `spawnAt` but I think it should and we can easily add it. What do you think?

On a side note, for heated rooms there's maybe a little more complication with the meaning of `spawnAt`, because passage back through a door in a heated room isn't generally free, but we aren't putting a `spawnAt` on all heated room doors. This doesn't matter for G-mode since heat damage is disabled, but it's something to think about more generally in cases where we have `resetRoom` requirements for instance. I guess for X-ray climb setups, we added a property "useFrames" to the "adjacentRunway" requirement for this, which is a little different since doing an X-ray climb setup will use more frames than simply going back through the door in the fastest way. The most tricky case I guess is returning from doors where you fall down, because here the time taken to return through the door depends on the room geometry. We were talking about annotating the height of the runway/jumpway/platform below such doors, so maybe this could be another application of that data (i.e., we could possibly infer heatFrames required to go back up through the door based on that height).

blkerby:
Ah, just realized that several of the Landing Site strats need to be split apart. For example, the "G-Mode Morph IBJ" strat from node 4 to 3 is not correct with the comeInWithGMode fromNodes=[1, 2, 4]. Even though it is true that the strat can be executed by coming in with artificial morph from any of those nodes, there will be a problem, say, if node 1 is inaccessible to us but is the only one with a matching G-mode setup. It would say that the G-mode setup in node 1 allows us to get from node 4 to 3, which would be wrong because we cannot reach node 1 and hence could not use it to get us into G-mode.

We had this same kind of mistake in Climb with canComeInCharged with multiple fromNodes. I think to avoid this, the only nodes we should put in `fromNodes` are the link's `from` node and nodes that are freely accessible from it. At some point we may still want to revisit the fundamentals of how we're doing cross-room strats (though I don't have anything concrete to propose at the moment).

blkerby:
Split the strats apart, so they should be correct now.

osse101:
`comeIn_blank_`  seemed to me that it would require visiting the node you are said to enter from, unless there's a case where that connection existing is important somewhere.  `canComeInCharged` has an implicit `resetRoom` to make sure you can enter from that node; as a way to retry the strat, to learn you need to try the strat, and of course to attempt the strat as intended.  `comeInGMode` should have the same `resetRoom`.  I was always fuzzy on how that's meant to be handled if you've just come from that door though.

Having real morph is a good catch.  Putting real morph in the `aritficialMorph` property does seem correct, and situations where artificialMorph is not possible would specifiy having Morph as part of the strat.  Gotta be careful  when using canIBJ and such at least until we're used to the change.

Making use of `spawnAt` sounds smart.  SpawnAts could use some extra looking into.  Mickey Mouse top and Crumble Shaft bottom have a little drop for extra heat damage but nothing big. Ridley room has the big fall that takes a tech.  Xray climbs would also have to watch for `spawnAt` and I don't think that they do.  In those cases you would need to worry about crumbles below your door as in IceGateRoom and MissionImpossibleLowerExit.

blkerby:
Yeah, there's still a problem that it's unclear how to feasibly implement a requirement like `resetRoom` in a randomizer (currently Map Rando just ignores it, treating it as free); this is why I'd still like to remove that requirement type, replacing it with simpler requirements (such as obstaclesNotCleared) that are tractable to implement. 

In any case, having an implicit `resetRoom` requirement wouldn't solve the problem in that Landing Site example. Say the only item we have is Bombs and that node 1 is the only node with a G-mode setup on the other side. Starting at node 4, there is no in-room path for us to get to node 1 to do the G-mode setup. But there could be some other path, e.g. going out through node 2 and looping around through a sequence of other rooms ending up at node 1 of Landing Site. I think the problem is solved by just have a G-mode strat going directly from node 1 to node 3. In general we just set up the links so the `fromNodes` in the requirement matches the `from` node; then nothing can go wrong and we don't need a `resetRoom` requirement.

The main drawback is that this results in more strats. Basically, if entering through a door with G-mode does something useful, then we need strats starting `from` that door (except that doors can be correctly lumped together in `fromNodes` in a single strat if they are freely accessible with each other). So in the case of a large room with almost every door being able to do something useful with G-mode, this can add up to a lot of strats. I think it's worth it in order to be able to correctly specify the logic . Also, in the worst case where we have `n` doors and `n` destinations, we don't need to have `O(n^2)` links, because we could add a new node to use as a "G-mode junction", meaning a junction where the only strats reaching it have `comeinWithGMode` requirements, so that being in the junction implies having G-mode. This can get it down to `O(n)` links instead of `O(n^2)`. In this room, I don't think a G-mode junction would help much, but it's something to keep in mind as an option for more complicated cases. We also have to keep in mind that not all G-mode states are the same: we could be in direct vs. indirect, in artificial morph vs. not, have overloaded PLMs already or not; if applicable, all those things would have to be specified in defining a junction. So in a lot of cases I think it's going to be easier to just consider all the possible useful (from, to) pairs and not add a junction, even if it does result in quite a few strats. Open to other ideas though.

schema/m3-region.schema.json
blkerby: Thanks, fixed.

blkerby:
Added a G-mode Morph junction at the Ship. This reduces the complexity of individual strats and cuts out a lot of duplication.

tech.json
```diff
@@ -861,6 +861,45 @@
...
+          "note": [
+            "Ability to enter R-mode.",
+            "Have some energy in reserves (set to Auto), take damage while going through a door, use X-Ray while entering the next room, and release X-Ray after reserves have finished filling."
+          ],
+          "extensionTechs": [
+            {
+              "name": "canEnterGMode",
+              "requires": [ "canEnterRMode" ],
+              "note": [
+                "Ability to enter G-mode starting with low reserve energy (the usual way of entering G-mode).",
```
osse101, line 877: The end bit still seems like it shouldnt be here.
"Ability to enter G-mode with little energy in reserve."

osse101 (commented):
Is it still useful to have `fromNodes`?  If `comeInGMode` is always on the door node then you know the node already.  Maybe it could help in the future, and does make reading changes on Github easier.

tech.json
blkerby: Agreed, fixed.

kjbranch:
Will this require us to add a `spawnAt` to all sand fall transitions?

kjbranch:
It looks like the landing site 1->4 transition should have a way to stop at the ship beforehand to refill. Is there a way we could add that, or should we leave it as a fixme?

kjbranch:
Sorry, I'm still catching up. 
- Another room with a drop off is the exit from Gravity suit room. 
- In order for us to prevent accidentally using things like `canIBJ` going forward, after removing its morph requirement, we could have two helpers: `h_canIBJ`, `h_canIBJArtificialMorph`. Then the only places `canIBJ` should be used are in the helpers and in the tech file.

helpers.json
```diff
@@ -308,6 +308,23 @@
...
+    },
+    {
+      "name": "h_canArtificialMorphMovement",
+      "requires": [
+        {"or": [
+          "SpringBall",
+          "h_canArtificialMorphIBJ",
+          "Morph"
+        ]}
+      ]
```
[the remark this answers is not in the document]
blkerby: Done. Just made it a regular note since this could be useful information for players too (if we ever display helpers in the UI somewhere)

logicalRequirements.md
```diff
@@ -343,6 +343,68 @@ __Additional considerations__
...
+  "mode": "any",
+  "artificialMorph": false
+}}
+```
+
+__Additional considerations__
+
+* A `comeInWithGMode` object implicitly requires X-Ray Scope and a Reserve Tank.
+* A `comeInWithGMode` object implicitly requires the `canEnterGMode` tech.
+  * If `artificialMorph` is `true` then it also requires the `canArtificialMorph` tech.
```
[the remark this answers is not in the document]
blkerby: Done.

kjbranch:
Should `h_canArtificialMorphMovement` require:
- `canArtificialMorph + (SpringBall or h_canArtificialMorphIBJ)`
- or `Morph`?

logicalRequirements.md
```diff
@@ -343,6 +343,68 @@ __Additional considerations__
...
+    * Samus must have non-zero reserve energy.
+    * Any additional requirements in the `requires` property of the `leaveWithGModeSetup`.
+  * A `leaveWithGMode` object must satisfy the following requirements in order to match:
+    * The `mode` in the `comeInWithGMode` object must be "indirect" or "any".
+    * If `artificialMorph` is `true`, then either the `leavesWithArtificialMorph` property of the `leaveWithGMode` object must be `true` or there is an additional requirement that the Morph item be collected.
+    * Any additional requirements in the `requires` property of the `leaveWithGMode` object.
+* In the case of direct G-mode, `comeInWithGMode` object implicitly requires an energy drain caused by the reserve trigger and preceding setup:
+  * If the tech `canEnterGModeImmobile` is enabled and the `gModeImmobile` on the corresponding door is satisfied, then Samus' regular energy will become whatever reserve energy she had before the transition, truncated to her maximum amount of regular energy (based on the number of ETanks collected).
+  * Otherwise, Samus' regular energy will become 4, or whatever reserve energy she had before the transition if it was less than 4.
```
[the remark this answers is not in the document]
blkerby: Hmm, how is it backwards? The way I'm thinking of it, before executing the strat to enter direct G-mode, reserve energy can be any positive value (doesn't have to be between 1 and 4), because it is always possible to drain it down into the correct range (4 or less) as part of the strat.

region/region-readme.md
```diff
@@ -145,6 +145,51 @@ Generating a shinespark charge using the door's runway (assuming the runway has
 
 Much like using runways, a `canLeaveCharged` can only be executed if the associated door can be interacted with.
 
+#### leaveWithGModeSetup
+
+Represents the ability to exit through the door while taking damage during the transition, in a pose such that X-Ray can be used on the first frame of control in the next room. Under certain conditions, taking damage in this way can be used to set up R-mode or G-mode in the next room. 
+
+The only known way to achieve this is to use an enemy that can follow Samus into the doorway during the transition. It will not work with enemy projectiles since these do not move during transitions, and environmental damage such as heat, lava, acid do not work as these are not active during the transition. Also note that the damage must happen *during* (not *before*) the transition, so being able to take a hit that knocks Samus into the door transition does not work.
+
+The node property `leaveWithGModeSetup` is an array of objects which each currently have a single property:
```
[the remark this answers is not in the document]
blkerby: I changed it now to say "each of which has one main property", since technically these can also have "note" and "devNote" properties.

blkerby:
> Will this require us to add a `spawnAt` to all sand fall transitions?

Thanks, added a sentence in the doc to clarify that we would exclude these transitions since they have "nodeType": "entrance" instead of "nodeType": "door".

blkerby:
> 



> It looks like the landing site 1->4 transition should have a way to stop at the ship beforehand to refill. Is there a way we could add that, or should we leave it as a fixme?

I added a FIXME for it. I think this is minor enough that we could skip over it for now.

blkerby:
> Should `h_canArtificialMorphMovement` require:
> 
> * `canArtificialMorph + (SpringBall or h_canArtificialMorphIBJ)`
> * or `Morph`?

The `canArtificialMorph` tech requirement is already included in the place where artificial morph is obtained, namely in `comeInWithGMode` for direct G-mode.

kjbranch:
Maybe a devNote to be careful when using `h_canArtificialMorphMovement`.
- Morph will not be a usable alternative to get up very constrained ledges, such as in a morph tunnel.
- Artificial IBJ is not usable for underwater rooms without gravity.

logicalRequirements.md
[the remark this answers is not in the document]
osse101: The way I see it, the `canEnterGModeImmobile` description would apply where reserve->regular energy.  And the main difference is that reserve energy is 1to4 at the point of activating GMode, which could use some stronger wording.  

Manually draining energy would fit under `leaveWithGModeSetup` tbh.
I don't think `reflecting the need` improves the description.

But its not at all an important difference.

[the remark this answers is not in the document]
blkerby: Ah, yeah "reflecting the need" was me starting to type something and then getting sidetracked, haha. Thanks. I added a little more above now to clarify how the reduced energy is coming from the setup in the previous room.

PR #671, opened 2023-04-07
Parlor G-mode strats

blkerby:
This builds on PR #668 and should not be merged until after it.

Adds G-mode strats in Parlor.


region/crateria/central.json
```diff
@@ -1661,6 +2020,31 @@
         {
           "from": 5,
           "to": [
+            {
```
osse101, line 2023: 5 to 2 would also be possible and has a piece of canOffScreenMovement

kjbranch:
This now relies upon `h_canArtificialMorphBombHorizontally`, which is currently in a pending PR into #685. (https://github.com/Bokyubi/sm-json-data/pull/7)

osse101:
👍

PR #672, opened 2023-04-09
Correct Main Hall elevator heat damage

blkerby:
I tested this and found that a suitless trip on Main Hall elevator takes exactly 109 energy in either direction (with no fast elevator patch).


osse101:
Timing it gives 440 frames too

PR #681, opened 2023-04-12
G mode Pit Room

kjbranch:


region/crateria/central.json
```diff
@@ -2599,6 +2648,66 @@
...
+                      "fromNodes": [2],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "PowerBomb",
+                    {"ammo": {"type": "PowerBomb", "count": 5}}
+                  ],
+                  "note": [
+                    "Use Power Bomb horizontal boosts to move toward the left side of the room in artificial morph.",
+                    "Roll off the second-to-last blue platform and lay a Power Bomb against the wall to get into range of the bomb blocks.",
```
osse101, line 2686: Could be helpful to say that this is underneath the platform pirate.
And to be careful bumping the wall pirate when bomb jumping over, or jumping into any spawned lazers.

osse101 (commented):
I think it should be acceptable to softlock yourself in with the missile item.  That would make the bomb blocks an obstacle.

PR #684, opened 2023-04-12
G mode final missile bombway

kjbranch:


osse101:
We're going to want explanations in GMode doc on why seemingly simple things like PBs and artificial morph through tunnel don't work.

kjbranch:
In case you weren't sure, the problem here is you can use artificial morph to place the PB, but the bomb blocks dont break until you leave gmode. But it requires unmorphing to leave gmode. 

I think it would be great to have a doc to explain everything

PR #687, opened 2023-04-13
Add itemNotCollected logical requirement

blkerby:


logicalRequirements.md
```diff
@@ -405,6 +405,21 @@ __Additional considerations__
 Please refer to the sections on `leaveWithGModeSetup`, `leaveWithGMode`, and `gModeImmobile` in [the Region documentation](region/region-readme.md) for a more detailed explanation of these objects.
 
 
+#### itemNotCollected object
+An `itemNotCollected` object represents the need to have not yet collected the item at a given node in the same room. For example, such
+an item could be used to overload PLMs in G-mode assuming the item has spawned. Note that any conditions for the item to spawn (e.g. for
+Wrecked Ship items) are not included in this requirement and would need to be specified separately if applicable. In many situations,
+an `itemNotCollected` requirement should be accompanied by a `canRiskPermanentLossOfAccess`, if it is possible to prematurely obtain
+the item and then get stuck from being able to do the strat.
```
osse101, line 413: "stuck by being unable to do"

logicalRequirements.md
```diff
@@ -405,6 +405,21 @@ __Additional considerations__
...
+
+__Example:__
+```json
+{"requires": [
+  {"itemNotCollected": 1},
```
osse101, line 418: itemNotCollectedAtNode is a more fitting name, if less what is desired by the new object

logicalRequirements.md
```diff
@@ -405,6 +405,21 @@ __Additional considerations__
 Please refer to the sections on `leaveWithGModeSetup`, `leaveWithGMode`, and `gModeImmobile` in [the Region documentation](region/region-readme.md) for a more detailed explanation of these objects.
 
 
+#### itemNotCollected object
+An `itemNotCollected` object represents the need to have not yet collected the item at a given node in the same room. For example, such
```
osse101, line 409: I'd like the explanation of what the integer represents in its own sentence so it stands out more.

logicalRequirements.md
```diff
@@ -405,6 +405,21 @@ __Additional considerations__
...
+an item could be used to overload PLMs in G-mode assuming the item has spawned. Note that any conditions for the item to spawn (e.g. for
+Wrecked Ship items) are not included in this requirement and would need to be specified separately if applicable. In many situations,
+an `itemNotCollected` requirement should be accompanied by a `canRiskPermanentLossOfAccess`, if it is possible to prematurely obtain
+the item and then get stuck from being able to do the strat.
+
+__Example:__
+```json
+{"requires": [
+  {"itemNotCollected": 1},
+  "canRiskPermanentLossOfAccess"
```
osse101, line 419: Could this be an inferred connection?  Collecting an item is always going to be a one time event and it follows any strat using the item can be permanently lost.
Until spacetime beam.

osse101: There are setups for strats like moat CWJ and mission impossible that use the item but you can still do the strat without it.  Maybe a case like as an ammo refill needed to get out it wouldn't be hard required.

osse101 (commented):
Can we support visiting a node containing an item without collecting it and does that matter anywhere?

kjbranch:
> Can we support visiting a node containing an item without collecting it and does that matter anywhere?

I believe that is on the randomizer's side, and is/can be done automatically without explicit logic on here. It is used to go through nodes with a lock on them (Gravity Suit), and in a similar way is used by going to a door node without going through the door.

It will have many g-mode applications, but few others. The only one that comes to mind is GT with fewer missiles.

logicalRequirements.md
kjbranch: > Could this be an inferred connection? Collecting an item is always going to be a one time event and it follows any strat using the item can be permanently lost. Until spacetime beam.

The main concern is that Map Rando is currently converting `canRiskPermanentLossOfAccess` into `never`. In rare cases with an item this won't be true:

In the Crateria Supers Room, going from bottom to top with g-mode using PBs should be in logic, but can't if the item is behind `canRiskPermanentLossOfAccess`. The only reasons to go this route is to obtain the item or to go to the top door. Getting to the top door could be done with x-ray climb. So I wanted to add a strat from the bottom door to the item which requires `canRiskPermanentLossOfAccess` or the x-ray climb. Then this strat could still be in logic if the player has x-ray climb tech turned on.

kjbranch: > There are setups for strats like moat CWJ and mission impossible that use the item but you can still do the strat without it. Maybe a case like as an ammo refill needed to get out it wouldn't be hard required.

We could add a way to represent that an item cannot be avoided doing a strat. It could apply in: Moat, Botwoon E-Tank Room, Brinstar Reserve, Green Bubbles Missile Room, etc.

PR #688, opened 2023-04-13
Gmode helpers - Split tech to no longer require morph

kjbranch:
This will require double checking currently open PRs to not be using tech assuming they include morph

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -75,7 +75,7 @@
       "name": "h_canFly",
       "requires": [
         {"or": [
-          "canIBJ",
+          "h_canIBJ",
           "SpaceJump"
         ]}
       ]
--- a/helpers.json
+++ b/helpers.json
@@ -324,6 +324,196 @@
       ]
     },
     {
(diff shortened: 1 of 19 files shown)


helpers.json
```diff
@@ -309,12 +309,209 @@
...
+    },
+    {
+      "name": "h_canSandIBJ",
+      "requires": [
+        "canSandIBJ",
+        "h_canJumpIntoIBJ"
+      ]
+    },
+    {
+      "name": "h_canArtificialMorphSandIBJ",
```
osse101, line 391: Sand is overloaded in GMode so this is not needed.

osse101 (commented):
helpers could be written as `h_canIBJ: requires[ "h_canArtificialMorphIBJ", "Morph"]` for grouping dependancies a little closer.  But it may be safer to leave them separate.

It is possible to `canSpringBallJumpMidAir` in artificial morph, as well as double springballjump when in liquid physics, but its not like exactly the same skill.  

I'm trusting `FindAll` captured all instances and that's what's been changed here.

PR #691, opened 2023-04-14
add HiJump + SpaceJump strat in East Cac Alley

blkerby:
imyt did this in the race last night: https://clips.twitch.tv/HumbleJoyousSwordTebowing-iWdruU5JdsnF1Q3p


osse101:
Strange this isn't in already since its what you do in RBO

PR #699, opened 2023-04-17
Add leaveWithGMode for Crateria elevators

blkerby:
Elevators can be used while in G-mode (unlike normal doors which cannot be shot open), so they are a way to bring G-mode into a third room.


osse101:
Red Brinstar elevator has the use case of going up and down the elevator to remove the super gate.  Is that shown on the `CaterpillarRoom` side as a free transition back down the elevator, or here with a comeInGMode from [2]?

blkerby:
Good point. I added node 2 to the fromNodes now to take that case into account.

PR #706, opened 2023-04-17
Speedkeep chain up climb

kjbranch:


tech.json
```diff
@@ -801,7 +801,20 @@
...
+          "extensionTechs": [
+            {
+              "name": "canSpeedKeepChain",
+              "requires": [
+                "canSpeedKeep",
+                "canLateralMidAirMorph"
+              ],
+              "note": [
+                "Moving after obtaining a speedkeep by jumping into an air ball and then soft unmorph while holding an angle to maintain the speedkeep.",
+                "After morphing, hold forward until unmorphing to prevent losing the speedkeep."
```
osse101, line 814: It can be mentioned here that with XRay you can speedKeepChain in the opposite direction.

tech.json
```diff
@@ -801,7 +801,20 @@
...
+          "devNote": "This is not a momentum continuing action like canBlueSpaceJump",
+          "extensionTechs": [
+            {
+              "name": "canSpeedKeepChain",
+              "requires": [
+                "canSpeedKeep",
+                "canLateralMidAirMorph"
+              ],
+              "note": [
+                "Moving after obtaining a speedkeep by jumping into an air ball and then soft unmorph while holding an angle to maintain the speedkeep.",
```
osse101, line 813: Its easier to control if you morph while falling, so "air ball" is confusing.
Move instead of Moving.

tech.json
osse101: until unmorphed

tech.json
osse101: Holding up or down is ok too.  That's why you can aim down and then press down again to morph.

osse101 (commented):
Also a version for shineCharge, plus a version for shineCharge with a broken wall.

kjbranch:
I rewrote the speedkeep chain description completely. Let me know what you think of it now.

tech.json
```diff
@@ -801,7 +801,21 @@
...
+          "devNote": "This is not a momentum continuing action like canBlueSpaceJump",
+          "extensionTechs": [
+            {
+              "name": "canSpeedKeepChain",
+              "requires": [
+                "canSpeedKeep",
+                "canLateralMidAirMorph"
+              ],
+              "note": [
+                "The ability to chain multiple speedkeeps in a row. This can be done by jumping and doing a stationary midair morph on the descent, then soft unmorph, then holding an angle during and after landing.",
```
osse101, line 813: Stationary midair morph sounds similar to `canStationaryLateralMidAirMorph` and this isn't that.

The tech should mention aim down pose being safe and you need to transition between angle to down.  That detail can be tricky.

"holding an angle during ... landing" would prevent morphing so you have to release it.  But it is true that during the unmorph angle should be held.  The way its written can be misunderstood.

osse101 (commented):
Also  it is all considered one speedkeep that is being continued.

PR #708, opened 2023-04-18
G-mode Crateria Kihunter Room

kjbranch:


osse101 (commented):
An upward GMode Immobile entry seems possible to me.  Crabs are global so supers will knock them down.

tech.json
```diff
@@ -898,6 +898,15 @@
...
                 }
               ]
+            },
+            {
+              "name": "canDownwardGModeSetup",
+              "requires": [ "canEnterRMode" ],
+              "note": [
+                "Ability to setup an R-mode or G-mode through a downward door.",
+                "Samus needs to be in a standing or walking position with no vertical speed on the first frame in the next room, while taking damage through the transition.",
+                "This is typically done using quicksand, clipping into the floor, or using a thawing frozen wall crawler."
```
osse101, line 908: Pretty sure any type of frozen enemy could work.  Covern, Atomic, a Gamet would be worth testing.

region/crateria/east.json
```diff
@@ -347,6 +387,25 @@
...
+                {
+                  "name": "Get Hit by Thawing Sciser",
+                  "notable": false,
+                  "requires": [
+                    "canTrickyUseFrozenEnemies",
+                    "canDownwardGModeSetup",
+                    {"ammo": {"type": "Super", "count": 1}}
+                  ],
+                  "note": [
+                    "Run from the ledge into the open doorway while hitting the frozen Crab as it thaws and the door transition simultaneously.",
```
osse101, line 403: More precisely, if I understand it right: You hit the frozen crab, then the transition, and the crab unfreezes during the transition.

osse101 (commented):
👍

tech.json
kjbranch: True. I would think any enemy would work, probably anything besides a crawler will end up notable tho. I am particularly curious about the covern/atomic though

PR #717, opened 2023-04-19
G mode Bowling Alley Path & Crab Maze

kjbranch:


region/crateria/east.json
```diff
@@ -2119,6 +2159,17 @@
               "openEnd": 0,
               "usableComingIn": false
             }
+          ],
+          "leaveWithGModeSetup": [
```
osse101, line 2163: A devnote on `leaveWithGModeSetup` top door would be ok

region/crateria/east.json
```diff
@@ -2171,13 +2248,50 @@
...
+                        "h_canArtificialMorphIBJ"
+                      ]},
+                      {"and": [
+                        "SpringBall",
+                        "HiJump",
+                        "Bombs"
+                      ]}
+                    ]}
+                  ],
+                  "note": "Place bombs at the far right, next to the low underwater ceiling."
```
osse101, line 2275: Fighting crabs with bombs sounds fun

region/crateria/east.json
```diff
@@ -2171,13 +2248,50 @@
...
                   "name": "Base",
                   "notable": false,
-                  "requires": []
+                  "requires": [
+                    {"or": [
+                      "canCarefulJump",
+                      "HiJump",
+                      "canWalljump",
+                      "Gravity",
+                      "h_canFly",
```
osse101, line 2291: Gravity alone isn't helpful.

Walljump doesn't seem helpful without grav.
IBJ, HiJump cannot make it out of the water.
SpringballJumpMidAir is pretty hard and I'd give it a canTrickyJump.  Or pair it with grav.

Add `canUseFrozenEnemies`.

But I'd be ok leaving this without requirements.  There's no time pressure and you can retry right away.

region/crateria/east.json
osse101: There's also a global crab for GModeImmobile top door

region/crateria/east.json
kjbranch: I could see inexperienced players thinking this isn't possible. and the careful jump is a _little_ tricky

kjbranch: Hijump/wall jump/gravity are from the upper ledge. Its a lot less tricky to get part way up compared to getting all the way onto the ledge while bootless/suitless. I removed spring ball jump midair, because it can be done with careful jump instead

region/crateria/east.json
```diff
@@ -2120,7 +2120,20 @@
...
+              {"enemyDamage": {
+                "enemy": "Sciser",
+                "type": "contact",
+                "hits": 1
+              }}
+            ]
+          },
+          "devNote": [
+            "FIXME Add leaveWithGModeSetup.",
+            "FIXME Requires Samus to be completely against the left or right side of the doorframe. Cuttently all setups can do this."
```
osse101, line 2135: currently

PR #722, opened 2023-04-20
Tighten acidFrames in Metroid Room 1

blkerby:
The acidFrames for the Fearless Dive here were overly lenient.

This is why, for example, the logic thought SpeedBooster was needed for progression at the start of cheesyboatride's recent seed (https://maprando.com/seed/6m8Q6m2UMFjlzpbTu-57lg/).

### DIFF
--- a/region/tourian/main.json
+++ b/region/tourian/main.json
@@ -642,6 +642,14 @@
                   "requires": [
                     {"acidFrames": 80}
                   ]
+                },
+                {
+                  "name": "Bounce Ball",
+                  "notable": false,
+                  "requires": [
+                    {"acidFrames": 60},
+                    "canBounceBall"
+                  ]
                 }
               ]
             }
--- a/region/tourian/main.json
+++ b/region/tourian/main.json
@@ -629,7 +629,7 @@
                   "name": "Fearless Dive",
                   "notable": false,
                   "requires": [
-                    {"acidFrames": 30},
+                    {"acidFrames": 20},
                     {"or":[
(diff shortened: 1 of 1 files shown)


osse101:
20 seems right for trying to get out of the acid right away with room to minmax for more.
Having Gravity reduces acid frames by about 5 but damage counts so fast in acid.

Right to left will also need a `canCarefulJump` added.  Possibly with a small acidFrame cost as well, or trickyJump.

blkerby:
Just realized the wall jump from right to left is fairly tight so I changed that to canPreciseWalljump; that probably makes it unnecessary to add a canCarefulJump?

Also realized the "Acid Run" strats were also overly lenient. I think this strat shouldn't mean they literally just run through the acid (since that wouldn't make sense) but rather do spinjumps to try to minimize the time in the acid. So I reduced those too. The one going left-to-right is higher because you bonk the lower platform which makes you spend a little more time in the acid.

osse101:
Right, precise walljump better represents what I wanted.

A bounceball left to right could get around the head bonk problem, saving about 30 health.  It is harder than walljump is the only thing.

Then for new strats outside the scope of this PR is the right to left trickyjump into walljump to stay above the acid.
And the left to right full run CWJ + metroid avoid.

blkerby:
Added a bounce ball strat for left-to-right. Confirmed that it can be done with 85 energy on a good attempt (90 more realistically), so 60 acidFrames seems good.

PR #724, opened 2023-04-21
Gmode Moat IBJ

kjbranch:


osse101:
All the various bomb jump across the moat strats would still work, but with the extra step of bomb jumping again to reach the moat pedestal. But this add is still good.

PR #726, opened 2023-04-21
G-Mode West Ocean Rework and Add PB Item Strats

kjbranch:


region/crateria/east.json
```diff
@@ -1658,6 +1665,76 @@
                   ]
                 }
               ]
+            },
+            {
+              "id": 17,
+              "strats": [
+                {
+                  "name": "G-mode Morph",
```
osse101, line 1673: I'd rather these were written as 1 to 17, 4 to 17, and 5 to 17

region/crateria/east.json
```diff
@@ -1833,6 +1878,43 @@
...
+                  ],
+                  "obstacles": [
+                    {
+                      "id": "B",
+                      "requires": []
+                    }
+                  ],
+                  "note": [
+                    "To overload the PLMs, place a PB precisely to the right of the bottom of the second overhang above the door to the Moat.",
+                    "This is at the max jump height without HiJump. Placing the PB higher or lower will not overload the PLMs without many PBs."
```
osse101, line 1911: This should work by placing the PB in range of most of the bomb blocks.

region/crateria/east.json
```diff
@@ -991,13 +991,39 @@
...
+        {
+          "id": 16,
+          "name": "G-Mode Morph Overloaded PLMs Junction (South West)",
+          "nodeType": "junction",
+          "nodeSubType": "junction",
+          "note": "Represents being at the Moat door with G-Mode, PLMs overloaded, and morph or artificial morph"
+        },
+        {
+          "id": 17,
+          "name": "G-Mode Morph Junction Direct (South West)",
```
osse101, line 1011: Indirect isn't possible here

region/crateria/east.json
```diff
@@ -991,13 +991,39 @@
...
       ],
       "obstacles": [
         {
           "id": "A",
           "name": "Shoot Respawning Super Block",
           "obstacleType": "abstract"
+        },
+        {
+          "id": "B",
+          "name": "G-Mode Overload PLMs with the Item",
```
osse101, line 1025: Unclear if this is by powerbombing it or by collecting it.  But it'd have to be powerbombing it.

region/crateria/east.json
```diff
@@ -1734,6 +1784,137 @@
...
+                }
+              ]
+            }
+          ]
+        },
+        {
+          "from": 17,
+          "to": [
+            {
+              "id": 10,
```
osse101, line 1886: 17 would have to go to 14.  And I think you have to end GMode at that point.

osse101 (changes requested):
You need to end GMode before reaching the item at 10 so you might as well link GMode strats to 14.  Then you can use non GMode paths up through the super block or down through the crumble.

region/crateria/east.json
kjbranch: If doors are freely connected we have been using a central junction to represent them as they can share a gmode strat. (Like in the Parlor). We have been assuming a careful jump is free when doing g-mode, as that is a tech that is low-level and can't ever be "turned off" like wall jumps. I put 4 here with 1,5 to keep them together, but it would be equivalent if they were moved to their doors.

region/crateria/east.json
kjbranch: I worked on this a lot yesterday. Placing the PB in this precise location works with 1PB. placing it 1 tile lower takes ~15 PBs and placing it 1-2 tiles higher also takes ~15 PBs.

region/crateria/east.json
kjbranch: Junction 17 is for direct only as it requires the item to be loaded

region/crateria/east.json
kjbranch: The strats that use this obstacle PB the item. I didnt make the name longer since it was just an obstacle, but I could

region/crateria/east.json
kjbranch: I put this from 17->10 and then return from 10->17. 10->14 will either require that you didn't come from below or that you `canRiskPermanentLossOfAccess`. All of the strats from 17->10 require morph and assume you go to 14, cancel gmode, and grab the item. I think this is the most clear way to do it in order to make strats from 14 be more clear.

kjbranch:
I listed my reasoning for each of your suggestions. Let me know which ones you would still like changed.

region/crateria/east.json
osse101: This was just a speculation comment, I didn't mean to act like I knew how it worked.

region/crateria/east.json
osse101: 10 can't return to 17, GMode will have ended.

kjbranch: oh right. looks like its going to 13

region/crateria/east.json
osse101: All these extra requirements show this is not free movement.  So using 13 as a base common junction doesn't make sense to me.

region/crateria/east.json
osse101: Now I'm not sure what this obstacle shows as a concept.  It's an ordinary way to overload PLMs isn't it?

region/crateria/east.json
osse101: I get what you mean now.  13 can freely reach the doors.  So by trying to find a path into 14 its sensible to want to check from 13.  So to me, I'd want to do it this way.  But I still think it is going to check the door node requirements and come back and that has been wrong in the past.

kjbranch:
The logic is pretty complex in this room, unfortunately. I spent a fair bit of time here figuring out what is possible, so let me discuss some of it and hopefully it can be a bit more clear. It is not letting me reply to most of these comments directly, so I'll do it in one long comment here:

- 15 is in direct or indirect gmode. 17 is specifically in direct gmode, as it is used to PB the item. The item only spawns in direct gmode.
- It is possible to get into this room in indirect gmode by entering direct gmode into a room without doorcaps, and then entering this room. Rooms without door caps that come to mind are the Crateria and Maridia tubes (and sand falls, but those are vertical.)
- Going from the bottom to Junction 14 should only have `canRiskPermanentLossOfAccess` if it requires the item to be not obtained in order to do the strat. Otherwise using gmode to go up from the bottom to top half of this room should not have this risk. However, going to the item and returning should not have this loss of access, because there is no reason to go there if the item isnt there. Because of this, there are strats going from 17->10->13 without loss of access, but if you go from 10 up, (10->14->12), there is a risk of loss of access. So I put 10->14 is free if you didnt come from below `{"obstaclesNotCleared": ["B"]}` or, you can come from below and risk loss of access. Unfortunately this makes reading the logic much more obscure, because you would expect 10->14 to be free. However, this is part of the reason that I think it is better to put this from 10->14 instead of 14->12, as 14->12 is a bit more complex and would be more likely someone would read the logic to see what is possible here.
- 10->14 isn't requiring morph because there is no way to get to it without morph, but I can add it on.
- The way that the logic is set up, in order for two nodes to be linked in a single strat, the nodes have to be freely connected. This is because you need to be able to get to any of the doors in the strat in order for the strat to be in logic. G-mode is really tricky, so we have been assuming careful jumps are free, so that would make 1 and 5 freely connected, but not to 4, as 4 needs a wall jump, which could potentially be turned off. These nodes need not be freely connected once in the room, using the strat. There is a fair amount needed in order to get from the right door to the left door while in artificial morph, so that is where all of that is coming from.
    - You may recall, that there was recently an error in the bottom of the Climb. There was a shared strat that required a runway to break the bomb blocks. This strat said you need to have a way to shinecharge at the bottom left door or the bottom right door. There was a way to shinecharge at the bottom left door, but there was no way to get to the bottom left door to do this, so it said the strat was possible coming from the right when it was not.

region/crateria/east.json
```diff
@@ -1734,6 +1784,137 @@
...
+                  "notable": false,
+                  "requires": [
+                    "h_canArtificialMorphIBJ",
+                    "h_canArtificialMorphCeilingBombJump"
+                  ],
+                  "note": "Start the IBJ on the second downward tile from the door. The crumble block is the leftmost flat ceiling tile.",
+                  "devNote": "Without canCeilingBombJump, this would take more than 10 minutes."
+                },
+                {
+                  "name": "West Ocean G-mode Overload PLMs by PBing Missile Item Softlock Risk",
```
osse101, line 1826: This strat should be removed as it is part of 17 now.  Trying to hit the item from an indirect entry would break.

osse101:
The role of 17 makes sense now.  It could help to overexplain in notes what's happening seeing how this took me a while.

17 to 16 to 14 with `risk...access` seems ok.  
17 to 10:  Going directly to the item and requiring a fall to 13 works for avoiding `risk`.  I'm ok either way preventing access through 12 or through 14.

It's interesting how Obstacles A and B are being used like `previousNode` twice here when I don't think they are used in this way anywhere else.
The name for B isn't fantastic when the powerbomb strat which also has `risk...access` doesn't clear B.
`Doing Something Risky :eyes:`

For the smaller stuff:
I'd like `Morph` added to 10 and 14 unless there's a reason you need it not there.

I'd rather see 4->15, 4->17, 5->15, 5->17.  1 could be 1 or 13.  The big `or` looks like you're combining multiple strats into one.  And you don't have to assume techs.

I haven't heard if there is a substantial benefit in having multiple `fromNode`'s in a public junction vs putting the links right on the door yet.

`canComeInCharged` schema expects you to visit the door node before trying anything using that door using a `resetRoom`.

kjbranch:
I moved the starting point of the G-mode strats away from the junction and into the door nodes. I added the morph requirement to 10->14, although this is not something that is/can be done consistently throughout the codebase. 

`previousNode` is something that I am hoping to remove from the logic. It is difficult to set up and computationally intensive to compute and doesn't scale well with obstacles and previous nodes in the same room. 

Obstacles cannot be unset, so the obstacle B remains. however there isnt really a reason to return to Nodes 10/14, so it shouldnt be a problem.

Have you found a location in the schema where it says that `canComeInCharged` requires going to its door node before it can be used?

osse101:
The last thing I think should be changed is the strat in 15->16 requiring the item should be moved to 17->16.
And that would be the only change of substance after all this talk because I'm slow to understand GMode 😸 

I guess resetRoom only ensures you are capable of visiting a door, so maybe it was all in my head.  But that still wouldn't be right for counting resource costs.  Doesn't matter anymore with it getting phased out.

kjbranch:
I have been surprised how complicated g-mode is. I've spent a lot of time lately thinking about it, testing it, and discussing it with Maddo and Bokyubi. Your feedback on the g-mode PRs so far has been incredibly useful.

osse101 (approved):

PR #727, opened 2023-04-21
G-Mode Crateria Supers PB Super Item

kjbranch:


region/crateria/central.json
```diff
@@ -4985,6 +4987,47 @@
                   ],
                   "note": "Quickly grapple to then release grapple on the grapple blocks a few times until they stop working. Be careful not to fall into the spikes.",
                   "devNote": "Grapple cannot be used to get over the spikes, as the PLMs are overloaded too quickly."
+                },
+	        {
+                  "name": "G-Mode Morph Overload PLMs by PBing Super Item",
+                  "notable": false,
+                  "requires": [
```
osse101, line 4994: requires itemNotCollectedAtNode

osse101 (commented):
I'm not seeing the benefit of "B".  Coming from below, there is no reason to reach the top then go back down.  You either want the item or the door and you can't have either and return down.  And from above there is no reason to return back up top.

kjbranch:
I was thinking you could grab the item and go back down. I forgot you have to leave g-mode if the PLMs are already overloaded to grab the item. I'll fix it another way

region/crateria/central.json
```diff
@@ -5112,8 +5151,13 @@
                   "name": "Crateria Supers G-Mode Up with PBs",
                   "notable": true,
                   "requires": [
+		    {"itemNotCollectedAtNode": 3},
                     "canConsecutiveWalljump",
                     "Morph",
+		    {"or": [
+		      "canRiskPermanentLossOfAccess",
+		      "canRightFacingDoorXRayClimb"
```
osse101, line 5159: Adding this option is pretty sketchy.
Keep in mind direct GMode puts you at 4 health, so they need grapple or spacejump to cross the spike pit already so its better to rely on the xray climb strat as the way up here.

kjbranch: Adding this is a way to avoid `canRiskPermanentLossOfAccess`. This could put it in logic if the item in that location is grapple, space jump, e-tank, or ammo to CF.

osse101: I think this is negative E.V.
With the extra xray climb requirement, its not likely to break anything.  Its just a ton of requirements to get around a spike hit.  
It also doesn't follow the get item return to start rule 😢 /s

The model would like to have this added, however. So we'd want a schema object requiring the 2->1xray climb strat be possible as a requirement of this strat, I think.  It wouldn't be wrong to more broadly require a path from 2->1 be available even if you must leave the room, since it is 1->4 that this strat is avoiding.

region/crateria/central.json
```diff
@@ -4985,6 +4982,48 @@
...
+		    {"ammo": { "type": "PowerBomb", "count": 4 }},
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "direct",
+                      "artificialMorph": true
+                    }},
+                    {"or": [
+                      {"and": [
+                        "SpringBall",
+                        {"spikeHits": 3},
```
osse101, line 5001: These spike hits are going to need a crystal flash but I don't that is to be handled here.

osse101 (approved):

PR #732, opened 2023-04-24
Fix spawnAt in top of glass tunnel

blkerby:


osse101:
Does 6 have a PB requirement on it?  Meaning you're never expected to tube clip?

kjbranch:
After spending an hour on this, I realized 5 also has a PB unlock strat... I think that the unlock strat shouldnt be a problem, as this is only to unlock the yield. So traversal from 6->5 can still use tube clip strats, but the flag cannot be set unless a PB is used at 5, 6, or 7.

Does this sound right @osse101 @blkerby ?

blkerby:
The PB requirement on node 6 is only to unlock the node (yielding the tube broken flag). We can still pass through node 6 to get down to other nodes using the tube clip without needing a power bomb.

osse101:
A locked node sounds like you cannot pass through it.  I think there should be `bypassStrats` on 6 and 5 with no requirements.

blkerby:
Unlocking a node is only required in order to interact with it. Following a link from a node to another node is not an interaction on the node, so there is no need for bypassStrats.

osse101:
I can see that being how it works

PR #737, opened 2023-04-25
Revamp WS E-Tank Room, Add canPreciseGrapple

kjbranch:


kjbranch:
I was hoping to define this tech in such a way as to keep it away from other grapple tech, like `canGrappleJump`, and `canGrappleClip`

region/wreckedship/main.json
```diff
@@ -2854,8 +2910,33 @@
...
+                      "f_DefeatedPhantoon",
+                      "SpaceJump",
+                      {"thornHits": 2},
+                      {"and":[
+                        {"thornHits": 1},
+                        {"or":[
+                          "HiJump",
+                          "canWalljump",
+                          "canIframeSpikeJump",
+                          "Grapple"
```
osse101, line 2924: How does the grapple help?  Without a precise fling.

tech.json
```diff
@@ -228,6 +228,11 @@
             "The ability to perform non-trivial movement while Samus is off-camera.",
             "This does not include movement where visual feedback is unimportant, such as simply falling down or holding a single direction"
           ]
+        },
+        {
+          "name": "canPreciseGrapple",
+          "requires": [ "Grapple" ],
+          "note": "The ability to move precisely when grappling to or releasing from a grapplable object."
```
osse101, line 235: More advanced usage of Grapple.
Precisely aiming at grapple points while moving quickly.
Precise control of Samus' speed and trajectory when releasing from a grapplable object.

osse101 (commented):
You should be able to reduce heat frame count with this in some rooms, like croc escape.

region/wreckedship/main.json
kjbranch: A mediocre fling will put you far enough to the right to be able to only take 1 spike hit, even without optimizing iframes

region/wreckedship/main.json
```diff
@@ -2840,6 +2850,52 @@
                       "shinesparkFrames": 58
                     }}
                   ]
+                },
+                {
+                  "name": "Precise Grapple",
+                  "notable": false,
+                  "requires": [ "canPreciseGrapple" ],
+                  "note": "Fling Samus from the first set of grapple blocks to the second set without falling. Requires precise timing."
```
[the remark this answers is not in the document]
kjbranch: I could go either way on this one. I think it is a similar difficulty to the others, and in Map Rando, I'm expecting `canPreciseGrapple` to be in `Very Hard` difficulty. It is also just to avoid a wall jump and thorn hit.

kjbranch:
After this PR is merged, I want to take a second look through the Colosseum.

region/wreckedship/main.json
osse101: I think the expected fling takes you as far as jumping from the ledge.
`canDamageBoost` could be added too.
Or XMode lol.

kjbranch: - Added dboost
- I can grapple further than I can jump, fairly easily. The distance is enough that you can immediately jump again and be out of the spikes. 
- I know it was a joke, but Xmode would likely assume iframe spike jump, so I think it isnt needed here

osse101: I think getting a useful fling is a tech and default grappling is a swing and release.  It's possible that can be an assumed tech; I'm interested in adding more assumed techs like `canJumpMorph`.  In this case there are many intuitive ways to play the room to avoid the second hit that it hardly matters.

osse101: For XMode requirements would look like; `canCarefulJump`, `Morph`, and `canOffScreenMovement`.  It's a frame perfect unmorph on spikes but that's your only interaction with the spikes.  The bounce can be done on the upper ledge by the Etank chozo.

region/wreckedship/main.json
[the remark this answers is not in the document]
osse101: Yeah, a bunch of the other `preciseGrapples` are pretty hard too.

osse101 (approved):

Issue #700, opened 2023-04-17
Add SpaceJump strats in East Ocean


blkerby:
Strats from BobBob:
- Both directions with SpaceJump only: https://vimeo.com/818432456
- With SpaceJump + Ice: https://vimeo.com/818438096

PR #739, opened 2023-04-25
Add canSpaceJumpWaterEscape

kjbranch:


kjbranch:
This addresses #700

region/crateria/east.json
```diff
@@ -2452,6 +2455,29 @@
...
+              "id": 2,
+              "strats": [
+                {
+                  "name": "Space Jump Water Escape with Ice",
+                  "notable": false,
+                  "requires": [
+                    "canSpaceJumpWaterEscape",
+                    "canTrickyUseFrozenEnemies"
+                  ],
+                  "note": "Freeze the Choot and stand on it on the right side of the stalagmite. Jump to the water line and space jump to the Kamer platform."
```
osse101, line 2468: This has to be done in a very particular way.
Its a tricky jump to the right to get above the choot.
Break momentum to align with the wall.
And a small frame window shot timing on the freeze.

But attempts are repeatable without getting hit.
Could be made a notable.

osse101 (commented):
There is a huge skill gap between `bouncing on water` and `turning the bounce into normal spacejump`.  So the tech needs more description.

region/maridia/inner-yellow.json
```diff
@@ -857,7 +869,11 @@
                   "requires": [
                     "canSuitlessMaridia",
                     "HiJump",
-                    "canUseFrozenEnemies"
+                    "canUseFrozenEnemies",
+                    {"or": [
+                      "canWalljump",
+                      "h_canCrouchJumpDownGrab"
```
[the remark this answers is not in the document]
kjbranch: Yes. but you need either a wall jump, crouch jump down grab, or space jump to escape the waterline

kjbranch:
Thanks for the feedback. I think this was worth separating into two tech. I believe it is possible to get out of the water while space jumping without hjb, but it is extremely precise. I am pretty sure I got it in the Red Fish Room, but I can't replicate it - it may only be possible in tidal water.

region/crateria/east.json
```diff
@@ -2408,7 +2408,10 @@
                   "requires": [
                     "canSuitlessMaridia",
                     "SpaceJump",
```
osse101, line 2410: spacejumpWaterBounce?

region/maridia/outer.json
```diff
@@ -4376,6 +4376,25 @@
                     ]}
                   ]
                 },
+                {
+                  "name": "Suitless Space Jump with Wall Jump",
```
osse101, line 4380: Rename

region/maridia/inner-yellow.json
```diff
@@ -872,7 +872,8 @@
                     "canUseFrozenEnemies",
                     {"or": [
                       "canWalljump",
-                      "h_canCrouchJumpDownGrab"
+                      "h_canCrouchJumpDownGrab",
```
osse101, line 875: I still can't get the jump off the fish without `CrouchJumpDownGrab`

osse101 (commented):
I like the split into 2 techs.

`canSpaceJumpWaterEscape` makes sense as a tech but its too bad that in practice its just a much harder way of avoiding `canWalljump`

region/maridia/inner-yellow.json
kjbranch: Whoops. You're right. I was confusing myself

tech.json
```diff
@@ -51,6 +51,21 @@
...
+              "requires": [
+                "canSuitlessMaridia",
+                "SpaceJump"
+              ],
+              "note": "Using Space Jump to bounce along the water line. It can be used to escape the water with a wall jump, or to get onto low platforms just above the water level.",
+              "extensionTechs": [
+                {
+                  "name": "canSpaceJumpWaterEscape",
+                  "requires": [ "canSpaceJumpWaterBounce" ],
+                  "note": "Using Space Jump with precise timing at the water line to escape water without Gravity Suit."
```
osse101, line 66: I'd like a mention of `Space Jump Water Bounce` here.
Converting Space Jump Water Bouncing into regular space jumps without Gravity.  Often paired with HiJump.

PR #740, opened 2023-04-27
Fix requirements in Tourian Escape Room 3

blkerby:


osse101:
Springball also works as a way up.
Really hard to do this room damageless.  Takes running from beams until they go off camera.

PR #741, opened 2023-04-27
Fish Tank Fix Spawn At

kjbranch:
Added a new junction to spawn at when entering the fish tank. 
- The spawn at is necessary because there is no way back up through the door once entering.
- Adding a new node is necessary, because spawning at 1 prevents room entry strats used to go to 5 or 7.

Please double check my logic on if this is valid: There are no ways to get to Node 8, except from the spawn at. Because of this, this is an orphan node - there are no links to it, just from it. If this is valid, then the sanity checks will need to be updated for this case.


miketrethewey:
Notes for Sanity Checks:
346: Gather spawnAts
423: Check if in spawnAts

kjbranch:
I kept a free 4->1 strat, as 4->8 is only possible when entering the room. 
I was debating on the 4->8 arrow, but I thought it was a bit better to be there than to not be there.

osse101:
4->8 arrow should definately be there.  But it could be a different kind of line.  Change up the colour or make it dotted or something.
I was talking about a 4->8 strat.

kjbranch:
If we are going to add more colors, we should also consider event nodes and utility nodes. I can make an issue for it.
As for a 4->8 strat, I think it is slightly better without it, but I could go either way. Adding it would fix the sanity checks, so there's that.

PR #742, opened 2023-04-27
Tighten heatFrames for Space+Screw in Pillar Room

blkerby:
This can be done in both directions with no tanks and no adjacent runways. Right-to-left uses more energy because of the shorter in-room runway.

### DIFF
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -1100,7 +1100,7 @@
                     "h_canNavigateHeatRooms",
                     "ScrewAttack",
                     "SpaceJump",
-                    {"heatFrames": 400}
+                    {"heatFrames": 360}
                   ]
                 },
                 {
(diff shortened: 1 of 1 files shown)


blkerby:
Hmm, actually this was assuming a door-frame runway in the next room (requiring air physics).

kjbranch:
Anyone have a good way to measure the required adjacent runway length if the adjacent runway is submerged?

osse101:
Speed caps at 3 tiles it looks like, so there's a start.

blkerby:
Nice, I tried out that right-to-left strat (breaking the half pillar) and 350 heatFrames works.

Also tested entering the left side used a worst-possible connection (Aqueduct right door, water physics with only a door-frame runway) and was able to get it with 92 energy. So I think 380 heatFrames should be good.

osse101:
I stop in the room with angle held most of the time.  Or walk as little as possible into the room.  
Left side has a pit area that old route uses as a backup runway and I got about a similar time running from there.

PR #752, opened 2023-04-27
Update Acid Chozo Statue

kjbranch:
Coming up from the bottom of acid chozo statue with a CF grapple clip will enable Samus to get to the right door without needing `f_UsedAcidChozoStatue`. It may be possible to go to the statue to save some heat frames, but the CF grapple clip currently requires being `heatproof`


osse101:
I don't see going statue first saving heat frames.

This room should be changed up incase spacejump is no longer required to activate the statue, like in DASH recall.

PR #754, opened 2023-04-28
Colosseum Grapple Changes

kjbranch:


osse101 (commented):
Putting Grapple as an option for Spike Platforming creates a big difficulty difference between branches.  So I like how it was with spike platforms in Suitless Grapple Only.  But that bit of spike platforming could be a thing on its own.
Spike Platforming (Right to Left) is the harder one.  It may even be better to remove both of them because the leftmost jump is so bad.  But grapple would help here pretty much the same amount.

Removing Grapple Only as a notable is fine not that there is a preciseGrapple.

kjbranch:
In general, I would say using the spike platforms is much harder than using the precise grapple. This made it so having damage boost enabled bypassed the precise grapple strat, moving this tech from a Map Rando difficulty of Very Hard to Hard.

In it's new location, all of the same tech is required whether or not grapple is owned. Grapple is only used to bypass 2 spike hits. and the notable description is still the same, so players should toggle it independently of this bypass being added or removed.

osse101:
Right side spike platforming is very hard tech at least, so it's not getting represented correctly through techs.

`Colosseum Spike Platforming (Right to Left)` would be benefitted by grapple in the exact same way, so it should be added there for consistency.

`Colosseum Spike Platforming` is an extreme tech because of the frame perfect jump off spikes where a miss is a reset.  Adding grapple removes the hard part and that's too big a difference.  Making a new grapple+spikes strat would be more balanced.

PR #755, opened 2023-04-28
Slightly reduce heat frames in Screw Attack Room

kjbranch:


osse101 (commented):
My numbers:
screw +prepare = 90
screw = 120
PBs = 190
Bombs 180

Adding prepare is smart but it also means 3->4 wouldn't have the extra cost either.

kjbranch:
Is this how you're recommending it be done? 

I think the prepare is pretty important on the 2->3 because otherwise you fall on entry and then need to start another jump to break the blocks. I don't see that being as impactful at the top door (3->4)

It would be nice to make a 2->1 link for going through the room without needing the item, but that would require a lot of strat duplication or another node. I am keeping it in my todo list for now.

osse101:
3->4 expands into
3->2 `heatframes:200` + 2->4 `heatframes:90 + heatframes:30`
It's a smaller percentage of energy cost, but the same amount.  A direct 3->4 with screw may be how you work around that.
I see jumping through door, jumping in room, and moonfalling all about the same frame count.  And landing on the lower blocks costing 30frames.

And yeah, there is that extra 60frames being added at 4->1 that doesn't always need to be.

osse101 (approved):

PR #756, opened 2023-04-28
Remove duplicate and unnecessary helpers

kjbranch:


helpers.json
```diff
@@ -533,13 +528,6 @@
         "IBJ is not usable for underwater rooms without Gravity."
       ]
     },
-    {
-      "name": "h_canArtificialMorphCeilingBombJump",
```
osse101, line 537: Isn't this used in climb supers top?

PR #764, opened 2023-05-02
WS Main Shaft g-mode

kjbranch:


region/wreckedship/main.json
```diff
@@ -666,6 +681,51 @@
...
+                          "h_canArtificialMorphIBJ",
+                          "h_canArtificialMorphSpringBallBombJump"
+                        ]}
+                      ]}
+                    ]},
+                    "canOffScreenMovement"
+                  ],
+                  "note": [
+                    "Use the camera scroll blocks just right of the bomb blocks, in order to overload PLMs and go through them.",
+                    "Exit g-mode and travel to the item and back while off screen."
```
osse101, line 727: You can't pickup the item or see what it is with plms overloaded.

osse101 (commented):
Shouldn't LeavewithGModeSetup and GModeImmobile be added with this?

region/wreckedship/main.json
kjbranch: Right, but it will still be off screen, unless you have morph. 

I'm planning on adding gmode immobile and g-mode setup strats in the next PRs.

osse101: Ok, I didn't see the difference between strats.  The other one has morph.

The next question is how you avoid spikes blind.  The platform enemies won't move.  And I think the conveyers aren't active either in gmode.

osse101: There's a good chance regular 9->8 needs `canCarefulJump` with power on, something to test.

kjbranch: Dodging the spikes will be very annoying. I am thinking of having `off screen movement` be relatively complex stuff, and possibly be in Insane difficulty on Map Rando. It's either that or have many notable off screen strats. Even Basement right to left is challenging while off screen

osse101: Yeah that's fair.  Even simple offscreen movement is really hard.

PR #766, opened 2023-05-02
WS G-Mode Immobile

kjbranch:


region/wreckedship/main.json
```diff
@@ -1342,7 +1509,29 @@
...
+                    "enemy": "Covern",
+                    "type": "contact",
+                    "hits": 1
+                  }},
+                  { "not": "f_DefeatedPhantoon" },
+                  "canRiskPermanentLossOfAccess"
+                ]}
+              ]}
+            ],
+            "note": "Use a Covern or Kihunter. A flying Kihunter takes about 25 seconds to get to the door.",
```
osse101, line 1532: I timed 12 seconds.

osse101 (dismissed):

region/wreckedship/main.json
kjbranch: Good catch. Looks like I can get it to range between 11 and 27 seconds. I updated the description, I mostly just didnt want people to think it was broken because the 2 ground Kihunters don't hit Samus.

PR #767, opened 2023-05-02
WS G-Mode Setup

kjbranch:
This does not include any vertical doors. They will need to be done later

### DIFF
--- a/region/wreckedship/main.json
+++ b/region/wreckedship/main.json
@@ -367,7 +367,15 @@
                 {
                   "name": "Get Hit By Covern or Atomic",
                   "notable": false,
-                  "requires": []
+                  "requires": [
+                    {"or": [
+                      "h_canBombThings",
+                      {"and": [
+                        { "not": "f_DefeatedPhantoon" },
+                        "canRiskPermanentLossOfAccess"
+                      ]}
+                    ]}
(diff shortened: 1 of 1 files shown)


region/wreckedship/main.json
```diff
@@ -317,6 +361,17 @@
               "openEnd": 0
             }
           ],
+          "leaveWithGModeSetup": [
+            {
+              "strats": [
+                {
+                  "name": "Get Hit By Covern or Atomic",
```
osse101, line 368: The atomic takes going through the morph tunnel.

region/wreckedship/main.json
```diff
@@ -1552,6 +1673,17 @@
               "devNote":  "This is unlocked and yielded from killing the enemies, which are free to kill"
             }
           ],
+          "leaveWithGModeSetup": [
+            {
+              "strats": [
+                {
+                  "name": "Get Hit By Covern, Atomic, or Kihunter",
+                  "notable": false,
+                  "requires": []
```
osse101, line 1682: bottom doors use ice afaik
maybe you meant right door

kjbranch: I did indeed. Thanks, I believe those are both fixed now

PR #771, opened 2023-05-04
Add SpaceJump and Grapple options for going up Aqueduct

blkerby:


osse101:
There are a couple of other holes in this rooms' strat coverage.  
Right door with Gravity needs some tech/item to reach the door.  
Top and right door need shinesparks added.  
And all the other oddball ways to reach these places.

PR #772, opened 2023-05-06
Update Tourian Escape Room 4

kjbranch:
Eventually we can add some sort of tech or strats to killing and dodging the tedious space pirates in rooms 3 and 4.

### DIFF
--- a/region/tourian/main.json
+++ b/region/tourian/main.json
@@ -3915,7 +3915,10 @@
                   ]
                 }
               ],
-              "note": "The acid starts rising when Samus is at the bottom of the long climb."
+              "note": [
+                "The acid starts rising when Samus is at the bottom of the long climb.",
+                "The right-side pirates can be killed before going down to prevent having to dodge their lasers. Alternatively, it can help to have them shoot before starting the climb."
+              ]
             }
           ]
         }
--- a/region/tourian/main.json
+++ b/region/tourian/main.json
@@ -3711,13 +3711,11 @@
                   "name": "Base",
                   "notable": false,
                   "requires": [
-                    {
-                      "or": [
-                        "HiJump",
-                        "canWalljump",
-                        "h_canFly"    
-                      ]
-                    }
+                    {"or": [
+                      "HiJump",
+                      "canWalljump",
+                      "h_canFly"    
+                    ]}
                   ]
                 },
                 {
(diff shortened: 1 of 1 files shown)


region/tourian/main.json
```diff
@@ -3876,13 +3879,43 @@
...
+                      "HiJump"
+                    ]},
+                    {"or": [
+                      "canConsecutiveWalljump",
+                      {"and": [
+                        "SpaceJump",
+                        {"or": [
+                          "HiJump",
+                          {"and": [
+                            "canCarefulJump",
```
kjbranch, line 3903: is this doable without acid frames with space jump and no wall jumps?

region/tourian/main.json
```diff
@@ -3876,13 +3879,43 @@
...
+                      "canCarefulJump",
+                      "ScrewAttack",
+                      "Morph",
+                      {"enemyDamage": {
+                        "enemy": "Tourian Space Pirate (all)",
+                        "type": "contact",
+                        "hits": 1
+                      }}
+                    ]},
+                    {"or": [
```
osse101, line 3891: What part of the room is this `or` helping with?

region/tourian/main.json
osse101: With min maxing you could.  It's probably better to have the acid frames here.  Until there is a `canOptimizeMovement` tech.

region/tourian/main.json
kjbranch: that's for climbing the central shaft while avoiding pirate lasers

osse101: If you kill the right side pirates then stick to the right, I don't know that there are lazers to dodge.  But I like keeping these options here.

region/tourian/main.json
```diff
@@ -3876,13 +3879,43 @@
...
+                          ]},
+                          {"acidFrames": 100}
+                        ]}
+                      ]}
                     ]}
                   ]
                 }
               ],
-              "note": "The acid starts rising when Samus is at the bottom of the long climb"
+              "note": "The acid starts rising when Samus is at the bottom of the long climb."
```
osse101, line 3918: A practical note on baiting the right side pirates to shoot right before you go up main shaft would be nice, but logically you can just kill them 🤔.
And a strat for falling through left pirates depends somewhat on how fast and on which side you climb main shaft.

osse101 (approved):

PR #773, opened 2023-05-08
Hero shot

kjbranch:


region/norfair/east.json
```diff
@@ -423,6 +423,7 @@
                   "requires": [
                     "h_canNavigateHeatRooms",
                     "Plasma",
+                    "canHeroShot",
```
osse101, line 426: I see this now as  `OR:[ Plasma, canHeroShot]`
Plasma's value is in getting around heroShotting.
Hero shot shinespark loses value as a notable with this tech too.

osse101 (commented):
Trying to think of more instances:
`Red tower Missile+Spark to the top`
`Ocean Fly` - not implemented
Breaking some blocks with Hyper?
Shooting before sparking so the door opens during the shinespark crash animation is a sorta common thing.  It saves a few frames making it pretty minor for heat frame counts.

This is likely not the same as shooting then bringing an enemy on camera.  But that does have some fun hero vs villain energy.
So it seems like `canHeroShot` also has a timing component where you wait for the shot to be in the correct position before following it.

This could apply to using plasma to get double hits.  Its kinda different because plasma doubles is shooting a intentionally slow shot, then following it.

kjbranch:
I updated the description and fixed a couple examples with your feedback. The reasoning for adding this tech was for using it to determine if players can use hyper beam to break the bottom PB blocks from the left in Big Pink.

We could expand the description to include information about enemy shots, but I think those are typically a lot less difficult and would be better without the tech. Using it precisely to double hit enemies is interesting though.

tech.json
```diff
@@ -799,7 +799,7 @@
         {
           "name": "canHeroShot",
           "requires": [],
-          "note": "The ability to follow a shot to prevent it from despawning before hitting its target."
+          "note": "The ability to time and position a shot, then follow it to move the camera to prevent it from despawning before hitting its target."
```
osse101, line 802: I think I like the first description more.  More details could be added as extra.
```
The ability to follow a shot to prevent it from despawning before hitting its target.
This may include aiming between enemies or waiting for the shot to travel some distance before following it.
```

PR #775, opened 2023-05-08
Add canWaterShineCharge, Fix some strats in Aqueduct

kjbranch:


tech.json
```diff
@@ -894,6 +894,17 @@
...
+        },
+        {
+          "name": "canWaterShineCharge",
+          "requires": [
+            "canSuitlessMaridia",
+            "SpeedBooster"
+          ],
+          "note": [
+            "Starting from an air environment, run into water and charge a spark in the water. This is typically done while crossing a door transition from a room with air physics to a room with water.",
+            "This is easy to do; the tech just represents knowledge that it is possible."
```
osse101, line 906: Techs tagged with `obscure` are like correct to be where they are now, but we'll want to revisit once some amount of time has passed for this to become more normal.

osse101 (approved):
There's a way to spark from 2 to 3 by running past the snail to activate it, then jump down and over it before sparking.  I think I tried it before but had enough trouble that I just put it off, and it just avoids `canPlayInSand` for some complex movement.

PR #776, opened 2023-05-08
Add notable strat in Red Tower for R-mode frozen beetom x-ray climb

blkerby:
Video by Bobbob: https://vimeo.com/810399620

Oatsngoats did this on stream last night and got it on his first attempt.

### DIFF
--- a/region/brinstar/red.json
+++ b/region/brinstar/red.json
@@ -240,6 +240,7 @@
             "Gain R-mode while entering the room.",
             "Use the respawning bugs to refill reserve energy.",
             "Get grabbed by the Beetom and carry it to the 4-tile high gap one screen above the bottom-left door.",
+            "Position Samus one pixel to the right of being against the wall.",
             "Jump and aim down, reaching the ceiling at the same time that reserves are triggered.",
             "This will force Samus to stand up, briefly clipping into the wall above.",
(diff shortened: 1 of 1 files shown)


osse101 (commented):
A `canRModeForceStand` is reasonable to package together with the `canEnterRMode` tech.

This needs to indicate you leave the room with `Reserve Health minus a Beetom hit and some Ripper hits.`  
The note should clarify where you stand before jumping with the beetom because it is pixel perfect on X.
Bob indicated in the video description there is a way to do the strat by getting the beetom stuck in the wall and walk from the frozen ripper.  Something to test and verify.

blkerby:
Yeah I think forced stand can be assumed as part of canEnterRMode (similar to how with G-mode we assume they can overload PLMs, rather than having a separate tech).

I added enemyDamage requirements to give the correct final health. The comeInWithRMode includes reserve energy being transferred to regular energy, and since since the farm is freely available and can be used before doing the strat, reserves can be full at the start. So this makes it unnecessary to explicitly model the farm in the middle of the strat.

I added a clarification about the X position in the note.

I tried that other strat from the video description but wasn't able to get it to work (probably I've misunderstood part of it). Maybe we can just add this one for now?

osse101:
Yep, it would just be an alternate way to do the same thing.

PR #781, opened 2023-05-09
Add canStutterWaterShineCharge, use in Sponge Bath, Outer Maridia

kjbranch:


tech.json
```diff
@@ -908,6 +908,17 @@
           "note": [
             "Starting from an air environment, run into water and charge a spark in the water. This is typically done while crossing a door transition from a room with air physics to a room with water.",
             "This is easy to do; the tech just represents knowledge that it is possible."
+          ],
+          "extensionTechs": [
+            {
+              "name": "canStutterWaterShineCharge",
+              "requires": [ "canWaterShineCharge" ],
+              "note": [
+                "Performing a canWaterShineCharge in a shorter runway by using a stutter.",
```
osse101, line 917: `...using a stutter that is greatly extended by the water slowdown,`

tech.json
```diff
@@ -908,6 +908,17 @@
...
+                "Perform a stutter shortly before the water entrance, typically a door transition, by briefly releasing and then repressing forward. This allows Samus to enter the water with little momentum.",
```
osse101, line 918: ...repressing forward before, but as close to, the transition as possible while continuing the animation cycle.
This carries the stutter into the water where Samus will run in place while building the shinecharge.

osse101 (commented):
AdjacentRunway needs to be set to `physics:air`

region/maridia/outer.json
```diff
@@ -1001,9 +1002,35 @@
...
+                      "useFrames": 80
+                    }}
+                  ],
+                  "obstacles": [
+                    {
+                      "id": "A",
+                      "requires": []
+                    }
+                  ],
+                  "note": "This can be done with only a door-frame runway in the adjacent room. It can technically be done without a stutter, but it's not really easier."
```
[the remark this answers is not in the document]
osse101: This is more of a devnote.
And there will also be a canWaterShineCharge strat with possibly more runway.

region/maridia/outer.json
```diff
@@ -1994,6 +2074,40 @@
...
+                {
+                  "name": "Suitless Water Shinecharge",
+                  "notable": false,
+                  "requires": [
+                    "canWaterShineCharge",
+                    {"or": [
+                      {"canShineCharge": {
+                        "usedTiles": 33,
+                        "shinesparkFrames": 43,
+                        "excessShinesparkFrames": 26,
```
osse101, line 2087: We should add another Nodewith strats to deal with the pirate.

region/maridia/outer.json
```diff
@@ -4640,13 +4814,59 @@
...
+                      "useFrames": 80
+                    }}
+                  ],
+                  "obstacles": [
+                    {
+                      "id": "B",
+                      "requires": []
+                    }
+                  ],
+                  "note": "Start the dash as close to the door transition as possible in order to have a larger window for storing the shinecharge.",
```
osse101, line 4865: Its unclear what this means with a runway of 1.5 tiles.
And I can't do it without a water stutter.  I need 6 air tiles to shortcharge with.

osse101 (commented):
I'm not sure if the mamaturtle room additions is supposed to be `waterStutterShineCharge` or not, but where there is a stutter strat, there should also be a canComeInCharged strat using `canWaterShineCharge` with the amount of runway not yet determined.
Sponge bath is a funny one for that because half of the runway isn't underwater, but you could cross-room run into the water and shinespark up that way.

region/maridia/outer.json
kjbranch: the shinespark strat often avoids or kills the pirate, but it's a complex relationship. While doing a midair shinespark for example, sparking immediately at the peak of your jump, there is a way to kill or completely avoid the pirate when starting with 42-45 (sparking up left), 51, or 58+ energy. In general though, with low energy it is probably best to spark up left and careful jump past the pirate.

region/maridia/outer.json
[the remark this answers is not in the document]
kjbranch: I thought this could be useful to the player to know, but I'll move the second half to a dev note

kjbranch:
We were thinking at the adjacent runway's physics could be normal, as you can turn off gravity during the stutter. Although, in most cases, simply having gravity will make the runway long enough without the water shinecharge.

I wasn't planning on putting in the failures, but I can if you think it's necessary. I personally find the failures pretty useless, as there are 100 ways to fail any strat, particularly in heated rooms, and  I find it unlikely that data will ever be used.

osse101:
`physics:normal` works here if turning off gravity is explicitly mentioned in the tech description for waterstutter.  `canWaterShineCharge` will just about always work better leaving gravity on.  

`canComeInCharged` Doesn't support the transition to water extending the amount of time it takes to build a charge, but I would like if it were written this way.  `LogicalRequirements.md` would need to be updated to talk about runways in water.  Or you could add `adjacentRunway` entries with larger `usedTiles` amounts.  And the amount of shortcharge a player can do determines the needed runway, which can sortof be gotten from a full air runway length.  But some of the shortcharge gets cut off by the transition 🤔 

Failures are whatever, I just saw that if you fall you won't be able to get back up.  And it is a commonly missed spark but you can probably land on an upper platform if you aim poorly.

region/maridia/outer.json
osse101: There's some risky business going on here.  
Sparking to 6 avoids `stationaryLateralMidAirMorph` with springball, and lets you freeze the fish to 5 without jump assist.
Springball can jump around however so this is good for that; maybe a little scary around the pirate at 29 health.
To reach the top of 6 you need `crouchjumpdowngrab` and to pass through the pirate, when you're at 29 health.  From there you could freeze the fish, to save a few shinespark frames reaching 5.

The vertical spark is in the same situation.  Just needs to also not kill the fish.

region/maridia/outer.json
```diff
@@ -1994,6 +2102,84 @@
...
+                    ]},
+                    {"adjacentRunway": {
+                      "fromNode": 3,
+                      "usedTiles": 0.5,
+                      "physics": ["normal"],
+                      "overrideRunwayRequirements": true,
+                      "useFrames": 80
+                    }},
+                    {"or": [
+                      "canCarefulJump",
```
osse101, line 2171: Getting around the pirate is some advanced tech, a little more than a tricky jump, with crouch jump.  Or tricky+morph.

region/maridia/outer.json
```diff
@@ -1994,6 +2102,84 @@
...
+                      { "enemyDamage": { "enemy": "Pink Space Pirate (standing)", "type": "contact", "hits": 1 } },
+                      { "enemyKill": {
+                        "enemies": [ [ "Pink Space Pirate (standing)" ] ],
+                        "explicitWeapons": [ "Plasma", "ScrewAttack" ]
```
osse101, line 2175: Screw doesn't work underwater

region/maridia/outer.json
kjbranch: I now see why this is so difficult. There are only a handful of useful strats leaving 6, and they mostly all make this not important:
midair spring ball up left,
going up right and then using a midair spring ball or the insane jump or plasma+ice ...

The only strat that doesnt play nice is using the frozen fish to get up left. The reason this one is annoying is because it is far easier to get over the pirate if the fish is already killed.

kjbranch:
I dont think we are anywhere close to being able to represent using the air side for most of the runway and then the water side as a small portion of the runway. I think using the air side as the full runway will be correct, and probably the best we can do for now, but might be a little longer than necessary.

One hacky way we could represent this, although I wouldn't want to, is to have several options in the adjacent runway:
- adjacent runway tiles: `15`, can shinecharge: `16`
- adjacent runway tiles: `16`, can shinecharge: `18`
- adjacent runway tiles: `18`, can shinecharge: `20`
- ...

osse101:
I forgot these runways have a `Gravity` requirement.  So how it is written is fine.

kjbranch:
I couldn't think of a good way to add the pirate dodge strat. It seemed too hard for a `tricky jump`, and too niche for a notable. And the fish just made things more complex... So I just removed it for now.

kjbranch:
I couldn't think of a good way to add the pirate dodge strat. It seemed too hard for a `tricky jump`, and too niche for a notable. And the fish just made things more complex... So I just removed it for now.

region/maridia/outer.json
```diff
@@ -1994,6 +2102,84 @@
...
+                      ]}
+                    ]},
+                    {"adjacentRunway": {
+                      "fromNode": 3,
+                      "usedTiles": 0.5,
+                      "physics": ["normal"],
+                      "overrideRunwayRequirements": true,
+                      "useFrames": 80
+                    }},
+                    {"or": [
```
osse101, line 2170: A movement tech is needed here.  A no movement tech path would go to 5 and come back.
h_crouchjumpdowngrab
ice+plasma
springballjump
Hijump would use a different strat.

region/maridia/outer.json
```diff
@@ -4626,7 +4865,14 @@
                   "notable": false,
                   "requires": [
                     "h_canNavigateUnderwater",
-                    "canSpringwall"
+                    "canSpringwall",
+                    "canResetFallSpeed"
```
osse101, line 4869: I'm sure there are many places to add `canResetFallSpeed` following a springball jump or HBJ.
Here, it can be done without the unmorph but its a little tricky to do the jump well enough.  Or you could turn springball off again and springfling.
You can also reach the item with springballjump without a walljump 🤔 Even land on the pedestal with a springfling.

tech.json
```diff
@@ -917,7 +917,7 @@
                 "Performing a canWaterShineCharge in a shorter runway by using a stutter that is greatly extended by the water slowdown.",
                 "Perform a stutter shortly before the water entrance, typically a door transition, by briefly releasing and then repressing forward repressing forward before, but as close to, the transition as possible while continuing the dash.",
                 "This carries the stutter into the water where Samus will run in place while building the shinecharge, Allowing Samus to enter the water with little momentum.",
-                "After entering the water, dash should be held the entire time."
+                "After entering the water, dash should be held the entire time. It is possible to do this between two adjacent submerged rooms with Gravity Suit, by removing the suit just before the transition."
```
osse101, line 920: More generally, you would do the stutter before the door transition.

Turning off Gravity the frame before entering a door transition shouldn't be encouraged.  It raises trick difficulty a notch, and can give you the black screen which may be confusing.

PR #784, opened 2023-05-11
Suitless Plasma Spark

kjbranch:


osse101 (commented):
This is kinda close to `canShinechargeMovementComplex`.  Jump into position, jump up to platform.  Avoid owtch.

This is a 2->5 strat that stops early.  A 2->5 version should be added too, because that is useful for getting to 3 with nothing.

The note for standing closer to the transition would help here.  Standing 0.5 tiles away has a high chance of getting stopped by the owtch.

PR #787, opened 2023-05-15
Water Shortcharge - Pink/Green Maridia

kjbranch:


region/maridia/inner-pink.json
```diff
@@ -156,15 +156,28 @@
...
-                  ]
+                    }},
+                    {"or": [
+                      "canCrouchJump",
+                      "Gravity",
+                      "HiJump",
+                      "canSpringBallJumpMidAir"
+                    ]}
+                  ],
+                  "devNote": "This does not require canWaterShineCharge, as you can store the shinecharge before entering the room."
```
osse101, line 167: Oasis has the same situation.  Many places probably do.
`canWaterShineCharge` may need to be rolled into `canComeInCharged`.  
There is charging a spark before the room.  Charging the spark in room with gravity.  And Charging the spark in room with `canWaterShineCharge`.  And its a pain to separate those cases here.

It looks like Gravity+Shinespark without canSuitlessMaridia is not a working combination across strats too.

region/maridia/inner-green.json
```diff
@@ -324,6 +324,55 @@
...
+                {
+                  "name": "Suitless Water Shinecharge",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
+                      {"and": [
+                        "canWaterShineCharge",
+                        {"or": [
+                          {"canComeInCharged": {
+                            "fromNode": 1,
```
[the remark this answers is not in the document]
osse101: side note: "Diagonal Spark to the Top" is going to be used instead of this branch, other than the `excessShinesparkFrames`

region/maridia/inner-pink.json
```diff
@@ -156,15 +156,28 @@
...
                 },
                 {
                   "name": "Crab Shaft Crab Climb",
                   "notable": false,
                   "requires": [
                     "canSuitlessMaridia",
-                    "canTrickyUseFrozenEnemies"
+                    "canTrickyUseFrozenEnemies",
```
osse101, line 174: if you had to, could you use the crab to reach the door?

region/maridia/inner-pink.json
```diff
@@ -2736,6 +2816,7 @@
                   "name": "Shinespark",
                   "notable": false,
                   "requires": [
+                    "canSuitlessMaridia",
```
osse101, line 2819: I don't see this as a need to know tech to do the strat.
But there could be a divide on number of frames remaining because it does take longer to turn around without gravity.  But this looks like a vertical spark and not diagonal 🤔

region/maridia/inner-pink.json
```diff
@@ -4015,6 +4163,39 @@
                     "canSpringBallJumpMidAir"
                   ]
                 },
+                {
+                  "name": "Suitless Water Shinecharge",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
+                      {"and": [
+                        "canWaterShineCharge",
```
osse101, line 4172: I'd call this section a dupe of "Suitless Shinespark"

osse101 (commented):
`FramesRemaining: 180` is requiring a long dry room runway, but that should already be the case for the `canComeInCharged` that don't have a `canWaterShortcharge` requirment ... for now.

region/maridia/inner-green.json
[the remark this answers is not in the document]
kjbranch: Right. If it is possible to do the strat with a come in charged, the water shinecharge shouldnt be a strat. I will delete it here. I think the others I was more careful on, but I'll check as I go through the feedback

region/maridia/inner-pink.json
kjbranch: canSuitlessMaridia is a very weird tech. It probably should be added and removed from many locations. I dont even think some of the harder underwater tech in the Moat use it ... I'll remove it from here for now, to not introduce an unrelated and unsure change to this PR.

region/maridia/inner-pink.json
```diff
@@ -185,7 +206,7 @@
                     "canSpringBallJumpMidAir",
                     "canUseFrozenEnemies"
                   ],
-                  "note": "Differs from crab climb in that the crabs only serve as one-time stepping stone between platforms."
+                  "devNote": "Differs from crab climb in that the crabs only serve as one-time stepping stone between platforms."
```
osse101, line 209: This note can be removed. 
The note for "Frozen HiJump" as well.

osse101 (dismissed):

PR #788, opened 2023-05-16
Aqueduct Add Strats and Link

kjbranch:


osse101 (commented):
You should be able to grapple over.

You can also tricky jump using the snails to jump from structure to structure.  But I'd rather call that `canSnailClimb`.  I do use it with HiJump to get from center to top left; it could be put as an option instead of flatley.  But if you're doing that, you could fall and climb right side easily enough so maybe not worth.

The rest is good.

kjbranch:
Good catch. I had assumed there was a 2->5 with grapple, but it requires gravity as well.

PR #792, opened 2023-05-17
Add a leniency shot for gate glitches

kjbranch:
A second shot is not much, but it could be a problem to increase this much more, as the player could easily go out of logic if the leniency is too large.

I kept the original and secondary shots separate, so it is more clear what is going on, and which number to increase if the leniency was increased or removed.

Note that in Map Rando, these will still be multiplied by the user's resource multiplier.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -164,8 +164,10 @@
       "name": "h_canGreenGateGlitch",
       "requires": [
         "canGateGlitch",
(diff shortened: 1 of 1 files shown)


osse101:
For comfort Insane would get 1 Super, Extreme 2, Expert 4, VHard 7-8, and Hard probably 3 packs.

PR #793, opened 2023-05-17
Crab Hole Cross Room Jump

kjbranch:


osse101:
The morph slide is very obscure tech.  It would be better to call it a turnaround as well.  As a tech, its benefits are too limited to make it its own thing.  Functionally its the same idea as the `turnAround`.

`canMomentumConservingTurnaround` already have an implied timing component.  These are a little tighter but not by much.  Sliding past the wall to reach the opposite side isn't a common thing but I don't know that it's a big enough deal to make notable.  And there are two ways to do it.
The Kago is something you don't think about so shouldn't be a factor.  As a tech requirement, its role might even be just the knowledge that you can pass through the crab.

region/maridia/outer.json
```diff
@@ -5758,9 +5798,24 @@
...
+                      "fromNode": 3,
+                      "usedTiles": 3.5,
+                      "physics": ["air"]
+                    }}
+                  ],
+                  "note": "Only requires a runway of approximately 3.5 tiles in the adjacent room."
                 }
               ],
-              "note": "This link is only for the X-Ray climb, which skips the junction altogether."
+              "note": "This link is only for the X-Ray climb and cross room jump. All other strats should go 3 -> 2 -> 1 -> 4."
```
osse101, line 5818: This doesn't display anywhere, but `devNote`

region/maridia/outer.json
```diff
@@ -5710,6 +5710,46 @@
...
+                    "canLateralMidAirMorph",
+                    "canMomentumConservingTurnaround",
+                    {"adjacentRunway": {
+                      "fromNode": 2,
+                      "usedTiles": 6,
+                      "physics": ["air"]
+                    }}
+                  ],
+                  "note": [
+                    "Only requires a runway of approximately 3.5 tiles in the adjacent room.",
```
osse101, line 5750: I counted `5`, but I'm not good at counting.  Change the `3.5` though.

osse101 (approved):

PR #794, opened 2023-05-17
Make Green Hill Zone Wall Jump Notable

kjbranch:
I wonder if this should be notable? Because people keep asking about it.

It is a little harder than most precise wall jumps, because of the spawning bugs, and because it feels doable without the pipe.

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -3272,8 +3272,8 @@
                   ]
                 },
                 {
-                  "name": "Green Hill Zone Walljump",
-                  "notable": false,
+                  "name": "Green Hill Zone Wall Jump to Top Right Door",
+                  "notable": true,
                   "requires": [ "canPreciseWalljump" ],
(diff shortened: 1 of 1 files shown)


osse101:
It could be `Medium` notable, but not for hard.  I think you could take off `canPreciseWalljump` if we do that; it's not that precise.

kjbranch:
It's really not that precise. But the other parts of the strat are relatively annoying for that skill level, and it's obscure enough to be one of the most common strat help questions. 

I would be fine leaving it as is, or lowering it to a normal wall jump, but making it notable and keeping it in `Hard`. This would give it more visibility. What do you think?

osse101:
I just suggested removing precisewalljump to remove the `Hard` tech.  The strat is half way to precisewalljump, and also halfway to consecutivewalljump so I could see it being above `Medium` notable.  And I think it is not well known among `Hard` level runners.  That sounds like enough reasons to make a hard notable.
But I do still think a `Medium` runner can execute it alright.

PR #795, opened 2023-05-17
East Cac Space Jump Water Bounce

kjbranch:
This may be possible without a wall jump (space jump into spring ball midair) or without space jump (precise wall jump), but the careful jump is easier.


osse101:
You can springballjump directly out of the water too, but that's a tricky jump.  Having just spacejump there was incorrect.

PR #800, opened 2023-05-18
Add Suitless, Bootless Spring Ball Escape

kjbranch:
https://www.youtube.com/watch?v=LUa-F4WHQR8

### DIFF
--- a/region/maridia/inner-green.json
+++ b/region/maridia/inner-green.json
@@ -2561,6 +2561,22 @@
                     "canSpaceJumpWaterBounce"
                   ],
                   "note": "Space jump while partially submerged for more time to mid-air morph."
+                },
+                {
+                  "name": "Spring Ball Room Escape - Suitless, Bootless, Space Jump",
+                  "notable": true,
+                  "requires": [
+                    "Morph",
+                    "canSuitlessMaridia",
+                    "canSpringBallJumpMidAir",
+                    "canSunkenTileWideWallClimb",
+                    "canTrickyJump",
+                    "canSpaceJumpWaterBounce"
+                  ],
+                  "note": [
+                    "Position Samus to consecutive wall jump without moving horizontally to perform many wall jumps to get to the water surface.",
+                    "Use space jump while partially submerged then perform a tight mid-air morph to escape."
+                  ]
                 }
               ]
             }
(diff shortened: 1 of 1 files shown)


region/maridia/inner-green.json
```diff
@@ -2561,6 +2561,22 @@
...
                   ],
                   "note": "Space jump while partially submerged for more time to mid-air morph."
+                },
+                {
+                  "name": "Spring Ball Room Escape - Suitless, Bootless, Space Jump",
+                  "notable": true,
+                  "requires": [
+                    "Morph",
+                    "canSuitlessMaridia",
+                    "canSpringBallJumpMidAir",
```
osse101, line 2571: Springball isn't in the clip.
I think there is a strat that uses springball instead of spacejump, but I can't do it.

kjbranch: I believe spring ball is needed for the bottom left part of the room. It's just a very standard one. I went back and forth on mentioning it.

osse101: That makes sense then.  I don't think anyone would attempt to walljump morph to get back up to this higher point?

I'd like to see `can3HighMidAirMorph` here as a `canMorphQuickly` tech.

region/maridia/inner-green.json
```diff
@@ -2561,6 +2561,22 @@
...
+                    "canSunkenTileWideWallClimb",
+                    "canTrickyJump",
+                    "canSpaceJumpWaterBounce"
+                  ],
+                  "note": [
+                    "Position Samus to consecutive wall jump without moving horizontally to perform many wall jumps to get to the water surface.",
```
osse101, line 2577: Not sure what this means.  `canSunkenTileWideWallClimb` is done by holding forward into a wall.  Once you're waterBouncing you do have to stop walljumping I suppose.

region/maridia/inner-green.json
```diff
@@ -2561,6 +2561,22 @@
...
+                    "Use space jump while partially submerged then perform a tight mid-air morph to escape."
```
osse101, line 2578: I find trying to escape when the water is at its highest works best.

osse101:
Spacejumping when the water is low into a springball jump is probably a little easier.

region/maridia/inner-green.json
kjbranch: I dont think it should be necessary under water, but either way, this strat will be fine since it has spring ball

PR #802, opened 2023-05-19
Update Blue Hopper Screw Strats

kjbranch:


osse101:
The hoppers are going to follow you in screw attack until you make a mistake.  But I think there is enough supporting notes that someone can figure out how to handle it.

PR #805, opened 2023-05-19
Add Node to Red Fish Room

kjbranch:
This node will be useful when representing vertical cross room gravity jumps


osse101:
Could use some shinespark strats here.  2->1 but also 2->3.

region/maridia/outer.json
```diff
@@ -4640,6 +4634,39 @@
...
+              "strats": [
+                {
+                  "name": "Base",
+                  "notable": false,
+                  "requires": [
+                    "Morph",
+                    {"or": [
+                      "HiJump",
+                      "canWalljump",
+                      "h_canFly",
```
osse101, line 4652: IBJ would also need `canBombHorizontally`.  And then you have 2->1 IBJ+Gravity.

region/maridia/outer.json
```diff
@@ -4572,24 +4578,40 @@
...
                 },
                 {
-                  "name": "Suitless Jump Assists",
+                  "name": "Shinespark",
                   "notable": false,
                   "requires": [
                     "Morph",
+                    {"canComeInCharged": {
+                      "fromNode": 2,
+                      "framesRemaining": 0,
```
osse101, line 4595: I can't think of a door that has a problem with a right side spark exit.

osse101: I barely got a wasteland spark up left side.  But in that case you can diagonal spark, or would be able to gravity jump and the canLeaveCharged is hard enough to assume a lot of tech enabled.

PR #811, opened 2023-05-27
Add west ocean deep stuck x-ray climbs

blkerby:


region/crateria/east.json
```diff
@@ -1167,6 +1167,43 @@
...
             },
+            {
+              "id": 12,
+              "strats": [
+                {
+                  "name": "G-Mode Deep Stuck X-Ray Climb",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [4],
```
osse101, line 1178: Does the little bit of knockback keep you from falling?

tech.json
```diff
@@ -864,6 +864,15 @@
               "requires": [ "canShinespark" ],
               "note": "Manually triggering reserve energy during a shinespark to extend the distance travelled.",
               "devNote": "This tech shouldn't need to be added to strat requirements."
+            },
+            {
+              "name": "canShinesparkDeepStuck",
+              "requires": [ "canShinespark" ],
```
osse101, line 870: `canMidairShinespark` fits a little better.

osse101 (approved):

region/crateria/east.json
blkerby: Seems the knockback doesn't work. Added requirements to get back up. Rather than duplicating all 5 -> 4 strats, I just put two options in "canArtificialMorph" and "canWalljump" since if the player is enabling G-mode they're very likely enabling at least one of those two.

tech.json
blkerby: Good point, changed it to use that now

PR #812, opened 2023-05-27
West ocean ice clip

blkerby:
Clip: https://youtu.be/Bja36hFupMA

I tried this out and it doesn't seem bad, so I'm thinking we don't make it notable.

### DIFF
--- a/region/crateria/east.json
+++ b/region/crateria/east.json
@@ -1490,11 +1490,8 @@
                   ],
                   "note": [
                     "Freeze a bug a three tiles to the right of the morph tunnel entrance, directly under the tile where ceiling is higher.",
-                    "Spin jump onto the bug, crouch, and jump up.",
-                    "With HiJump it is more lenient, otherwise the bug must be at an exact pixel height.",
-                    "If you have XRay and Morph, it can be done more easily:",
-                    "unequip HiJump (if you have it) and freeze the bug high enough that you can roll under it, unmorph, jump to clip up through it;",
-                    "aim down as you jump up through the bug, then morph & unmorph, perform an XRay forced stand up, and jump.",
+                    "Jump onto the bug, crouch, and jump up.",
+                    "With HiJump it is more lenient, otherwise the bug must be at an exact pixel height (unless you have XRay and Morph).",
                     "Note that although the ceiling here is 3 tiles thick, only the bottom tile is solid; the other two are slopes pushing Samus up, allowing the trick to be done."
                   ],
                   "devNote": [
--- a/region/crateria/east.json
+++ b/region/crateria/east.json
@@ -1483,6 +1483,17 @@
                   "devNote": "Could use a failure definition?"
                 },
                 {
+                  "name": "West Ocean Bug Ice Clip",
+                  "notable": false,
+                  "requires": [
+                    "canPixelPerfectIceClip"
+                  ],
+                  "devNote": [
+                    "FIXME: add a more lenient alternative using HiJump",
+                    "We don't use canNonTrivialIceClip here since it includes an XRay standup option which doesn't seem helpful here."
+                  ]
+                },
+                {
                   "name": "G-mode Morph X-Ray Standup",
                   "notable": false,
                   "requires": [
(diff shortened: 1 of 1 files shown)


osse101:
The part that could be notable is that this clip goes through a ceiling that looks 3 tiles high.

osse101:
HiJump made it a lot easier.
Without HiJump, I liked having morph.  If you don't jump just right, then clipping into the bug to position yourself helps.
Having XRay helps a ton.  You need to make the bug rise higher to use it, so aim down when jumping.

blkerby:
Good point about it appearing to be 3 tiles high. I changed it to notable now.

Also changed it to use canNonTrivialIceClip since you're right that XRay can make it much easier. Added more detail to the note to explain what I found to be the easiest setup for the XRay version: freeze the bug high enough that you can roll under it and clip up onto it from below (I think it has to be this high in order to work anyway).

Created issue #817 for modeling how HiJump can help (which needs to be done for most other ceiling clips as well).

region/crateria/east.json
```diff
@@ -1482,6 +1482,24 @@
...
+                {
+                  "name": "West Ocean Bug Ice Clip",
+                  "notable": true,
+                  "requires": [
+                    "canNonTrivialIceClip"
+                  ],
+                  "note": [
+                    "Freeze a bug a three tiles to the right of the morph tunnel entrance, directly under the tile where ceiling is higher.",
+                    "Spin jump onto the bug, crouch, and jump up.",
+                    "With HiJump it is more lenient, otherwise the bug must be at an exact pixel height.",
```
osse101, line 1494: This is more of a `canNonTrivialIceClip` note, but since it is not mentioned there, it is ok here.

region/crateria/east.json
```diff
@@ -1482,6 +1482,24 @@
...
+                    "If you have XRay and Morph, it can be done more easily:",
```
osse101, line 1495: This is normal for `canNonTrivialIceClip` unless you are introducing the bug height needed for xray clip with this.

As an aside, Morph, XRay, and canPartialFloorClip can traverse this link with these items.

region/crateria/east.json
```diff
@@ -1482,6 +1482,24 @@
...
                   "devNote": "Could use a failure definition?"
                 },
+                {
+                  "name": "West Ocean Bug Ice Clip",
+                  "notable": true,
+                  "requires": [
+                    "canNonTrivialIceClip"
+                  ],
+                  "note": [
+                    "Freeze a bug a three tiles to the right of the morph tunnel entrance, directly under the tile where ceiling is higher.",
```
osse101, line 1492: I'd like it pointed out that it looks like a 3 tile high ceiling.  Less important would be to say its one solid tile and two slopes.

blkerby: Done.

region/crateria/east.json
```diff
@@ -1482,6 +1482,25 @@
...
+                    "Spin jump onto the bug, crouch, and jump up.",
+                    "With HiJump it is more lenient, otherwise the bug must be at an exact pixel height.",
+                    "If you have XRay and Morph, it can be done more easily:",
+                    "unequip HiJump (if you have it) and freeze the bug high enough that you can roll under it, unmorph, jump to clip up through it;",
```
osse101, line 1496: Unequipping HiJump can give a better setup for positioning the bug, but its not clear why you would turn it off from how this is written.

region/crateria/east.json
```diff
@@ -1482,6 +1482,25 @@
...
+                    "aim down as you jump up through the bug, then morph & unmorph, perform an XRay forced stand up, and jump.",
```
osse101, line 1497: I'm not sure what "aim down as you jump up through the bug" means.  
Is it to avoid a jump morph?
To get on the bug, an aim down jump seems more niche and not really required compared to jump morph.
And positioning for the clip is done before the jump to the bug and after the morph.

osse101 (commented):
I was thinking this note was too wordy.  But this clip is also different from most so what works is less obvious.

region/crateria/east.json
blkerby: True, that seems to take away most of the reason to detail how to use Morph & XRay in the ice clip. I simplified the note now by cutting that part out.

PR #825, opened 2023-05-31
Yellow Maridia Partial Update

kjbranch:


region/maridia/inner-yellow.json
```diff
@@ -829,51 +829,70 @@
...
-                    "Gravity"
+                    "h_canNavigateUnderwater",
+                    "canTrickyUseFrozenEnemies"
                   ],
                   "note": [
-                    "With Gravity, it's possible to move forward and mid-air morph into the morph passage.",
-                    "The momentum can push Samus into the tunnel, no need to press forward while in mid-air."
-                  ],
-                  "devNote": "FIXME This probably needs to be behind a tech as well?"
+                    "Freeze one of the bugs (called Menus) in a position where you can use it to get into the morph passage.",
```
osse101, line 892: I can't tell how this is meant to be done.
Using two menus you can make a full bridge and jump on top and roll over, but its pretty impossible.
With one menu you can freeze it a little low and do a high jump morph to bounce from the bug to the tunnel.
The way I liked was freezing the menu just a little above the dry patch and waiting for it to unfreeze to damage boost into the hole.

In all cases, it was really hard to do this damageless (didn't try to)

region/maridia/inner-yellow.json
```diff
@@ -829,51 +829,70 @@
...
-                    "That allows passage without having to gain height while being morphed."
+                    "With Gravity, it's possible to move forward and mid-air morph into the morph passage.",
+                    "The momentum can push Samus into the tunnel, no need to press forward while in mid-air."
                   ]
                 },
                 {
                   "name": "Maridia Bug Room Suitless MidAir Morph",
                   "notable": false,
                   "requires": [
                     "h_canNavigateUnderwater",
```
osse101, line 871: canNavigateUnderwater could be removed here too.  You're waiting for the water to go away, not using it.

region/maridia/inner-yellow.json
```diff
@@ -1020,7 +1039,8 @@
                   "notable": false,
                   "requires": [
                     "canSuitlessMaridia",
-                    "canSpringBallJumpMidAir"
+                    "canSpringBallJumpMidAir",
```
osse101, line 1042: max height springball jumps have been called `trickyJump` too.

osse101 (commented):
I don't think power beam is enough to comfortably deal with Menus.  So I could see a future change adding a middle node with strats for avoid/kill menus.

region/maridia/inner-yellow.json
```diff
@@ -588,13 +588,20 @@
                 {
                   "name": "Crouch Jump Down Grab Escape",
                   "notable": false,
-                  "requires": ["h_canCrouchJumpDownGrab"],
+                  "requires": [
+                    "canSuitlessMaridia",
+                    "h_canCrouchJumpDownGrab",
+                    "canTrickyJump"
```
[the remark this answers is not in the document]
kjbranch: That is true. I was debating between tricky and careful, but was thinking tricky fit a little better, as I have noticed a lot of people try and think that strat isn't possible (partly because it isn't on the left side).

region/maridia/inner-yellow.json
kjbranch: This strat was already in there, I just moved it. I never could get it, and thought it was going to be a lot harder than the midair morph option. I believe Map Rando ignores this notable - do you think we should remove the strat altogether?

region/maridia/inner-yellow.json
kjbranch: I don't really think it is as tight as some of the others, it just took me a couple tries. I could have gotten lucky, but I was thinking maybe it's not quite a max height one?

region/maridia/inner-yellow.json
osse101: Ah, strat order got shuffled so I couldn't tell.

The strat should exist imo.  
I like the version where you stand in the water, take a hit for positioning, and freeze the bug near but left of samus.  Then morph on top and damage boost on the unfreeze.  And that would be 2 hits if all goes well.

region/maridia/inner-yellow.json
osse101: Looks like a max height jump to me.  You can actually use the choot for a slightly less precise jump.

PR #826, opened 2023-06-01
Update Plasma Spark Room

kjbranch:


region/maridia/inner-yellow.json
```diff
@@ -1462,22 +1514,25 @@
...
+                },
                 {
-                  "name": "Base",
+                  "name": "Shinespark",
                   "notable": false,
-                  "requires": []
+                  "requires": [
+                    "canShinechargeMovementComplex",
+                    {"canComeInCharged": {
+                      "fromNode": 1,
```
osse101, line 1530: 3

region/maridia/inner-yellow.json
```diff
@@ -1500,6 +1558,33 @@
...
                     ]}
                   ]
+                },
+                {
+                  "name": "Shinespark",
+                  "notable": false,
+                  "requires": [
+                    "canShinechargeMovementComplex",
+                    {"canComeInCharged": {
+                      "fromNode": 1,
```
osse101, line 1568: 3

region/maridia/inner-yellow.json
```diff
@@ -1488,7 +1543,10 @@
                   "name": "Base",
                   "notable": false,
                   "requires": [
-                    "canConsecutiveWalljump",
+                    {"or": [
```
osse101, line 1546: There isn't a 4->5 anymore.  
The only real difference is that you could shinespark from 4, but that strat was missing already.
`powerbombs` was intended to be placed at the ledge at 4, but placing it in the air would work too.  That concept would open up using wide shots (wave or spazer) to hit the choot from safety and reclimbing.

region/maridia/inner-yellow.json
```diff
@@ -1587,6 +1595,25 @@
...
+              "id": 5,
+              "strats": [
+                {
+                  "name": "Shinespark",
+                  "notable": false,
+                  "requires": [
+                    "canShinechargeMovementComplex",
+                    "canMidairShinespark",
+                    {"canComeInCharged": {
+                      "fromNode": 3,
```
osse101, line 1609: 4

PR #830, opened 2023-06-02
Update Maridia Bug Room

kjbranch:


region/maridia/inner-yellow.json
```diff
@@ -814,23 +827,60 @@
...
+              "strats": [
+                {
+                  "name": "Kill the Menus",
+                  "notable": false,
+                  "requires": [],
+                  "obstacles": [
+                    {
+                      "id": "A",
+                      "requires": [
+                        {"and":[
```
osse101, line 855: Missing `or`?

region/maridia/inner-yellow.json
```diff
@@ -814,23 +827,60 @@
...
+                          "canCarefulJump",
```
osse101, line 856: is this a `canDodgeWhileShooting`? Or is this a full avoid strat?

region/maridia/inner-yellow.json
```diff
@@ -858,32 +904,22 @@
...
                   "requires": [
                     "h_canTwoTileJumpMorph",
                     {"or":[
-                      "HiJump",
-                      "h_canFly",
-                      "canWalljump",
-                      "canSpringBallJumpMidAir",
-                      "canTrickyDashJump",
-                      "canTrickyUseFrozenEnemies"
+                      {"enemyDamage": { "enemy": "Menu", "type": "contact", "hits": 1 }},
```
osse101, line 921: I think this should require the obstacle clear.
Meaning, put `enemyDamage` as a way to clear the obstacle.

osse101 (commented):
Good changes.  This is a hard room.

region/maridia/inner-yellow.json
```diff
@@ -907,8 +939,79 @@
...
+                  "requires": [
+                    {"or":[
+                      "h_canIBJ",
+                      "canWalljump",
+                      "canSpringBallJumpMidAir"
+                    ]},
+                    {"or":[
+                      {"enemyKill":{
+                        "enemies": [ [ "Menu", "Menu", "Menu" ] ],
+                        "explicitWeapons": [
```
osse101, line 978: Missile kill is my first choice.

region/maridia/inner-yellow.json
```diff
@@ -907,8 +939,79 @@
...
+                }
+              ]
+            },
+            {
+              "id": 2,
+              "strats": [
+                {
+                  "name": "Base",
+                  "notable": false,
+                  "requires": [ "Morph" ]
```
osse101, line 1012: Does this need something for dealing with Menus?
In testing I was blocked by left menu, but never in runs.
Right menu I always kill with a missile, but it looks like you can wait a second for it to pass.

kjbranch: I think it is simple enough to be techless. There are probably strats of this difficulty all over. 
The first menu can be jumped over, or lured and then avoided or killed pretty easily.
The right menu can be jumped over, or killed from the left water patch.

region/maridia/inner-yellow.json
kjbranch: Good catch. I was thinking this would be harder to aim. I guess it is really simple though

PR #839, opened 2023-06-05
More Updates to Yellow Maridia

kjbranch:


region/maridia/inner-yellow.json
```diff
@@ -2938,7 +2965,7 @@
                         "HiJump",
                         "canWalljump"
                       ]},
-                      "SpaceJump"
+                      "h_canFly"
```
osse101, line 2968: ScrewAttack would have to be separated out as an enemyKill method

osse101 (dismissed):

region/maridia/inner-yellow.json
kjbranch: Good catch

PR #843, opened 2023-06-09
Acid Chozo Statue Partial Update

kjbranch:
In vanilla, the only way to activate the statue is with Space Jump, so strats from 1->3 required space jump or the statue to have already been activated. 

A randomizer may wish to remove this space jump requirement (located on the statue lock), so common non-space jump strats from 1->3 are added.

### DIFF
--- a/region/lowernorfair/west.json
+++ b/region/lowernorfair/west.json
@@ -87,7 +87,7 @@
                     "h_canNavigateHeatRooms",
                     "SpaceJump",
                     "h_canUsePowerBombs",
-                    {"heatFrames": 800}
+                    {"heatFrames": 1000}
                   ]
                 }
               ]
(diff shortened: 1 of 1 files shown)


region/lowernorfair/west.json
```diff
@@ -164,6 +164,28 @@
                     }},
                     {"heatFrames": 125}
                   ]
+                },
+                {
+                  "name": "Acid Dive",
+                  "notable": false,
+                  "requires": [
+                    "canSuitlessLavaDive",
```
osse101, line 172: There would be another strat for having Gravity I guess.

kjbranch: Surprisingly, gravity didnt really change the acid frames, and it would still require `canSuitlessLavaDive`, since it is lava and gravity doesnt fully protect you

PR #853, opened 2023-06-14
Outer Maridia Gmode Immobile

kjbranch:


region/maridia/outer.json
```diff
@@ -739,7 +739,30 @@
...
+                  "enemy": "Sciser",
+                  "type": "contact",
+                  "hits": 1
+                }}
+              ]}
+            ],
+            "note": "Start on the right side and carefully land on the right edge. While in artificial morph, carefully bomb over the crab to avoid a second hit.",
+            "devNote": [
+              "It may be possible to get hit on the left side, but it would be dependent on the strats coming in, sometimes you start too high to get hit.",
+              "All known gmode strats in this room require artificial morph, otherwise the crab could easily be jumped over."
```
osse101, line 763: Springball.  I think your movement options are Grav+IBJ, or HJB+pseudo double springballjumps.
Bombing over the crab would use Gravity too but maybe you could kill it.

And I saw a clip of GCC getting the missile by standing inside the speed blocks.

osse101 (commented):
https://youtu.be/53O3udS18hA
There are other ways to overload PLMs but here's the clip I saw.
Triggering reserves should work too instead of using Ice.

kjbranch:
This is very cool. I was trying to do something like this, but didn't think to use a frozen crab. I would love to see SpeedBooster in this item location 😆

region/maridia/outer.json
kjbranch: I dont believe we have a way to represent this immobile strat, so I am just going to delete it for now. This immobile strat only has extra requirements if the strats that use it need artificial morph, and only those that remain in the room. (Staying up, in morph, will require dealing with the crab, everything else won't)

In any case, there arent even any upward gmode strats yet.

kjbranch:
This should be addressing all of your feedback - overloading the speed blocks will need to be added to #858

PR #855, opened 2023-06-16
Outer Maridia Partial Gmode Update

kjbranch:
This should include all in-room strats for using gmode in Outer Maridia for rooms that have door caps on all doors.
(Gmode Setup, Glass Tunnel, Main Street, Everest, Crab Hole will be in a different PR)

### DIFF
--- a/region/maridia/outer.json
+++ b/region/maridia/outer.json
@@ -2494,7 +2494,7 @@
                   "notable": false,
                   "requires": [
                     {"comeInWithGMode": {
-                      "fromNodes": [4],
+                      "fromNodes": [2],
                       "mode": "any",
                       "artificialMorph": true
                     }},
@@ -5143,6 +5143,18 @@
                       "requires": ["h_canOpenGreenDoors"]
                     }
                   ]
+                },
+                {
+                  "name": "Indirect G-Mode Gate Despawn",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "indirect",
+                      "artificialMorph": false
+                    }}
+                  ],
+                  "note": "The gate will not spawn in indirect g-mode, and can be freely passed."
                 }
               ]
             }
@@ -5164,6 +5176,18 @@
                       "requires": ["h_canGreenGateGlitch"]
                     }
                   ]
+                },
+                {
+                  "name": "Indirect G-Mode Gate Despawn",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [2],
+                      "mode": "indirect",
+                      "artificialMorph": false
+                    }}
+                  ],
+                  "note": "The gate will not spawn in indirect g-mode, and can be freely passed."
                 }
               ]
             }
@@ -5269,6 +5293,26 @@
                   "name": "Base",
                   "notable": false,
                   "requires": [ "Morph" ]
+                },
+                {
+                  "name": "G-Mode Morph",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    {"or": [
+                      "SpringBall",
+                      "Bombs",
+                      {"and": [
+                        "canBombHorizontally",
+                        {"ammo": { "type": "PowerBomb", "count": 3 }}
+                      ]}
+                    ]}
+                  ],
+                  "note": "Use spring ball to cross the room, or use bombs to kill the Zebbos so they stop respawning. One PB can be used below the corner in the ceiling to bomb across the farm while killing the Zebbos."
                 }
               ]
             }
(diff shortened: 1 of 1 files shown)


region/maridia/outer.json
```diff
@@ -2423,6 +2485,40 @@
...
+                        "h_canArtificialMorphSpringBallBombJump",
+                        "h_canArtificialMorphSpringBallBombJump"
+                      ]}
+                    ]}
+                  ],
+                  "note": [
+                    "With spring ball, get over the first pirate then spring ball bomb boost up to the ledge on the right, spring ball onto the platform with the pirate then up and go as far right as possible.",
+                    "One final spring ball bomb boost is needed to get to the top right door. Note that the Pirate's hitbox is taller than it looks.",
+                    "Without spring ball, using IBJ is trickier. IBJ over the pirate while ensuring that it doesnt fire invisible stationary lasers too close to the right.",
+                    "Get over the pirate while it isn't too close to the left wall in order to have time to IBJ up before it touches Samus. Then IBJ multiple times to get to the top right door."
```
osse101, line 2518: I think I take the 2->5 path then kill the fish with bombs; go under the rightish pirate and IBJ up.

region/maridia/outer.json
```diff
@@ -2423,6 +2485,40 @@
...
+                    "Gravity",
+                    {"or": [
+                      {"and": [
+                        "h_canArtificialMorphIBJ",
+                        "canTrickyJump",
+                        "canBombHorizontally"
+                      ]},
+                      {"and": [
+                        "h_canArtificialMorphSpringBallBombJump",
+                        "h_canArtificialMorphSpringBallBombJump"
```
osse101, line 2510: I see this as 3 springballbombjumps.
Or 1 sbbj and landing near two pirates.  But standing near pirates wouldn't need GMode.

region/maridia/outer.json
```diff
@@ -2454,6 +2550,45 @@
...
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [4],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "h_canArtificialMorphIBJ",
+                    "Gravity",
+                    "h_canArtificialMorphBombHorizontally"
```
osse101, line 2586: You've added the tech `canBombHorizontally` in some places and the helper in others, all next to `h_canArtificialMorphIBJ`.

region/maridia/outer.json
```diff
@@ -5134,6 +5293,26 @@
...
+                    {"or": [
+                      "SpringBall",
+                      "Bombs",
+                      {"and": [
+                        "canBombHorizontally",
+                        {"ammo": { "type": "PowerBomb", "count": 3 }}
+                      ]}
+                    ]}
+                  ],
+                  "note": "Use spring ball to cross the room, or use bombs to kill the Zebbos so they stop respawning. One PB can be used below the corner in the ceiling to bomb across the farm while killing the Zebbos."
```
osse101, line 5315: I like bomb jumping from the bottom leftmost plant to jump over the zebbos.  Starting from under the corner gets me hit most of the time.
You could use an extra pb instead of bomb horizontally too.
Killing them with bombs is very not safe.  But bouncing on the bombs to avoid the zebbos helps.

region/maridia/outer.json
```diff
@@ -5722,6 +5921,30 @@
                     }
                   ],
                   "devNote": "Can save an additional 4 energy with HiJump, and a bit more excess frames if not getting on top of the block to go to 2."
+                },
+                {
+                  "name": "G-Mode Morph",
```
osse101, line 5926: You can IBJ without grav from mama Turtle by bombing a baby turtle and springballing up to her back.

region/maridia/outer.json
kjbranch: I didnt even think of landing on the platform with the last pirate. This is just a careful jump, which I have been counting as free in G-Mode

kjbranch: This whole strat is not useful either, as it can be done with gravity alone

region/maridia/outer.json
kjbranch: I'll standardize them. The helper is going to be a bit slower to compute, and adds ammo requirements (which wont matter since theyre with IBJs)

region/maridia/outer.json
```diff
@@ -2240,6 +2240,40 @@
...
+                    }},
+                    "h_canArtificialMorphIBJ",
+                    "Gravity"
+                  ],
+                  "note": "Kill the first fish with bombs then dodge or kill the second one."
+                }
+              ]
+            },
+            {
+              "id": 7,
```
osse101, line 2261: 5 to 7 can be done with grav+careful jump

region/maridia/outer.json
```diff
@@ -2240,6 +2240,40 @@
...
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "h_canArtificialMorphIBJ",
+                    "Gravity"
+                  ],
+                  "note": "Kill the first fish with bombs then dodge or kill the second one."
```
osse101, line 2256: Killing fish 2, to me, would be with a power bomb.  But being able to IBJ up and retry if the fish in a bad spot is fine here, unless you are really slow with your IBJ rhythm, and that shouldn't be the case in a GMode strat.

osse101 (dismissed):

PR #857, opened 2023-06-16
Add G-Mode Strats to Glass Tunnel, Crab Hole

kjbranch:


region/maridia/outer.json
```diff
@@ -120,6 +120,117 @@
...
+                  "name": "Carry G-Mode Through Tube (Top to Left)",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [4],
+                      "artificialMorph": false,
+                      "mode": "any"
+                    }},
+                    "f_MaridiaTubeBroken",
+                    {"or":[
```
osse101, line 161: These items should be in the bottom to left strat?

region/maridia/outer.json
```diff
@@ -120,6 +120,117 @@
...
+                    ]}
+                  ]
+                },
+                {
+                  "name": "Carry G-Mode Morph Through Tube (Bottom to Left)",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [2],
+                      "artificialMorph": false,
```
osse101, line 210: artificial morph needs to be true in this section

region/maridia/outer.json
```diff
@@ -120,6 +120,117 @@
...
+                  "name": "Carry G-Mode Morph Through Tube (Top to Left)",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [4],
+                      "artificialMorph": false,
+                      "mode": "any"
+                    }},
+                    "f_MaridiaTubeBroken",
+                    {"or":[
```
osse101, line 226: The requirements in `from[4]` and `from[2]` look swapped again.
Moving from 2 would need artificialmorph movement items too.

region/maridia/outer.json
```diff
@@ -179,6 +290,117 @@
...
+                  "name": "Carry G-Mode Through Tube (Top to Right)",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [4],
+                      "artificialMorph": false,
+                      "mode": "any"
+                    }},
+                    "f_MaridiaTubeBroken",
+                    {"or":[
```
osse101, line 331: These are requirements coming from bottom

region/maridia/outer.json
```diff
@@ -179,6 +290,117 @@
...
+                    "f_MaridiaTubeBroken"
+                  ]
+                },
+                {
+                  "name": "Carry G-Mode Morph Through Tube (Top to Right)",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [4],
+                      "artificialMorph": false,
```
osse101, line 392: same comments as above

region/maridia/outer.json
```diff
@@ -6225,6 +6503,89 @@
...
+                      "mode": "any"
+                    }}
+                  ]
+                },
+                {
+                  "name": "Carry G-Mode Through Tube (Bottom to Left)",
```
osse101, line 6523: Passing through the middle involves blind movement.  Its not that bad, the hard part would be jump morphing into the turnnel at the correct height.  And being aware that all crabs are active off camera.

region/maridia/outer.json
```diff
@@ -6225,6 +6503,89 @@
...
+                      ]}
+                    ]}
+                  ],
+                  "devNote": "There are many ways to get up with Gravity, but it can be done with no other items with a gravity jump."
+                }
+              ]
+            },
+            {
+              "leavesWithArtificialMorph": true,
+              "strats": [
```
osse101, line 6559: You can artificial morph into the room from below and IBJ up to either tube entrance?

region/maridia/outer.json
```diff
@@ -6307,6 +6668,89 @@
...
+                {
```
osse101, line 6725: Same comment about IBJing from below without morph

region/maridia/outer.json
```diff
@@ -6331,6 +6775,17 @@
                   "name": "Base",
                   "notable": false,
                   "requires": [ "Morph" ]
+                },
+                {
+                  "name": "G-Mode Morph",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1,4],
```
osse101, line 6784: You have to time a roll pretty well coming from 1 to not hit a crab.

region/maridia/outer.json
```diff
@@ -6524,6 +6971,20 @@
...
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [2,3],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "h_canArtificialMorphIBJ",
+                    "Gravity"
+                  ],
+                  "note": "Kill or avoid the crabs."
```
osse101, line 6987: not trivial

region/maridia/outer.json
```diff
@@ -6524,6 +6971,20 @@
...
+                {
+                  "name": "G-Mode Morph IBJ",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [2,3],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "h_canArtificialMorphIBJ",
```
osse101, line 6984: Should the non GMode strat require blind movement because Samus is covered for part of the IBJ?

region/maridia/outer.json
osse101: Doing the strat right away worked for avoiding crabs.  Fast crab is kinda hard to kill, I wanted two passes for that one.

region/maridia/outer.json
osse101: With GMode the camera is an issue for bringing morph into another room.  If you are ending GMode after reaching the top half of the room, you just need a way to know you are doine IBJing.

region/maridia/outer.json
kjbranch: Because of the way the logic works, this strat technically starts at the left door. So in order to have the gmode setup work, you also need to ensure you can get to the door that has the setup. So these are here to describe how to get to the top door (getting to the bottom and back is free)

region/maridia/outer.json
kjbranch: You can, but because of the limitations of how the logic allows multi room strats, these arent possible to represent. It requires starting at the node in question, and then going to the doors below, which requires morph, so no artificial strats can be added currently

region/maridia/outer.json
kjbranch: I'm on the fence about this one. It is blind, but not too challenging. The places with off screen movement currently are all harder than this, and those skipping it are all easier than this.

region/maridia/outer.json
kjbranch: I think the blind movement here is simple enough. I'll put a comment to help give a strategy for this.

region/maridia/outer.json
```diff
@@ -179,6 +301,128 @@
...
+                          "Bombs",
+                          {"ammo": { "type": "PowerBomb", "count": 1 }}
+                        ]}
+                      ]},
+                      {"and":[
+                        { "not": "f_MaridiaTubeBroken" },
+                        "canRiskPermanentLossOfAccess"
+                      ]}
+                    ]}
+                  ]
```
osse101, line 383:
```suggestion
                  ],
                  "devNote": [
                    "Bomb jumping across the broken tube uses a second bomb to climb up to the tube.",
                    "FIXME: Moving before the bomb explodes gives enough momentum to only need one power bomb."
                  ]
```

region/maridia/outer.json
```diff
@@ -120,6 +120,128 @@
...
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [4],
+                      "artificialMorph": true,
+                      "mode": "any"
+                    }},
+                    "f_MaridiaTubeBroken",
+                    {"or":[
+                      "Gravity",
+                      "HiJump"
```
osse101, line 239: "canSpringBallJumpMidAir" is in the other `Top to Left`.  It should be here too?
Same with the artificial morph `Top to Right`

region/maridia/outer.json
```diff
@@ -492,6 +774,28 @@
...
+                {
+                  "name": "G-Mode Morph Breaking the Maridia Tube Gravity Jump",
+                  "notable": true,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1,3],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    {"ammo": { "type": "PowerBomb", "count": 1 }},
```
osse101, line 787: This cost is part of the breaking glass event.  But then I'm not sure how that interacts with `not: f_MaridiaTubeBroken`.
I'd guess you could follow these gravity jump strats to reach top door repeatedly, and get back down with maridiaTubeClip, and never really break the tube to get the flag. Because that would add underwater movement requirements for future trips.
Seems like the `yield` needs to be attached to these strats.

region/maridia/outer.json
osse101: I would put blind movement on `no movement items` and probably springball only too.  It can be hard to know exactly where the morph tunnel is -- its always higher than I think.  Jumping to the open tube isn't an issue.  Crabs being active off camera should be a note.

region/maridia/outer.json
```diff
@@ -6225,6 +6525,89 @@
...
+                      "Morph",
+                      "SpringBall",
+                      {"and":[
+                        "Gravity",
+                        "h_canArtificialMorphIBJ"
+                      ]},
+                      {"and":[
+                        "Gravity",
+                        "h_canArtificialMorphBombHorizontally",
+                        {"or": [
```
osse101, line 6601: You can turn off gravity and roll to the middle to save a PB.

To PB jump the entire way with the momentum trick, you have to back up for more roll speed making it really hard.

region/maridia/outer.json
kjbranch: That would require morph, and this is a duplicate from above, except this one is leaving with artificial morph. If you have morph, I _think_ that you should be able to use non-artificial morph strat above and then use morph in the next room and it will work out the same. I'll check with Maddo tomorrow

region/maridia/outer.json
```diff
@@ -120,6 +120,128 @@
...
+                          "Bombs",
+                          {"ammo": { "type": "PowerBomb", "count": 1 }}
```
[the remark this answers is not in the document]
kjbranch: That's cool. It looked possible so I tried it for a bit and never got it.

region/maridia/outer.json
kjbranch: If I'm understanding this right, that does seem to be a problem ... It's the same as the non g-mode version of the same strat, which Map Rando doesn't yet use because of `canRiskPermanentLossOfAccess`. I'll talk to Maddo about this as well. I know that currently we can't include a `yields` on a strat, but that would be nice in several places and possibly required here.

region/maridia/outer.json
kjbranch: With the G-Mode starting position from the left door, it looks like you wont have enough momentum to do the same. You could back up, but the crabs will be headed for Samus (unless immobile is being used). I think I'll keep it only on the `R->L` for now. We should probably add a way to know if a strat is coming from immobile or not.

region/maridia/outer.json
kjbranch: Yeah, this is how he says it works.

region/maridia/outer.json
kjbranch: Apparently there are problems with having a `yields` on a link. This new commit should fix these problems.

PR #859, opened 2023-06-17
G-Mode Main Street/Everest Morph Tunnel

kjbranch:
In order to put in `leaveWithGMode`, you need to include strats to go to the door node that sets up the gmode and return. This ended up being a pain for Everest.

### DIFF
--- a/region/maridia/outer.json
+++ b/region/maridia/outer.json
@@ -1323,7 +1323,7 @@
                   ]
                 },
                 {
-                  "name": "G-Mode Morph Get Item and Return",
+                  "name": "Direct G-Mode Morph Get Item and Return",
                   "notable": false,
                   "requires": [
                     {"comeInWithGMode": {
@@ -1359,7 +1359,7 @@
                   "requires": [
                     {"comeInWithGMode": {
                       "fromNodes": [5],
-                      "mode": "indirect",
+                      "mode": "any",
                       "artificialMorph": true
                     }},
                     {"or":[
@@ -1376,7 +1376,7 @@
                   "note": "The crab will not come through the whole tunnel, so wait for it to pass before going through the vertical portion."
                 },
                 {
-                  "name": "G-Mode Morph",
+                  "name": "Direct G-Mode Morph",
                   "notable": false,
                   "requires": [
                     {"comeInWithGMode": {
@@ -1431,7 +1431,19 @@
                 {
                   "name": "G-Mode Morph Return",
                   "notable": false,
-                  "requires": [ {"obstaclesCleared": ["B"]} ]
+                  "requires": [
+                    {"obstaclesCleared": ["B"]},
+                    {"or":[
+                      "SpringBall",
+                      {"and":[
+                        "Gravity",
+                        {"or": [
+                          "Bombs",
+                          {"ammo": { "type": "PowerBomb", "count": 1 }}
+                        ]}
+                      ]}
+                    ]}
+                  ]
                 }
               ]
             }
@@ -3636,6 +3648,13 @@
                     {"or": [
                       "h_canArtificialMorphIBJ",
                       {"and": [
+                        "SpringBall",
+                        {"or": [
+                          "canGravityJump",
+                          "HiJump"
+                        ]},
+                      ]},
+                      {"and": [
                         "h_canArtificialMorphSpringBallBombJump",
                         {"or": [
                           "Bombs",
@@ -4186,6 +4205,13 @@
                     }},
                     "Gravity",
                     {"or": [
+                      {"and": [
+                        "SpringBall",
+                        {"or": [
+                          "canGravityJump",
+                          "HiJump"
+                        ]},
+                      ]},
                       "h_canArtificialMorphIBJ",
                       "h_canArtificialMorphSpringBallBombJump"
                     ]}
--- a/region/maridia/outer.json
+++ b/region/maridia/outer.json
@@ -1359,7 +1359,29 @@
                   "requires": [
                     {"comeInWithGMode": {
                       "fromNodes": [5],
-                      "mode": "any",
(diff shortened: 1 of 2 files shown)


region/maridia/outer.json
```diff
@@ -845,6 +845,25 @@
...
+                {
+                  "name": "Get Hit By Sciser",
+                  "notable": false,
+                  "requires": [
+                    "h_EverestMorphTunnelExpanded",
+                    "Morph",
+                    {"ammo": { "type": "Super", "count": 1 }}
+                  ],
+                  "note": [
+                    "Get to the item while avoiding the crab, note that it doesn't go all the way through the tunnel.",
```
osse101, line 860: I don't think you can avoid a crab hit without some items but I'm not testing for it.  
You can pause abuse through but then it changes your reserve tank numbers and thats annoying.
Its mostly a problem because crabs are 1 hit kill when suitless.

osse101 (commented):
I can't tell if it is added under another PR, but there should be strats for indirect artificialmorph room entry through these tunnels.  And entering the everest side in artificialmorph from any door.

region/maridia/outer.json
```diff
@@ -3008,7 +3091,191 @@
...
+                {
+                  "name": "Carry G-Mode Morph Through Tunnel From Top Left",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "artificialMorph": false,
+                      "mode": "any"
+                    }},
+                    "h_EverestMorphTunnelExpanded",
```
osse101, line 3129: This doesn't seem required.  You would want the space when entering but its not needed leaving, and already in GMode.  (Leaving and not it GMode you would need the space to stand up)

osse101: There are strats for reaching 6 in artificial morph, but they could also be used to leave in GMode, no?

osse101 (commented):
My problem with `h_EverestMorphTunnelExpanded` is that there are other ways to make the connection not stick you in a wall.  But I do think every randomizer needs a way to handle these transitions.
Open randomizer made the entire tunnel walkable iirc.  Maybe you could be forced into Morph, or placed in a spot you can manually morph.
It looks out of place.  Maybe there should be a section for optional edit type things.

region/maridia/outer.json
kjbranch: Yeah, I tried it with nothing. it's not too bad

region/maridia/outer.json
kjbranch: Correct. The `leaveWithGMode` section, from lines 3116-3277 are for strats getting direct G-Mode in Everest and carrying it through the tunnel. These don't actually need `h_EverestMorphTunnelExpanded`

The other strats in the links section will(?) require it, because it technically has you exit G-Mode before touching the room transition tiles (although, I don't think that actually matters, so maybe it could be removed there too?)

blkerby: If I understand, you're proposing having a separate notation like "e_EverestMorphTunnelExpanded" ("e" for edit) which behaves equivalently to the helper, just written differently to make it clear that this is only a randomizer-level toggle and not something with arbitrary requirements (such as tech, items. etc.). I think that would be fine but don't see too much difference from just using a helper, which has the advantage that it doesn't require adding complexity to the schema. Maybe for now we could use a helper, to avoid blocking adding these strats, and then potentially migrating them to a different format could be a follow-up discussion? 

I think we can consider this helper as representing the edits to these rooms specifically the way that Map Rando did it. This doesn't prevent us from potentially supporting alternative types of edits to these tunnels in the future; supporting another one would just require adding another helper for it.

I do like the idea of having randomizer-specific tweaks being something handled within the repo if possible. Requiring each randomizer to either fork the data or apply heavy patches to it would be harder to maintain. For example, if we renamed a tech or restructured nodes in a room, it would be a lot harder if you have to track down multiple repos affected and change them all.

kjbranch: I agree that it would be nice to move them to their own file, so it is easier for people adopting the logic to easily see what they should/shouldn't edit. Currently it's just the first ~5 helpers. Similarly, we could pull out a separate file for leniency factors. I can make an issue for this topic.

osse101: I think I do only want it organized differently so its ok to stay as a helper for now.  The camera fixes feel like they make a better default state to build from even though they are changes.  We do want to make it more clear what kinds of things can be readily changed.

region/maridia/outer.json
```diff
@@ -3306,6 +3620,34 @@
...
+                    {"comeInWithGMode": {
+                      "fromNodes": [2],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "Gravity",
+                    {"or": [
+                      "h_canArtificialMorphIBJ",
+                      {"and": [
+                        "h_canArtificialMorphSpringBallBombJump",
```
osse101, line 3639: Springball+hijump
springball+gravity jump
bomb+gravity jump

region/maridia/outer.json
```diff
@@ -3829,6 +4172,28 @@
...
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [4],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "Gravity",
+                    {"or": [
+                      "h_canArtificialMorphIBJ",
```
osse101, line 4189: same comment as 2->6

osse101 (commented):
More movement options while balled 🤽

region/maridia/outer.json
kjbranch: bomb + gravity jump is really interesting, but I'm going to leave it off, since there's IBJ

PR #860, opened 2023-06-17
Most Outer Maridia G-Mode Setups

kjbranch:
This is based off of #853, so it should be merged after.


osse101 (commented):
No issues with the PR except this first main street gModeImmobile isn't comparing against the correct file version.  You wanted to keep it at ["never"] right?

PR #861, opened 2023-06-17
Add Missing G-Mode Strat in Tourian Escape 3

kjbranch:


region/tourian/main.json
```diff
@@ -4227,6 +4227,23 @@
                   ],
                   "note": "IBJ or spring ball bomb jump to avoid the wall jump. Using a total of 3 PBs will allow Samus to roll through the bottom 4 pirates."
                 },
+                {
+                  "name": "G-Mode Morph Pirate Tank",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "any",
```
osse101, line 4236: This would require immobile for health which would make this indirect.  Maybe the `enemyDamage` would ensure that is how it plays out?
And a CF would try to kill the 4 pirates with pbs first.  Unless you need the health for future rooms 🤷‍♂️

kjbranch: Yes, I think the logic should force it to be indirect. CF would force Samus to stand up, so that wouldn't work.

PR #862, opened 2023-06-18
More Lenient Kraid Super Kill

kjbranch:


blkerby:
One thing to keep in mind is that Map Rando applies the resource multiplier on top of these Super requirements. The result is that, with this PR, being required to farm one Super during the fight (without canBePatient) will only be in logic on Very Hard and above, and farming two or more Supers will always require canBePatient. I think that works out nicely.

PR #865, opened 2023-06-19
Add canWallJumpIntoMorph to fix East Sand Hole and use throughout

blkerby:
Doing wall jumps here (with only Gravity and Morph) seems like it should require canPlayInSand, so this moves it out of Base into the "Gravity on Sand" strat.

See recent discussion: https://discord.com/channels/1053421401354285129/1053436236628496454/1120345710278234193


kjbranch:
Play in sand is used for needing to get a clean higher jump off of sand. A mediocre jump works in this situation. The tricky part feels like it is the midair morph before bonking on the ceiling.

blkerby:
Yeah, going up left side needs a clean jump off of the sand, while going up the right side needs the walljump into morph. Both options feel like they should be in Hard rather than Medium. Maybe we need a new Hard tech canWalljumpIntoMorph? In that case we could probably remove Alcatraz Escape as a notable strat (having it use canConsecutiveWalljump and canWalljumpIntoMorph). We would make canSpringwall and can3HighMidAirMorph be extensions of canWalljumpIntoMorph

osse101:
Well, I like the idea of adding canWalljumpIntoMorph, and canNonTrivialMidAirMorph.  
This jump to the power bombs fits under what the current techs expect imo.  With a second semi precise walljump on the right wall you have time for the morph.  Or walljump morphing from the left wall for the instant morph.  Either way, it expects a little bit better than `canWalljump` because you have to move into another action right away.

kjbranch:
I decided on the capital `J` in `canWallJumpIntoMorph` because I think eventually it should be used like that in all wall jump tech. It currently only is in `canWallJumpBombBoost`.

tech.json
```diff
@@ -156,14 +163,39 @@
...
+          "requires": [ "Morph" ],
           "note": [
             "The ability to enter Morph Ball while in the air.",
             "There can be an expectation to Morph quickly as well, usually following a jump from the ground.",
             "Out of water physics this means jumping and morphing in a 4 tile high space.",
             "In water physics this means jumping and morphing in a 3 tile high space (Samus takes up 3 tiles when standing)."
           ],
           "extensionTechs": [
+            {
+              "name": "canWallJumpIntoMorph",
```
osse101, line 175: I was thinking `canWallJumpInstantMorph`
The name and description still sound like this is nothing special.

Pressing down right away is pretty important.  Turning around would break the instant morph effect and you'd need two down presses again.

region/maridia/inner-pink.json
```diff
@@ -6470,6 +6470,7 @@
                     "Morph",
                     "canPlayInSand",
                     {"or": [
+                      "canWalljump",
```
osse101, line 6473:
```suggestion
                      "canWalljump",
                      "canGravityJump",
```

region/norfair/west.json
```diff
@@ -2092,8 +2093,7 @@
                   "notable": false,
                   "requires": [
                     "h_canNavigateHeatRooms",
-                    "Morph",
-                    "canWalljump",
+                    "canWallJumpIntoMorph",
```
osse101, line 2096: This really isn't that precise of a morph

region/brinstar/blue.json
```diff
@@ -1219,10 +1219,40 @@
...
+                          "ScrewAttack"
+                        ]}
+                      ]
+                    }
+                  ],
+                  "note": "Base strat when entering the room from below.",
+                  "devNote": "canWallJumpIntoMorph is not needed here, as the precision is very low."
+                },
+                {
+                  "name": "Power Bomb, Wall Jump, and Bombs",
```
osse101, line 1234: Walljump isn't needed here either.  The key feature is placing a bomb right against the bomb block.

region/brinstar/red.json
```diff
@@ -539,7 +539,7 @@
                   "name": "Red Tower Diagonal Bomb Jump",
                   "notable": false,
                   "requires": [
-                    "canWalljump",
+                    "canWallJumpIntoMorph",
```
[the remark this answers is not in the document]
kjbranch: Oh. The positioning of node 7 is a little odd. I wouldve thought that wall jump would be in 6->7

region/crateria/central.json
```diff
@@ -2103,7 +2103,8 @@
                     "canPreciseWalljump",
                     {"or": [
                       "HiJump",
-                      "canConsecutiveWalljump"
+                      "canConsecutiveWalljump",
```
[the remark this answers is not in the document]
kjbranch: I think I put that in the wrong place. the `canWallJumpInstantMorph` should be required regardless of `HiJump`

PR #870, opened 2023-06-22
Green Maridia G-Mode Strats

kjbranch:


helpers.json
```diff
@@ -61,6 +61,10 @@
       "name": "h_canActivateAcidChozo",
       "requires": ["SpaceJump"]
     },
+    {
+      "name": "h_ShaktoolVanillaFlag",
+      "requires": []
```
osse101, line 66: Add a note saying this is the shaktool room unmodified behaviour when entering from the right.

region/maridia/inner-green.json
```diff
@@ -1392,6 +1484,29 @@
...
+                      "Gravity",
+                      "HiJump",
+                      "canSpringBallJumpMidAir",
+                      "canTrickyJump"
+                    ]}
+                  ],
+                  "note": [
+                    "The sand does not impede Samus in G-Mode.",
+                    "It is possible to jump to the tall pillar with nothing. It requires a very precise jump as late as possible.",
+                    "Alternatively, starting on the last pixel facing left, turning around and very rapidly jumping may be easier."
```
osse101, line 1508: I don't know what this means

region/maridia/inner-green.json
```diff
@@ -2903,10 +3091,47 @@
...
+                        "Gravity",
+                        "h_canArtificialMorphIBJ"
+                      ]},
+                      "SpringBall"
+                    ]},
+                    {"ammo": { "type": "PowerBomb", "count": 1 }}
+                  ],
+                  "note": [
+                    "The Snails will dig through the sand without showing any visual difference. Each block they pass through gets closer to overloading PLMs.",
+                    "In order to PB the left wall before PLMs are overloaded, it is necessary to kill one of the Snails with a bomb before it digs any sand.",
```
osse101, line 3129: Can the snails travel in the same path or will that count as both eating sand blocks?  They'll be in different positions if you use gimmobile. 

With springball you can stunlock one then move it off camera.

osse101 (commented):
Evirs are a menace but not as bad as when you also have to deal with sand.  You may need to know approximately where they spawn to know how to move through the room but that's not well defined.

There were a couple places that you can fall through the sand and not be able to jump back out.  Just west hall really, maybe that needs some `or: [canCarefulJump, h_canCrouchJumpDownGrab, ...]`

kjbranch:
In that conflict resolution, I added a devnote to the shaktool camera fix helper. It might be a bit hidden.

region/maridia/inner-green.json
```diff
@@ -1505,7 +1519,7 @@
                   "note": [
                     "The sand does not impede Samus in G-Mode.",
                     "It is possible to jump to the tall pillar with nothing. It requires a very precise jump as late as possible.",
-                    "Alternatively, starting on the last pixel facing left, turning around and very rapidly jumping may be easier."
+                    "Alternatively, facing left while on the right-most pixel, then turning around and very rapidly jumping may be easier."
```
osse101, line 1635: This is subpixel dependant.  But you can adjust to a good value, and the tells are there if you have sharp eyes.
Feels like a 3 frame jump window to me.  Could be 2.
The problem is "very rapidly jumping"  sounded like like mashing jump helped somehow.

I learned running before jumping doesn't matter at all so you can stand one pixel from the right edge and do a forward spinjump.  But that also only works when jumping from the high subpixels of the farthest pixel. One frame jump.

I like the forward jump more.

kjbranch: I think this is what you were saying?

osse101: Right.  There is one subpixel precise spot that works and you can approach it facing forwards or backwards.

I think there is a `from[3]` strat that would use the same note change.

kjbranch: Good catch. Thanks!

PR #876, opened 2023-06-25
Combine and Fix Mickey Mouse Spring Ball/Wall

kjbranch:


region/lowernorfair/east.json
```diff
@@ -6086,13 +6086,22 @@
                   "note": "Requires one jump off a crumble block"
                 },
                 {
-                  "name": "MickeyMouse Crumbleless MidAir Spring Ball",
+                  "name": "Mickey Mouse Spring Ball Escape",
```
osse101, line 6089: This name is too similar to the crumblejump + springball escape.  HiJump+Springball is similar too.

osse101 (commented):
I think you can remove this as notable since it has `canTrickyJump`.   And with springball on, there's no risk of falling so failure just means heat damage.

It's a little questionable if lateralmorph branch is trickyjump since you have the option of timing the pause or timing the springball jump.  Pressing all the buttons at once is tricky, so I can give it that.

Is springwall branch supposed to have trickyJump?  Its pretty similar to most springwalls.  precisewalljump makes sense because you want to walljump pretty low.

I found another way that I like - `canSpringBallJumpMidAir` + `can3HighMidAirMorph`.  A lot like the springwall, but just a jump morph.

blkerby:
I tried all three methods for just a few minutes and they all seemed pretty difficult. Much harder than a typical springwall. I couldn't get the can3HighMidAirMorph but it could make sense to throw it in as a option if you like it most. Lateral mid-air morph felt easiest to me but also hard enough to need tricky jump.

osse101:
I can see the input timings on all three being awkward.  

Thinking about which strat I would use here; It'd either be the springwall or straight jump.  In the past I went for the crumble jump, but I like how these strats are safer.

kjbranch:
I'm not sure if this many strats should be combined like this, but I think I prefer it, as the description can mention all of the options. 

I hadn't really processed that you were working on this room. Feel free to close this PR if it's going to interfere with what you're doing - this room is a beast.

PR #879, opened 2023-06-27
Add and update subregion maps

kjbranch:


osse101:
Very nice maps.
I was thinking that the colored arrow would indicate which region it connected to, by color.  But with all of the other colors being used here I don't think that would look great.

PR #905, opened 2023-07-09
Helper Structure

kjbranch:
Mostly helpers have only been moved into categories. Only Leniency helpers have been created and those that use them have been edited. 

Eventually more Leniency helpers would be added, so that all leniency resources are centrally located

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -60,6 +60,59 @@
       ]
     },
     {
+      "name": "Leniency",
+      "description": "Helpers with built in leniency where extra Energy or Ammo are used in a failure.",
+      "helpers": [
+        {
+          "name": "h_BlueGateGlitchLeniency",
+          "requires": [
+            {"or": [
+              {"ammo": { "type": "Missile", "count": 1 }},
+              {"ammo": { "type": "Super", "count": 1 }}
+            ]}
+          ],
+          "devNote": "Ammo for a second shot for leniency."
+        },
+        {
+          "name": "h_GreenGateGlitchLeniency",
+          "requires": [
+            {"ammo": { "type": "Super", "count": 1 }}
+          ],
+          "devNote": "Ammo for a second shot for leniency."
+        },
+        {
+          "name": "h_HeatedBlueGateGlitchLeniency",
+          "requires": [ 
+            "h_GreenGateGlitchLeniency",
+            {"heatFrames": 60}
+          ],
+          "devNote": "One ammo and 60 heatFrames for a second shot for leniency."
+        },
+        {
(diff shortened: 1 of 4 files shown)


helpers.json
```diff
@@ -1,502 +1,574 @@
...
+            {"ammo": { "type": "Super", "count": 1 }},
+            "h_GreenGateGlitchLeniency"
+          ]
+        },
+        {
+          "name": "h_canHeatedBlueGateGlitch",
+          "requires": [ 
+            "h_canNavigateHeatRooms",
+            {"heatFrames": 60},
+            "canGateGlitch",
```
osse101, line 229: Either h_canBlueGateGlitch
or h_BlueGateGlitchLeniency

helpers.json
```diff
@@ -1,502 +1,574 @@
...
+              "h_canTwoTileJumpMorph",
+              "can3HighMidAirMorph"
+            ]}
+          ]
+        },
+        {
+          "name": "h_canOpenZebetites",
+          "requires": [
+            {"or": [
+              {"ammo": { "type": "Missile", "count": 12 }},
```
osse101, line 283: This is 11 plus 1 leniency.  
I think I'd like to add a `or:[canDodgeWhileShooting, more_leniency]` too
Not that it really matters.

osse101 (dismissed):
Looks all in order, and much more organized now.

I don't know that leniency helpers can cover all leniency cases but I can try them out.  They work well when used inside of these other helpers.

helpers.json
kjbranch: I think this should be correct as is. It has its own leniency a few lines down. If I call `h_canBlueGateGlitch`, it will use the `h_BlueGateGlitchLeniency` as well - giving 3x ammo and 2x heat frames

helpers.json
kjbranch: I partially fixed it. I'm not sure how reliably you can open them with 11 missiles, if so, I could add an extra leniency helper that only is used when using only missles.

helpers.json
osse101: I must have been thinking that `h_heated...` was only giving heat frames

helpers.json
osse101: Looks like it heals every 2 seconds, with the first tick being instant.  Two rinka/turret hits would be enough delay, or getting hit while jumping a gap.  To bring the ammo costs to what is used here.  That makes it a tighter window when using more missiles.  Also when using more missiles it is more likely to waste one in overkill.  
Should be alright as it is.

PR #906, opened 2023-07-09
Wall jump to reset fall speed upon entry of Fish Tank

kjbranch:
https://discord.com/channels/1053421401354285129/1053436236628496454/1100564018080514060

### DIFF
--- a/region/maridia/outer.json
+++ b/region/maridia/outer.json
@@ -3876,10 +3876,8 @@
                     "canSuitlessMaridia",
                     "canPreciseWalljump",
                     "canPrepareForNextRoom",
-                    {"or":[
-                      "canMidairWiggle",
-                      "canCarefulJump"
-                    ]},
+                    "canMidairWiggle",
+                    "canCarefulJump",
                     {"resetRoom": {
                       "nodes": [ 4 ],
                       "nodesToAvoid": [ 1 ]
--- a/region/maridia/outer.json
+++ b/region/maridia/outer.json
@@ -3870,6 +3870,21 @@
                   "devNote": "Strat starts before entering this transition. An enemy can be used for this strat instead of morph."
                 },
                 {
+                  "name": "Fish Tank Wall Jump Entry",
+                  "notable": true,
+                  "requires": [
+                    "canSuitlessMaridia",
+                    "canPreciseWalljump",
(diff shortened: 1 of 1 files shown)


osse101:
From how I remember trying this, I'd call it `canTrickyJump` or `canMidAirWiggle`.

I call the shot to break spin and instantly turnaround a wiggle.

osse101:
I think you are able to reach the right wall on all vertical doors.
Water->water with a careful jump should be ok for lining up the walljump on the right wall even with the awkwardness.
Air->water should have carefuljump too because you move kinda fast.
Because of the hitbox extend touching the transition, its not really trickyJump.
water->water should have wiggle.

region/maridia/outer.json
```diff
@@ -3869,6 +3869,21 @@
                   "note": "Morph and unmorph then drift to the ledge.",
                   "devNote": "Strat starts before entering this transition. An enemy can be used for this strat instead of morph."
                 },
+                {
+                  "name": "Fish Tank Wall Jump Entry",
+                  "notable": true,
+                  "requires": [
+                    "canSuitlessMaridia",
+                    "canPreciseWalljump",
+                    "canPrepareForNextRoom",
```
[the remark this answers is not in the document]
osse101: I'm going with `canCarefulJump` now.

And I don't think you can physics check with `adjacentRunway` on a vertical door since it won't have a defined runway.
So give them `canMidairWiggle` with a fixme for later.

The physics check would be a `adjacentJumpway` thing, but those don't have that yet either.  So it'd mean adding `doorFrameAbove` and `physics` as fields and values to jumpways.

region/maridia/outer.json
```diff
@@ -3876,12 +3876,19 @@
                     "canSuitlessMaridia",
                     "canPreciseWalljump",
                     "canPrepareForNextRoom",
+                    {"or":[
```
osse101, line 3879: Sorry, I meant I now think both `canCarefulJump` and `canMidairWiggle` should be required.

kjbranch: You were clear, that was my bad

kjbranch: Whoops sorry, I guess I'm a little out of it today. Hopefully this has all the necessary information now.

PR #912, opened 2023-07-10
Edit Mickey Mouse Runways

kjbranch:


osse101:
I think the desgeega hit the floor in my save state, and didn't move forward.  But now it looks like his expected behaviour is to jump forward.

region/lowernorfair/east.json
```diff
@@ -6204,9 +6222,13 @@
...
                       ]}
                      ]},
-                    {"heatFrames": 45}
+                    {"heatFrames": 100}
                   ],
-                  "note": "The Multiviola can be killed by aiming diagonally down when entering the room and shooting a strong enough beam shot (not ice)."
+                  "note": "The Multiviola can be killed by aiming diagonally down when entering the room and shooting a strong enough beam shot (not ice).",
+                  "devNote": [
+                    "Taking a hit by the Multiviola isn't an option, as it will stop Samus' run and then the Dessgeega will be in the way.",
+                    "FIXME This strat could be used while entering - heatFrames could be lowered to 45 and runway extended to 13 tiles. This is not currently possible to model."
```
osse101, line 6230: The desgeega should also jump forward making the runway a size 7 so I don't think this fixme is needed.

osse101 (dismissed):

region/lowernorfair/east.json
kjbranch: you can have 9-10 tiles before turning blue and be safe, but that's not a big enough difference to keep the fixme. Currently all heated runways are too lenient, as they require round trip heat frames. Oh well for now.

PR #914, opened 2023-07-11
Increase Croc ammo for low difficulties

kjbranch:


osse101 (approved):
With resource scaling, these numbers look pretty good across difficulties.

I think the 5 missile fight could be added as a notable strat since there probably won't be a `canDelayDropPickup` tech.

kjbranch:
Map Rando doesn't multiply `ammo capacity`, which is why this was such a problem. This will fix the problem significantly, but if it's still a problem, we could add the multiplier there to the capacity as well, but I think that would maybe be too far ...

I added a notable 5 Missile Croc kill:
Because it requires patience, it can only be notable insane ... We could lower it with the new lower-level patience tech, once that is made, but I don't anticipate this being a loved strat. xD

osse101:
I don't think I've done a 5 missile Croc before, but it wasn't much trouble now.  You just need a plan for fitting in missiles during the farm phase but before collecting drops so you effectively have 7 capacity.  Took about 4 minutes.

I'd rather see `Hard` croc at 15 missiles and that may still be low, but acceptable.

PR #915, opened 2023-07-11
Update Sky Missiles

kjbranch:
This should prevent tedious Sky Missiles from low difficulty settings and wall jump-less challenges

### DIFF
--- a/region/crateria/east.json
+++ b/region/crateria/east.json
@@ -1789,15 +1789,15 @@
                     "canUseEnemies",
                     {"or": [
                       "canCarefulJump",
-                      "canUseFrozenEnemies"
-                    ]},
-                    {"or": [
-                      "canWalljump",
-                      "canSpringBallJumpMidAir",
-                      "h_canCrouchJumpDownGrab"
+                      "canUseFrozenEnemies",
+                      "canPreciseWalljump"
                     ]}
                   ],
-                  "note": "Freezing the Trippers can make this much easier."
+                  "note": [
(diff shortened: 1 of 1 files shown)


osse101 (commented):
I think the strats are 
HiJump to jump from 1 ->2 -> item. Quick.

And the rest lead 2 to right, ride left and jump to item:
canCarefulJump
canUseFrozenEnemies

Jumping 1 to 2 without hijump would be carefuljump at least.
IBJ would want a better way to reveal the item while IBJing but it doesn't seem that important in this case.

region/crateria/east.json
```diff
@@ -1772,8 +1772,51 @@
...
+                {
+                  "name": "Climb the Trippers Bootless",
+                  "notable": false,
+                  "requires": [
+                    "canUseEnemies",
+                    {"or": [
+                      "canCarefulJump",
+                      "canUseFrozenEnemies"
+                    ]},
+                    {"or": [
```
osse101, line 1794: I think this is to jump into the item.  But you don't need any extra height gain to reach it?

kjbranch: These are ways to climb the platforms. I would say jumping from the middle platform to the item would probably be a `tricky jump`, because the way you will likely encounter it has the top tripper in the way and requires a somewhat precise shot and then somewhat precise jump, where a failure will surely waste a lot of time.

kjbranch: The tricky jump strat could be added, but I think the careful + cjdg is more intuitive and easier.

Issue #817, opened 2023-05-28
Rework ceiling clip tech


blkerby:
We need some way to take into account that (at least in most situations) HiJump helps with ice clips as much as XRay+Morph. We could put the HiJump logic into the (already complicated) requirements for the tech canNonTrivialIceClip, or we could pull it out of the tech and put it into individual strats. My initial thinking is the second option seems better: though more verbose, it's more flexible, and being more explicit can make it easier for someone to understand when reading the logic.

Also, maybe it's not necessary to have separate ice/snail clip tech: e.g. instead of canNonTrivialIceClip and canPixelPerfectIceClip we could have canNonTrivialCeilingClip and canPixelPerfectCeilingClip, and we can get rid of canSnailClip? There's also canMochtroidIceClip, which we could also get rid of (becoming covered by canCeilingClip), ending up with a hierarchy like this:

  canCeilingClip < canNonTrivialCeilingClip < canPixelPerfectCeilingClip

Then we could also get rid of all the enemy-specific ice clipping tech ("canBeetomIceClip", etc.); particularly tricky cases could be reflected as notable strats.

osse101:
snailClip is pretty much exactly the same as wallcrawlerIceClip.  One difference is that people are used to having a visual cue for the snail clip placement and haven't had a reason to learn cues on wallcrawlerClips.  Another difference is you can buffer XRay to stop a snail to better position it, but using XRay will not allow you to use ice right after.

MochtroidClip would fall under notable ceiling clip in that case.  The inputs are only trivial after you know what they are.  You probably have to be a little fast with button presses too.

PixelPerfectIceClip is essentially an extension of NonTrivialIceClip and for those I've been putting the helpful items into the strats themselves.  But you could do it with a helper.  HiJump vs Morph is a little different in that morph lets you position on top of the enemy easier.  Maybe that is a bigger problem for certain clips.  They help a similar amount, but I'd rather have morph+xray.

Enemy specific clip techs aren't that helpful.  I think it was written that way for enemizer.

osse101:
I am considering renaming `canSnailClimb` to `canManipulateSnails` then replacing `canSnailClip` with `canManipulateSnails` + `canNonTrivialCeilingClip`.

osse101:
I was aware no HJB crouch jump had a two pixel range, and that HJB helped some, but the rest is new to me.

[...]

PR #918, opened 2023-07-12
Ceiling and Ice Clip Rework

kjbranch:
This addresses #817 

What are your thoughts on this? 

There are currently only two different ceiling clip tech, one for low level strats (`canMochtroidIceClip`) and then it can be paired with other tech to increase its difficulty (`canPartialFloorClip`). Then actual tricky strats are all notable, so they could be increased in difficulty. Then the `canPreciseCeilingClip` is for the very precise stuff, which would be able to be used on all of the lower difficulty notables without Morph+XRayTurnAround. 

Alternatively, we could make a mid level clip for standard wall crawlers and snails, and make those no longer notable.

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -787,8 +787,15 @@
                 },
                 {
                   "name": "Green Brinstar Main Shaft Ice Clip",
-                  "notable": false,
-                  "requires": ["canWallCrawlerIceClip"]
+                  "notable": true,
+                  "requires": [
+                    "h_canNonTrivialCeilingClip",
+                    "canTrickyUseFrozenEnemies"
+                  ],
+                  "note": [
+                    "Freeze the wall crawler at a precise location in order to jump through the Power Bomb Blocks.",
+                    "The pixel window is larger and higher with Morph and an X-Ray Stand Up."
+                  ]
                 }
               ]
             }
@@ -2569,9 +2576,10 @@
               "strats": [
                 {
                   "name": "Etecoon E-Tank Beetom Clip",
-                  "notable": false,
+                  "notable": true,
                   "requires": [
-                    "canBeetomIceClip",
+                    "h_canNonTrivialCeilingClip",
+                    "canTrickyUseFrozenEnemies",
                     "Morph",
                     {"enemyDamage": {
                       "enemy": "Beetom",
@@ -2579,7 +2587,11 @@
                       "hits": 2
                     }}
                   ],
-                  "note": "Two Beetom hits are assumed, but with failed attempts it could take more, possibly requiring leaving the Beetom and going back to the right to farm.",
+                  "note": [
+                    "Jump and freeze the Beetom at a precise location in order to jump through the crumble blocks.",
+                    "The pixel window is larger and higher with Morph and an X-Ray Stand Up.",
+                    "Two Beetom hits are assumed, but with failed attempts it could take more, possibly requiring leaving the Beetom and going back to the right to farm."
+                  ],
                   "devNote": "Morph is required to lure a Beetom. Morphless tunnel crawl should not be required."
                 },
                 {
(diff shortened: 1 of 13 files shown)


kjbranch:
I added `Morph` to `Mickey Mouse Viola Clip (XRay Standup Bomb Block Clip)`. Is that correct?

tech.json
```diff
@@ -1313,8 +1244,21 @@
...
+          ],
+          "devNote": "This often uses enemies, so is required for enemy clip enemies that are extensions of canUseEnemies.",
+          "extensionTechs": [
+            {
+              "name": "canPreciseCeilingClip",
+              "requires": [
+                "canCeilingClip",
+                "canUseEnemies"
+              ],
+              "note": "Setting up a an enemy positioning to perform a very precise ceiling clip. The enemy positioning will require a 2-3 pixel range, so the clip can work without the leniency allowed by an X-Ray standup."
```
osse101, line 1259:
```suggestion
              "note": "Setting up an enemy positioning to perform a very precise ceiling clip. The enemy positioning will require a 2 pixel range, so the clip can work without the leniency allowed by an X-Ray standup."
```

osse101: Hijump isn't being modelled for the 3rd pixel

osse101 (commented):
This way of organizing it looks really good.

I really like that snail clip can be contained in one notable.

Wall crawler ice clips shouldn't be notable.  But I don't think it is obvious you can ice clip at those locations yet.

Mochtroid ice clip does take specific knowledge in how to freeze and climb the mochtroid.  Its mainly the botwoon hall room where you mochtroid clip so it could be a notable there.  

Pants ice clip with gravity is probably ok as not notable.  

I would if `canCeilingClip` had a more impactful word, but it is correctly named.  Making the gif an ice clip should adequately explain what it means.  Same with `canPreciseCeilingClip`

osse101:
> There are currently only two different ceiling clip tech, one for low level strats (`canMochtroidIceClip`) and then it can be paired

🤔

tech.json
kjbranch: I got 2-3 pixels without HJB and 3 pixels with

kjbranch:
Lots of changes, but mostly I addressed your feedback and added another mid tier of ceiling clips with xray+morph

region/brinstar/green.json
```diff
@@ -2569,17 +2576,22 @@
...
                     {"enemyDamage": {
                       "enemy": "Beetom",
                       "type": "contact",
                       "hits": 2
                     }}
                   ],
-                  "note": "Two Beetom hits are assumed, but with failed attempts it could take more, possibly requiring leaving the Beetom and going back to the right to farm.",
+                  "note": [
+                    "Jump and freeze the Beetom at a precise location in order to jump through the crumble blocks.",
+                    "The pixel window is larger and higher with Morph and an X-Ray Stand Up.",
```
osse101, line 2592: I just noticed that you need another way to gain height to get on top of the beetom once frozen.  With the precise clip, you can rub the frozen enemy to gain height, and conveniently put you in position to clip through the crumbles.  For Morph+Xray you could use a second beetom to stand on but that would also come with more beetom hits, and maybe you farm that health back anyway.

My old setup walljumped up, but this walljumpless setup is working better 😺

osse101 (approved):

region/brinstar/green.json
```diff
@@ -2585,7 +2585,20 @@
...
                       "hits": 2
-                    }}
+                    }},
+                    {"or": [
+                      "canPreciseCeilingClip",
+                      "canWalljump",
+                      "HiJump",
+                      "SpaceJump",
+                      "h_canJumpIntoIBJ",
+                      "canSpringBallJumpMidAir",
```
osse101, line 2595: IBJ doesn't work; its not fast enough and has a big chance of killing the beetom.
Springball jump would be tedious.  I think I'd rather leave that out and take the two extra hits.

PR #919, opened 2023-07-13
Yellow Maridia G-Mode Setup and Immobile

kjbranch:
This is all of Yellow Maridia, except quicksand rooms.

### DIFF
--- a/region/maridia/inner-yellow.json
+++ b/region/maridia/inner-yellow.json
@@ -1990,7 +1990,21 @@
                 {
                   "name": "Get Hit By Zoa",
                   "notable": false,
-                  "requires": []
+                  "requires": [
+                    {"or":[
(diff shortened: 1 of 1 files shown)


region/maridia/inner-yellow.json
```diff
@@ -1946,7 +1983,27 @@
...
             }
-          ]
+          ],
+          "leaveWithGModeSetup": [
+            {
+              "strats": [
+                {
+                  "name": "Get Hit By Zoa",
+                  "notable": false,
+                  "requires": []
```
osse101, line 1993: I think this has an implied h_canCrouchJumpDownGrab 
or 
jump into room and this room has normal physics.

Because you fall in the water entering this room.

osse101 (dismissed):

PR #929, opened 2023-07-15
Yellow Maridia East G-Mode

kjbranch:


region/maridia/inner-yellow.json
```diff
@@ -1572,6 +1573,23 @@
...
+                {
+                  "name": "G-Mode Morph",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "h_canArtificialMorphIBJ"
```
osse101, line 1588:
```suggestion
                    "h_canArtificialMorphIBJ",
                    {"or": [
                      "h_canArtificialMorphSpringBallBombJump",
                      "h_canArtificialMorphBombHorizontally"
                    ]}
```

region/maridia/inner-yellow.json
```diff
@@ -1631,6 +1649,24 @@
...
+                  "name": "G-Mode Morph",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    {"or": [
+                      "h_canArtificialMorphIBJ",
```
osse101, line 1663:
```suggestion
                      {"and": [
                        "h_canArtificialMorphIBJ",
                        "h_canAritficialMorphBombHorizontally"
                      ]},
```

region/maridia/inner-yellow.json
```diff
@@ -3206,7 +3322,139 @@
...
+                  "name": "G-Mode Morph IBJ, PB",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [2],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "h_canArtificialMorphIBJ",
+                    {"ammo": { "type": "PowerBomb", "count": 1 }},
```
osse101, line 3447: 1) The `h_canArtificialMorphSpringBallBombJump` strat is exactly this with an extra tech so won't be used.
2) Its hard to place the power bomb and avoid damage.  I could do it with double bomb jumps, or damage boosting on the ripper.  But you also need to know precisely where to put the PB.

kjbranch: I figured the spring ball variant was a bit easier and would be chosen first since it was higher in the list - it would have a different title and was considering a different description. I didn't notice they ended up having the exact same requirements. I'll update it to have a description that mentions spring ball instead of being two identical strats.

PR #933, opened 2023-07-16
Rename lavaPhysicsFrames, Add Heat and Acid Variants

kjbranch:


osse101:
suitless lava dive is not possible with gravity equipped.  You would have to turn off gravity even if you have it.  Plus all the lava movement numbers are done with no gravity.  Should strats like this be changed to gravitylessLavaFrames?

All cases should have a lower frame count following the gravity strat, and the gravityjump should be easier than some precise walljumps with other requirements being the same.  So maybe its ok as is.

In amphitheater there is only the triple springballjump that you cannot do with gravity on.  And the gravity jumps would need these gravityless frames added still.

kjbranch:
I would vote yes, as it would help make it more clear, and more accurate. But if the easier and less damaging versions are with gravity, it might not really matter much.

Reviewing Amphitheater was the inspiration for this PR - these should be added there as well

blkerby:
I like the improved accuracy here. For gravity jump Lava Dive, I think the gravitylessHeatFrames in this PR helps it be more correct in a scenario where Gravity suit affects heat damage (like in the vanilla game); in Map Rando it happens to not matter since removing Gravity doesn't affect heat damage, but for other randomizers it could matter.

PR #936, opened 2023-07-17
G-Mode in Quicksand rooms and Bug Sand Hole

kjbranch:


region/maridia/inner-yellow.json
```diff
@@ -2315,7 +2315,55 @@
...
+          "nodeAddress": "0x001a624",
+          "leaveWithGMode": [
+            {
+              "leavesWithArtificialMorph": false,
+              "strats": [
+                {
+                  "name": "G-Mode",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
```
osse101, line 2327: I think you need requirements for reaching the door, and for crossing the room.

kjbranch:
This isnt perfect, but I think its a reasonable simplification

PR #937, opened 2023-07-17
G-Mode Sand Holes

kjbranch:


region/maridia/inner-pink.json
```diff
@@ -7100,6 +7189,78 @@
...
+                  "name": "G-Mode Morph Suitless Through Crumbles",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "canSuitlessMaridia",
+                    "HiJump",
```
osse101, line 7256: A Morphless strat isn't relevant unless it can reach the item nodes but there is a shot block in the way.  Jumping from sand and a springballjump should be easier than gmode but it is a different set of tech 🤷‍♂️ 

There is value in a hjbless+springball+morph strat to 7.

osse101: Spinjumping out of the sand works.  And springballjumps are enough to get to the items.

GMode+HiJump+Morph would avoid a `canBreakFree` 🤔

osse101:
I like this max springballjump helper

PR #940, opened 2023-07-18
Pink Maridia G-Mode Setup and Immobile

kjbranch:


osse101 (approved):
There is a `leaveWithGModeSetup` on both Draygon doors using turrets.  But with the door locks its not something we can use in the vanilla room layout.

PR #941, opened 2023-07-18
Add Majority of Missing G-Mode in Pink Maridia

kjbranch:
Rooms that are still missing: 
- East Cac Alley, 
- Crossing the Colosseum with artificial morph IBJ (If it's possible, it sounds like a huge pain)
- Going down the sand pits in Botwoon E-Tank Room (currently a huge pain to represent and it can't work with this schema)

### DIFF
--- a/region/maridia/inner-pink.json
+++ b/region/maridia/inner-pink.json
@@ -3012,6 +3012,35 @@
               ]
             },
             {
+              "id": 3,
+              "strats": [
+                {
+                  "name": "G-Mode Morph",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    {"or": [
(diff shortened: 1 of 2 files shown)


region/maridia/inner-pink.json
```diff
@@ -3240,6 +3398,21 @@
                   "name": "Base",
                   "notable": false,
                   "requires": [ "Morph" ]
+                },
+                {
+                  "name": "G-Mode Morph Exit Tunnel Before Leaving G-Mode",
+                  "notable": false,
+                  "requires": [ {"obstaclesCleared": ["B"]} ]
```
osse101, line 3405: The strats which don't clear the obstacle don't appear to have a way out.  
4->5->1 is still useful for avoiding a tricky jump.

4->5->1 also sounds very not safe with the puyos

kjbranch: Right, that is the only option for indirect mode. The only way out would be if the item is Morph

PR #942, opened 2023-07-19
Tests

kjbranch:


tests/asserts/keywords.py
```diff
@@ -372,8 +475,97 @@ def process_strats(src, paramData):
...
+                            # Validate Requires Nodes
+                            requiresErrors = search_for_valid_keyvalue(
+                                [
+                                    "fromNode",
+                                    "fromNodes.",
+                                    "inRoomPath.",
+                                    "resetRoom.nodes.",
+                                    "nodesToAvoid.",
+                                    "itemNotCollectedAtNode",
+                                    "entranceNodes."
```
osse101, line 552: entranceNodes is part of the room description
`rooms: roomEnvironments: entranceNodes: []`

The node value would be a door (way of entering) node for the room.

tests/asserts/keywords.py
```diff
@@ -398,6 +591,17 @@ def process_strats(src, paramData):
                                                             msg = f"🔴ERROR: Destination node not defined:{fromNodeRef}"
                                                             messages["reds"].append(msg)
                                                             messages["counts"]["reds"] += 1
+                                                        if str(fromNode) in roomData["links"]["from"]:
```
osse101, line 594: Does `fromNode` update to become the previous "destinationNode"?

tests/asserts/keywords.py
miketrethewey: I've moved `entranceNodes` to it's own check and have it check against nodes that have `nodeType` `door` or `entrance`.
It'll be included in an upcoming commit.

tests/asserts/keywords.py
miketrethewey: I changed this to leverage `search_for_path()` which will try to find a multi-node pathway.
It'll be included in an upcoming commit.

osse101 (dismissed):

PR #943, opened 2023-07-19
Restructure East Cac Alley

kjbranch:
This node will be useful when adding g-mode strats, and it can help make the strats cleaner to help prevent missing things (the uwu strat was missing morph)


osse101:
But you didn't actually add the node 🧐

osse101:
Good add. Whoops about missing morph again.

kjbranch:
I ended up adding some G-Mode strats. Hopefully they wont be too hard, theyre all artificial morph ones.

region/maridia/inner-pink.json
```diff
@@ -6222,12 +6235,86 @@
...
+                  "name": "G-Mode Morph",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }},
+                    "Gravity",
+                    "h_canArtificialMorphJumpIntoIBJ"
```
osse101, line 6252: Getting to 4 looks like spike hits or diagonal IBJs

kjbranch: It's actually quite easy with spring ball and I think it could be assumed for this difficulty. I'll add the diagonals and a note

kjbranch: JK, the platform before this is the main problem. I could add an HBJ to pass those spikes, then a spring ball into IBJ ..

PR #944, opened 2023-07-19
X-Ray Rework

kjbranch:
Before I go through and convert everything, I was hoping to get feedback on this tech, specifically if you think it would be better, and if the tech and helper names seem reasonable.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -481,7 +481,14 @@
           "name": "h_canRightDoorXRayClimbFromWater",
           "requires": [
             "canXRayClimb",
-            "canRightSideDoorStuckFromWater"
+            {"or": [
+              "canRightSideDoorStuckFromWater",
+              {"and": [
+                "Gravity",
+                "canRightSideDoorStuck"
+              ]}
+            ]}
+
           ]
         }
       ]
--- a/helpers.json
+++ b/helpers.json
@@ -462,6 +462,13 @@
             "canSpringBallJumpMidAir",
             "canTrickyJump"
           ]
+        },
+        {
+          "name": "h_canRightwardXRayClimbSetup",
+          "requires": [
+            "canXRayClimb",
+            "canStationarySpinJump"
+          ]
         }
       ]
     },
--- a/helpers.json
+++ b/helpers.json
@@ -488,7 +488,6 @@
                 "canRightSideDoorStuck"
               ]}
             ]}
-
           ]
         }
       ]
--- a/helpers.json
+++ b/helpers.json
@@ -478,10 +478,10 @@
           ]
         },
         {
-          "name": "h_canSubmergedRightDoorXRayClimb",
+          "name": "h_canRightDoorXRayClimbFromWater",
           "requires": [
             "canXRayClimb",
-            "canSubmergedRightSideDoorStuck"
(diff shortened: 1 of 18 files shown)


tech.json
```diff
@@ -1109,38 +1109,25 @@
               "name": "canXRayClimb",
               "requires": [ "canXRayStandUp" ],
               "note": [
-                "All-purpose tech for X-Ray climbing by repeatedly using X-Ray's forced standup trick to climb up through walls.",
-                "This tech is uncoupled from different setups that place Samus in a position to do this and only refers to the actual climbing portion."
+                "Repeatedly using X-Ray's forced standup trick to climb up through in-bound walls.",
+                "This requires Samus to be partially clipped into the wall to begin climbing, which is often set up by getting stuck in a closing door.",
+                "The simplest setup is getting stuck in a right facing door in order to climb the left wall of a room.",
+                "To do this, enter this room from the room to the left with a stationary spinjump moving away to touch the transition with momentum that will get you stuck in the destination door."
```
osse101, line 1115:
```suggestion
                "To do this, enter this room from the room to the left with a stationary spinjump moving away to touch the transition with momentum that will get you stuck in the destination door.",
                "X-Ray climbs can also be setup by allowing solid blocks to reform on Samus, or by clipping into a wall by some other means."
```

helpers.json
```diff
@@ -462,6 +462,27 @@
...
+          "requires": [
+            "canXRayClimb",
+            "canRightSideDoorStuck"
+          ]
+        },
+        {
+          "name": "h_canRightDoorXRayClimbFromWater",
+          "requires": [
+            "canXRayClimb",
+            "canRightSideDoorStuckFromWater"
```
osse101, line 484:
```suggestion
            {"or": [
              "canRightSideDoorStuckFromWater",
              {"and": [
                "Gravity",
                "canRightSideDoorStuck"
              ]}
            ]}

```

tech.json
```diff
@@ -1109,38 +1109,25 @@
...
-                    "canStationarySpinJump"
-                  ],
+                  "name": "canRightSideDoorStuck",
+                  "requires": [ "canStationarySpinJump" ],
                   "note": [
-                    "Though using the same techs, the setup is a bit different for left-facing doors.",
-                    "It requires a dashing stationary spinjump into a doorcheck, then holding down as Samus arrives in the other room.",
-                    "Then use normal X-Ray climbing."
+                    "Getting stuck in a right side (left facing) door.",
+                    "Starting from the room to the right, perform a dashing stationary spinjump into a doorcheck, then hold down as Samus arrives in the desired room.",
```
osse101, line 1123:
```suggestion
                    "Starting from the room to the right, perform a dashing stationary spinjump into a doorcheck.",
                    "Then,  as Samus arrives in the desired room, hold down to automatically fall forward into the closing door.",
```

osse101 (commented):
Looks great.
`canEnterGMode` and `canShinesparkDeepStuck` can cover other ways to get doorstuck.
Also `canEnterGMode` should say you can use it to get doorstuck.

An adjacent topic is how XRay climb strats looks so close to being turned into entirely helpers, except you need that one node ID of input 😭 You could assume it starts on the door node in almost, but not, all cases too.

Issue #932, opened 2023-07-16
Crystal Flash Node Property


kjbranch:
Add a new property to nodes which indicates the requirements to CF at that node.

Things to consider: [Some of these are probably not worth modeling]
- Requirements should be able to include general requirements (ways to kill enemies or have enemy obstacles cleared)
- Damaging environmental effects need to apply before and after the CF (heat/acid/lava frames before, and CF might not fill Samus with a lot of tanks in lava or acid)
- CF can also clear obstacles (enemies, bomb blocks, yellow doors)
- CF can be done with 1 less PB in some instances

osse101:
CF with 10pbs would need some kind of tech indicating advanced farming strat.  Its tricky to get the sequence of motions without practice.  But I've seen it done in frog speedway, Kronic Boost room, Caterpillar room.  Any room with near guaranteed pb drops, and a way to stay aligned with the exploding power bomb, by jumping straight up and down, or placing it in a corner. 

More nodes should be added to some rooms.  Lava dive mainly, also long straight rooms.  

Non heated rooms would need to allow CFs; as ways to gain full health or setup a GMode trick.

Can put it on the GT fight in canSafeSpotGT notable strats.

kjbranch:
Closing this, as the CF strats were added in a different way. Some of these issues could still apply, but may or may not be worth adding. A new issue could be created for some of the remaining missing strats.

PR #947, opened 2023-07-20
Add 'refill' logical requirement to schema & docs

blkerby:
This adds a new logical requirement type to represent a process that refills certain resource types, potentially in the middle of a strat. This is proposed as a generic, flexible building block applicable to many situations, for example:
- farming in the middle of a boss fight (e.g. GT)
- using a Crystal Flash in the middle of a heat run (or lava/acid dive)
- non-infinitely spawning enemy farms (in conjunction with some other requirement to indicate ability to reset the room through an accessible door)

This would also allow us to represent re-spawning farms (and uses of the Ship or recharge stations) as normal strats, e.g. in a link from a node to itself, eliminating the need for some special handling.

With refills being combined with other resource requirements, it matters what order requirements are applied in: for example you could have heat damage at start of GT room, followed by a refill during the fight, followed by some heat damage at the end of the fight; and in order for the logic to be correct these requirements must be applied in their chronological order. So we update the docs to clarify that requirements in an "and" are to be applied in the order listed.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -408,7 +408,8 @@
             {"ammo": { "type": "PowerBomb", "count": 1 }},
             {"ammo": { "type": "Missile", "count": 10 }},
             {"ammo": { "type": "Super", "count": 10 }},
-            {"ammo": { "type": "PowerBomb", "count": 10 }}
+            {"ammo": { "type": "PowerBomb", "count": 10 }},
+            {"refill": ["Energy"]}
           ]
         },
         {
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -24,6 +24,9 @@ __Example:__
 ]}
 ```
 
+Requirements are applied in the order in which they are listed; this can matter in cases where "refill",
+"ammoDrain", or "energyAtMost" requirements are involved with other resource usage requirements.
+
 #### or object
 An `or` object is fulfilled if at least one of the logical elements it contains is fulfilled
 
@@ -72,10 +75,6 @@ __Example:__
   "count": 75
 }}
 ```
-
-__Additional considerations__
-
-Whenever an `ammoDrain` object is part of a strat, it should be applied after all other ammo costs.
 
 #### enemyKill object
 An `enemyKill` object communicates the need to kill a given set of enemies, and is satisfied by having the necessary items to use one of the valid [weapons](weapons/weapons-readme.md) that will kill each of the enemies (as well as enough ammo, if applicable).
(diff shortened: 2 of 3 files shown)


kjbranch (approved):

osse101:
I see the usage looking like nodes linking to themselves, with a strat for `refill:[X]` for what we decide are good drops from local enemies.  Strats would have path to door <-> enemy requirements , and enemies have a drop requirement that each would be paid per farm.  Indicating how `refill` means farm local enemies repeatedly, vs other enemies, vs search for a refill isn't clear.  But it should be this room's enemies, and the other options would be the randomizers problem.

"Passively" refilling with on path enemies, with an effecient weapon, would be accurate player behaviour, except for all the people I see skip easy farms.

You can CF anywhere so that would be better written on the node.  But midfight CFs could be added to strats, and we script generally how the fight goes.  And cost order is needed here.  A GMode `refill` needs to be a CF so that may still be written as `do a CF`.

Structure to the order of requirements is good.

Farming bosses could turn into many similar strats for various equipment sets.  Or maybe we get a nice and accurate boss scenario, I'm not really sure.

Replacing farm strats with regular strats sounds nice.

blkerby:
Yeah, the way you describe using it sounds like what we have in mind too. We can throw in "canBePatient" (hopefully soon to be split into "canBePatient" and "canBeVeryPatient") requirements on obnoxious farms. The time requirements will depend on the total amount of farming, which means it won't necessarily align perfectly with "3-minute" definition of "canBePatient". We thought about a different way of doing it where we would include a farm cycle time and have the randomizer figure out how much farming can be afforded within the limits of the enabled patience tech; but that's a lot more work, leads to some problems, and could be hard for players to understand (e.g. less transparent when trying to understand the logic through the logic pages). So yeah, it seems simpler or more appealing to just define by hand which farms are "good" ones (for specific resources) and which ones need patience tech.

I added the "refill" to the "h_canCrystalFlash" helper now. There are edge cases that will have to be handled carefully and might not be fully expressible yet (e.g. crystal flashing in acid with so many tanks that you no longer get a full refill). But I think this will support almost all the farms and Crystal Flash applications that are currently missing from logic.

As far as how to encode regular Crystal Flash uses (i.e. ones not in the middle of a strat), I'm not sure which approach is best: we could represent them as regular strats from a node to itself with a "h_canCrystalFlash" requirement, or we could have some new specific structure inside the node instead. That could maybe be a separate discussion under #932, which hopefully wouldn't need to block this change and getting started on non-respawning enemy farm strats?

PR #954, opened 2023-07-21
Farms in Central Crateria

blkerby:
This PR starts with a small number of rooms, to get feedback on the approach before doing more.

To decide which patience tech (if any) to require for a farm, I try to estimate roughly the expected amount of time to collect a certain amount of the given resource, and compare that with thresholds of 1.5 minutes (canBePatient), 3 minutes (canBeVeryPatient), and 12 minutes (not in logic):

Energy: 100 (a full ETank)
Missile: 2
Super: 2
Power Bomb: 2

For Missiles they come in units of 2 anyway, while for Supers it's possible they will be for a gate glitch so a second one is for leniency. For Power Bombs it's because there's a significant possibility of needing two to break blocks in two different places. It's all a little arbitrary but the goal is just to get things roughly on the right scale for the amount of time/patience required.

For now, to handle the worst-case scenario, I'm assuming an expected loss of 10 energy per cycle in case the neighboring room is heated. We could add requirements for a non-heated neighboring room in cases where this would bring a strat into a lower patience tier, but that's a possible refinement for later.

Considering a scenario where you spend 8 seconds in-room to do the farm, plus 7 seconds to reset the room (let's assume no fast doors), for a total of 15 seconds cycle time, the numbers above translate into the following thresholds for the total drop rate on Supers or Power Bombs (summed over all the enemies killed per cycle):

No patience required: 34% or more
canBePatient: 17% - 33%
canBeVeryPatient: 4% - 16%
Out of logic: <4%

If a room has a cycle time in that 15-second ballpark (as seems to be common) I'll just use these ranges. And if its cycle time seems a little higher or lower, I'll try to adjust accordingly, but without trying to grind the room to optimize the times. It's all guesswork anyway so no point trying to make it super precise.

### DIFF
--- a/region/crateria/central.json
+++ b/region/crateria/central.json
@@ -2374,6 +2374,42 @@
                   "requires": []
                 }
               ]
+            },
+            {
+              "id": 8,
+              "strats": [
+                {
+                  "name": "Basic Farm",
+                  "notable": false,
+                  "requires": [
+                    "f_ZebesAwake",
+                    {"resetRoom": {
+                      "nodes": [3, 4, 6, 7],
+                      "mustStayPut": false
+                    }},
+                    {"refill": ["Energy", "Missile"]}
+                  ],
+                  "devNote": "FIXME: Other nodes could be used to reset the room, with additional requirements."
+                },
+                {
+                  "name": "Very Patient Farm",
+                  "notable": false,
+                  "requires": [
+                    "canBeVeryPatient",
+                    "f_ZebesAwake",
+                    {"resetRoom": {
+                      "nodes": [3, 4, 6, 7],
+                      "mustStayPut": false
+                    }},
+                    {"refill": ["Energy", "Missile", "Super", "PowerBomb"]}
+                  ],
+                  "note": [
+                    "The three Skrees each have a 2% Super and 2% Power Bomb drop rate.",
+                    "The Geemers each have a 2% Super drop rate."
+                  ],
+                  "devNote": "FIXME: Other nodes could be used to reset the room, with additional requirements."
+                }
+              ]
             }
           ]
         }
@@ -3490,6 +3526,27 @@
                   "requires": []
                 }
               ]
+            },
+            {
+              "id": 6,
+              "strats": [
+                {
+                  "name": "Basic Farm",
+                  "notable": false,
+                  "requires": [
+                    "f_ZebesAwake",
+                    {"resetRoom": {
+                      "nodes": [1, 5],
+                      "mustStayPut": false
+                    }},
+                    {"refill": ["Energy", "Missile"]}
+                  ],
+                  "devNote": [
+                    "The Space Pirates (wall) drop Supers and Power Bombs each at a 0.4% rate, too low to put into logic.",
+                    "FIXME: Other nodes could be used to reset the room, with additional requirements."
+                  ]
+                }
+              ]
             }
           ]
         }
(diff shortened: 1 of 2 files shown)


osse101:
I did a round of farming in parlor and it took 7.5 minutes for 2pbs and 4 supers, using the 3 geemers and 3 screes at the top.  It took slightly more than 15s round trip, but then I got better at it and time spent went down probably under 15s.

Refilling to full pbs would be dreadful.  Reaching 2 is ok for veryPatient, and if you really know the drop table.  I guess you are meant to browse the wiki when an uncommon farm spot comes up?  Maybe you even need a quick enemy drop reference sheet now.
`Missile 2` is much lower than `Health 100`.  100 health is like 6 big drops if all in one room, or 11 big drops if its 1 enemy per room reset.  Missiles, splitting with health is like 5 drops / 5 drops with that setup.  So 10 missiles would be a better equivalent.  And energy would still be the slower refill because of small health drops.  Missiles are used in a lot of kill strats, but also are really convenient for farming, like with alcoons.  Alcoon farm is 20s (no transitions) with power beam and about 7s with missiles.

It says `refill` so I take that to mean to full.  What you've written is good for a round of farming, but they need to be repeated to reach a full refill.  And that could just mean adding an `__Additional considerations__` to `refill` that non farm point enemy refills only account for refilling a small amount with regards to `canBePatient`.  The randomizer would know how many resources are needed to be filled and would convert that to expected time.  You could need 2 pbs to break a wall, or 20 to jumpIntoCrystalFlashClip.
Randomizer could decide if its possible to refarm this room more times, with some more patience logic, or to look for a refill room instead at that point.

blkerby:
Unfortunately, the way the graph traversal works with Map Rando, the randomizer doesn't know how many resources need to be filled. In theory maybe it could be overhauled to make it be able to do that, but there's no easy obvious way, so I'd say it's impractical for now. So we need the `refill` requirements to handle the worst-case scenario of how much refill might be required.

I hadn't been thinking about farming for Crystal Flash (or even worse, CF clips, like you mention). We'll probably be adding a bunch more of those to logic soon, so that could be a problem. Like you said, farming 20 PBs with some of these slow farms would be terrible. Even for Extreme/Insane difficulties, adding very slow farms to logic is probably not something that would make the game more fun. 

So I'm thinking maybe it would make sense to keep things simpler and not try to include any "Patient" or "Very Patient" farms. For Supers/PBs we could just include the ones where you expect at least 1 per cycle. It seems like the drop rates are bimodal anyway, where they're close to 100% or else very low, so that would make it a lot more cut-and-dry to decide what is in logic and what is not, with probably no need to measure cycle times. This could also make it easier for players to get a handle on which farms are in logic. 

If that sounds like a better approach, we may want to consider also reconsidering the re-spawning bug farms. For example, Zebbos are pretty slow to farm PBs with (2% drop rate) so maybe they shouldn't be considered in-logic either for PBs.

kjbranch:
There's not a good way to represent only being able to farm a small amount of a resource, because if you can farm a few PBs, then the logic could have you do that over and over until you're full. This type of problem already exists and probably always will - if you have `canBePatient` turned off, it could give you many slow strats in a row that each almost require it and together certainly should. Similarly, there is already a problem for farming slow, infinitely respawning farm, like Green Hill Zone and Red Tower for `Supers` and `PBs`.

I think that the cutoff for these looks reasonable, as the time required to farm 2 PBs is probably enough to make progress on the seed or to get to another, better farm. I would, however, increase the number of missiles required from 2 to 10 ish. It probably won't change just about anything, but it could be particularly annoying to farm missiles at a place where getting more than 2 would require patience.

kjbranch:
That sounds fine to me as well. Simply adding the good PB/Super farms will significantly improve the logic as is, maybe that's a good enough step.

Although, presumably if you need to do a strat to CF with 10 leniency PBs, it is probably safe to assume that it's faster for the player to find a save and reload, instead of farming for PBs.

osse101:
I like keeping it to the very high drop rate items.  There are cases where you need one more power bomb to do a thing, but most of the time you are not happy farming low% drop rate pbs.  And its a lot easier to remember which enemies have a high rate for certain items.

Just about everything drops energy and missiles so there's a bit there to think about regarding number of drops per reload and small vs big energy chance.

There's always the risk of a farm being between two ammo intensive steps with no alternative, but leniency does help a lot with that.  Farm spots are so much nicer than reloading rooms and still its not fun past 1 or 2 rare drops.

osse101 (approved):

PR #957, opened 2023-07-22
Rest of Crateria and Brinstar farms

blkerby:


region/brinstar/green.json
```diff
@@ -2082,6 +2143,47 @@
...
+                            {"or": [
+                              {"ammo": {"type": "Missile", "count": 1}},
+                              {"ammo": {"type": "Super", "count": 1}},
+                              {"ammo": {"type": "PowerBomb", "count": 1}}
+                            ]}    
+                          ]}
+                        ]}
+                      ]}
+                    ]},
+                    {"refill": ["Energy", "PowerBomb"]}
```
osse101, line 2175: Beetoms aren't known for their energy drops.  Is 4 smalls at a time ok?

Missile/Super would give a power bomb to use to kill the rest, and then refill more pbs.  That's a neat interaction.

region/brinstar/green.json
```diff
@@ -3778,6 +3938,22 @@
...
           "from": 1,
           "to": [
+            {
+              "id": 1,
+              "strats": [
+                {
+                  "name": "Kihunter Farm",
+                  "notable": false,
+                  "requires": [
+                    {"resetRoom": {
```
osse101, line 3948: I think I would like a beam/screw for this farm.

region/brinstar/kraid.json
```diff
@@ -1631,6 +1691,44 @@
...
+                      "Ice",
+                      "Wave",
+                      "Spazer",
+                      "Plasma",
+                      {"and": [
+                        "Bombs",
+                        "canCarefulJump"
+                      ]}
+                    ]},
+                    {"refill": ["Energy", "Missile", "Super"]}
```
osse101, line 1717: I think the pirate farm should be separate.  And you are mainly gaining missiles there.

region/brinstar/kraid.json
```diff
@@ -1631,6 +1691,44 @@
...
+                  ]
+                },
+                {
+                  "name": "Mini-Kraid Farm",
+                  "notable": false,
+                  "requires": [
```
osse101, line 1723: mini kraid should have some requirements.
2 spike hits, or supers, or missiles, or speed, or beams.  There are a lot of options, but with nothing you'll get hit first.

osse101 (commented):
Farming Kihunterse with basic beam would be annoying.  They do die in few hits, but are harder to hit than most enemies and fight back.  And I don't think of them as an enemy to farm for drops but maybe that could change.

Beetoms are going to drop small energies so its not very effecient.  Its also pretty easy to take some damage even with ice when trying to shoot them.

Moving farm point refills into strats is logical.

region/brinstar/green.json
blkerby: Right, the "Energy" refill was a mistake here, removed that now.

region/brinstar/green.json
blkerby: Done. Should this also be added to the Crateria Kihunter farm?

region/brinstar/kraid.json
blkerby: The reason for combining them is so that we could include the option to use Supers (farmed from Mini-Kraid) to kill the Pirates

region/brinstar/kraid.json
blkerby: I took another pass this. Got rid of the separate Mini-Kraid farm and added those requirements to the combined farm. Removed the option of using Bombs to kill the Pirates since that is not fun. With the slow methods (everything except Screw, Plasma, or Super), assumed either a careful jump or a stone hit from Mini-Kraid; spike hit is also possible but just wanted to include the worst case.

region/brinstar/green.json
osse101: I'm more ok with that one.  A bit easier to fight them, and crabs are there for a better health farm.
However crabs are on the line of tedious and worth with just power beam.

region/brinstar/kraid.json
osse101: On second thought, it'd be organized to have two separate farms, but that's not actually a good reason to.  A couple things are simpler with them grouped.

blkerby: Cool, and this can always be refined later if you have a different idea for how to split up the strats. I suppose Mini-Kraid could still be a useful energy farm even with no way to kill the Pirates, and that isn't expressed with the combined strat here

Issue #952, opened 2023-07-21
Add canPauseAbuse tech and apply to heated farms


blkerby:
A pause abuse tech could be used to make the heat frame requirements more precise for re-spawning enemy farms in heated rooms (e.g. as came up recently in Acid Snakes Tunnel). It should also possibly be a requirement for using the farm in Lower Norfair Farming Room (with Grapple but no heat protection).

PR #958, opened 2023-07-22
Add canPauseAbuse tech, refine heated farms

blkerby:
This adds a new tech canPauseAbuse for pausing to prevent death just before collecting an energy drop. (For the moment at least, manual reserve usage is out of scope of the tech.)

Farming strats are added for heated rooms, using the new tech. I took the opportunity to tighten some of the heat frames as well. Bat Room could still use some tightening for the walljumpless strats.

Closes issue #952.

### DIFF
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -6437,7 +6437,7 @@
                         {"heatFrames": 50}
                       ]}
                     ]},
-                    {"refill": ["Energy", "Missile", "Super"]}
+                    {"refill": ["Energy", "Super"]}
                   ]
                 }
               ]
@@ -6503,7 +6503,7 @@
                         {"heatFrames": 50}
                       ]}
                     ]},
-                    {"refill": ["Energy", "Missile", "Super"]}
+                    {"refill": ["Energy", "Super"]}
                   ]
                 }
               ]
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -6421,6 +6421,27 @@
               ]
             },
             {
+              "id": 3,
+              "strats": [
+                {
+                  "name": "Zebbo Farm",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
+                      {"and": [
+                        "canPauseAbuse",
+                        "Grapple"
+                      ]},
+                      {"and": [
+                        "h_heatResistant",
+                        {"heatFrames": 50}
+                      ]}
+                    ]}
+                  ]
+                }
+              ]
+            },
+            {
               "id": 4,
               "strats": [
                 {
@@ -6460,6 +6481,27 @@
                   "requires": [
                     "h_canNavigateHeatRooms",
                     {"heatFrames": 210}
+                  ]
+                }
+              ]
+            },
+            {
+              "id": 4,
+              "strats": [
+                {
+                  "name": "Zebbo Farm",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
+                      {"and": [
+                        "canPauseAbuse",
+                        "Grapple"
+                      ]},
+                      {"and": [
+                        "h_heatResistant",
+                        {"heatFrames": 50}
+                      ]}
+                    ]}
                   ]
                 }
               ]
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -6436,7 +6436,8 @@
                         "h_heatResistant",
                         {"heatFrames": 50}
                       ]}
-                    ]}
+                    ]},
+                    {"refill": ["Energy", "Missile", "Super"]}
                   ]
                 }
               ]
@@ -6501,7 +6502,8 @@
                         "h_heatResistant",
                         {"heatFrames": 50}
                       ]}
-                    ]}
+                    ]},
+                    {"refill": ["Energy", "Missile", "Super"]}
                   ]
                 }
               ]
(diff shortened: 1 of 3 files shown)


region/norfair/east.json
```diff
@@ -3348,14 +3380,24 @@
...
-                        "canWalljump",
-                        "HiJump",
-                        "SpaceJump",
-                        "canUseFrozenEnemies"
-                      ]},
+                    {"or": [
+                      "canWalljump",
+                      "HiJump",
+                      "SpaceJump",
+                      "canTrickyUseFrozenEnemies",
```
blkerby, line 3387: Note this was "canUseFrozenEnemies" before, and we're changing it to "canTrickyUseFrozenEnemies" here. Also added a Spring Ball jump strat.

region/lowernorfair/east.json
```diff
@@ -6420,6 +6420,27 @@
...
+              "id": 3,
+              "strats": [
+                {
+                  "name": "Zebbo Farm",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
+                      {"and": [
+                        "canPauseAbuse",
+                        "Grapple"
```
osse101, line 6433: I assume we're trying to refill here, but not gaining health here is accurate for below expert, even with grapple.

region/lowernorfair/east.json
```diff
@@ -6420,6 +6420,28 @@
...
+                      ]},
+                      {"and": [
+                        "h_heatResistant",
+                        {"heatFrames": 50}
+                      ]}
+                    ]},
+                    {"refill": ["Energy", "Missile", "Super"]}
```
osse101, line 6440: `refill: missile` would need heatproof

region/lowernorfair/east.json
blkerby: Whoops, yeah forgot the "refill" part. Added that now. The idea is that "canPauseAbuse" will go into Expert so you won't be able to logically gain health from this farm without that (or heat resistance). Aside from the precision needed to farm quickly with Grapple, the problem was that you could run out of energy while getting ammo drops, with the health bomb not being enough to reliably save you without pause abuse

region/lowernorfair/east.json
osse101: and `refill: Energy` is something like `canBeVeryPatient`.  I think you get like 20 health back per cycle if done really well.  But doing that increases the risk of mistakes and those almost like you 100.

blkerby: Good catch, removed that now. I'm planning on doing a separate PR for heatproof and/or non-respawning Norfair farms and can include the Missile one there (along with one to farm Power Bombs from the Violas)

blkerby: Just saw the comment about energy. Yeah, that's a good point. Maybe we could add a link from each farm to the opposite door node with no heat frames, as a (admittedly hacky) way to represent that you can farm just enough to get through the room without a net loss. And then we could include a canBeVeryPatient alternative that fully refills

osse101: It does decently match how you try to go through the room while saving health.  Drop RNG becomes a potential problem though.  Usually only when not full on supers.

blkerby: Ok I took another pass at this. The assumption is that refilling on Supers & Power Bombs (but not energy) should be reasonable with Expert tech; it may take a while to do (and farm back up to the original energy level) but not to an extent requiring patience tech? Even with "bad" RNG (e.g. lots of Super drops quickly), you can still survive with health bomb and pause abuse, right?

osse101: At health bomb its about every third farm cycle you gain enough health to get ammo drops again.  Pbs are still a rare drop from the bugs, but the Violas are right there.  At health bomb its hard to get any pbs from violas too.

blkerby: I was imaging the PB drops would come from resetting the room and killing Violas again, though you lose health with that so it would be slow. Do you think Supers & PB refill should just be removed here (or only included with canBeVeryPatient)?

region/lowernorfair/east.json
```diff
@@ -6420,6 +6452,57 @@
...
+                      "mustStayPut": false
+                    }},
+                    {"refill": ["PowerBomb"]}
+                  ]
+                },
+                {
+                  "name": "Zebbo and Viola Farm",
+                  "notable": false,
+                  "requires": [
+                    "h_heatProof",
```
osse101, line 6482: heatproof and pauseabuse aren't needed together

region/norfair/east.json
```diff
@@ -3372,7 +3414,18 @@
                   "notable": false,
                   "requires": [
                     "h_canNavigateHeatRooms",
-                    {"heatFrames": 250}
+                    {"or": [
+                      {"heatFrames": 230},
+                      {"and": [
+                        "canDamageBoost",
+                        {"heatFrames": 200},
+                        {"enemyDamage": {
```
osse101, line 3422: fwiw I make it out with more health with a contact damage boost by dropping faster.  By waiting for the particles to despawn I can get the same health on exit as a particle dboost, but the boost would be easier to time.

osse101: 160 heatframes + contact
185 heatframes + particle
250 heatframes  by waiting
And there's a neat double skree farm that is 210 heatframes + 2 drops

osse101 (approved):

region/norfair/east.json
blkerby: Oh cool, I was able to reproduce that for the contact method; best I could get was 48 damage in the room total, so 38 heat damage = ~152 heat frames. 160 heat frames seem good to be doable consistently. With the particle boost I'm leaving with about 5 less energy. Seems good to put in the contact method, updated that now.

PR #963, opened 2023-07-23
Other Norfair Farms

blkerby:


region/norfair/east.json
```diff
@@ -308,6 +308,25 @@
...
+                  "requires": [
+                    "h_heatProof",
+                    {"or": [
+                      "canWalljump",
+                      "SpaceJump",
+                      "h_canUseSpringBall"
+                    ]},
+                    {"refill": ["Energy", "Missile", "Super"]}
+                  ],
+                  "note": "You can stay on the left part of the room to avoid interacting with the Gerutas."
```
osse101, line 326: `Sm. Dessgeega`

region/norfair/east.json
```diff
@@ -4298,6 +4451,34 @@
...
               ]
+            },
+            {
+              "id": 6,
+              "strats": [
+                {
+                  "name": "Alcoon and Multiviola Farm",
+                  "notable": false,
+                  "requires": [
+                    {"or": [
```
osse101, line 4462: Screw, heatproof

region/lowernorfair/west.json
```diff
@@ -2787,6 +2787,25 @@
...
+                  "requires": [
+                    "h_heatProof",
+                    "ScrewAttack",
+                    {"resetRoom": {
+                      "nodes": [1],
+                      "mustStayPut": false
+                    }},
+                    {"refill": ["Energy", "Super"]}
+                  ],
+                  "devNote": "Resetting using node 2 would also be possible, with Supers."
```
osse101, line 2805: 👎

osse101 (commented):
KiHunters with full beam would be ok to add.
Gold pirates with full beam+charge would work in this context.  They might get a shot off before you can charge a beam, but when resetting the room you are ready.

Curious how fast you get supers from bubble mountain wavers.  One per reset probably.

blkerby:
Farming just the bottom 2 Wavers each with 24% Super drop rate, it would be an expected 0.5 Supers per cycle, so a little slow I guess. Should we keep it?

blkerby:
Makes sense, removed the Waver farm from the bottom of the room now. Like you said, it's slow, and with canDelayedWalljump you can reach the Cacatac/Waver farm above anyway. I'm on the fence about the ripper: it's a bit far away down in the corner, and for low difficulty settings it may not be well known that it's a good source of supers/energy. On higher settings, again the Cacatac/Waver farm on top is reachable so I think we could probably say that covers it.

PR #974, opened 2023-07-25
Fix Beetom farms

blkerby:
This is to handle the situation where you may come into the room with low energy, where health bomb would prevent you from getting Power Bomb drops from Beetoms initially. We could refine this further if we had a requirement type to check for a minimal amount of a resource (energy), i.e. >= 30 energy, without consuming it. In practice, 4 Missiles will probably be available anyway, so not much is lost by doing it this way, by assuming you just need to reload the room a few times to get your health up first.

We also change the "PowerBomb" requirement to a "resourceCapacity" requirement: this will be more clear for readers of the logic; the meaning of a "PowerBomb" requirement is not intuitive and kyleb is working on phasing it out.

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -2135,18 +2135,7 @@
                       "ScrewAttack",
                       {"and": [
                         "Ice",
-                        "Morph",
-                        {"or": [
-                          "Bombs",
-                          {"and": [
-                            {"resourceCapacity": [ {"type": "PowerBomb", "count": 1} ]},
-                            {"or": [
-                              {"ammo": {"type": "Missile", "count": 4}},
-                              {"ammo": {"type": "Super", "count": 4}},
-                              {"ammo": {"type": "PowerBomb", "count": 4}}
-                            ]}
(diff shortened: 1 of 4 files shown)


osse101:
Clearing a room of Beetoms damageless with one missile needs a tech of some kind (or notable).  You need to let the Beetoms attach to Samus in a very controlled way and freeze them before damage happens.  And that's not obvious.
The old way with one missile could assume you use that missile to get a pb to clear the other 3 Beetoms for more PBs.

osse101:
```
or: [
  {enemydamage: Beetom, contact, 4}
  canEscapeEnemyGrab
]
```
?

blkerby:
Contact damage won't model it correctly because that is affected by suits whereas the health bomb threshold is not. If you meant taking 4 contact hits on every reload, then you would lose health instead of gaining health. canEscapeEnemyGrab is for something different, wiggling them off of you, right? Maybe canDodgeWhileShooting? Though it seems like this is getting too complicated and maybe we should just remove this case

PR #975, opened 2023-07-25
Partial Lower Norfair East Obstacle Rework

kjbranch:
This is the file with the most obstacles, by far, so I figured it would be best split in two.

### DIFF
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -4764,7 +4764,7 @@
                   "notable": false,
                   "requires": [
                     "h_canNavigateHeatRooms",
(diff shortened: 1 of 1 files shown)


region/lowernorfair/east.json
```diff
@@ -4648,6 +4540,15 @@
             {
               "id": 7,
               "strats": [
+                {
+                  "name": "Base",
+                  "notable": false,
+                  "requires": [
+                    "h_heatProof",
```
osse101, line 4547:
```suggestion
                    "h_canNavigateHeatRooms",
```

PR #981, opened 2023-07-27
Tourian Escape 1 Right Door Helper

kjbranch:


blkerby (commented):
I think the runways should be one tile less (length 2 and 20). The tile in front of the transition (where a door cap normally would be) isn't counted as part of the runway because the transition triggers as soon as any part of Samus touches the transition tile, skipping almost all of the tile in front of it (and when Samus enters the room, she spawns beyond that tile too).

kjbranch:
I think the runway stuff is confusing. The runway to the left door is 21 tiles, but I guess that's because there's no door cap. I think runways on nodes assume the door is closed and cant be opened, so that reduces it by 1 tile or something. 

I may have added all of the cross room jumps incorrectly (off by 1). I think using a door ledge in the adjacent room is supposed to be 0 tiles in the adjacent room, where I counted it as 1. But if the adjacent room requirement was 0 tiles that would seem weird to me. I also don't think rooms with just a doorsill runway would have a runway on their node of 0 length, so maybe they cant even be used as an adjacent runway...?

osse101:
One less runway tile to account for door shell looks in line with how other runways, including shellless doors, are done.

When entering a room you are placed an extra tile away from the transition and that cuts into how much space you have to run.  In fact, I recently found out `NorthwestMaridiaBugRoom` places you farther forward than other rooms.

When entering, you can shoot the first shutter's bottom half to run through.  Probably need wave.  Its what you do in speedruns to run through the room without slowing down; Shoot down then shoot forward with hyper.  Entering with run speed changes the interaction and would be something to test.

kjbranch:
I knew about the Bug room, and was worried it was a common thing that would need to be measured for every room. I didnt have a good way to test with speed, but looks like wave and spazer work well

PR #983, opened 2023-07-28
Add More Patience Reqs

kjbranch:
I added some more options for the `Snail Walk`. I didn't add the super method, because it brings the fast snail instead, and it would be close to also requiring patience. In the end, after trying it for a bit and then simply jumping across with an angle button, I am wondering if this should just be removed and the simpler option be added - thoughts?

### DIFF
--- a/region/lowernorfair/east.json
+++ b/region/lowernorfair/east.json
@@ -4140,7 +4140,7 @@
                   "notable": false,
                   "requires": [
                     "h_heatProof",
-                    "canBeVeryPatient",
+                    "canBePatient",
                     {"enemyKill":{
                       "enemies": [[ "Kihunter (red)", "Kihunter (red)" ]]
                     }}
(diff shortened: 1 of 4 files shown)


blkerby:
To me it seems like the angle-jump method is simpler, easier, faster and has no requirement of taking damage. So I don't see any reason to keep a snail walk strat. What do you think, @osse101 ?

kjbranch:
I mostly feel like `canPlayInSand` can add a description of how to walk across sand easily with the angle method. Then if this strat can bypass `canPlayInSand` it might be weird, especially if this doesnt remain notable

osse101:
Snail walk is moreso a gimmicky thing you can do than a real strat.

osse101 (dismissed):

kjbranch:
Let me know what you think of these proposed changes

osse101:
I checked if GMode would have a use case, but you clip down the best between the two sandpits and a shot block prevents going down the far right like you normally would.  
And I don't think we're trying to be completionist here if the strats aren't serving a purpose.

PR #989, opened 2023-07-29
Damage Boost Rework

kjbranch:
Split damage boost tech into two: `canNeutralDamageBoost` and `canHorizontalDamageBoost`. Remove `canContinuousDboost`, as its 3 uses were notable horizontal ones.

Looking for help with the `canHorizontalDamageBoost` description

### DIFF
--- a/region/maridia/inner-pink.json
+++ b/region/maridia/inner-pink.json
@@ -4950,7 +4950,7 @@
                         "canSuitlessMaridia",
                         "canPlayInSand",
                         "HiJump",
-                        "canDamageBoost",
+                        "canHorizontalDamageBoost",
                         "canSpringBallJumpMidAir",
                         {"spikeHits": 2}
                       ]},
@@ -4968,7 +4968,7 @@
                     {"spikeHits": 1},
                     {"or": [
                       {"spikeHits": 1},
-                      "canDamageBoost"
+                      "canHorizontalDamageBoost"
                     ]}
                   ],
                   "note": [
@@ -4983,7 +4983,7 @@
                   "requires": [
                     "canWalljump",
                     "canTrickyJump",
-                    "canDamageBoost",
+                    "canHorizontalDamageBoost",
                     "canIframeSpikeJump",
                     {"spikeHits": 3},
                     {"or": [
@@ -5305,7 +5305,7 @@
                     "canIframeSpikeJump",
                     {"spikeHits": 3},
                     {"or": [
-                      "canDamageBoost",
+                      "canHorizontalDamageBoost",
                       {"spikeHits": 1}
                     ]}
                   ],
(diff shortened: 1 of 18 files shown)


tech.json
```diff
@@ -729,15 +729,17 @@
               ]
             },
             {
-              "name": "canDamageBoost",
+              "name": "canNeutralDamageBoost",
               "requires": [ "canUseEnemies" ],
-              "note": "Using the knockback from taking enemy damage to assist Samus in her movement.",
-              "devNote": "FIXME This could use a better description",
+              "note": "Using the small vertical knockback from taking enemy damage, while not holding any directional inputs, to assist Samus in her movement.",
```
osse101, line 734: While morphed you can hold directions.  `canDamageBoost` was a better name for this in that sense.

tech.json
```diff
@@ -729,15 +729,17 @@
...
               "extensionTechs": [
                 {
-                  "name": "canContinuousDboost",
-                  "requires": [ "canDamageBoost" ],
-                  "note": "A tech that involves multiple damage boosts with precise timing. No room for missing one of them."
+                  "name": "canHorizontalDamageBoost",
+                  "requires": [ "canNeutralDamageBoost" ],
+                  "note": [
+                    "Using the knockback from taking enemy damage to give Samus a large horizontal knockback.",
+                    "?"
```
osse101, line 741:
```suggestion
                    "Using the knockback from taking enemy damage to give Samus a large movement boost by holding the jump and backwards buttons.",
                    "By changing the timing of the directional input, the way that Samus is knocked back can be changed and has situational uses.",
                    "Holding forward will knock Samus backwards and down.",
                    "Holding jump but delaying the back button delays the horizontal damage boost to gain some neutral boost first.",
                    "Very quickly releasing back and pressing forward will give a forward boost."
```

region/maridia/inner-pink.json
```diff
@@ -4950,7 +4950,7 @@
                         "canSuitlessMaridia",
                         "canPlayInSand",
                         "HiJump",
-                        "canDamageBoost",
+                        "canHorizontalDamageBoost",
```
osse101, line 4953: This would be a neutral damage boost.
This one knocks you above the water, the spike hits that follow are neutral for positioning but aren't used in a way to move Samus.
And the bottom `or`'d spike it is the horizontal damage boost.

region/maridia/inner-pink.json
```diff
@@ -5305,7 +5305,7 @@
                     "canIframeSpikeJump",
                     {"spikeHits": 3},
```
osse101, line 5306: I do a controlled neutral boost on the far left spike for positioning, maybe you can uninteruptable animation frames to avoid that.

tech.json
kjbranch: I thought `canNeutralDamageBoost` was better because I was worried that people would think they already know what a _damage boost_ is, and toggle it based on that.

region/maridia/inner-pink.json
```diff
@@ -5305,7 +5305,7 @@
...
                     {"or": [
-                      "canHorizontalDamageBoost",
```
osse101: This horizontalDamageBoost refers to the two right side platforms where you boost from one to the other.

tech.json
osse101: Sounds good, I'm cool with leaving it `canNeutralDamageBoost`

region/maridia/inner-pink.json
```diff
@@ -5305,7 +5305,7 @@
...
+                      "canNeutralDamageBoost",
```
osse101, line 5308: This horizontalDamageBoost refers to the two right side platforms where you boost from one to the other.

kjbranch: did this fix it?

Issue #988, opened 2023-07-29
Adjust reqs for Double Chamber walljump up left side


blkerby:
This strat seems to be one that people frequently struggle with. Right now with HiJump it's in Map Rando logic in Basic, otherwise Medium, and there's not much lenience (on Basic the logic says it could be done with just 1 tank, on Hard with no tanks). If the player doesn't know what they're doing, it's pretty brutal. A couple examples:

- https://youtu.be/flqjV0FBokU?t=1326
- https://clips.twitch.tv/CrispyWildRamen4Head-3VcGtoIo6pI1OYaM (3rd attempt, after which they started using save states)

Many things can go wrong here: getting fireballed by the Namihe, bonking the Namihe while trying to reach the Kamer, or the Kamer being at the wrong height. I think this strat would also be a good one to make notable: bringing visibility to it with a detailed note could help, since the easier way to do this strat (running under the Kamer to get it to come down) is not very intuitive. 

To get up there optimally (not waiting for the Kamer to come down) I think also requires canDelayedWalljump. Heat frames could probably be tightened in that case.

On Medium settings and lower, I think it should require h_heatProof. So we could just split off the suitless version as a notable and put that in Hard or Very Hard. I guess Hard probably makes more sense, so that it can describe the easier method without the delayed walljump.

osse101:
Traditionally left side double chamber wasn't worth the trouble at very hard level, vanilla map randomizers. With or without hijump.  Wild that its in Basic.

Yeah optimally you delayed walljump off the very bottom of the fune head, but that should only come up if you enter the room in spinjump, otherwise you can lower the kamer.
I've heard killing the Fune helps, but that was in reference to climbing up right side.

PR #992, opened 2023-07-30
Double Chamber Wall Jump

kjbranch:
I know this doesn't address all of the concerns in #988, but this is how I would write the strat.
- It's still in `Basic` with HiJump, but with ~50% more frames. 
- It's also still in `Medium` without HiJump, but now requires `canPreciseWalljump` in addition to `canConsecutiveWalljump`, and either more than 2x the frames or `canCarefulJump` and 1.5x the frames.
- It's now tighter with `canDelayedWalljump`

I wouldn't be opposed to moving `canPreciseWalljump` into `Hard`

### DIFF
--- a/region/norfair/east.json
+++ b/region/norfair/east.json
@@ -4746,18 +4746,73 @@
               "id": 1,
               "strats": [
                 {
-                  "name": "Walljump",
+                  "name": "Walljump on the Kamer",
                   "notable": false,
                   "requires": [
                     "h_canNavigateHeatRooms",
+                    "canPreciseWalljump",
+                    "canConsecutiveWalljump",
+                    "canUseEnemies",
                     {"or":[
-                      "canConsecutiveWalljump",
-                      {"and": [
-                        "HiJump",
-                        "canWalljump"
+                      {"heatFrames": 560},
+                      {"and":[
+                        "canCarefulJump",
+                        {"heatFrames": 384}
+                      ]},
+                      {"and":[
+                        "canTrickyJump",
+                        {"heatFrames": 276}
+                      ]},
+                      {"and":[
+                        {"or":[
+                          "canUseFrozenEnemies",
+                          {"ammo": { "type": "Super", "count": 1 }}
+                        ]},
+                        {"heatFrames": 432}
+                      ]},
+                      {"and":[
+                        {"ammo": {"type": "PowerBomb","count": 1}},
+                        {"heatFrames": 456}
                       ]}
-                    ]},
-                    {"heatFrames": 250}
+                    ]}
+                  ],
+                  "note": [
+                    "Wall jump up the left wall and then on the moving platform (Kamer) while avoiding the Fune's fireball.",
+                    "The Kamers will temporarily move down if Samus is below them, so it is best to walk under them before climbing the wall.",
+                    "It is possible to freeze or kill the Fune with a Super or Power Bomb to make things easier."
+                  ]
+                },
+                {
+                  "name": "Delayed Walljump",
+                  "notable": false,
+                  "requires": [
+                    "h_canNavigateHeatRooms",
+                    "canDelayedWalljump",
+                    "canConsecutiveWalljump",
+                    "canUseEnemies",
+                    {"heatFrames": 200}
+                  ],
+                  "note": "Wall jump up the left wall, perform a delayed wall jump on the Fune to reach the door."
+                },
+                {
+                  "name": "Walljump with HiJump",
+                  "notable": false,
+                  "requires": [
+                    "h_canNavigateHeatRooms",
+                    "HiJump",
+                    "canUseEnemies",
+                    "canWalljump",
+                    {"or":[
+                      {"heatFrames": 360},
+                      {"and":[
+                        "canCarefulJump",
+                        {"heatFrames": 250}
+                      ]}
+                    ]}
+                  ],
+                  "note": [
+                    "Wall jump up the left wall and then on the moving platform (Kamer) while avoiding the Fune's fireball.",
+                    "The Kamers will temporarily move down if Samus is below them, so it is best to walk under them before climbing the wall."
                   ]
                 },
                 {
--- a/region/norfair/east.json
+++ b/region/norfair/east.json
@@ -4822,6 +4822,7 @@
                     "h_canNavigateHeatRooms",
                     "canDelayedWalljump",
                     "canConsecutiveWalljump",
+                    "canTrickyJump",
                     "canUseEnemies",
                     {"or": [
                       {"heatFrames": 190},
--- a/region/norfair/east.json
+++ b/region/norfair/east.json
@@ -4706,6 +4706,16 @@
           "dropRequires": ["h_heatProof"]
         }
       ],
+      "reusableRoomwideNotable": [
+        {
+          "name": "Double Chamber Walljump Climb Using the Kamer",
+          "note": [
+            "Wall jump up the left wall and then on the moving platform (Kamer) while avoiding the Fune's fireball.",
+            "The Kamers will temporarily move down if Samus is below them, so it is best to walk under the first Kamer before climbing the wall.",
+            "It is possible to freeze or kill the Fune with a Super or Power Bomb to make things easier."
+          ]
+        }
+      ],
       "links": [
         {
           "from": 1,
@@ -4746,8 +4756,8 @@
               "id": 1,
               "strats": [
                 {
-                  "name": "Walljump on the Kamer",
-                  "notable": false,
+                  "name": "Double Chamber Walljump Climb Using the Kamer",
+                  "notable": true,
                   "requires": [
                     "h_canNavigateHeatRooms",
                     "canPreciseWalljump",
@@ -4761,7 +4771,7 @@
                       ]},
                       {"and":[
                         "canTrickyJump",
-                        {"heatFrames": 276}
+                        {"heatFrames": 240}
                       ]},
                       {"and":[
                         {"or":[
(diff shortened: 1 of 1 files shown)


region/norfair/east.json
```diff
@@ -4746,18 +4746,73 @@
...
+                },
+                {
+                  "name": "Delayed Walljump",
+                  "notable": false,
+                  "requires": [
+                    "h_canNavigateHeatRooms",
+                    "canDelayedWalljump",
+                    "canConsecutiveWalljump",
+                    "canUseEnemies",
+                    {"heatFrames": 200}
```
blkerby, line 4793:
```suggestion
                    {"or": [
                      {"heatFrames": 200},
                      {"and": [   
                        "canPrepareForNextRoom",
                        {"heatFrames": 120}
                      ]}
                    ]}
```

region/norfair/east.json
```diff
@@ -4746,18 +4746,79 @@
               "id": 1,
               "strats": [
                 {
-                  "name": "Walljump",
+                  "name": "Walljump on the Kamer",
                   "notable": false,
```
blkerby, line 4750:
```suggestion
                  "notable": true,
```

region/norfair/east.json
```diff
@@ -4746,18 +4746,79 @@
...
+                  ],
+                  "note": [
+                    "Wall jump up the left wall and then on the moving platform (Kamer) while avoiding the Fune's fireball.",
+                    "The Kamers will temporarily move down if Samus is below them, so it is best to walk under them before climbing the wall.",
+                    "It is possible to freeze or kill the Fune with a Super or Power Bomb to make things easier."
+                  ]
+                },
+                {
+                  "name": "Delayed Walljump",
+                  "notable": false,
```
blkerby, line 4787:
```suggestion
                  "notable": true,
```

region/norfair/east.json
```diff
@@ -4746,18 +4746,79 @@
...
+                        "canPrepareForNextRoom",
+                        {"heatFrames": 120}
+                      ]}
+                    ]}
+                  ],
+                  "note": "Wall jump up the left wall, perform a delayed wall jump on the Fune to reach the door."
+                },
+                {
+                  "name": "Walljump with HiJump",
+                  "notable": false,
```
blkerby, line 4805:
```suggestion
                  "notable": true,
```

region/norfair/east.json
```diff
@@ -4746,18 +4746,79 @@
               "id": 1,
               "strats": [
                 {
-                  "name": "Walljump",
+                  "name": "Walljump on the Kamer",
```
blkerby, line 4749:
```suggestion
                  "name": "Double Chamber Left Side Walljumps",
```

region/norfair/east.json
```diff
@@ -4746,18 +4746,79 @@
...
+                    ]}
+                  ],
+                  "note": [
+                    "Wall jump up the left wall and then on the moving platform (Kamer) while avoiding the Fune's fireball.",
+                    "The Kamers will temporarily move down if Samus is below them, so it is best to walk under them before climbing the wall.",
+                    "It is possible to freeze or kill the Fune with a Super or Power Bomb to make things easier."
+                  ]
+                },
+                {
+                  "name": "Delayed Walljump",
```
blkerby, line 4786:
```suggestion
                  "name": "Double Chamber Delayed Walljumps",
```

region/norfair/east.json
```diff
@@ -4746,18 +4746,79 @@
...
+                      {"and": [   
+                        "canPrepareForNextRoom",
+                        {"heatFrames": 120}
+                      ]}
+                    ]}
+                  ],
+                  "note": "Wall jump up the left wall, perform a delayed wall jump on the Fune to reach the door."
+                },
+                {
+                  "name": "Walljump with HiJump",
```
blkerby, line 4804:
```suggestion
                  "name": "Double Chamber Walljumps with HiJump",
```

blkerby (commented):
I think it would be good to make all three of these notable, allowing us to bump up their difficulty tiers in the randomizer, as well as add visibility for them in the UI:

with HiJump: Basic -> Medium
without HiJump: Medium -> Hard
delayed walljump: Hard -> Very Hard

kjbranch:
Not sure I agree. A lot of the reason I wanted to work on this issue is because I was thinking these strats wouldnt necessarily need to be notable if they were logically represented correctly. 

If we do need to increase their difficulties (which I'm not sure we do), I would think the notable aspect would just learning how the kamers work. So with and without hijump should use the same notable. This could put them both into the difficulties you suggested if `canPreciseWalljump` is moved to hard. As for the delayed version, how is this harder than other delayed wall jumps, meriting it to be notable?

osse101:
I'm not seeing HiJump make this any easier unless I lower the kamer, and that would go in the kamer strat... killing the Fune actually does make it easier to get up. (Can't get up at all without HJB or the Kamer while killing the fune).

I think the use of Ice is if you mess up, or overjump?  So that would be `OR`'d with carefuljump.

Walljumping up to the Kamer is under 240 frames for me, from door.  Its not really a careful or tricky jump though.  Full jump left, full jump right, jump to door.  Technically not a consecutivewalljump either, but I think two full jumps counts as a conescutivewalljump.  

HJBless Karmerless left side you can also jump up into two walljumps and looks really reliable so I'll switch to that from now on.  Full jump->walljump->full jump puts you are the exact right spot for an easy walljump off the fune.  Its hardly a delayedwalljump done this way, when normally it definately is.

Manipulating the Kamer is not obvious and could be notable for that reason.
HJBless delayed walljump is off a spot lower than where you think it possible to walljump so I could go either way with it.
HiJump delayed walljump seems normal.

kjbranch:
I find HiJump to help a ton. At the very minimum it makes the wall jump off of the kamer to be a normal wall jump, whereas without it, its a pretty tight wall jump. If the kamer is in the best position it's even possible to completely avoid that wall jump, but that's not something that should be required at this level.

To me, ice feels similar to killing the fune. You no longer need to worry about the fireballs.

Two wall jumps is not consecutive. This would be the only location I know of where it would be used with two, but I figure it was ok, since without higher level strats youll probably want to get some extra wall jumps to lower the kamer, but it is not required. It could be removed here if you would like it to be.

kjbranch:
The version with and without HiJump both use the Kamer. Only the delayed walljump strat skips it

kjbranch:
I made them notable even tho I mostly disagree with them being notable. They could be notable for knowing to walk under the kamer to lower it, but then some of those strats would be usable without the notable turned on, which would be confusing (the ones without a careful jump). The strats without a careful jump are to represent the player climbing the wall to lower the kamer and then reclimbing, so those ones (which should be in a lower difficulty) are the ones that should have consecutive wall jump while the others dont.... This is why I just chose for all the ones without hijump to require consecutive and all those with to not require it.

osse101:
Jumping towards the Kamer is enough to lower it, then just adding an extra attempts worth of heat frames would make it not notable 🤔

osse101 (commented):
I don't know what I'm supposed to do differently to differentiate carefuljump and trickyjump.  If carefuljump lands on the kamer and avoids a fireball, the framecount is a smidge high.

From my experience I think the left side climb is harder than it looks, or feels harder than it looks.  So I'd like `canTrickyJump` added to the delayed walljump strat.
A lot of that could be from not knowing that well how the Kamer worked and misaligning fireballs with the low platform.

Coming from the right, HJBless gets a worse fireball, so you do still want the careful/ tricky / neither split.  With HiJump, you take an extra jump to check its in the right place, then go?

kjbranch:
Looks like everyone wants something different here. I'm going to close this for now and maybe come back to it later with some new ideas, or anyone else can work on it.

blkerby:
I think how this PR handles it seems good. I like the idea of having at least one notable in order to call out to the player what the expectations are and to show them how to do it (using the animated GIFs & descriptions), and to give them an option to toggle it off. Single reusable notable seems good here, and I like the idea of canTrickyJump to limit the hardest version (not covered by the notable) to Very Hard.

blkerby:
I pushed a commit to the branch, to add the canTrickyJump to the delayed walljumps version (though since the PR is currently closed I guess it doesn't show here?). @osse101 If you have ideas for how to refine the requirements, maybe you could make a code suggestion or create a new PR based on this branch; I would try to incorporate your other feedback but I'm having some trouble following some of the details to interpret exactly what you mean.

osse101:
I liked what was here.  
I just wanted `canTrickyJump` added to the delayed walljump strat.
And 
```
{"and":[
  "canCarefulJump",
  {"heatFrames": 384}
]},
```
changed to 360.

Then my last comment was to say 2->1 looks like the main focus, but 5->2->1 changes how the Kamer acts but it can be treated similarly.  `canPrepareForNextRoom` Wouldn't apply in that case, so the numbers would be different 🤔

kjbranch:
Added those. 
The 5->2->1 is a common problem in a lot of rooms. For example, 2->1->4 would also have this same problem, as it assumes you go to the door. Hopefully 5->2->1 is not too far off, and hopefully it's conservative. Can anyone test it?

A more extreme example that might be worth adding is going 3->5->4 with climbing the right wall; that would save a lot of frames relative to the current logic that goes through 2 and 1.

osse101:
5->2->1 is about 1 second faster than 5->2+2->1, so the `canPrepareForNextRoom` strat uses about the same heatframes and the other strats have extra.
5->2->1->4 is 3 seconds slower than 5->4.  Going fast is kinda hard because of a fune flame so 2 seconds may be a more average number.  
Testing just the HJBless delayedwalljump strats.

osse101:
Then right side with hijump is about 2.5s faster, but can get up to 3.5s faster than 5->2->1->4 if you grind it out, and use a super on the fune.

But I'm cool with putting that change in a separate PR.  These left side changes look good without reworking the room.

PR #994, opened 2023-07-31
Bump up requirement for Frozen Pirate Plasma Escape

blkerby:
This upgrades the `canUseFrozenEnemies` requirement to `canTrickyUseFrozenEnemies` and adds more details to the description of the strat. This only affects the 3 -> 1 link, the one that requires killing all the pirates, which is what makes the strat tricky.

### DIFF
--- a/region/maridia/inner-yellow.json
+++ b/region/maridia/inner-yellow.json
@@ -4994,12 +4994,14 @@
                   "name": "Frozen Wall Pirate Plasma Escape",
                   "notable": false,
                   "requires": [
-                    "canUseFrozenEnemies",
+                    "canTrickyUseFrozenEnemies",
                     "canPreciseWalljump"
                   ],
                   "note": [
                     "It's possible to climb up with just ice by freezing the wall space pirates (who can be frozen and killed without Plasma).",
-                    "Clearing the grey door lock would require killing the frozen pirate after using it to climb."
+                    "Clearing the grey door lock would require killing the frozen pirate after using it to climb.",
+                    "To do this most easily, freeze the Pirate as high as possible on the left wall.",
+                    "After reaching the top, it can help to reset the camera by moving to the right on the center platform and then back left."
                   ]
                 },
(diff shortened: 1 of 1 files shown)


blkerby:
Some things which potentially make this strat "tricky":
- The wall pirate jumps back and forth from wall to wall, making it annoying to position
- Its hitbox when frozen is much larger than what is visible, making the wall jump less intuitive,
- You have to get it higher than you might expect, or it will be too far off camera to shoot once you're on top.
- Once on top, you probably have to manipulate the camera to get the pirate into range to shoot.

osse101:
No question thats its a trickyFreeze.

Positioning the pirate is bad.  If the pirate gets low, you can get locked in a bad cycle if you don't know how to handle it.
It looks a lot like `canConsecutiveWalljump`, but there are ways around it, like riding the wall pirate like a crab climb.  Or freeze it as it jumps across.
Then the camera manip is more involved than most camera manips.

region/maridia/inner-yellow.json
```diff
@@ -4994,12 +4994,14 @@
...
                   "requires": [
-                    "canUseFrozenEnemies",
+                    "canTrickyUseFrozenEnemies",
                     "canPreciseWalljump"
                   ],
                   "note": [
                     "It's possible to climb up with just ice by freezing the wall space pirates (who can be frozen and killed without Plasma).",
-                    "Clearing the grey door lock would require killing the frozen pirate after using it to climb."
+                    "Clearing the grey door lock would require killing the frozen pirate after using it to climb.",
+                    "To do this most easily, freeze the Pirate as high as possible on the left wall.",
```
osse101, line 5003:
```suggestion
                    "To do this most easily, freeze the Pirate as high as possible on the left wall.",
                    "Bait the pirate to jump across the room early so that it stays high on the wall, or bring it on camera by jumping below the platform if the wall pirate gets too low.",
```

PR #1000, opened 2023-08-01
Blue Brinstar: G-Mode Immobile and Setups

kjbranch:


region/brinstar/blue.json
```diff
@@ -1007,6 +1112,66 @@
...
+                    "h_canUsePowerBombs",
+                    {"or": [
+                      {"canShineCharge": {
+                        "usedTiles": 33,
+                        "shinesparkFrames": 43,
+                        "openEnd": 2
+                      }},
+                      {"and": [
+                        "canShinechargeMovementComplex",
+                        "canFastWalljumpClimb",
```
osse101, line 1153: This would consecutive walljump up instead of spark... A will be destroyed by the spark, so the check for `obstacleCleared` can be removed here.

region/brinstar/blue.json
```diff
@@ -1007,6 +1112,66 @@
...
+                          "shinesparkFrames": 17,
+                          "openEnd": 2
+                        }}
+                      ]}
+                    ]}
+                  ],
+                  "note": "Falling down the shaft and breaking the crumble block does not require Morph."
+                },
+                {
+                  "name": "Get Hit By Geemer - Blocks Already Broken",
```
osse101, line 1165: Would it be possible to remove the other strats and only use this one?  Have it find the path where you go down and come back up, or start down and come up.

kjbranch: I dont believe that obstacles can be included in cross room strats with the way the logic is currently implemented. It definitely cant be in Map Rando. I will try to work with Maddo to put together a doc about the limitations and some ideas for a new structure soon.

PR #1003, opened 2023-08-02
Blue Brinstar G-Mode Uses

kjbranch:


region/brinstar/blue.json
```diff
@@ -238,6 +261,95 @@
...
+                        {"ammo": { "type": "PowerBomb", "count": 1 }}
+                      ]},
+                      {"and": [
+                        "Morph",
+                        "ScrewAttack",
+                        "canTrickyJump"
+                      ]}
+                    ]}
+                  ],
+                  "clearsObstacles": ["D"],
```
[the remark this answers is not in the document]
kjbranch: Right, this shouldnt break A or C. The power bomb is immediately after entry to avoid 2 more hopper hits

region/brinstar/blue.json
```diff
@@ -1733,6 +2051,40 @@
...
+                      "canTrickyJump",
+                      "canMoonwalk",
+                      "canXRayTurnaround",
+                      "Morph"
+                    ]}
+                  ],
+                  "note": "In direct g-mode, the door will not close, and the in-room doorsill gives enough running room to make it up to the invisible ledge."
+                },
+                {
+                  "name": "G-Mode Morph Spring Ball IBJ",
```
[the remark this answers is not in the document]
kjbranch: gravity makes this free

region/brinstar/blue.json
```diff
@@ -1336,6 +1593,66 @@
...
+                  "requires": [
+                    "f_ZebesAwake",
+                    {"ammo": {"type": "PowerBomb","count": 1}},
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "any",
+                      "artificialMorph": true
+                    }}
+                  ],
+                  "clearsObstacles": ["A","B"]
```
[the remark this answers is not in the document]
kjbranch: I was going back and forth on how to solve this, but it looks like I kind of went half way between my 2 ideas. I'll change it to have the PBs be an obstacle, that will be the most clear way to solve it.

region/brinstar/blue.json
```diff
@@ -1239,6 +1416,198 @@
...
+                    "h_canArtificialMorphMovement",
+                    {"or": [
+                      "canConsecutiveWalljump",
+                      "h_canFly"
+                    ]}
+                  ],
+                  "note": [
+                    "Overload PLMs with a single Power Bomb by hitting the item at the right end of the room.",
+                    "This can be done with a single precisely placed Power Bomb. There should be 4 empty tiles between Samus and the left of the blocks when the bomb is placed."
+                  ]
```
osse101, line 1582:
```suggestion
                  "note": [
                    "Touch the item to overload PLMs so that Samus can pass through the crumble block and bomb block.",
                    "Ending gmode will collect this item so this strat can only be used once."
                  ]
```

region/brinstar/blue.json
```diff
@@ -1239,6 +1416,198 @@
...
+                      "canConsecutiveWalljump",
+                      "h_canFly"
+                    ]}
+                  ],
+                  "reusableRoomwideNotable": "Blue Brinstar Energy Tank G-Mode Flashing Lights",
+                  "note": [
+                    "The eye scanners are particularly annoying while in g-mode. They scan Samus with bright flashing lights which remain for a further distance.",
+                    "This is notable so a player can disable having to enter these flashing lights. If disabled, Samus will only require being in g-mode in this room if Zebes is awake.",
+                    "Overload PLMs with a single Power Bomb by hitting the item at the right end of the room.",
+                    "This can be done with a single precisely placed Power Bomb. There should be 4 empty tiles between Samus and the left of the blocks when the bomb is placed."
```
osse101, line 1606:
```suggestion
                   "Touch the item to overload PLMs so that Samus can pass through the crumble block and bomb block.",
                   "Ending gmode will collect this item so this strat can only be used once."
```

region/brinstar/blue.json
```diff
@@ -1239,6 +1416,198 @@
...
+                  ]
+                },
+                {
+                  "name": "G-Mode Morph Touch the Item",
+                  "notable": false,
+                  "requires": [
+                    {"itemNotCollectedAtNode": 4},
```
osse101, line 1565: This doesn't visit 4 to get the item so I don't think the logic tracks this as a one time only strat.

region/brinstar/blue.json
```diff
@@ -1239,6 +1416,198 @@
...
+                      "fromNodes": [1],
+                      "mode": "direct",
+                      "artificialMorph": true
+                    }},
+                    {"ammo": {"type": "PowerBomb","count": 1}},
+                    {"or": [
+                      "canConsecutiveWalljump",
+                      "SpaceJump"
+                    ]},
+                    {"or": [
```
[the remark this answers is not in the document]
kjbranch: I think the duplication is a problem, and will be in many room as we get more detail on strats. I think I can fix it here by adding 2 more nodes.

region/brinstar/blue.json
kjbranch: Technically, you can go through the door in order to not pick up the item.

osse101: The door can't be opened in gmode

region/brinstar/blue.json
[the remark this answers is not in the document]
osse101: While there is a lot of duplication in the room, I do just mean within the strat with this comment.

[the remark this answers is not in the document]
kjbranch: I started working on reducing the duplication before seeing this, but it looked like it was going really well, so I finished it off

region/brinstar/blue.json
```diff
@@ -1338,6 +1539,146 @@
...
+                      {"enemyDamage": {
+                        "enemy": "Geemer (blue)",
+                        "hits": 1,
+                        "type": "contact"
+                      }},
+                      {"comeInWithGMode": {
+                        "fromNodes": [1],
+                        "mode": "any",
+                        "artificialMorph": true,
+                        "mobility": "immobile"
```
osse101, line 1569: Not that its technically wrong, but immobile would require direct and should lead to the direct junction.

osse101 (approved):
Looks a lot cleaner

region/brinstar/blue.json
```diff
@@ -1494,6 +1858,103 @@
...
+                  "note": [
+                    "Overload PLMs with a single Power Bomb by hitting the item at the right end of the room.",
+                    "This can be done with a single precisely placed Power Bomb. There should be 4 empty tiles between Samus and the left of the blocks when the bomb is placed."
+                  ]
+                },
+                {
+                  "name": "G-Mode Morph Touch the Item",
+                  "notable": false,
+                  "requires": [
+                    {"itemNotCollectedAtNode": 4},
```
osse101, line 1931: I don't think this is modelled correctly still.
1) Logic wouldn't know the player has the item and can be at the top door.
2) Logic doesn't know that this is a one time use strat.  

It needs some kind of `{remoteItemCollection: 4}` object or something.

kjbranch: I could add an obstacle to enable it to grab the item on the way, but there's no way to force the item collection. I was discussing a remote collection with Maddo and he had major concerns with it. I'll ask him to respond on that.

I wonder if we could have a strat to or from a node that forces item pickup at that node (like going through Botwoon E-Tank tunnel cant skip the item). We could also add a way to indicate that a strat can only be done once (like the tube gravity jump)

blkerby: We wouldn't have to force item collection if we just be sure that the strat can only be used to collect the item and nothing else. That could be done by having the strat go from the junction (6) to the item node (4) while clearing a new abstract obstacle; then there can be a strat from 4 back to 1 just requiring that obstacle.

I don't think I would try to model strats using the item to pass through the room to the other door (while collecting it). We could consider that later but it would seem to require some schema changes and not be usable in randomizer logic anyway due to the risk of permanent loss of access.

kjbranch: Currently I've been putting in all strats that I can think of. Those that go to the item and back are listed without `canRiskPermanentLossOfAccess`, and those that do require it.

blkerby: We could add a new strat-level property "itemsCollected" (similar to "obstaclesCleared"), to represent items that are (and must be) collected as a side-effect of performing the strat. I don't think Map Rando would be able to consume this data in the foreseeable future, but it could be useful anyway to help make the model more complete, since it could be useful later in some way that we haven't thought of

PR #1012, opened 2023-08-09
Migrate repo to organization

miketrethewey:
This appears to be less-breaking than I thought.
Waiting for a time when @blkerby is available to update stuff when I pull the lever.
@tewtal is less-reliant on having stuff re-pointed immediately.
Redirects *should* function appropriately when that happens.


blkerby:
Cool, sounds good. I don't foresee any issues on my end. Feel free to go ahead whenever you like

PR #1020, opened 2023-08-21
Update Thread The Needle Room

kjbranch:
Make it a little more lenient for lower difficulties by adding `canDodgeWhileShooting`, or other ways to dodge or kill the enemies. 

It's assumed that some of the enemies can be killed easily (mostly the Choots) and can be used to refill missiles enough for the trickier enemies in the middle of the room.

### DIFF
--- a/region/maridia/inner-yellow.json
+++ b/region/maridia/inner-yellow.json
@@ -3597,8 +3597,15 @@
                   "name": "Choot and Puyo Farm",
                   "notable": false,
                   "requires": [
+                    {"or": [
+                      "h_hasBeamUpgrade",
+                      "canDodgeWhileShooting",
+                      {"resourceCapacity": [
+                        {"type": "Missile", "count": 1}
+                      ]}
+                    ]},
                     {"resetRoom": {
-                      "nodes": [1, 2],
+                      "nodes": [1],
                       "mustStayPut": false
                     }},
                     {"refill": ["Energy", "Missile", "Super", "PowerBomb"]}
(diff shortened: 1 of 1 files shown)


osse101 (commented):
I'll have to test to see what I think about 2 puyo hits to cross the room.  At that number you would stop to shoot everything and they can still get out of control in that case.

I'm thinking Morph strats should be made their own strat.  Then I guess the idea is you enter the water right away and roll to the middle to lay your power bomb.  And the 1 hit option is for crossing the middle but I'm not sure on which side you take the hit.

CanDodgeWhileShooting is a little weak for what we want imo, But I think power beam only is about a very hard strat so its close, and there's nothing better.  I was thinking of canBePatient as an option.

Missile only has the issue of missing shots, but power beam is there as backup and is probably fine.

Farming supers and PBs with basic beam can't be fast enough to be enjoyable.

Add ScrewAttack and Grappling Beam as options.

kjbranch:
I still think it's reasonable to cross the room with only 2 hits. Without `CanDodgeWhileShooting` will still attempt to kill the enemies that are safe to kill. Most of the challenge is in the center of the room. Most of the danger is getting on top of the middle platform; after that it's usually easier to shoot down diagonally while staying safe. 

`Grapple` should definitely be added. I will look into `canBePatient` and `Screw Attack` as other options.

I can look into also adding `canBePatient` to the farms. I'll talk to Maddo and see where the threshold for patience was definied for farms. 

I was also debating adding `Gravity` in places, as that makes the room easier, but it didnt seem to really lower the threshold for any of the strats - maybe it could be used as an alternative to lower 2 Puyo hits down to 1. I can look into that as well.

kjbranch:
Screw is a little scary through the middle of the room, but it makes the room easy enough to farm, so even if a hit is taken it's not a big deal.

kjbranch:
I think these are a lot easier than `canTrickyJump`, especially since you can farm to full before/after if you really mess up. I think the `canDodgeWhileShooting` is enough to be included with no hits.

osse101:
The `canTrickyJump` idea would be for killing no enemies, and there's no logical reason to do that.  But I was thinking about learning that route, cause its fast.

I like how this is written now.

osse101 (approved):

PR #1022, opened 2023-08-21
Maridia - Max height spring ball, tricky frozen enemies

kjbranch:
Added a few more uses of `h_canMaxHeightSpringBallJump`
Updated some `Ice` strats to use `canTrickyUseFrozenEnemies`
Added a new strat to climb Main Street: https://discord.com/channels/1053421401354285129/1053436236628496454/1129587005311951009

### DIFF
--- a/region/maridia/outer.json
+++ b/region/maridia/outer.json
@@ -1792,14 +1792,24 @@
                   "notable": true,
                   "requires": [
                     "canSuitlessMaridia",
-                    "h_canMaxHeightSpringBallJump",
+                    "canSpringBallJumpMidAir",
                     "canTrickyUseFrozenEnemies",
-                    "canStationaryLateralMidAirMorph"
+                    {"or": [
+                      "canCrazyCrabClimb",
+                      {"and": [
+                        "h_canMaxHeightSpringBallJump",
+                        "canStationaryLateralMidAirMorph"
+                      ]}
+                    ]}
                   ],
                   "note": [
-                    "The tricky part is getting to the ledge below the missiles. Stand on a frozen crab with a frozen fish at the lowest height of its cycle to the right.",
-                    "Perform a max height spring ball jump, with a stationary lateral midair morph to get onto the fish before it thaws.",
-                    "Press against the fish while moving upwards in order to gain more height in order to get on top of it.",
+                    "The tricky part is getting to the ledge below the missiles. There are two ways to do this:",
+                    "1. Stand on a frozen crab with a frozen fish at the lowest height of its cycle to the right. Further to the right is better as long as it can be stood upon without the ledge above.",
+                    "Perform a stationary lateral midair morph into a max height spring ball jump to get onto the fish before it thaws.",
+                    "Press against the fish for the maximum amount of time possible while moving upwards in order to gain more height in order to get on top of it.",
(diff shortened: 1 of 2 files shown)


region/maridia/outer.json
```diff
@@ -2915,8 +2931,7 @@
                   "notable": false,
                   "requires": [
                     "canSuitlessMaridia",
-                    "canSpringBallJumpMidAir",
-                    "canCrouchJump",
+                    "h_canMaxHeightSpringBallJump",
```
osse101, line 2934: I think I'd change the strat to recommend using Global crab to assist with all of the jumps to the top.  You wouldn't need `maxHeight`, `stationaryLateral`, `trickyJump`, and probably not `crouchJump` either.
The only hard part is quickly freezing the above Missiles Crab so that you can use two crabs together.

region/maridia/outer.json
```diff
@@ -3229,9 +3244,9 @@
                   "notable": false,
                   "requires": [
                     "canSuitlessMaridia",
-                    "canUseFrozenEnemies",
+                    "canTrickyUseFrozenEnemies",
                     "HiJump",
-                    "canSpringBallJumpMidAir"
+                    "h_canMaxHeightSpringBallJump"
```
osse101, line 3249: This jump doesn't need to crouch so its kinda lenient

region/maridia/outer.json
```diff
@@ -2933,14 +2916,27 @@
                     "canSuitlessMaridia",
                     "h_canMaxHeightSpringBallJump",
                     "canTrickyUseFrozenEnemies",
+                    {"or": [
+                      {"and": [
+                        "h_canMaxHeightSpringBallJump",
+                        {"or": [
+                          "canTrickyJump",
```
osse101, line 2923: `h_canMaxHeightSpringBallJump` includes canTrickyJump.

osse101:
If we GMode or temp blue to the missile item, we can't climb back up to top door 🤔

region/maridia/outer.json
kjbranch: It does, but I think it would be nice to make it it's own tech which doesn't rely on tricky jump after this.

kjbranch:
g-mode should be fine, since it wont break the blocks, but it uses the global crab, so you probably can't catch back up to it - so it will need to cycle the room, or maybe with a super.

temp blue will break the blocks. the only way this could be used with the global crab to go back up is possibly with a super, but that could be a pain.

region/maridia/outer.json
```diff
@@ -1784,14 +1784,40 @@
...
+                },
+                {
+                  "name": "Main Street Climb with Ice and Spring Ball",
+                  "notable": true,
+                  "requires": [
+                    "canSuitlessMaridia",
+                    "canSpringBallJumpMidAir",
+                    "canTrickyUseFrozenEnemies",
+                    {"or": [
+                      "canCrazyCrabClimb",
```
osse101, line 1798: I'd call this either not crazy crab climb or not notable.  Its not that hard of a crab freeze.  You can freeze it right as it passes a corner on the wall so you're not going to fall through the crab.  But balancing the fish freeze and the two crab freezes are something.

osse101 (approved):

PR #1027, opened 2023-08-22
Add canMaxHeightSpringBallJump

kjbranch:


region/maridia/inner-pink.json
```diff
@@ -8325,9 +8325,11 @@
                     ]},
                     {"or": [
                       {"and": [
-                        "canSpringBallJumpMidAir",
-                        "canTrickyJump",
-                        "h_canCrouchJumpDownGrab"
+                        "canMaxHeightSpringBallJump",
+                        {"or": [
+                          "canWalljump",
+                          "h_canJumpIntoIBJ"
```
osse101, line 8331:
```suggestion
                          "canWalljump",
                          "can3HighMidAirMorph",
                          "h_canJumpIntoIBJ"
```

osse101 (commented):
`canMaxHeightSpringBallJump` should be made a requirement for `canDoubleSpringBallJumpMidAir`.  Just has that oddity of requiring crouch jump when not all variations of double springball jumps use crouch jumps.

And if the key point of the tech is watching for the maximally high point to jump, then the delayed morph springball jumps would fall under this tech too.  Where you jump out horizontally as far as you can before pausing and morphing.  But those also don't use crouch jumps.

kjbranch:
The tech could be renamed and described differently if we need.

Some aspects that it either could or doesn't need to include:
1. The tight timing/delay of the second jump - which could be used in double spring ball jumps and some springwalls. 
2. The crouch jump - we could add it manually to the places that need it, and add it to the description, similar to how a regular spring ball jump mentions crouch jump typically being helpful and often paired with the tech.
3. The tight initial jump and pause - there are quite a few somewhat tight spring ball jumps out of shallow water that it could help with. (East Cac, Blue Boulder, E Sand Hole)

We could have 3 tech: `canSpringBallJump`, `canTrickySpringBallJump`, `canDoubleSpringBallJump`. Where the middle tier includes springwall and all 3 of these aspects listed above.

We could instead simply add a single new tech which is the tighter pause for delayed aspect for the higher (or more horizontal) second jump. We could keep `canSpringwall`, and add this to some of the tighter strats that use it. Then we could use `can3HighMidAirMorph` to get out of shallow water.

kjbranch:
If we do the second option we could add a new helper `h_canMaxHeightSpringBallJump` to as we had before this PR. And we should consider reconsidering the difficulty tier of `can3HighMidAirMorph`.

osse101:
I like keeping `canSpringwall` separate from `canSpringBallJump`, in particular in how `canSpringBallJump` is being used as an easier tech and Springwall is easyish but has situations which make it challenging.  So I could see a modifier `canTrickySpringballJump` being used for both ground and wall springball jumps.

`h_canMaxHeightSpringBallJump` works well as an inline descriptor.

`can3HighMidAirMorph` fits at Very Hard.  East Cac, where the bottom left cac sits, is a `can2HighWallMidAirMorph`, even if it can be done as a pure jump morph.  Really tight jump morphs are RJumps.

And jump morphing out of shallow water feels appropriate to me, maybe slightly hard, but its a little hard for me to tell.
`canBombJumpWaterEscape` should be in Very Hard, even.  But some water pools are deeper than others and are actually really hard to bomb jump out of.

osse101 (approved):
I like not having canTrickyJump as part of maxheightspringballjump even if it kinda fit.  

The trickyspringballjumps in Outer maridia, and crocs lair and whereever else can be added in a separate PR.

PR #1029, opened 2023-08-22
Add uses of canDisableEquipment

kjbranch:


region/maridia/inner-pink.json
```diff
@@ -7451,16 +7489,7 @@
                     "canSuitlessMaridia",
                     "canSpringBallJumpMidAir",
                     {"or": [
-                      "HiJump",
                       "canCarefulJump",
-                      {"and": [
```
osse101, line 7456: Why remove all of these?

osse101 (commented):
Assuming the worst case of having all the movement items does align with how we've been doing logic so far.

It is harder to follow what a strat means now.  Usually it only adds HiJump, and that cause the tech changes.  Gravity/Suitless strats have some weird duplication, but I can get used to it, or it may just be a handful of strats.  The Kraid mouth Speed+Walljump vs TrickyDashJump is a little hard to notice.

Speed and HiJump would break every `canTrickyJump` that doesn't require them, most likely.  Maybe some `canCarefulJump`s too.  Heat rooms are written for minimum itemsets, but more movement items may only reduce heatframe counts.  Speed would have to be removed in lava rooms.

Turning off HiJump for 3high and 2high morphs is not required, but does make them harder to do than I'd like.  RJump with hijump isn't much different, but it is important to turn gravity off if you're underwater.  
Crab Supers and Botwoon Etank can be reached with Gravity HiIump + jump morphs.  But if you can do the jumps then you can be assumed to know about disabling equipment, and this is a knowledge check tech.

Speed reducing jump height probably fits at `Very Hard` level knowledge.

region/maridia/inner-pink.json
kjbranch: the HiJump one was already listed in the strat above, and you can get out of the water with just a spring ball jump .... But now I'm testing it more, I think its a bit harder than should be expected with just a normal spring ball jump, so I'll revert some of these.

kjbranch:
I'm hoping this will be a low enough level tech that we won't need to hammer out all the possible ways to avoid it. i.e. it's still probably possible to jump morph in Botwoon E Tank tunnels with HiJump/Gravity, but I don't think it should be necessary to add. I'll rework the PR and see what you think.

kjbranch:
I think this is moving in the right direction. Let me know what you think.

region/maridia/inner-pink.json
```diff
@@ -7464,52 +7526,75 @@
...
+                  "name": "Suitless",
+                  "notable": false,
                   "requires": [
-                    "Gravity",
-                    {"or":[
-                      "SpaceJump",
-                      "canWalljump",
+                    "canSuitlessMaridia",
+                    {"or": [
+                      {"spikeHits": 1},
```
osse101, line 7561: or `canBombJumpWaterEscape` if you want.

region/maridia/inner-pink.json
```diff
@@ -6801,10 +6809,17 @@
...
           "enemyName": "Cacatac",
           "quantity": 2,
           "homeNodes": [4]
+        },
+        {
+          "id": "e4",
+          "groupName": "East Cactus Alley Top Left Cacatac",
+          "enemyName": "Cacatac",
+          "quantity": 1,
+          "betweenNodes": [5, 6]
```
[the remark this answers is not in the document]
kjbranch: This is the left cac that wasn't being represented before. its above the morph tunnel

region/maridia/inner-pink.json
```diff
@@ -7373,20 +7388,62 @@
...
                   "requires": [
                     "Gravity",
-                    "Morph",
                     {"or": [
                       "SpaceJump",
-                      "canCarefulJump",
-                      "h_canResetFallSpeed",
-                      "canSuitlessMaridia",
                       "h_canHBJ",
+                      "h_canResetFallSpeed",
```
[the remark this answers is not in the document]
kjbranch: That's what I was thinking as well, but then i realized it can be used to make the jump with gravity from the bottom as well. jump and midair morph and press against the left wall, then unmorph after to move left a bit more

osse101 (approved):

PR #1031, opened 2023-08-22
Add uses of canCameraManip

kjbranch:
I didn't touch any of the strats going up the Three Musketeer's Room, because I wasn't sure how they would be done without `canCameraManip`

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -1205,6 +1205,7 @@
                       "mustStayPut": true
                     }},
                     "canCrouchGateClip",
+                    "canCameraManip",
                     "canHorizontalDamageBoost",
                     {"enemyDamage": {
                       "enemy": "Zeb",
--- a/region/brinstar/pink.json
+++ b/region/brinstar/pink.json
@@ -296,6 +296,7 @@
                     "canTrickyUseFrozenEnemies",
                     "canEnemyStuckMoonfall",
                     "canBePatient",
+                    "canCameraManip",
                     {"obstaclesCleared": ["A"]}
                   ],
                   "note": [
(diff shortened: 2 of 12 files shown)


osse101 (commented):
-Plasma Room Shinespark strats as a way to kill pirates would camera manip the bottom pirate.  Instead of that you would use morph or take damage or I think you could jump over its head as it shoots and then stay crouched while it walks away.  And nontrivially walking through an enemy with iframes, I was thinking of calling canHitbox.

-Tourian Piratesin Escape Room 3: the idea is that you use heavy camera manipulation to fight them itemless.  Fighting them that way should have `canBePatient` added as well.

-West Ocean trippers?
-Crateria Kihunters going down would camera manip to know where the crabs are.
-East Ocean Choots when coming from left door, possibly.
-Early Supers crouchGateClip Damage Boost has issues with the bug despawning and you have to move with that in mind.
-Reverse Botwoon fight.  Mostly the Charge+Wave versions I think?
-Draygon has it, but maybe its optional.
-Crossing Colosseum with Ice Mochtroid climbing; I would add it to.
-HiJumpless Lava Dive

kjbranch:
looks like the slow tripper ride is still a bit over 3 minutes with camera manip. so it will be `very patient` either way.

kjbranch:
Thanks for listing all of these additional places this is used! I added most of them and a couple in Crumble Shaft.

Notes:
- Plasma Room: We can add that to the description/use of `canHitbox` in a later PR.
- East Ocean seems to be fine without camera manip.
- Draygon/Reverse Botwoon with Charge+Wave also shouldnt need it.

osse101 (approved):

PR #1032, opened 2023-08-25
Green Hill X-Ray Wiggle - Notable

kjbranch:
This is a bit harder than a normal x-ray turnaround, so it should be in a slightly higher difficulty tier (Map Rando: `VH` -> `Expert`)


osse101 (approved):
I would say `canXRayTurnaround`  is a `Hard` tech that is in `VeryHard` for being not well known.  The strat is fine at VH, and gets more visibility as notable, but could go in Expert for being tedious.

PR #1062, opened 2023-09-11
Wrap-up obstacles clean-up

blkerby:
- Remove deprecated strat property "obstacles" from schema (All occurrences were previously removed).
- Remove deprecated property "requires" on obstacles from the schema and from the last two places where they occurred.
- Clean up the docs to remove references to the now-removed properties.
- Remove the tests corresponding to the now-removed properties (and also to a strat property "clearedObstacles" which doesn't exist). I checked that the remaining tests do cover the strat property "clearsObstacles" and logical requirements "obstaclesCleared" and "obstaclesNotCleared".

### DIFF
--- a/tests/asserts/keywords.py
+++ b/tests/asserts/keywords.py
@@ -268,31 +268,6 @@ def process_strats(src, paramData):
                     messages["counts"]["reds"] += 1
                 else:
                     # valid strat
-                    # if it's got obstacles
-                    if "obstacles" in strat:
-                        for obstacle in strat["obstacles"]:
-                            # make sure the obstacle exists in the room
-                            if obstacle["id"] not in roomData["obstacles"]["ids"]:
-                                msg = f"🔴ERROR: Invalid Obstacle ID:{stratRef}:{obstacle['id']}"
-                                messages["reds"].append(msg)
-                                messages["counts"]["reds"] += 1
-                            # check additionalObstacles too
-                            if "additionalObstacles" in obstacle:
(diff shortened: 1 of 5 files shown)


tests/asserts/keywords.py
```diff
@@ -268,31 +268,6 @@ def process_strats(src, paramData):
...
-                            # check additionalObstacles too
-                            if "additionalObstacles" in obstacle:
-                                for addtlObstacle in obstacle["additionalObstacles"]:
-                                    # make sure it exists in the room
-                                    if addtlObstacle not in roomData["obstacles"]["ids"]:
-                                        msg = f"🔴ERROR: Invalid Additional Obstacle ID:{stratRef}:{obstacle['id']}:{addtlObstacle}"
-                                        messages["reds"].append(msg)
-                                        messages["counts"]["reds"] += 1
-                    # check cleared obstacles too
-                    if "clearedObstacles" in strat:
```
osse101, line 288: clearedObstacles looks ok to keep.

blkerby: "clearedObstacles" isn't an actual strat property though, so this test will never do anything. Possibly this was intended to be "clearsObstacles", but a check later in the file already covers that.

osse101: Maybe I don't remember what clearedObstacles did, there aren't anymore uses of it.  and obstaclesCleared is tested elsewhere.

kjbranch (commented):
The json changes look ok to me. (I didn't look through the python changes.)

osse101 (approved):

PR #1063, opened 2023-09-12
G-Mode: Pink Brinstar part 2

kjbranch:
Adds several g-mode strats in Pink Brinstar. The remaining strats that are not included in this PR:
- Big Pink uses
- Big Pink leave with g-mode from door 4
- Pink Hoppers: Left to Right

### DIFF
--- a/region/brinstar/pink.json
+++ b/region/brinstar/pink.json
@@ -568,6 +568,18 @@
                 }
               ]
             }
+          ],
+          "leaveWithGModeSetup": [
+            {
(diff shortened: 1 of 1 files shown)


osse101 (commented):
Just needs to be able to reach the bugs from the door.  
And for Zero's the technique I've seen is to use XRay buffers and watch for the enemy to reach a certain pixel before entering the transition.

"Big Pink leave with g-mode from door 4" looks to be here?

PR #1065, opened 2023-09-13
Split (excess)shinesparkFrames into separate requirements

blkerby:
Currently `shinesparkFrames` occurs in several places in the schema, namely in `canShineCharge`, `canLeaveCharged`, and `canComeInCharged`. Also `excessShinesparkFrames` occurs in both `canShineCharge` and `canComeInCharged`.

This PR simplifies the schemas by pulling `shinesparkFrames` and `excessShinesparkFrames` into stand-alone logical requirements (similar to `heatFrames`, `metroidFrames`, etc.), and removing them from `canShineCharge`, `canLeaveCharged`, and `canComeInCharged`.

I believe this is cleaner for a few reasons: 
- `canShineCharge` is primarily a tech/skill requirement (a measure of short-charging ability). It is used in many places where there are no shinespark frames required (e.g. if only blue speed is needed). So it makes sense to split off any shinespark frame requirements, because those are different from the tech/skill required to short charge.
- Properties relating to cross-room strats (`canLeaveCharged` and `canComeInCharged`) are some of the trickiest to consume (by randomizers and other tools that may want to use this data). Keeping these properties as lean as possible by stripping out non-essential parts can make them easier to deal with.

A couple important things to note with this change:
- With `shinesparkFrames` and `excessShinesparkFrames` being separate requirements, their total value should be the total shinespark frames required. Previously we used `shinesparkFrames` for the total, and `excessShinesparkFrames` was a part of that total. 
- `shinesparkFrames` and `excessShinesparkFrames` must appear in that order in the requirements, or it will not be correct.

So far the only region file changed is Blue Brinstar. I wanted to get feedback and continuing to do the rest.

### DIFF
--- a/region/crateria/central.json
+++ b/region/crateria/central.json
@@ -1326,7 +1326,7 @@
                     {"or":[
                       {"canComeInCharged": {
                         "fromNode": 1,
-                        "framesRemaining": 180,
+                        "framesRemaining": 180
                       }},
                       {"and": [
                         {"canComeInCharged": {
--- a/region/crateria/central.json
+++ b/region/crateria/central.json
@@ -46,7 +46,6 @@
               },
               "usedTiles": 33,
               "framesRemaining": 0,
-              "shinesparkFrames": 0,
               "openEnd": 1,
               "strats": [
                 {
@@ -73,7 +72,6 @@
               },
               "usedTiles": 33,
               "framesRemaining": 0,
-              "shinesparkFrames": 0,
               "openEnd": 1,
               "strats": [
                 {
@@ -101,7 +99,6 @@
               },
               "usedTiles": 33,
               "framesRemaining": 0,
-              "shinesparkFrames": 0,
               "openEnd": 2,
               "note": [
                 "It's actually possible to use this without going to 1 to open the door first, with a wraparound shot at 3.",
@@ -340,10 +337,12 @@
                     "canMidairShinespark",
                     {"canComeInCharged": {
                       "fromNode": 1,
-                      "framesRemaining": 0,
-                      "shinesparkFrames": 160,
-                      "excessShinesparkFrames": 107
-                     }}
+                      "framesRemaining": 0
+                    }},
+                    {"shinespark": {
+                      "frames": 160,
+                      "excessFrames": 107
+                    }}
                   ],
                   "clearsObstacles": ["A"]
                 },
@@ -459,9 +458,11 @@
                   "requires": [
                     {"canComeInCharged": {
                       "fromNode": 3,
-                      "framesRemaining": 180,
-                      "shinesparkFrames": 125,
-                      "excessShinesparkFrames": 33
+                      "framesRemaining": 180
+                    }},
+                    {"shinespark": {
+                      "frames": 125,
+                      "excessFrames": 33
                     }}
                   ],
                   "clearsObstacles": ["A"]
@@ -494,7 +495,6 @@
                     "canTrickyDashJump",
                     {"canShineCharge": {
                       "usedTiles": 22,
-                      "shinesparkFrames": 0,
                       "openEnd": 2
                     }}
                   ],
@@ -516,7 +516,6 @@
                     "canLateralMidAirMorph",
                     {"canShineCharge": {
                       "usedTiles": 30,
-                      "shinesparkFrames": 0,
                       "openEnd": 2
                     }}
                   ],
@@ -571,9 +570,11 @@
                       "usedTiles": 19,
                       "steepUpTiles": 2,
                       "steepDownTiles": 1,
-                      "shinesparkFrames": 125,
-                      "excessShinesparkFrames": 33,
                       "openEnd": 2
+                    }},
+                    {"shinespark": {
+                      "frames": 125,
+                      "excessFrames": 33
                     }}
                   ],
                   "clearsObstacles": ["A"]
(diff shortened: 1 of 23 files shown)


blkerby:
I updated the PR now to keep `frames` and `excessFrames` grouped together into a single `shinespark` object. This way the numbers won't have to change.

One thing I realized while going through this is that the order of requirements can matter in terms of where `shinespark` requirements appear in relation to other energy requirements (for example `heatFrames`). Usually putting `shinespark` requirements at the end is the more conservative choice (since it increases the possibility of hitting 29 energy and hence not being able to satisfy the shinespark requirement). On the other hand, with `excessFrames` it can also go the other way around.  The ideal would be if we could put the requirements in chronological order of what happens in the strat. There are a number of heated rooms where the heat frames before and after the spark should probably be split apart from each other in order to make the logic fully accurate, but I didn't try to do that in this PR and will create a separate issue for it.

region/lowernorfair/east.json
```diff
@@ -9521,7 +9532,8 @@
                         }},
                         {"heatFrames": 620}
                       ]}
-                    ]}
+                    ]},
+                    {"shinespark": {"frames": 35}}
```
kjbranch, line 9536: I don't think this belongs here.
```suggestion
                    ]}
```

region/maridia/inner-pink.json
```diff
@@ -3303,7 +3321,6 @@
                     ]},
                     {"canShineCharge": {
```
kjbranch, line 3322: I know this is out of scope, but can this be shinecharge be deleted? If not, should it be moved into the `canStutterWaterShineCharge` strat?

region/maridia/inner-pink.json
```diff
@@ -3462,9 +3478,9 @@
...
```
kjbranch, line 3479: same here

kjbranch (commented):
This review is turning my brain to mush, so I'm going to take a break. Here's what I've found so far.

region/norfair/east.json
```diff
@@ -2799,10 +2828,12 @@
                         {"canComeInCharged":{
                           "fromNode": 3,
                           "inRoomPath": [3, 9],
-                          "framesRemaining": 35,
-                          "shinesparkFrames": 35,
-                          "excessShinesparkFrames": 0
-                         }}
+                          "framesRemaining": 3
```
kjbranch, line 2831:
```suggestion
                          "framesRemaining": 35
```

schema/m3-requirements.schema.json
```diff
@@ -609,28 +593,40 @@
...
-              "shinesparkFrames": {
-                "$id": "#/definitions/logicalRequirement/items/properties/canShineCharge/properties/shinesparkFrames",
+              }
+            }
+          },
+          "shinespark": {
+            "$id": "#/definitions/logicalRequirement/items/properties/shinespark",
+            "type": "object",
+            "title": "Shinespark",
+            "description": "Fulfilled by entering the room in a specified node, and avoid visiting barred nodes and destroying barred obstacles.",
```
osse101, line 603: \#resetRoom

schema/m3-requirements.schema.json
```diff
@@ -609,28 +593,40 @@
...
-                "minimum": 0,
+                "minimum": 1,
                 "title": "Shinespark Frames",
-                "description": "The duration (in frames) of the shinespark that must be executed. This defines the energy cost of the shinespark."
+                "description": "The duration (in frames) of the shinespark that must be executed. This defines the energy cost of the shinespark assuming it is not interrupted by running down to 29 energy."
               },
-              "excessShinesparkFrames": {
-                "$id": "#/definitions/logicalRequirement/items/properties/canShineCharge/properties/excessShinesparkFrames",
+              "excessFrames": {
+                "$id": "#/definitions/logicalRequirement/items/properties/shinespark/properties/excessShinesparkFrames",
```
osse101, line 617:
```suggestion
                "$id": "#/definitions/logicalRequirement/items/properties/shinespark/properties/excessFrames",
```

schema/m3-requirements.schema.json
```diff
@@ -609,28 +593,40 @@
...
+            "type": "object",
+            "title": "Shinespark",
+            "description": "Fulfilled by entering the room in a specified node, and avoid visiting barred nodes and destroying barred obstacles.",
+            "required": [
+              "frames"
+            ],
+            "additionalProperties": false,
+            "properties": {
+              "frames": {
+                "$id": "#/definitions/logicalRequirement/items/properties/shinespark/properties/shinesparkFrames",
```
osse101, line 610:
```suggestion
                "$id": "#/definitions/logicalRequirement/items/properties/shinespark/properties/frames",
```

kjbranch (commented):
Just 1 more thing in the rest of my review.

region/maridia/inner-pink.json
osse101: The `canShineCharge` doesn't actually mean anything logically.  A shinecharge is still be performed with either `canWaterShineCharge` or `canStutterWaterShineCharge`.  And those techs are together with the needed requirements for a shineCharge to work, so we can say that there is a shineCharge occurring.  And that makes the `shinespark` in the other strats not out of place, since there is still a shineCharge here.

This `canShineCharge` should be able to be deleted.

region/crateria/central.json
```diff
@@ -46,7 +46,6 @@
               },
```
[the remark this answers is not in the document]
blkerby: Yeah we should be able to improve this soon when we move cross-room strats into links

osse101 (approved):

PR #1070, opened 2023-09-14
Split Indiana Jones Grapple

kjbranch:
This is probably worth splitting, as the two directions are fairly different difficulties. In Map Rando, I think we will want to increase the difficulty of the reverse direction from `VH` -> `Expert`

### DIFF
--- a/region/norfair/crocomire.json
+++ b/region/norfair/crocomire.json
@@ -1450,12 +1450,6 @@
           "homeNodes": [ 2 ]
         }
       ],
-      "reusableRoomwideNotable": [
-        {
-          "name": "PCJR Indiana Jones Grapple",
-          "note": "Involves grappling off several Rippers, where falling may lead to a soft lock."
-        }
-      ],
       "links": [
         {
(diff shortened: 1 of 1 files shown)


osse101:
I think the reverse is easier.  But the reverse should be less practiced making it overall harder.

With reverse you can jump out to the first ripper when it is far left in its cycle and then cross all of the rippers with small movements.  The forward strat has a mostly blind jump to the 4th ripper and it can be anywhere in its cycle since it is hard to see.

I don't think you could blindly, no practice do this strat so it should be in Expert anyway.  Getting used the the heights each one is at is too important.  But maybe you could drop onto the next ripper in line going reverse, and make VH work.  
Going forward, timing the jump to the final ripper without a setup and missing being a softlock, would put it in extreme to me.  A setup would be making the same movements from room load, and you can sortof know where the ripper will be with that.  Or flinging the third jump is usually good enough to hover in place and catch him.  And thats good enough for expert.

But I've practiced the room too much to properly rank the difficulty.

PR #1071, opened 2023-09-15
G-Mode: Big Pink Top Third

kjbranch:
This includes strats to get between the top three doors, to get to the new G-Mode Morph junction, and to get from the junction to the top left door. 

Still need the rest of the room, as well as possibly Spring Ball + Power Bomb strats, which would require PLMs to be overloaded first, to prevent the PB from hitting the block.

The hardest part was figuring out where to put the junction :)

### DIFF
--- a/region/brinstar/pink.json
+++ b/region/brinstar/pink.json
@@ -2129,7 +2129,7 @@
                   "note": [
                     "Carefully dodge the Sidehoppers, or kill them with a Power Bomb. Be careful not to hit the bomb block with the Power Bomb or it will remain solid.",
                     "With IBJ alone, it is easiest to IBJ from the bug pipe below to the ceiling, luring the Sidehoppers to the left.",
-                    "At the top, over PLMs using the camera scroll block just in front of the block, then roll through it."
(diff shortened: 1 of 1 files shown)


region/brinstar/pink.json
```diff
@@ -1967,6 +2096,45 @@
...
+                        "hits": 1,
+                        "type": "contact"
+                      }},
+                      {"ammo": { "type": "PowerBomb", "count": 1 }}
+                    ]}
+                  ],
+                  "note": [
+                    "Carefully dodge the Sidehoppers, or kill them with a Power Bomb. Be careful not to hit the bomb block with the Power Bomb or it will remain solid.",
+                    "With IBJ alone, it is easiest to IBJ from the bug pipe below to the ceiling, luring the Sidehoppers to the left.",
+                    "At the top, over PLMs using the camera scroll block just in front of the block, then roll through it."
```
osse101, line 2132:
```suggestion
                    "At the top, overload PLMs using the camera scroll block just in front of the block, then roll through it."
```

osse101 (approved):
Harder to dodge than your average hoppers, but I found some tricks that keep it reasonable.

PR #1077, opened 2023-09-26
G-Mode Pink Hopper Room

kjbranch:


region/brinstar/pink.json
```diff
@@ -3181,6 +3186,37 @@
                     "Its acceleration will cause it to clip into the blue gate off-screen."
                   ],
                   "devNote": "The reverse gate glitch pretty much requires killing the enemies first, so it's required here."
+                },
+                {
+                  "name": "G-Mode Indirect to Despawn Gate",
+                  "notable": false,
+                  "requires": [
+                    {"comeInWithGMode": {
```
osse101, line 3194: A strat for avoiding hoppers should be added.  It would help save health for this GMode strat.
`canPrepareForNextRoom` with consecutivewalljump, spacejump, precise grapple, or speed+runway.

Also this strat needs order priority.  And a 1->1 clear A would happen after since most of those take damage.

region/brinstar/pink.json
```diff
@@ -3194,15 +3230,54 @@
...
-                  "note": "The reverse gate glitch may be executed earlier from node 1 -> 1."
+                  "devNote": "The reverse gate glitch may be executed earlier from node 1 -> 1."
+                },
+                {
+                  "name": "Speedy Jump Onto Grapple Block",
+                  "notable": false,
+                  "requires": [
+                    "HiJump",
+                    "canTrickyDashJump",
+                    "canDisableEquipment",
```
osse101, line 3256:
```suggestion
                    "canTrickyJump",
                    "SpeedBooster",
```

osse101: This feels easier than a trickyDashJump to me.  And the runway length sets itself up for the jump so you don't need to think about pixel positions.
Jumping from the grapple block can be made without turning anything off, and `canTrickyJump` is already required.

kjbranch: When you suggested this in the past, I didn't realize you meant to still have a `canTrickyJump` on it. I think this will work fine.

region/brinstar/pink.json
kjbranch: This strat will use the `Base` 1->2 afterwards. Or it will do another 1->1 that clears the enemies (`A`) after and then get up with speed.

region/brinstar/pink.json
```diff
@@ -3181,6 +3186,37 @@
...
+                    {"comeInWithGMode": {
+                      "fromNodes": [1],
+                      "mode": "indirect",
+                      "artificialMorph": false
+                    }}
+                  ],
+                  "clearsObstacles": ["B","C"]
+                },
+                {
+                  "name": "G-Mode Morph Power Bomb Kill",
```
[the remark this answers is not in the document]
kjbranch: I removed it and put it as a note. it's now only used while in `indirect`

kjbranch:
I can't get grapple working in g-mode .... if it works it will need to be it's own notable strat.

region/brinstar/pink.json
kjbranch: Ah, I see now. That does seem like a large problem. I'll work on it.

kjbranch: Talking to Maddo, this sounds like it won't actually be a problem. In order to use a `comeInWithGMode` it will require all of the obstacles to not be cleared.

blkerby: Right, here's the relevant code in case anyone is curious: https://github.com/blkerby/MapRandomizer/blob/857b168c602923e586a4f7996f9e77994ab48b76/rust/src/game_data.rs#L1320. If there's an obstacle already cleared, then there's no possibility of satisfying a comeInWithGMode in that state.

Once we finish reworking the cross-room schema, it will only allow entrance conditions (e.g. comeInWithGMode) to be used in conjunction with a strat that enters the room, which should make this even simpler to think about.

PR #1078, opened 2023-09-26
G-Mode Through the Spikeway

kjbranch:
Mostly just saves a thorn hit

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -3189,27 +3189,6 @@
                   "name": "Base",
                   "notable": false,
                   "requires": []
-                },
-                {
-                  "name": "G-Mode Morph Overload PLMs with the Item Blind Movement Through Spikeway",
-                  "notable": false,
-                  "requires": [
-                    {"comeInWithGMode": {
-                      "fromNodes": [1],
-                      "mode": "direct",
-                      "artificialMorph": true
-                    }},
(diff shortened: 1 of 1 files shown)


region/brinstar/green.json
```diff
@@ -3184,6 +3189,27 @@
...
+                    {"itemNotCollectedAtNode": 5},
+                    "canRiskPermanentLossOfAccess",
+                    "canOffScreenMovement"
+                  ],
+                  "clearsObstacles": ["B"],
+                  "note": [
+                    "Roll through the item to overload PLMs, then fall in the pit and through the crumble blocks.",
+                    "Carefully fall to not touch the spikes on the right, then roll through the spikeway.",
+                    "After the camera stops scrolling, move another 2-3 tiles then exit G-Mode.",
+                    "Make small jumps to reset the camera then quickly kill the Beetom as it jumps at Samus."
```
osse101, line 3211: I don't see the camera resetting to Samus so maybe you need to touch the scroll block before overloading PLMs.
Blindly going through the spiky platforming section is hard enough that I wouldn't include it as a strat.

kjbranch: Just when I think I know how G-mode works .... It is possible to almost overload PLMs with the item then fall into the pit and finish it with the camera scroll blocks, but let's not include that either.

Issue #1059, opened 2023-09-07
Cross-Room Strat Cleanup


kjbranch:
Due to the increasing number of types of cross-room strats, and the complex ways in which they need to be represented in order to describe in every instance, cross-room strats have become difficult to understand and to implement. 

Cross-room strats are represented in a very different way to in-room strats and to each other, making them difficult to understand by people reading the logic. 
-	`canComeInCharged`: Is a requirement on a strat, which requires either: `canLeaveCharged` on a node in the previous room, a long enough runway at the node in the current room, the adjacent room, or combined between the two rooms.
-	R- and G-mode strats are setup and exit immobile in the nodes, while uses are implemented like normal strats, in the links section.
-	The remaining cross-room strats are only modeled in the second room, while requiring runways in the first room. Adjacent runways have their own sets of problems, such as assuming there are no enemies and making their heat frames hard to interpret.  

The logic implicitly assumes that the system using it can maintain a history of previous nodes traversed through the map, as well as Samus’ states upon room entry. We have been working towards eliminating these, because maintaining cross-room states and previous nodes in the graph traversal of this data is not really possible with its size, and if it were, it would be very computationally intensive. It’s also not very clear how to use the remote setups or remote uses of cross-room shinespark strats. Because of this it is implemented in Map Rando that you have to start the cross-room strat in the destination room. This, and other problems have caused many limitations, such as:
-	Heated room strats require starting from the door, so leaving with speed requires moving away from the door and running back towards it. The heat frames could be reduced if the strat could be started before getting to the door. This is what `initiateRemotely` is attempting to fix.

All of the examples in Crateria that I can think of:
-	Landing Site: Using a shinespark to break the blocks to the Gauntlet Entrance, and then sparking through the door.
-	Parlor: Using g-mode artificial morph to crystal flash, damage back down, then break the blocks to the top left door to bring in a Geemer to setup g-mode going left.
-	Parlor: Using speed to break the blocks and leave through the left door with speed.
-	Gauntlet Entrance: Using Speed or artificial morph to break the blocks, releasing the Waver, to then use for a G-mode setup on the same or opposite door.
-	Gauntlet Entrance: Use speed to break the central blocks, making a runway that can be used to spark out the opposite door.
-	Gauntlet E-Tank: Come in with speed or g-mode, break the left blocks to make a larger runway to go out with after farming.
-	Climb: Break the bomb blocks then use the full runway and the adjacent room’s by running out of the door.
[...]

PR #1079, opened 2023-09-28
Cross-room strat restructure (part 1)

blkerby:
This PR begins work on issue #1059, implementing the necessary schema changes to support the new way of handling cross-room strats involving runways. This includes:
- Deprecating the node property "runways".
- Deprecating logical requirements "canComeInCharged" and "adjacentRunway".
- Introducing new strat property "exitCondition" with three possible keys:
  - leaveWithRunway
  - leaveCharged
  - leaveWithSpark
- Introducing new strat property "entranceCondition" with eight possible keys:
  - comeInRunning
  - comeInJumping
  - comeInCharging
  - comeInCharged
  - comeInWithSpark
  - comeInStutterCharging
  - comeInWithBombBoost
  - comeInWithDoorStuckSetup
- Updating the Blue Brinstar region file as an example of how to apply the new schema.

This new approach aims to provide a more general, extensible way of expressing cross-room strats. Strats that exit a room in a special way (e.g. with a shinecharge) will now be expressed as ordinary strats inside of a room's "links" (instead of as properties on the node). An "exitCondition" is specified on the strat to describe the special way in which it leaves the room. A strat with an exit condition is only eligible to be used only in conjunction with a strat in the next room having a compatible "entranceCondition". There are several benefits to this approach:
- Strats that exit the room in a special way are now in the same place as all other strats. This makes it simpler for randomizers, logic viewers, and other tools to consume and present this data (e.g. with this change, strats that leave the room charged or that exit with a runway can now automatically show up in the logic pages of Map Rando alongside other strats).
- This structure makes it natural for strats with an "exitCondition" to start from any node in the room (i.e. possibly a node other than the door that the strat exits through). This makes it possible to encode many strats that up to now have not been expressible.
- By introducing finer-grained entrance conditions that specifically describe the action that occurs through the transition, it reduces the amount of boilerplate required on such strats. This also eliminates the need to "override" requirements in the other room's runway, which was confusing and error-prone. Finally, by more explicitly expressing the specific action happening (as opposed to just expressing a geometric constraint on a runway), this makes it possible for the corresponding heat frames to be calculated automatically and with greater accuracy.
- By organizing how cross-room strats are encoded (e.g. putting them in just one place under the "entranceCondition"/"exitCondition" strat properties, and making the data inside there as lean as possible), it makes the task easier for randomizers or other tools to implement them. Based on my experience with Map Rando, dealing with cross-room strats were the most difficult part of consuming the sm-json-data. Making that experience more straightforward can help a lot in the future for other consumers.

For details on how the new schema works, see the docs which are updated in this PR.

A couple other things worth noting in this PR:
- We've had a lot of confusion about runway lengths and whether or not they should include the door shell tile. This stems from the fact that when you run out of a room, the door shell tile forms part of the runway, but when running into the room it does not. In the new schema of this PR, this problem goes away because outgoing runways are completely separated from incoming ones: runways are no longer pieces of geometry attached to nodes; instead they become strats, and strats that exit the room with a runway are different strats from ones that enter with one. My opinion is that the concept of shared runway data (used for both incoming/outgoing strats) is not adding a lot of value and is not worth the trouble it brings. Very often incoming strats needed to override the runway with "unusableTiles"; it seems simpler to just specify the usable incoming runway geometry for these strats directly.
- I noticed that we've been putting in 200 heat frames for all the heated X-ray climb setups. I checked this out and it seems not hard to do in 100 heat frames if you use X-Ray to buffer the positioning. So I reduced that accordingly. Fortunately this will now be specified in one place (in the docs) instead of needing to be copied across all the strats, so it could easily be adjusted later.

This PR does not address the remaining cross-room strats, involving jumpways and G-mode. Those can be addressed in a separate PR afterwards and will involve similar schema changes (and those will be less complex than the cases handled here).


strats.md
```diff
@@ -45,3 +47,411 @@ Execution of a strat may have an effect of clearing one or more [obstacles](regi
...
+{
+  "name": "Leave With Spark",
+  "notable": false,
+  "requires": [
+    {"canShinecharge": {
+      "usedTiles": 20,
+      "openEnd": 0
+    }},
+    {"shinespark": {
+      "frames": 200
```
kjbranch, line 157: I think it's good to encourage the use of excessFrames
```suggestion
      "frames": 200,
      "excessFrames": 50
```

strats.md
```diff
@@ -45,3 +47,411 @@ Execution of a strat may have an effect of clearing one or more [obstacles](regi
...
+- _comeInWithSpark_: This indicates that Samus must shinespark into the room.
+- _comeInWithBombBoost_: This indicates that Samus must come into the room with a horizontal bomb boost.
+- _comeInStutterCharging_: This indicates that Samus must run into the room with a stutter immediately before the transition.
+- _comeInWithDoorStuckSetup_: This indicates that Samus must enter the room in a way that allows getting stuck in the door as it closes.
+
+Each of these properties is described in more detail below.
+
+### Come In Running
+
+An `comeInRunning` entrance condition represents the need for Samus to be able to run into the room with speed in a certain range. It has the following properties:
```
kjbranch, line 183:
```suggestion
A `comeInRunning` entrance condition represents the need for Samus to be able to run into the room with speed in a certain range. It has the following properties:
```

strats.md
```diff
@@ -45,3 +47,411 @@ Execution of a strat may have an effect of clearing one or more [obstacles](regi
...
+* _speedBooster_: If true, then Speed Booster must be used while running into the room. If false, then Speed Booster must not be used. If "any", then Speed Booster may or may not be used.
```
kjbranch, line 184: if false, should that imply `canDisableEquipment`?

strats.md
```diff
@@ -45,3 +47,411 @@ Execution of a strat may have an effect of clearing one or more [obstacles](regi
...
+* _minTiles_: The minimum horizontal speed that will satisfy the condition, measured in effective runway tiles with dash held.
+* _maxTiles_: The maximum horizontal speed that will satisfy the condition, measured in effective runway tiles with dash held.
+
+A `comeInRunning` condition can be satisfied only by a matching strat on the other side of the door with `leaveWithRunway` exit condition: a match is valid if all of the following are true:
+- The effective runway length of the `leaveWithRunway` is at least as long as the `minTiles` in the `comeInRunning` condition.
+
+Where applicable, a `comeInRunning` condition also includes implicit requirements for actions to be performed in the previous room, which are effectively prepended to the start of the strat's `requires` (or equivalently but more properly, onto the end of the `requires` of the `leaveWithRunway` strat in the other room):
+- If `speedBooster` is true, then there is an implicit "SpeedBooster" item requirement.
+- If the previous room is heated, then `heatFrames` are required based on the time needed to perform the run.
```
kjbranch, line 193: Might be worth clarifying - the heat frames are implicit, and gravity below is not?

strats.md
```diff
@@ -45,3 +47,411 @@ Execution of a strat may have an effect of clearing one or more [obstacles](regi
...
+| Action                                                                  | Heat frames             |
+| ----------------------------------------------------------------------  | ----------------------- |
+| Run from a standstill in one direction, without Speedbooster, 0-6 tiles | `9 + 4 * tiles`         |
+| Run from a standstill in one direction, without Speedbooster, 7+ tiles  | `13 + 64 / 19 * tiles`  |
+| Run from a standstill in one direction, with Speedbooster, 0-6 tiles    | `9 + 4 * tiles`         |
+| Run from a standstill in one direction, with Speedbooster, 7-16 tiles   | `15 + 3 * tiles`        |
+| Run from a standstill in one direction, with Speedbooster, 17-42 tiles  | `32 + 2 * tiles`        |
+| Run from a standstill in one direction, with Speedbooster, 43+ tiles    | `47 + 64 / 39 * tiles`  |
+| Turn around                                                             | `8`                     |
+| Position using moonwalk (and shoot open the door)                       | `12`                    |
```
kjbranch, line 207: what if the runway is longer than the max tiles? should these still be added in as if Samus were not running?

kjbranch (commented):
I'm really excited for this to be changed! Here is a partial review. I'll continue on it later.

osse101 (commented):
Reducing xray setup frames is fine if there is heat frame scaling still.  If it takes more than one attempt, it was still unfun at 200 frames.  And some players can just be slow at setting it up.  

I can see this being expanded to include more cross room state conditions.  
-comeInWithSpark often would like to spark along the top of the door

There may be situations where it would help to have 2 conditions, Screw attack room mid->top  would like to comeInCharged and also jumping but the jump may not require an entry condition.  Running into everest top-left door with comeInCharged might need the run speed to spark in the correct spot to hit the middle grapple blocks.  

I haven't looked at how comeInCharged:180 or unusableFrames going into the previous room would work yet.  But I'm thinking about fast mockballs.

Assuming water rooms require Gravity seems right except for very specific cases.  But 3 room shinecharges/shinesparks wouldn't need gravity.

`comeInCharged` sounds like using charge beam, preparing for a pseudoscrew or buffered charge shot.  But by context it means shineCharge.

strats.md
osse101: `excessFrames` would mean the comeInWithSpark would fail the health check on the following strat, and that is fine for this strat doubling as a way to reach the door.

But there probably needs to be a field for preopening doors.  Or these can all link to self on the door to show it was opened first.

region/brinstar/blue.json
```diff
@@ -413,6 +374,19 @@
...
+                  "notable": false,
+                  "entranceCondition": {
+                    "comeInWithSpark": {}
+                  },
+                  "requires": [
+                    {"shinespark": {
+                      "frames": 40,
+                      "excessFrames": 9
+                    }}
+                  ]
```
osse101, line 388: This kills all of the hoppers unless you get a normalized RNG desync, and enter at the top of the door.

strats.md
```diff
@@ -45,3 +47,411 @@ Execution of a strat may have an effect of clearing one or more [obstacles](regi
...
+
+A `comeInWithDoorStuckSetup` condition must match with a `leaveWithRunway` condition on the other side of the door. A match comes with the following implicit requirements for actions to be performed in the previous room:
+- If the previous room is heated, a requirement of `{"heatFrames": 100}` is included, for positioning and executing the door stuck setup. Note that if X-Ray Scope is available (and it always should be for a strat with this condition, since the only known application of door stuck setups is for X-Ray climbing), then minimizing heat damage is made easier by using X-Ray to buffer the positioning near the door and to turn around in place.
+- If the previous room is to the left, then a tech requirement `canStationarySpinJump` is included.
+- If the previous room is to the right, then a tech requirement `canRightSideDoorStuck` is included.
+- If the previous door environment is water and is to the right, then a requirement `{"or": ["Gravity", "canRightSideDoorStuckFromWater"]}` is included.
```
osse101, line 443: `canShinesparkDeepstuck` doesn't have that many combinations of rooms that allow it to work, without spikesuit/bluesuit.
`GMode` deepstuck would be setup in room without this entrance condition.

strats.md
```diff
@@ -45,3 +47,411 @@ Execution of a strat may have an effect of clearing one or more [obstacles](regi
...
+
+When a `leaveWithRunway` condition is used to charge a shinespark, it implicitly includes a `canShinecharge` requirement with `usedTiles` based on the runway length (and also based on a continuation of the runway in the destination room where applicable). Again, this is described in more detail under the applicable entrance conditions below.
+
+When a `leaveWithRunway` conditions occurs on a door in a water environment, it implicitly includes a `Gravity` requirement.
```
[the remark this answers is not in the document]
blkerby: Yeah East Pants Room could maybe use a rework; I think it would work better if it were just treated as a separate room with its own nodes, instead of how it is now where there's a fake "transition" from Pants Room to East Pants Room when going through Pants Room. So then the door environment on the right door of Pants Room and East Pants Room could be different from each other (one of them water and one of them air).

strats.md
blkerby: I don't think we want excessFrames on a leaveWithSpark strat. A cross-room strat needs to satisfy the requirements on both sides of the transition or it is not applicable; and if it runs down to 29 energy in the first room then it can't make it to the next room, so the combined strat doesn't apply. So stopping in the middle at the door would not be an option; that would have to be a separate strat.

strats.md
blkerby: I guess so? That could also go as a requirement in the strat though, so we wouldn't have to make it implicit. I could go either way. Might be worth checking first if there are any examples

strats.md
blkerby: These are all implicit requirements. Clarified the Gravity one though now, to make it clear that it depends on the entrance condition in the next room (since Gravity isn't always needed)

strats.md
blkerby: In heated rooms at least, we will need to provide the accurate runway length no matter how long.

schema/m3-region.schema.json
```diff
@@ -67,6 +66,299 @@
...
+              "type": "object",
+              "title": "Come in Charged",
+              "description": "Fulfilled by coming in through this door with a shinespark charge that has at least the given number of frames remaining.",
+              "required": [
+                "framesRequired"
+              ],
+              "additionalProperties": false,
+              "properties": {
+                "framesRemaining": {
+                  "$id": "#/definitions/strat/properties/entranceCondition/properties/comeInCharged/properties/framesRequired",
```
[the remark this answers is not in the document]
blkerby: The typo here was on the "framesRemaining". The intent was that the exit condition provides a "framesRemaining" whereas the entrance condition has a "framesRequired".

strats.md
kjbranch: Oh, I was thinking this was the frames post transition. that makes sense

strats.md
```diff
@@ -45,3 +47,411 @@ Execution of a strat may have an effect of clearing one or more [obstacles](regi
...
+
+      $$\begin{align*}
+      T &= \text{total time} = 85 \text{ frames} \\
+      C &= \text{initial speed} = 0.125 \text{ tiles per frame} \\
+      L &= \text{combined effective runway length (in tiles)} \\
+      L_o &= \text{effective runway length in other room} \\
+      a &= \text{acceleration} = 2(L - CT) / T^2 \\
+      T_o &= \text{time in other room} = \frac{\sqrt{C^2 + 2ax} - C}{a} \\
+      T_c &= \text{time in current room} = T - T_o \\
+      \end{align*}$$
```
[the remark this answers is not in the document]
blkerby: Ah, good catch. It looked fine in my code editor's preview, not sure why GitHub is having a problem with it. Anyway, best I could figure out now is to put each equation in a separate display; the alignment gets messed up but it's better than not getting formatted at all.

blkerby:
> Reducing xray setup frames is fine if there is heat frame scaling still. If it takes more than one attempt, it was still unfun at 200 frames. And some players can just be slow at setting it up.

Yeah I think this could be a good time to add a randomizer lenience setting for number of attempts on door stuck setups, similar to what we did with gate glitch leniency. Maybe 3 attempts on Expert, 2 attempts on Extreme, 1 on Insane?

> There may be situations where it would help to have 2 conditions, Screw attack room mid->top would like to comeInCharged and also jumping but the jump may not require an entry condition. Running into everest top-left door with comeInCharged might need the run speed to spark in the correct spot to hit the middle grapple blocks.

This gets tricky fast though. Would it mean that on every "leaveCharged" strat we need to provide a corresponding "leaveWithRunway" to describe how much run speed you can leave with? I mean if we want to go down that road, I suppose it's a possibility. My initial thinking was to restrict to just having a single condition, to try to keep things more manageable and easier for the randomizer (and other tools) to process. But it would be an option to relax the "maxProperties: 1" constraints on entrance/exit conditions. An alternative which I think I might prefer a little more would be to add another property inside of the "leaveCharged" (and "comeInCharged") to describe run speed, if we decide that's something we need.

> I haven't looked at how comeInCharged:180 or unusableFrames going into the previous room would work yet. But I'm thinking about fast mockballs.

Yeah it seems likely we'll need another set of entrance/exit conditions for that but I didn't tackle it yet since I wasn't sure what it should look like. 

> Assuming water rooms require Gravity seems right except for very specific cases. But 3 room shinecharges/shinesparks wouldn't need gravity.

Yeah we'll probably need some new exit condition (e.g. "leaveCharging"?) to fully support 3-room shinecharges. Though with the schema in this PR we could already have a middle room with a "comeInCharging" entrance condition and a "leaveCharged" exit condition, and this could work with the middle room being water and no Gravity requirement.

> `comeInCharged` sounds like using charge beam, preparing for a pseudoscrew or buffered charge shot. But by context it means shineCharge.

Yeah, any suggestions for a better alternative? I didn't think of anything better so was just following the terminology of the existing schema ("canComeInCharged"). I guess maybe "leaveWithShinecharge"/"comeInWithShinecharge" is an option?

strats.md
```diff
@@ -8,10 +8,12 @@ Strats are usually presented within an array of strats, where all individual str
...
   * _notable:_ Indicates whether the strat is notable.
+  * _reusableRoomwideNotable:_ The name of the reusable roomwide notable strat. This must share an identical name with an entry in the `reusableRoomwideNotable` array and the other strats that are connected to this one. This is only applicable for strats where `notable` is `true`.
   * _requires:_ The [logical requirements](logicalRequirements.md) that must be fulfilled to execute that strat.
   * _clearsObstacles:_ An array containing the ID of obstacles that will be cleared by executing this strat (if they are not already cleared).
-  * _reusableRoomwideNotable:_ The name of the reusable roomwide notable strat. This must share an identical name with an entry in the `reusableRoomwideNotable` array and the other strats that are connected to this one. This is only applicable for strats where `notable` is `true`.
-
+  * _exitCondition_: Indicates that this strat leaves through the door transition in a special way that combines with a strat in the next room. 
+  * _entranceCondition_: Indicates that this strat requires entering through the door transition in a special way, combining with a strat in the previous room.
```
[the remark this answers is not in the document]
blkerby: Reordered it now so that `requires` is sandwiched between `entranceCondition` and `exitCondition`. That's the order that makes most sense to me to write these properties in, to keep things in chronological order with how they're executed.

region/brinstar/blue.json
```diff
@@ -244,19 +170,54 @@
...
             {
               "id": 5,
               "strats": [
                 {
                   "name": "Room Entry Speedball",
                   "notable": false,
+                  "entranceCondition": {
+                    "comeInCharging": {
+                      "length": 2,
+                      "openEnd": 1
```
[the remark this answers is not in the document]
blkerby: The full runway here is length 3 with an open end. So the length of 2 here already has 1 subtracted, corresponding to the "unusableTiles: 1". I think this is okay, since it looks like you wouldn't hit the hopper until almost at the end of the last tile (which isn't being counted as part of the runway). If it sounds good, I could add a devNote to make it more clear about the 1 being subtracted and the reason why?

region/brinstar/blue.json
blkerby: Thanks, forgot to copy over that `"clearsObstacles": ["C"]`. Did that now.

strats.md
blkerby: I assume we'll probably use a `comeInCharged` entrance condition for shinespark deep stuck strats. We could make something more specific but I'm not sure if it's needed?

osse101: `comeInCharged` would be a good way to handle it.

strats.md
osse101: It would be better to need `canDisableEquipment` tech if speed needs to be off since that is the most confusing equipment item.  Those situations are harder to find though, probably short runways.  
I turned off speed for the tube gravity jump yesterday, when connected to a save room.  That would also have a crossroomJumpIntoWater tech on it.  Maybe the boulder room entry jump, or maybe not, for an example without a harder tech with it.
I expect most of the short runway jumps to need HiJump too and then it doesn't matter.

region/brinstar/blue.json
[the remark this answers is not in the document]
osse101: The openEnd should be 0, from how I see it.  Since you cannot use it as part of shinecharge.

strats.md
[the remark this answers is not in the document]
osse101: Sounds good.  Could you reorder `_entranceCondition_` here in strat structure too.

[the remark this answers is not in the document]
blkerby: Ah, right. Done.

region/brinstar/blue.json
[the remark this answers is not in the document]
blkerby: Sure, changed it to 0 now. To be clear, this has nothing to do with the open end at the end of the platform, since with length 2 that last tile already wasn't considered part of the runway. Changing it to 0 just makes the usable runway even shorter (by 9 pixels), indicating that Samus needs to get blue before moving into the space above that last platform tile at all.

region/brinstar/blue.json
```diff
@@ -2257,6 +2226,66 @@
...
+                  "requires": [],
+                  "exitCondition": {
+                    "leaveWithRunway": {
+                      "length": 1,
+                      "openEnd": 1
+                    }
+                  }
+                },      
+                {
+                  "name": "Leave Charged (with Gravity)",
```
kjbranch, line 2245: I think this and the next strat should be moved from `1->2`

region/brinstar/blue.json
```diff
@@ -1422,6 +1384,50 @@
...
+                  }
+                },
+                {
+                  "name": "Leave Charged - Power Bomb Blocks Broken",
+                  "notable": false,
+                  "requires": [
+                    {"obstaclesCleared": ["C"]},
+                    {"canShineCharge": {
+                      "usedTiles": 33,
+                      "openEnd": 0
```
kjbranch, line 1421: With the new ways to calculate runway lengths, 33 tiles and 0 open ends won't be in logic if the player selects 33 tiles. We should eventually go through and increase the number of tiles a bit, maybe to 35 or 36, because 33 can be a bit tight for beginners.
```suggestion
                      "openEnd": 1
```

region/brinstar/blue.json
osse101: ☝️

region/brinstar/blue.json
osse101: I can see that; being able to stop when you turn blue is hard at any point before you can 3 tap.  Knowing when to stop should come with a visual indicator that is the end of the runway.  Converting to spinjump if you run into a wall would help to have more runway leeway... and properly storing the shinespark is difficult early on.

`usedTiles` in schema should be updated then.   In this case you could fit an openend on both sides of a 33 tile runway but it is clearer if the usedTile count is increased instead.

strats.md
blkerby: Ok I added to the docs that there is an implicit `canDisableEquipment` when `speedBooster` is false.

region/brinstar/blue.json
blkerby: Done.

kjbranch (approved):
I'm satisfied with this. I'll summarize the concerns that I see that have been raised but not resolved.

`startingDownTiles` can be confusing: how do we count steep vs gentle tiles and early but not starting tiles.
- This proposal is intended to be a migration and isn't intended to fix everything. Once we determine how we want this to work, this schema structure can relatively easily be adjusted to accommodate.

How to deal with `leaveWithRunway` implicitly requiring `Gravity` with East Pants Room.
- This can probably be resolved by edits to the room. Otherwise, we can figure something out when we get there.

Situations with coming in with 2 entry conditions.
- Without this, I think we do actually lose something here relative to the current schema, because we can no longer use adjacent runways. I think we will need to figure something out.

Entrance mockballs.
- These are going to be a bit complex, but I think we can add something similar once we get there.

Rename `comeInCharged`.
- I'm fine either way. It could be `comeInShineCharged`.

osse101:
> Entrance mockballs
Some of the existing mockballs can't be represented I don't think, so something would need to be added.  Almost all of the `canSlowShortCharge` speedballs are in room, maybe not WS basement - the tricky entry speedballs that would use a comeInFalling requirement.  Entering while falling would give more usable tiles when jumping while fast before mockballing.  
But the new structure is better for modelling disconnected runways like what's used for UN Ice Shutters escape in vanilla.  So this is moving in the right direction.

> Rename `comeInCharged`
Where we are expanding the possible states Samus is in when entering a room, I think we do want this renamed now.  `comeInShinecharged` is good.

osse101 (approved):

PR #1089, opened 2023-10-09
Cross room strat restructure (part 2) - G-Mode

blkerby:
This PR continues the work on #1059, implementing the schema changes to support the new way of handling cross-room strats for G-mode:
- Deprecating node properties "leaveWithGModeSetup", "leaveWithGMode", and "gModeImmobile".
- Deprecating logical requirements "comeInWithRMode" and "comeInWithGMode".
- Adding two new exit conditions:
    - leaveWithGModeSetup
    - leaveWithGMode
- Adding two new entrance conditions
    - comeInWithRMode
    - comeInWithGMode
- Adding a new strat property: gModeRegainMobility
- Updating the Blue Brinstar region file as an example of how to apply the new schema.


strats.md
```diff
@@ -163,6 +165,62 @@ The `leaveWithSpark` object currently has no properties. If needed, properties m
...
+  "requires": [],
+  "exitCondition": {
+    "leaveWithGMode": {
+      "morphed": false
+    }
+  }
+}
+```
+
+and another where the `false` values for `morphed` are replaced with `true`. Here we are referring to nodes with `"nodeType": "door"`, which excludes sand entrances (which instead have `"nodeType": "entrance"`). These implicit strats have the same door node for both their `from` and `to` nodes.
```
kjbranch, line 220: You should be able to mix the true/false of these two conditions, right? i.e. if you needed artificial morph upon entry, but then went through a door with no cap, that you couldn't get to unless you unmorphed along the way? (I don't believe any of these exist in the vanilla tho)

strats.md
```diff
@@ -454,3 +513,103 @@ A `comeInWithDoorStuckSetup` condition must match with a `leaveWithRunway` condi
...
+- Requirements for regaining mobility (there may be multiple options here, and they should be treated either as separate strats or the requirements should be joined as though with an `or`):
+  - G-mode mobile: Samus uses the knockback from the previous room to retain mobility. Here the `knockback` property of the `leaveWithGModeSetup` must be true, and reserve energy is assumed to have been drained to at most 4 before doing the strat.
+  - G-mode immobile: Samus uses knockback from a hit in the current room to regain mobility. The possibility of doing this is indicated by the presence of a strat with a `gModeRegainMobility` property with `from` and `to` node matching the `from` node (entrance door node) of the current strat. Any requirements of the `gModeRegainMobility` strat should be included as requirements to execute the current strat; it is important that these requirements be applied *after* the requirement to damage down to 0 energy and trigger reserves. If there are multiple such `gModeRegainMobility` strats, then these should be treated as multiple options.
```
kjbranch, line 575: `from` and `to` matching the `from` sounds a bit weird since `from` will always be matching itself.

strats.md
```diff
@@ -454,3 +513,103 @@ A `comeInWithDoorStuckSetup` condition must match with a `leaveWithRunway` condi
...
+{
+  "name": "G-Mode Morph",
+  "entranceCondition": {
+    "comeInWithGMode": {
+      "mode": "any",
+      "morphed": true
+    }
+  },
+  "requires": []
+}}
```
kjbranch, line 591:
```suggestion
}
```

strats.md
```diff
@@ -454,3 +513,103 @@ A `comeInWithDoorStuckSetup` condition must match with a `leaveWithRunway` condi
...
+```
+
+
+## G-Mode Regain Mobility
+
+A `gModeRegainMobility` property on a strat indicates that the strat allows Samus to regain mobility (by taking knockback from an enemy or projectile) after entering with G-mode immobile. In all strats with a `gModeRegainMobility` property, the `from` node of the strat must be the same as the `to` node and must be a door node. A strat with `gModeRegainMobility` should normally include an `enemyDamage` requirement in its `requires`.
```
kjbranch, line 597: also consider removing projectile here

kjbranch (commented):
Looks good to me

osse101 (commented):
I don't think `gModeRegainMobility` is better than the old `gModeImmobile` object.  Its still a 'look somewhere else to handle comeInWithGMode' setup, so it works differently from all the other strats.  Does keep all the strats in one place tho.

Another way to implement it could be to make `RegainMobility` link to a new GMode node and then do the GMode strat from that node.  But adding nodes and branching on `artificialMorph` sounds messy too.  Making `RegainMobility` produce an "exitCondition" that could be used in the same room would work too lol.  Some way to show that this link is being passed through and that game state is being changed.

schema/m3-region.schema.json
```diff
@@ -346,9 +397,50 @@
...
+                  "description": "If true, then this strat results in leaving the room in a morphed state, either by maintaining artificial morph or by having the Morph item."
+                }
+              }
+            }
           }
         },
+        "gModeRegainMobility": {
+          "$id": "#/definitions/strat/properties/gModeRegainMobility",
+          "type": "object",
+          "description": "Indicates that this strat allows Samus to regain mobility (by taking knockback from an enemy or projectile) after entering with G-mode immobile.",
```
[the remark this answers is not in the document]
blkerby: So the thing is that these strats describe how to get out of the immobile state while staying in G-mode, i.e. how to regain mobility by taking an enemy hit. They don't describe that you can enter G-Mode Immobile at that location, because that is something that could be done at any door (hence doesn't need to be explicitly described in the data); it's just that entering with G-Mode Immobile isn't useful without a way to get out of it while retaining G-mode. So to me "RegainMobility" still seems more accurate for what these strats describe. But I don't care too much about the name either way; what do you think @kjbranch ?

(Also updated the description here to remove the reference to projectile, and clarify about staying in G-mode.)

strats.md
blkerby: Yes, that is the idea. The implicit strats don't mix them, but explicit strats could.

blkerby:
> I don't think `gModeRegainMobility` is better than the old `gModeImmobile` object. Its still a 'look somewhere else to handle comeInWithGMode' setup, so it works differently from all the other strats. Does keep all the strats in one place tho.

Putting all the strats in one place is one of the main goals of this migration. The old way also didn't have a way to mark a `gModeImmobile` strat as notable.

> 
> Another way to implement it could be to make `RegainMobility` link to a new GMode node and then do the GMode strat from that node. But adding nodes and branching on `artificialMorph` sounds messy too. 

Yeah needing to add extra nodes to represent `gModeRegainMobility` would seem complicated. This also needs to work with the implicit `leaveWithGMode` strats that turn around and go back out through the same door. 

> Making `RegainMobility` produce an "exitCondition" that could be used in the same room would work too lol. Some way to show that this link is being passed through and that game state is being changed.

Yeah it's an interesting idea, but the way it's defined now, "exitCondition" describes a condition leaving the room, so it would only combine with strats in the next room, not the same room.

I guess the way I'm thinking, the `gModeImmobile` property as it currently exists is working nicely, in terms of its semantics of how it combines with the explicit & implicit G-mode strats; it's nice in that it doesn't require much boilerplate, and I don't think we've had a scenario that we couldn't model accurately. So the `gModeRegainMobility` is just trying to migrate it to the new structure without making significant changes to how it works.

strats.md
kjbranch: This sounds odd and is very confusing to me. Maybe something along these lines would be more clear:
'''
... with a `from` and `to` node matching the entrance door node, as well as the `from` node of the current strat.
'''

kjbranch:
Exit/entrance conditions currently are where a setup happens in room 1 and a strat happens in room 2. These immobile strats are a bit weird in how there are three things that need to happen: setup, regain mobility, strat. 

Are there any other cross-room strats that will also require more than 2 things that we will need a more complex structure for? 

Possibly cross room mock/speed balls? The only way I see these being represented easily with 2 steps would be to continue using something like unused tiles, or to have several types of exit/entrance conditions: ranging from come jumping to come in already morphed and with speed.

osse101:
>Making RegainMobility produce an "exitCondition" that could be used in the same room would work too lol. Some way to show that this link is being passed through and that game state is being changed.

So `exitConditions` are cool because you run the strat and a flag gets set.  So `RegainMobility` would look good if it produced a `removeImmobile` type of flag.  And `RegainMobility: {}` actualy does that.  Maybe the weird looking part is how `comeInGMode` is used on the `1->5 strat link` while `RegainMobility` is back on `1->1`.  Like a `comeInGModeImmobile` condition is converted into the `comeInGMode` before that one can be used, while still keeping the immobile setting for strats that rely on enemy movement.  

Looking at it from the perspective of 'GMode Immobile is possible on any door', its pretty much ok this way.  Just a little less explicit compared to the old system.  And it has become a property to handle on all doors where the interaction is possible, except there isn't an indicator saying that all door nodes should have a self linking strat to handle Immobile.  So there could be a `Strats.md` category  for invisible flags like `Immobile` if more are added.  Since it seems like something you have to just remember to do.

strats.md
osse101: >and must be a door node

Pretty minor, but with the Entrance/Exit Condition schema you can GModeImmobile into an entrance node.  That would just be the fall through sand rooms and the sand halls probably.

blkerby:
So there's not a requirement that every door have a "gModeRegainMobility" strat. That would only happen at door nodes where there's a suitable enemy that can damage you, which is a small fraction of doors. The way I'm seeing it, this isn't any different from how not every door has a "leaveShinecharged" strat. Like all other strats in the game, you have to remember to put them in if they are applicable. The schema can't force you to remember it, regardless of whether it's a node property vs. a strat inside of links.

strats.md
blkerby: Good point, updated that now here and in a couple other places, to reflect that

osse101 (dismissed):
I think we do want the name `gModeRegainMobility` in order to keep Regain Mobility strats separate from "G-Mode Immobile Morph Do A Thing" strats.  Clearer that they are the GoTo way to end immobile.  And I think "G-Mode Regain Mobility" is identifiable enough among strats.  For notable strats, there is only the parlor offscreen Geemer drop that comes to mind.

Getting the randomizer to find which strats has the `gModeRegainMobility` sounds a little inefficient compared to placing them in a section on the door node itself.  But the randomizer also already looks through all of the strats everywhere to find which path unlocks a flag or does some specific thing so that's probably fine.

So I don't have any other issues with this PR

PR #1090, opened 2023-10-10
Update Cross Room Runways in West Crateria

kjbranch:


osse101:
```
"leaveWithSpark", # validated by schema
```
should be added to `badKeys` in keywords.py.

Looks good to me.

PR #1093, opened 2023-10-12
Cross Room Runways in Central Crateria, Pt 3

kjbranch:
Things of note:
- I kept the strats that had more unusable tiles than their in-room runway had. And I kept their node's runways. This is not yet representable.
- I deleted some strats in the Climb that climb with temporary blue that come from a doorway. I think the 28 tiles in room should be plenty for strats this complex.
- There are 3 strats in Crateria Supers that I had to split, called `SpeedKeep for Temporary Blue`. I'm not I properly renamed them and edited their notes.

### DIFF
--- a/region/crateria/central.json
+++ b/region/crateria/central.json
@@ -4851,21 +4851,7 @@
           "nodeType": "door",
           "nodeSubType": "blue",
           "nodeAddress": "0x0018c82",
-          "doorEnvironments": [{"physics": "air"}],
-          "runways": [
-            {
-              "name": "Base Runway - Final Missile Bombway Left Door (to Final Missile)",
-              "length": 3,
-              "strats": [
-                {
-                  "name": "Base",
-                  "notable": false,
-                  "requires": []
-                }
-              ],
-              "openEnd": 0
-            }
-          ]
+          "doorEnvironments": [{"physics": "air"}]
         },
         {
           "id": 2,
@@ -4896,6 +4882,22 @@
           "from": 1,
           "to": [
             {
+              "id": 1,
+              "strats": [
+                {
+                  "name": "Leave with Runway",
+                  "notable": false,
+                  "requires": [],
+                  "exitCondition": {
+                    "leaveWithRunway": {
+                      "length": 4,
+                      "openEnd": 0
+                    }
+                  }
+                }
+              ]
+            },
+            {
               "id": 2,
               "strats": [
                 {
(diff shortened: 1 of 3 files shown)


region/crateria/central.json
```diff
@@ -2999,24 +2940,46 @@
...
-                          "usedTiles": 19,
+                          "usedTiles": 20,
                           "openEnd": 0
                         }},
-                        {"obstaclesCleared": ["A"]}
+                        {"doorUnlockedAtNode": 2}
                       ]}
                     ]}
                   ],
                   "clearsObstacles": ["A"],
```
osse101, line 2958:
```suggestion
```

region/crateria/central.json
```diff
@@ -4104,6 +4086,17 @@
             {
               "id": 2,
               "strats": [
+                {
+                  "name": "Leave with Runway",
+                  "notable": false,
+                  "requires": [ {"obstaclesNotCleared": ["A"]} ],
```
osse101, line 4092:
```suggestion
                  "requires": [],
```

osse101 (approved):
It would have been a good idea to create a release before we started deleting runways.  `adjacentRunway` should be broken now, or just incomplete.

region/crateria/west.json
```diff
@@ -988,6 +988,30 @@
                       "framesRemaining": 130
                     }
                   }
+                },
+                {
+                  "name": "Leave Charged, In-Room Shortcharge",
```
osse101, line 993:
```suggestion
                  "name": "Leave Shinecharged, In-Room Shortcharge",
```

region/crateria/central.json
```diff
@@ -4448,6 +4411,38 @@
...
+                  "requires": [],
+                  "exitCondition": {
+                    "leaveWithRunway": {
+                      "length": 3,
+                      "openEnd": 1
+                    }
+                  }
+                },
+                {
+                  "name": "Leave Charged",
```
osse101, line 4430:
```suggestion
                  "name": "Leave Shinecharged",
```

region/crateria/central.json
```diff
@@ -4411,6 +4347,33 @@
...
```
osse101, line 4362:
```suggestion
                  "name": "Leave Shinecharged",
```

blkerby:
> It would have been a good idea to create a release before we started deleting runways. `adjacentRunway` should be broken now, or just incomplete.

While we're in the middle of the migration, I think I can set up the randomizer so that `adjacentRunway` (for example) can be satisfied by either a `runway` (old style) or a `leaveWithRunway` with `from`=`to` node, so it should still have good coverage. 

Probably I won't be able to use exit conditions with different `from` and `to` until the migration is complete, but those types of strats haven't ever been used yet by the randomizer anyway.

PR #1094, opened 2023-10-13
New Runway Cleanup - Misc Changes

kjbranch:
Converted references from `Charged` to `Shinecharged`
Added old `unusable tiles` as a dev note, because I think they're still somewhat useful to be able to see
There was a strat in the climb that used to just say 19 tiles. I wasnt sure if that was including the the door already unlocked or not, so I included the opposite option now, not sure which is correct.

### DIFF
--- a/region/crateria/central.json
+++ b/region/crateria/central.json
@@ -1164,7 +1164,7 @@
                   "requires": []
                 },
                 {
-                  "name": "Leave Charged",
+                  "name": "Leave Shinecharged",
                   "notable": false,
                   "entranceCondition": {
                     "comeInShinecharging": {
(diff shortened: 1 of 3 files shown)


region/crateria/central.json
```diff
@@ -2943,12 +2943,12 @@
                     {"obstaclesCleared": ["A"]},
                     {"or": [
                       {"canShineCharge": {
-                        "usedTiles": 19,
+                        "usedTiles": 18,
```
osse101, line 2946: The old values were right at 19/20 because you jump straight up and land on the left edge of the lower left platform as the first movement.
However there should be some space given to slide with Blue when doing the crouch, so 18/19 tiles is better.  Adding an openend would bring it closer to being more accurate except it would be a virtual openend so could be confusing.  But shortcharging is never going to be extremely precise and the player can just run it a little better.

osse101 (approved):

PR #1100, opened 2023-10-15
Fix a few typos/errors in the new docs

blkerby:


strats.md
```diff
@@ -403,7 +403,9 @@ A `comeInStutterShinecharging` condition must match with a `leaveWithRunway` con
   "name": "Come In Stutter Shinecharging",
   "notable": false,
   "entranceCondition": {
-    "comeInStutterShinecharging": {}
+    "comeInStutterShinecharging": {
+      "minTiles": 2
```
osse101, line 407: This reminds me that maxTiles would be around 2 or 3.  Precious Room is barely more than 1 tile + openend.  And I don't know of any places 2 vs 3 tiles makes a difference.
The technique to the stutter is also weird at fast speeds.  You want to do a poor door transitions so you don't run in too far while slowing down.

osse101 (approved):

strats.md
blkerby: Hmm, the schema isn't set up to support maxTiles here because I didn't think they were needed. But it could be added. I'm pretty sure I got Precious Room to work using the full length of runway at the top of Draygon's Room (in practice hack)

osse101: Well, it would be max effective tiles.  Where it would be better to kill your run speed mid runway.

Draygon->Precious works if you waste some frames before the door, but a perfect stutter will run into the wall before slowing down.

blkerby: So I always thought that longer runways made the stutter easier, because there would be a larger window to do the stutter while you're sliding. I guess it may not help in the same way when entering a room without much space, like Precious Room. I still wonder if a longer run makes the frame window smaller or not (or does it just make the timing/visual cue different), and whether or not there's a maximum speed above which you can't get it at all?

osse101: I would guess there isn't a max speed problem since you can move more of the slide into the dry side which may end up just being easier.  And at some point a dry side shortcharge becomes possible.  If we find an underwater runway that is shorter than a screen but still long, then you may need more exact inputs for it to work.

PR #1102, opened 2023-10-16
Update Cross Room Runways in Wrecked Ship

kjbranch:


region/wreckedship/main.json
```diff
@@ -2225,6 +2167,65 @@
...
+                  "requires": [],
+                  "exitCondition": {
+                    "leaveWithRunway": {
+                      "length": 2,
+                      "openEnd": 1
+                    }
+                  }
+                },
+                {
+                  "name": "Leave Shinecharged, Second Closest Runway",
```
osse101, line 2182: Strange that these `canLeaveCharged` didn't have any requirements on killing the kihunters before.

region/wreckedship/main.json
```diff
@@ -3084,6 +3192,43 @@
...
+                    }
+                  }
+                }
+              ]
+            },
+            {
+              "id": 2,
+              "strats": [
+                {
+                  "name": "Come in with Runway, Leave Charged",
```
osse101, line 3215: Technically requires the item to be collected.

region/wreckedship/main.json
```diff
@@ -3632,56 +3798,75 @@
...
-                    }},
-                    "SpeedBooster",
+                    "canCarefulJump",
                     {"or": [
                       "canWalljump",
                       "HiJump"
                     ]}
                   ],
-                  "note": "A bit finicky for shorter runways, but exactly 3 tiles works."
+                  "note": "Requires at least 3 tiles to work, but some longer lengths don't work."
```
osse101, line 3869: Exactly 3 tiles looked like it doesn't work, I needed a few more pixels.  I didn't hit any of the longer length anti-trickyDashJumps; not sure that that needs to be mentioned.

region/wreckedship/main.json
```diff
@@ -4332,12 +4541,14 @@
                 {
                   "name": "Shinespark",
                   "notable": false,
+                  "entranceCondition": {
```
osse101, line 4544: There should be `comeInShinechargeding` options too.

osse101 (commented):
Mostly issues with the runways from before this cross-room update.

region/wreckedship/main.json
```diff
@@ -100,6 +72,38 @@
...
+                  "name": "Leave with Runway",
+                  "notable": false,
+                  "requires": [],
+                  "exitCondition": {
+                    "leaveWithRunway": {
+                      "length": 40,
+                      "openEnd": 0
+                    }
+                  },
+                  "devNote": "This runway is shorter because of the moving floor."
```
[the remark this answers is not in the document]
kjbranch: I didn't realize that. That's really annoying lol.

region/wreckedship/main.json
kjbranch: This does seem like a problem. Fortunately, the enemies won't appear until Phantoon is defeated, so it will be very unlikely to matter. It also looks like you can farm energy from the left or right door atomic spawns (assuming it's not heated), then once full, easily kill a Kihunter while taking a hit, then farm the rest of the room to survive. Coming in from the bottom door is less forgiving though, if you enter with very low energy.

I'll put a redo of this room in my to-do list.

region/wreckedship/main.json
kjbranch: We could add something like `itemCollectedAtNode`

region/wreckedship/main.json
kjbranch: The old way to define runways was very ambiguous, and I've noticed a lot of variation based on who/when it was added. With the new schema we adopted Bobbob's way of calculating it, which is: tiles of movement that Samus has. So 3 tiles of runway in the next room is the same as 3 tiles + an open end (including the door cap), so Samus can traverse 48 pixels before entering the room.

region/wreckedship/main.json
kjbranch: Yeah true. A `leaveShineCharged` can fulfil a `comeInWithSpark`, so adding charged strats will just reduce the spark frames. I'll put it as a fixme for now

PR #1106, opened 2023-10-18
Update Cross Room Runways in Tourian Pt.1

kjbranch:
This is for the runways and the adjacent runways. Come in / leave charged is not included.


region/tourian/main.json
```diff
@@ -3737,54 +3813,6 @@
...
-                {
-                  "name": "Base",
-                  "notable": false,
-                  "requires": []
-                }
-              ],
-              "openEnd": 0
-            },
-            {
-              "name": "Diagonal Shot Runway - Tourian Escape Room 1 Right Door (to Mother Brain)",
```
osse101, line 3754: These runways would still be used for `canLeaveShinecharged` out the bottom door.

region/tourian/main.json
```diff
@@ -3494,11 +3537,44 @@
...
+              "strats": [
+                {
+                  "name": "Leave with Runway",
+                  "notable": false,
+                  "requires": [],
+                  "exitCondition": {
+                    "leaveWithRunway": {
+                      "length": 5,
+                      "openEnd": 1
+                    }
```
osse101, line 3573: Pretty easy to get hit by a rinka while taking this runway.

region/tourian/main.json
osse101: I see now that shinecharges are meant for a separate PR

kjbranch: I was actually debating about this and took it out. I felt like getting through was harder than getting the 21 tile short charge. But that might not be the case with some beams. I'll look into it more

osse101: That may be the appropriate way to look at it.  It is the same way of doing the room as speedruns, but they would have hyper beam.  And even then I wouldn't expect them to know the movements until `VH`

PR #1107, opened 2023-10-19
Update Cross Room Runways in Tourian Pt.2

### LINKED ISSUE #1060: canLeaveCharged out of Metroid rooms need some reqs
It's possible to leave charged out of Metroid Rooms 1 and 3. Currently the logic has no requirements for this. It seems like it should have requirements such as "canMetroidAvoid", Ice to freeze them, damage from tanking, or the flag for them to already be killed.

kjbranch:
Should finish off the migration in Tourian, as well as close #1060.


region/tourian/main.json
```diff
@@ -3565,23 +3609,69 @@
...
               "strats": [
                 {
-                  "name": "Base",
+                  "name": "Speed Zebetite Skip",
                   "notable": false,
-                  "requires": []
+                  "entranceCondition": {
+                    "comeInShinecharging": {
+                      "length": 4,
+                      "openEnd": 0
```
osse101, line 3644:
```suggestion
                      "openEnd": 1
```

region/tourian/main.json
```diff
@@ -3565,23 +3609,69 @@
...
+                    }
+                  },
+                  "requires": [
+                    "canSpeedZebetitesSkip",
+                    {"shinespark": {"frames": 15}}
```
osse101, line 3649: 5 frames + rinka hit

region/tourian/main.json
```diff
@@ -3565,23 +3609,69 @@
...
+                  "name": "Speed Zebetite Skip, Come in Shinecharged",
+                  "notable": false,
+                  "entranceCondition": {
+                    "comeInShinecharged": {
+                      "framesRequired": 100
+                    }
+                  },
+                  "requires": [
+                    "canSpeedZebetitesSkip",
+                    {"shinespark": {"frames": 15}}
```
osse101, line 3662: 5 frames + rinka hit

osse101 (approved):
The new acid obstacle should be added to the diagram update.

kjbranch:
I couldnt decide where to include the obstacle on the room diagram, but I think I am liking this approach of showing it in 3, since that's where it is triggered, and then an arrow pointing to the acid. 

The obstacle could potentially be renamed, since it's shown at 3, but I'm not too concerned with that unless it will eventually be needed for more things in the room. I was also having a hard time deciding if it should be an `abstract` or `inanimate`. It moves, so inanimate seemed odd, but it might be better?

PR #1108, opened 2023-10-20
Kraid runway migration, and comeInSpeedballing schema addition

blkerby:
For the new "comeInSpeedballing" entrance condition, I created a "runway geometry" object to hold the common runway properties ("length", "openEnd", etc.) as a nested structure. I'm anticipating we're going to be using this in many more places going forward (e.g. in future exit conditions like "leaveWithSpaceJump", "leaveWithSpringBallBounce", etc.), so I thought it seemed nicer for it to be bundled together into a common structure. If this is considered to be a good idea, we could go back and make the existing uses (in "leaveWithRunway" and "comeInShinecharging") likewise take a "runway" property instead of having all those fields loose at the top level; alternatively, if we like the fields being at the top level I could modify this PR to do the new entrance condition the same way.

One thing that's not addressed in this PR is that strats requiring coming into the room with "canSlowShortCharge" (including implicitly in "comeInSpeedballing") will need some way for that to be reflected in the required heat frames in the previous room. One way would be to add "minTiles"/"maxTiles" properties to "comeInShinecharging" to allow constraining the final run speed after obtaining blue. In this case it might make sense to combine "comeInShinecharging" and "comeInRunning" into a single entrance condition (probably "comeInRunning") and just have a boolean "blue" property indicating whether obtaining blue speed is required.


blkerby:
The `comeinSpeedballing` is starting simple here, just something minimal to support this part of the runway migration.

Some thoughts on possible future extensions:
- A mockball version would also be helpful and could be structured the same, just not requiring blue speed: `comeInMockballing`? Or maybe they can be the same exit condition (`comeInMockballing`) and just have a boolean `blue` property? The new setup for entrance conditions is scalable in that it's easy now to add more (not requiring much work in the randomizer), so it's not necessarily important to keep the amount of entrance conditions small; my original thought was lots of simple entrance conditions was better than a few highly complicated ones. It's just a question of how to organize them best.
- `comeInSpeedballing` could match with not only `leaveWithRunway` but also other, more specific exit conditions that could be added later, e.g. `leaveWithSpeedball`, `leaveWithBlueJump`, for cases where the jump happens from a different runway that is not connected to the door. Probably in general we wouldn't want to have to write both a `leaveWithSpeedball` and `leaveWithMockball`, so we'd want to set up a way for the blue/speedball version to be implicitly generated based on the runway length and the user's shortcharge difficulty setting. Similarly, aside from setting up speedballs there could be some other applications for jumping through a door without blue (from a disconnected runway), e.g. into a water room, though this would put you in a high position through the transition, so it's not going to satisfy `comeInJumping` in general; we could add some property to `comeInJumping` to constrain the position, or add a different entrance condition for this type of jump.
- We'll probably eventually want to add properties to constrain the range of allowable horizontal speed (measured in runway tiles before the jump) for `comeInSpeedballing`. I'm thinking of strats with mockball/speedball + springball, like crossing Everest or Sand Halls, where specific speed ranges would be required. Also the mockball through Ice Beam Gate Room has a speed requirement.

region/brinstar/kraid.json
```diff
@@ -1667,6 +1572,39 @@
...
+                      "openEnd": 2
+                    }},
+                    "canShinechargeMovement"
+                  ],
+                  "exitCondition": {
+                    "leaveShinecharged": {
+                      "framesRemaining": 120
+                    }
+                  },
+                  "note": "We won't worry about obstacles as requirements here, because achieving blue suit will destroy them anyhow."
```
kjbranch, line 1604: This is no longer true. There are some requirements to get over them and to kill them with speed now, to keep this out of `Basic`
```suggestion
                    "canShinechargeMovement",
                    {"obstaclesCleared": ["A"]}
                  ],
                  "exitCondition": {
                    "leaveShinecharged": {
                      "framesRemaining": 120
                    }
                  }
```

region/brinstar/kraid.json
```diff
@@ -2328,6 +2270,38 @@
...
+                },
+                {
+                  "name": "Leave Shinecharged",
+                  "notable": false,
+                  "requires": [
+                    {"canShineCharge": {
+                      "usedTiles": 22,
+                      "openEnd": 0
+                    }},
+                    "canShinechargeMovement"
```
kjbranch, line 2295:
```suggestion
                    "canShinechargeMovement",
                    "f_DefeatedKraid"
```

region/brinstar/kraid.json
```diff
@@ -2313,6 +2223,38 @@
...
+                  ],
+                  "exitCondition": {
+                    "leaveShinecharged": {
+                      "framesRemaining": 90
+                    }
+                  }
```
kjbranch, line 2254:
```suggestion
                  },
                  "devNote": "An X-Mode strat is likely possible but not included. It would require the doors unlocked and canRiskPermanentLossOfAccess."
```

region/brinstar/kraid.json
```diff
@@ -2328,6 +2270,38 @@
...
```
kjbranch, line 2301:
```suggestion
                  },
                  "devNote": "An X-Mode strat is likely possible but not included. It would require the doors unlocked and canRiskPermanentLossOfAccess."
```

strats.md
```diff
@@ -516,6 +520,23 @@ A `comeInWithDoorStuckSetup` condition must match with a `leaveWithRunway` condi
...
+
+Note that a `comeInSpeedballing` entrance condition can always be satisfied by coming into the room while already in a speedball. Therefore, a different entrance condition must be used if the strat specifically requires obtaining the speedball after entering the room (either by jumping through the transition, or by running through the transition and jumping afterward).
+
+A `comeInSpeedballing` entrance condition must match with a `leaveWithRunway` condition on the other side of the door. A match with a `leaveWithRunway` comes with implicit requirements:
+- A `canShineCharge` based on the combined runway length, minus the amount of tiles needed to perform the jump into the speedball. The amount of tiles needed for the jump depends on the player's shortcharge ability (as well as ability to short-hop mockball): obtaining blue speed with lower run speed means less space needed to perform the jump. This can be approximated in a simple way with the following assumptions:
+  - If the tech `canSlowShortCharge` is enabled, then 5 tiles are needed for the jump into the speedball.
+  - Otherwise 14 tiles are needed.
```
kjbranch, line 536: Seems slick, and can easily be expanded to include more precision if we ever want it. But I'm no expert on this.

kjbranch (approved):

kjbranch:
Kraid's section looks good to me. I'm not as good at reviewing the schema, but it also looks fine.

osse101:
The Kraid region changes and schema changes look correct.  

A Common `Runway` object in schema is smart.  Region seems like a good place for it.  Existing runways can be updated to reference it.
Usage in entrance/exit conditions is tied to the meaning of `usedTiles` that are part of those conditions.  If they are geometrically important or just used for gaining run speed.  So a "previous room" condition wouldn't know the structure and would be an equivalent `usedTiles` count; unstructured.  And a "current room" condition could be a `runway`.  I think we'll want runway info available to strats that are grounded on tiles but also to clarify which `usedTiles` is being used.

`comeInSpeedballing` as it is will always end with a vertical wall and its doubtful slopes will affect speed when using this condition.  But Steep Up Slopes shortening your fall time is interesting.  And `canTemporaryBlue` out of a speedball is something you could do to make use of openEnds.

The handling of various instances of `comeInSpeedballing` might need a rework as they come up.  Like how Statues Hallway isn't supported yet.  Or Brinstar Reserve - you can jump into the room to speedball, but cannot roll into the room.

osse101 (approved):

PR #1109, opened 2023-10-20
Check node types

blkerby:
The new cross-room strat schema has specific requirements on node types of the `from` and `to` nodes:
- In strats with an entrance condition, the `from` node must have `nodeType` either `door` or `entrance`.
- In strats with an entrance condition, the `to` node must have `nodeType` either `door` or `exit`. 

In the old style, cross-room strats often started at junctions instead of a door node, which makes it easy to violate the new constaints by forgetting to change the `from` node as part of the migration. To prevent problems like this, this PR adds tests that will fail if the new constraints are not satisfied.

This PR also fixes all the existing strats that were caught failing the new tests.

### DIFF
--- a/region/crateria/east.json
+++ b/region/crateria/east.json
@@ -86,6 +86,43 @@
                     }}
                   ],
                   "devNote": "To avoid making another obstacle, it is assumed the item at 3 is collected by other means."
+                },
+                {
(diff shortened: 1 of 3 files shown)


kjbranch (approved):
Sheesh. Whoever did the Crateria migration should be fired. :)

PR #1111, opened 2023-10-23
Add helpers for X-mode leniency

blkerby:
This standardizes all the X-mode strats to have a leniency of 1 extra spike hit, easily adjustable by being centralized to leniency helpers.

I'm the assuming the varying amount of hits in the existing logic were just representing different amounts of lenience and had no significance. If some of these strats actually have a required second (or third) hit for some reason, these will need to be added back.

This PR also removes the leniency from the Upper Norfair Farming Room gate glitch strat, due to the availability of the good farm right next to the gate.

### DIFF
--- a/region/brinstar/red.json
+++ b/region/brinstar/red.json
@@ -1287,7 +1287,7 @@
                   "requires": [
                     "canMidairShinespark",
                     "h_canXMode",
-                    "h_XModeSpikeHit",
+                    {"spikeHits": 1},
                     {"canShineCharge": {
                       "usedTiles": 33,
                       "openEnd": 2
--- a/region/brinstar/red.json
+++ b/region/brinstar/red.json
@@ -1052,7 +1052,9 @@
                   "name": "XMode",
                   "notable": false,
                   "requires": [
-                    "h_canSpikeXMode",
+                    "h_canXMode",
+                    "h_XModeSpikeHit",
+                    "h_XModeSpikeHit",
                     "canIframeSpikeJump",
                     {"or": [
                       "SpaceJump",
@@ -1061,7 +1063,8 @@
                     {"shinespark": {
                       "frames": 50
                     }}
-                  ]
+                  ],
+                  "devNote": "Two spike hits are expected per attempt (with any additional leniency hits being multiplied by this amount)."
                 }
               ]
             }
@@ -1116,13 +1119,16 @@
                   "notable": false,
                   "requires": [
                     "canShinechargeMovement",
-                    "h_canSpikeXMode"
+                    "h_canXMode",
+                    "h_XModeSpikeHit",
+                    "h_XModeSpikeHit"
                   ],
                   "note": [
                     "Jump from the entry runway to bounce onto spikes for XMode.",
                     "Store a spark in XMode and use that to shinepark across the room.",
                     "Bumping a solid tile before activating XMode will remove dash state, preventing shinecharging."
-                  ]
+                  ],
+                  "devNote": "Two spike hits are expected per attempt (with any additional leniency hits being multiplied by this amount)."
                 }
               ]
             }
@@ -1281,7 +1287,7 @@
                   "requires": [
                     "canMidairShinespark",
                     "h_canXMode",
-                    {"spikeHits": 1},
+                    "h_XModeSpikeHit",
                     {"canShineCharge": {
                       "usedTiles": 33,
                       "openEnd": 2
--- a/region/brinstar/red.json
+++ b/region/brinstar/red.json
@@ -1058,11 +1058,15 @@
                     "canIframeSpikeJump",
                     {"or": [
                       "SpaceJump",
-                      {"spikeHits": 2}
+                     "h_XModeSpikeHit"
                     ]},
                     {"shinespark": {
                       "frames": 50
                     }}
+                  ],
+                  "note": [
+                    "Use the Spikes in the center of the room to build a Shinespark with X-Mode.",
+                    "There are sections of the room where using more complex movement can be used to save a Spike hit when X-Moding."
                   ],
                   "devNote": "Two spike hits are expected per attempt (with any additional leniency hits being multiplied by this amount)."
                 }
@@ -1125,7 +1129,7 @@
                   ],
                   "note": [
                     "Jump from the entry runway to bounce onto spikes for XMode.",
-                    "Store a spark in XMode and use that to shinepark across the room.",
+                    "Store a spark in XMode and use that to leave with a Shinecharge.",
                     "Bumping a solid tile before activating XMode will remove dash state, preventing shinecharging."
                   ],
                   "devNote": "Two spike hits are expected per attempt (with any additional leniency hits being multiplied by this amount)."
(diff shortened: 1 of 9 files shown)


osse101:
There shouldn't be many cases of xmode leniency.  Certain strats in shaft supers probably does, like the super jump strats.  But adding leniencies is fine - I find I can't hit XMode or speedkeep or any of those on command if I need to in a seed.  There is a question of whether leniency should be multiples of 1 2 or 3 spike hits.

XMode = bounce ball + unmorph in spikes + do a thing while standing in the spikes.
The first spike hit depends on the room.  If you can bounce on a safe patch or not.  For Speedbooster strats you need more space for the bounce.
Second spike hit is to activate xmode with the unmorph
Third spike hit is when exiting XMode because you are still in the spikes.  Storing a shinespark stays in the spikes.  Jumping with blue does not stay in the spikes.  Moving in xmode lets you jump out of the spikes before ending xmode.  Superjumping leaves the spikes but may fall back in if the goal is only to get a bluesuit.  

XModing in thorns may reset your iframes before you leave xmode as a way to avoid the final thorn hit, but its hard to rely on that.

blkerby:
I see, it looks like I've misunderstood the purpose of those extra spike hits. Sound like they're generally not really for leniency, in which case it isn't correct to remove them like this PR does. We could have leniency in multiples of 2 or 3 by just repeating the "h_canSpikeXMode" in the strat; it's messy but a better alternative isn't coming to mind.

blkerby:
Ok, made a second pass at this. At @kjbranch's suggestion, made it now so that the helpers for spike/thorn hits (e.g. "h_XModeSpikeHit") no longer have the tech/item requirements baked into them, only the damage requirements for the hits (including leniency). And then in each strat I included as many copies of those as you had spike/thorn hits required before. Only now those requirements will get effectively multiplied by the leniency factor (which I assume for Extreme we'd include 1 leniency attempt, while for Insane none would be included).

I'm not familiar/skilled enough with X-mode to be able to tell which strats might have already had some leniency baked in before. Those ones should probably have their number of hits reduced, to avoid doubling up the leniency. So if you notice any of those feel free to make/suggest edits.

region/brinstar/red.json
```diff
@@ -1283,21 +1287,22 @@
...
-                    {"spikeHits": 1}
+                    }}
                   ],
                   "note": [
                     "Jump from the entry runway to bounce onto spikes for XMode.",
                     "Store a spark in XMode and use that to shinepark across the room.",
                     "Bumping a solid tile before activating XMode will remove dash state, preventing shinecharging."
-                  ]
+                  ],
+                  "devNote": "Omitting leniency spikeHits, since the firefleas provide an opportunity to farm after a failed attempt."
```
kjbranch, line 1305: Confused here, since h_XModeSpikeHit has leniency built in

region/norfair/east.json
```diff
@@ -1838,10 +1838,15 @@
...
+                    {"heatFrames": 220},
+                    "canGateGlitch",
+                    {"or": [
+                      {"ammo": { "type": "Missile", "count": 1 }},
+                      {"ammo": { "type": "Super", "count": 1 }}
+                    ]}
                   ],
-                  "clearsObstacles": ["A"]
+                  "clearsObstacles": ["A"],
+                  "devNote": "Omitting gate glitch leniency, since the nearby Gamet farm allows refilling energy and ammo."
```
kjbranch, line 1849: This is interesting. The farm is good, but the room is heated. This will be the only gate glitch in logic if you only have a small number of ammo, which can be a bit confusing. It also only works because its in the same room - if there is a great farm in a neighboring room it won't be in logic. 

Something to compare to, should the green hill zone gate be in logic if the right door can be used to reset room? those hoppers can be farmed for missiles.

kjbranch (approved):

region/brinstar/red.json
blkerby: Fixed, thanks!

region/norfair/east.json
blkerby: I would be fine going either way on this; we could keep the leniency here so that the ammo requirements stay consistent across all gates. Or we could add more strats with reduced leniency based on good neighboring farms.

The Green Hill Zone sidehoppers have 34% Missile drop rate (60% with full everything else), so with having to reset the room every time they're not a great farm but also not bad. Neighboring good farms would be in the realm of possibility of something we could model with the new style of cross-room strats. There's still a delay from having to reset the room, but if you don't have Varia or tanks that still might be easier than in-room farming in Upper Norfair Farming Room. And like you said, there's a question of whether all of this makes it harder for the player to understand how gate glitch leniency works.

osse101: The UN farm does become harder to follow the logic if you are basing it on GGG leniency value, which is what I've been doing.  Looking at it as something you could do, as a `X Difficulty` runner, then you would count the farm as an extension of equipment capacity.  I think I would rather ammo requirements stay consistent.  

GGG does feel a lot better with a nearby farm spot.  But that's better off just being a bonus.  Reloading rooms to farm is still going to feel bad, and most of the random enemy farm spots are still pretty slow as ammo per reload.

Including the farm spot here; 
Expert - no Etanks, 1 missile pack is fine
VH - 1 tank, 1 pack.  This gate glitch comes up in other randomizers and its pretty uncomfortable to attempt the gate and then refill on no tanks.
Hard - 1 tank for the room, and 2 missile packs seems ok.  I expect a lot of missed shots at hard.

region/wreckedship/main.json
```diff
@@ -4513,10 +4520,12 @@
                   "notable": false,
                   "requires": [
                     "h_canXMode",
+                    "h_XModeThornHit",
+                    "h_XModeThornHit",
+                    "h_XModeThornHit",
```
osse101, line 4525:
```suggestion
```

region/wreckedship/main.json
```diff
@@ -4536,7 +4545,10 @@
                   "exitCondition": {
                     "leaveWithSpark": {}
                   },
-                  "devNote": "FIXME: There is likely a fast wall jump climb strat with more spark frames."
+                  "devNote": [
+                    "Three thorn hits are expected per attempt (with any additional leniency hits being multiplied by this amount).",
```
osse101, line 4549:
```suggestion
                    "Two thorn hits are expected per attempt (with any additional leniency hits being multiplied by this amount).",
```

region/norfair/east.json
blkerby: Makes sense, I'll put the strat back how it was and we could think about it more in a separate issue or PR.

For this strat I think the ammo requirement is less of an issue than the energy requirement. On Hard, the 14 leniency shots are 60 heat frames each, plus the base 100 heat frames for heated gate glitch on top of 120 heat frames in the strat. Adds up to 1060 heat frames. With the Hard multiplier, that becomes 1590 heat frames, or 397 energy. So it would need 3 tanks, or 4 tanks if you need to be able to make it back to the farm after opening the gate. 

Agree that 1 tank seems like enough for this to be in logic on Hard. And with only 1 tank, I'm not sure that a second missile pack really makes the strat easier. Probably 5 attempts is enough that you would want to be going back to farm energy anyway by that point, and this will refill your Missiles. The way I wrote though it would put it in logic with no tanks, which seems unforgiving for Hard (and maybe Very Hard)

osse101 (approved):

PR #1115, opened 2023-10-23
Run autoformatter to standardize JSON format

blkerby:
This runs the script created in PR #1114. This PR should not be merged until after PR #1114 is merged first. Also if any other PRs are merged before this one, then this PR should be redone (rerunning the script again) before being merged. In general this PR should probably be redone before merging, in order to create a cleaner history (since it has been updated after changes to the autoformatting script).

After this PR is merged, conflicts can be avoided with pending PRs, by having each pending PR merge PR #1114 into it and run the autoformatting independently.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,1 +1,2 @@
 9faa7c1d54e11f578188d9bd9fea645eb12505e7  # Run autoformatter
+08dbc21764e9e01c1f50bf93a94298fe735d4a09
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,1 @@
+9faa7c1d54e11f578188d9bd9fea645eb12505e7  # Run autoformatter
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Auto detect text files and perform LF normalization
 * text=auto
 .vscode/*
+scripts/.idea/*
 
 __pycache__
 *.pyc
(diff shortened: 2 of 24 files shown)


blkerby:
Listed the formatting commit in `.git-blame-ignore-revs`, so GitHub will ignore it in the "blame" view. I was hoping this would also prevent it from counting in GitHub Insights, so that people casually looking at the project stats wouldn't be misled into thinking I contributed more than I did; but it sounds like that functionality might not be supported yet. If this PR is accepted, I'd be happy if one of the primary logic maintainers (@kjbranch or @osse101) wanted to run the script and remake the PR, so that credit would go more to the right place

kjbranch:
I think this looks nice, and I'm ok with it. The main thing that's different that feels weird is that if a `requires` has a single value in it, it is always forced to a new line. But, maybe this is better anyway, as it makes adding another item to the requires easier - and a lot of strats may get very low-level tech added to them eventually (i.e. `canMidAirMorph`).

osse101 (commented):
Requires is pretty important.  It is probably better to always have it multiline, unless the list is empty.

Shinespark is fine single line to me.  You mostly want to know that there is a health cost occurring and the amount isn't super important, but is something you can look at.  Then excess cost is even less important.

enemyDamage seems better single line except that opens it up to some confusion like if there are multiple potential damaging enemies, or number of hits isn't 1, then you would want to see that information.  So multiline might be correct for that.

I like how enemyKill is done, more elements = more lines.

droprequires, betweenNodes, homeNodes and all that isn't important so can stay 1 line.

ammo looks better as 1 line even thought count matters.  type is usually clear from the strat.

cleared/clears obstacles looks nicer as 1 line.

So I don't see anything that I'd change.

blkerby:
Closing since @osse101 will create a new PR incorporating the latest merged changes

PR #1117, opened 2023-10-23
Update Cross-Room X-Ray Climbs

kjbranch:
This was prioritized because currently new and old x-ray climbs have different leniency factors.

I did get rid of some `canNavigateUnderwater` (not really affecting x-ray movement and wont matter at this difficulty) and `canHeatRun` (implied by the heat frames and wont matter at this difficulty).

### DIFF
--- a/region/maridia/inner-pink.json
+++ b/region/maridia/inner-pink.json
@@ -3670,26 +3670,6 @@
                   "notable": false,
                   "requires": [ "f_DefeatedDraygon" ]
                 }
-              ],
-              "bypassStrats": [
-                {
-                  "name": "Halfie Climb Room Xray Climb Grapple Clip",
-                  "notable": true,
-                  "entranceCondition": {
-                    "comeInWithDoorStuckSetup": {}
-                  },
-                  "requires": [
-                    "canXRayClimb",
-                    "canGrappleClip"
-                  ],
-                  "note": [
-                    "Perform an Xray Climb until you are just above the door into Cacatac Alley.",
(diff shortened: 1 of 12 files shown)


region/maridia/inner-pink.json
```diff
@@ -3717,21 +3675,12 @@
                 {
                   "name": "Halfie Climb Room Xray Climb Grapple Clip",
                   "notable": true,
+                  "entranceCondition": {
+                    "comeInWithDoorStuckSetup": {}
```
osse101, line 3679: There's probably an issue here, being on a node lock.
It should start on node 2 to clip into node1 door, however that should be written.

osse101 (approved):

region/maridia/inner-pink.json
blkerby: I was thinking we could add a `bypassesLock` boolean as a strat property, to be able to represent strats like this (and then deprecate the `bypassStrats` lock property). If that sounds like a good solution, that could be added now. Otherwise, I guess postpone for later and just leave the strat where it was, inside the `bypassStrats` in the lock.

osse101: It does fit the structure better if this strat were moved to links.  
Bypassing a door lock then makes the most sense as a strat property, yeah.  
IDK if there are other lock type bypasses, but you could just say it means bypassing door locks.  Like Draygon you can clip through the left side door with grapple and some other items, but grapple would let you do the fight too.

PR #1118, opened 2023-10-24
Revert "Update Cross-Room X-Ray Climbs (#1117)"

kjbranch:
This reverts commit 091d7fa6907cdbcb4026d1240b6a9536d9c36c3c.


kjbranch:
Reverting this because it created unusual and incorrect conflicts with #1110

PR #1119, opened 2023-10-24
Re-Open Update Cross-Room X-Ray Climbs

kjbranch:


blkerby:
For some reason, the results of auto-merging when trying to pull in upstream/master (with UN West changes) into the xroom-xray branch were super messed up, no idea what was happening really but it looked like a bug in the merge driver I'd guess. Happened when I ran it locally too. Was able to get it sorted out using VSCode 3-way merge tool.

PR #1122, opened 2023-10-24
Change room name of green brin main shaft

blkerby:
The slash in the name was going to create a problem when we split region files into room files. Also, this way the room name now matches the name in the Wiki (https://wiki.supermetroid.run/Green_Brinstar_Main_Shaft).

### DIFF
--- a/region/brinstar/green.json
+++ b/region/brinstar/green.json
@@ -3,7 +3,7 @@
   "rooms": [
     {
       "id": 44,
-      "name": "Green Brinstar Main Shaft / Etecoon Room",
+      "name": "Green Brinstar Main Shaft",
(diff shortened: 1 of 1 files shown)


kjbranch (approved):
Seems fine to me. 

I think it would be nice to eventually get all of our room names to match the wiki - either by changing things here or on the wiki.

blkerby:
I think the room names are almost all matching the wiki, but would be good to find if there are more exceptions. I checked Crateria  and the only one that wasn't exactly a match was "Lower Mushrooms": the Wiki redirects that to "Mushroom Kingdoom", though it's still named "Lower Mushrooms" in the "List of Rooms" page

PR #1123, opened 2023-10-25
Add scripts to split region files into room files with flattening

blkerby:
This PR adds a script to split all the region files apart into individual room files, while flattening the nested structure of `links/from/to/strats` into a flat list of `strats` per room, along with supporting changes to schemas and tests. 

To run the script locally, simply run `python split_region_files.py` from a working directory of `sm-json-data/scripts`.

Some things to note:
- Any notes and devNotes inside of `links` are not preserved.
- Git history does not connect the new room files with their source region files (I tried to find a way to do this but it doesn't seem possible to do in a practical way).
- This PR fails tests, since the tests are updated to target the new file/schema structure. The tests will pass after running the script.

The plan is that if this PR is accepted, then after being merged, a new PR can be created to commit the output of running the script. For other pending PRs that are not merged before that point, they can avoid conflicts by running the same script independently (similar to what we did for auto-formatting).


osse101 (commented):
Schema and tests seem to be correct.  But its also a lot of text :eyes:
I wasn't ready for the rooms array removal and links change, when of course it would be different 😸 

Script does split regions into room files, and passes tests.

blkerby:
> Schema and tests seem to be correct. But its also a lot of text 👀 I wasn't ready for the rooms array removal and links change, when off course it would be different 😸

Is it ok doing it like this, all these changes at once? Or would you prefer I go back and do it one step at a time?

tests/asserts/keywords.py
```diff
@@ -398,541 +346,441 @@ def process_strats(src, paramData):
...
-                            "fullarea": fullarea,
-                            "links": {
+                room = roomJSON
+                roomName = room["name"]
+                area = room["area"]
+                subarea = room["subarea"]
+                subsubarea = room["subsubarea"] if "subsubarea" in room else ""
+                showArea = False
+                fullarea = f"{area}/{subarea}" + ((subsubarea != "") and f"/{subsubarea}" or "") + f"/{roomName}"
+                print(fullarea)
```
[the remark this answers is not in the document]
blkerby: Makes sense, removed this print now.

tests/asserts/validate.py
```diff
@@ -197,65 +197,69 @@
...
-                            errorLine.replace("\n",""),
-                            ("-" * (errorCol - 3)) + "^"
-                        ])
-                        bail = True
+            if os.path.isdir(os.path.join(".", "region", region, subregion)):
+                if "roomDiagrams" not in subregion:
+                    for roomFileName in os.listdir(os.path.join(".", "region", region, subregion)):
+                        if ".json" in roomFileName:
+                            roomName = roomFileName.replace(".json", "")
+                            print("   " + roomName)
```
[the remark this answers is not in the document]
blkerby: Went back to only printing subregion now.

osse101:
Its alright.  

I looked at the individual commits, since you reminded me I can look at the changes in steps, and that helped a little to focus on the real changes, but gave a pretty poor diff overall.

schema/m3-room.schema.json
```diff
@@ -0,0 +1,1922 @@
...
+          "title": "Steep Down Tiles",
+          "description": "Number of tiles in the runway that steeply slope (one vertical tile) downwards"
+        }
+      }
+    },
+    "strat": {
+      "$id": "#/definitions/strat",
+      "type": "object",
+      "required": [
+        "name",
```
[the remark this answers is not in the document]
blkerby: Had to revert this: we can't make "from" and "to" required in the schema yet, since there are strats inside runways (etc.) that don't have them. After the cross-room strat migration is complete, we could add this to make those fields required.

In the meantime, I added checks in the keywords test now to give a nice error if "from" or "to" is missing for strats in the room "strats" array (previously the test would have just crashed if one of these were missing).

blkerby:
> Its alright.
> 
> I looked at the individual commits, since you reminded me I can look at the changes in steps, and that helped a little to focus on the real changes, but gave a pretty poor diff overall.

Yeah as you saw I tried renaming the region file as a separate commit, to make the diff show up for the changes, but for some reason the diff on GitHub looks terrible. I found that pasting the two files into Sublime Text gives nice diff results (and VSCode does too).

blkerby:
I noticed that the schema validation slowed down significantly after splitting the region files into room files. Looking into this a bit, found that it's because `jsonschema.validate` validates the schema itself on each call, in addition to the JSON instance being validated. So by making more calls to `validate` the overhead of validating the schema itself was starting to dominate. Pushed a commit to resolve the problem by constructing the validator just once (which validates the schema) and then calling it on each instance. This was based on a suggestion in the docs (https://python-jsonschema.readthedocs.io/en/stable/api/#module-jsonschema):

> If you know you have a valid schema already, especially if you intend to validate multiple instances with the same schema, you likely would prefer using the [jsonschema.protocols.Validator.validate](https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/protocols/#jsonschema.protocols.Validator.validate) method directly on a specific validator

osse101:
The validator change makes it much faster, faster than it was before this PR by a little bit.

kjbranch:
I haven't looked through the schema in detail, but I think the result output looks really nice.

Are the notes on the links the only information lost from this migration? I think that is not a big deal to lose, because it is mostly just describing that some links are bypassing junctions and shouldnt be used otherwise. (i.e. `This link is only used for Shinesparks and G-Mode, all other strats should go 1->2->3.`) 
- To remedy this, it may be worth defining how to choose the to/from node on a strat in the schema somewhere, but it might be tricky to describe.
- If no information is lost, I would say lets go for it and if we decide there's something we would like structured a bit differently, it will be easy to alter it again. 
  - One thing that does come to mind is if it might make sense to make the to/from node be better as an array of some sort, i.e. `"path": [1, 2]`, as it may be easier to search. Alternatively, is the schema able to force that the strats are ordered numerically by `from` then by `to`?

blkerby:
If it would be useful to keep the note/devNotes from the links, I could make a spot to retain it in the schema, probably by just keeping the old "links" schema but with the "strats" body removed. Then it could be removed over time, checking that important info (if any) in the notes has been taken into account elsewhere.

An array property like `path` instead of `from` and `to` could work. I'd suggest the name `link`, as in `"link": [1, 2]`. This name could have an advantage in that it's more clear that it describes a pair of nodes. With `path` it could sound like more or less than two nodes could be provided, which I don't think we'd want to support. Either way we could add a schema validation to require that the array has exactly two elements. Having the `from` and `to` together like this seems appealing from the standpoint of making it easier to search, and also making the strats slightly more compact. I like it. Thoughts @osse101 ?

kjbranch:
Good call. I would definitely want it to have exactly 2 elements, otherwise it would be a huge pain to search for.

osse101:
I like the links array, looks better.  A 3 node link doesn't make sense since that would be equivalent to travelling directly to the last node.  Touching a middle node to unlock a door or collect an item but still continue on with the same strat could be neat.  But I don't know that we'd want to farm or CF mid strat if that were possible somewhere.

`Link: [1,1]` looks just as odd since you don't go anywhere.  But its understandable.  What I'm worried about right now is that the room diagrams aren't becoming more confusing.  Nested links can match up to the picture and is easy to see; flattened links stand out less which nodes are connected, but that also means the diagram can clarify the connections.  

There should be a handful of links->devnotes for strats which haven't been added yet.  And it would be a place to describe a group of strats who follow the same node link but definately hasn't been used much for that.

blkerby:
So one thing we could do is add tests to enforce that the list of strats remains sorted by `from` node and then `to` node. Then strats with the same `from` and `to` nodes would stay next to each other in the list.

It could make sense for the concept of "link" to stick around for the purpose of describing room diagram edges. I'm hoping that eventually we could get it to where these diagrams could be automatically generated. The main things we'd need is for each node to define a bounding box (or more generally, a polygon) and each link to define a curve connecting the two nodes (e.g. by a sequence of points which are to be smoothly interpolated). Links could have tags such as "G-Mode" to allow drawing them in a different color or being toggleable in a logic viewer. In this case there could be multiple links with the same from/to node but different tags. We would want a way to say that a strat is associated with a specific link (if more than one link with that from/to node exists). Given that we would no longer be stuffing the strats inside of the link, a "link" property inside the strat would be a natural way to make that association via a link `id`. The link `id` may only need to be unique among links with the same `from` & `to` nodes. This is all going beyond the scope of this PR, but seems worth thinking ahead about: it may mean that "link" isn't a great name for the from/to node array inside a strat, if we want to reserve the "link" property to be able to reference a specific link. We could try to encode the whole link identifier as a 3-element array with the third element being a link name to disambiguate links with the same from & to nodes, e.g.

```
"strats": [
  {
    "link": [1, 2, "gmode"],
    ...
  }
]
```

We would want that third array element to be optional in case there is a unique link between those nodes or if we want to implicitly use a default link; so that could get a little messy. An alternative is to keep the components as separate fields:

```
"strats": [
  {
    "from": 1,
    "to": 2,
    "link": "gmode",
    ...
  }
]
```

Or something in between, where just "from" and "to" stay combined into a single property, e.g. "fromTo" (or a better name if we can think of one).

kjbranch:
After talking with @blkerby, we're thinking we keep a `links` section which can house the notes, and they can automatically check that strats exist with the same `from/to` address (called `link`). If you add a strat with a `to/from` that isn't in the `links`, it will give an error, which is hopefully a useful reminder to edit the room diagram.

As for room diagrams, we're thinking it would be nice to use multiple colors to show different types of links that can exist between nodes. This could help users more easily see what types of strats are possible in a room. Eventually we can make all of this automated, so a player can toggle tech and links will be added/removed from the room diagrams so they can see what is in logic with their settings ... but we're a long way off that for now. In the mean time, I have updated a room diagram as an idea of how we can utilize multiple colors: 
- 1 line will exist between a pair of nodes, prioritizing the type closer to the top of the next list. If the two directions need to be different colors, it will instead exist as two arrows.
  - Red: spawn at
  - Yellow: normal strats
  - Blue: cross room
  - Purple: g-mode
room diagram update FishTank

osse101:
Using a links structure to describe and generate the room diagram connections sounds cool.  And coincidently works better with json files being one per room.

And adding more details to the map seems sensible, its not really important that they be kept clean.  I was wondering about the self linking strats too.  If farm spots should be listed, since the ones we use are arbitrary at this point.  Obstacle clearing strats is probably fine left as it is, just listing the obstacle.  G-Immobile could be marked.  CF nodes may be too numerous.

Prioritizing line colours depends on what the room diagram is being used for.  Its fine for showing which nodes can connect to others.  I was thinking coloured marks along the priorty colour line could keep more information, if we wanted.  And GMode links are still weird.

How do spawnAt and crossRoom strats interact?  3 and 4 are the door nodes, but spawnAt says to begin the room at 7 and 8.
8->5 has a comeInShinecharged as a crossroom strat.  And a walljump just before the transition, enter with charge beam stored, and unmorph just before transition, plus enter while damage boosting which hasn't been added.  So they are all crossroom strats but just the one has entranceConditions defined right now.

blkerby:
Is there anything else that needs to be addressed before this can be merged and executed? The way the PR now retains the `links` skeleton should help prevent accidentally forgetting to update the room diagram when adding a new link; and I assume that further enhancements to the room diagrams would be a separate issue?

osse101:
- The flattened array feels more uncomfortable since there are no big organizing structures anymore.
- Its not ideal for notes referencing strats to be in links instead of in strats
- Its like we are better set up to reference strats from other sections of the file now.  `reusableRoomwideNotable` is the only thing doing that now though.
- It feels a lot faster scrolling through strats.  Code review will definetely go faster.  
- I like the changes.  It all looks to work

schema/m3-room.schema.json
```diff
@@ -0,0 +1,1985 @@
...
+              "additionalProperties": false,
+              "properties": {
+                "speedBooster": {
+                  "$id": "#/definitions/strat/properties/entranceCondition/properties/comeInJumping/properties/speedBooster",
+                  "type": ["boolean", "string"],
+                  "title": "Speed Booster",
+                  "description": "Whether or not this run should be performed with Speed Booster.",
+                  "enum": [
+                    "any"
+                  ]
```
osse101, line 176: This enum is different from comeInRunning

schema/m3-room.schema.json
```diff
@@ -0,0 +1,1985 @@
...
+              "properties": {}
+            },
+            "comeInWithGMode": {
+              "$id": "#/definitions/strat/properties/entranceCondition/properties/comeInWithGMode",
+              "type": "object",
+              "title": "Come in With G-Mode",
+              "description": "Represents that Samus must have or obtain G-mode coming through this door.",
+              "required": [
+                "mode",
+                "morphed"
```
osse101, line 340: I don't think `morphed` should be required unless a majority of GMode strats use `morphed`.  And that may be the case.

schema/m3-room.schema.json
```diff
@@ -0,0 +1,1985 @@
...
+            "default": 0,
+            "examples": [
+              1
+            ]
+          },
+          "to": {
+            "$id": "#/properties/links/items/properties/to",
+            "type": [
+              "array",
+              "null"
```
osse101, line 1653:
```suggestion
              "array"
```

osse101: An empty array looks better and doesn't break tests.

And probably also means a clean links section would have a link entry for every node as a `from` value.  And a `to` array to be filled in with the in-room node connections..

osse101 (approved):

schema/m3-room.schema.json
blkerby: Yeah, it needs to be fixed. Maybe it can be done in a separate PR, to keep the history cleaner, since there's so much going on in this one. @kjbranch already found it needed to be fixed in #1128 so maybe it can be adapted there after this is merged

schema/m3-room.schema.json
blkerby: I think it's fine either way. If we change it to be not required, again maybe it can be done in a separate PR to keep the history cleaner?

schema/m3-room.schema.json
blkerby: Agree, the "null" option doesn't make sense. I can remove it in a separate PR after this is merged & executed, if that sounds ok?

PR #1130, opened 2023-11-02
A Couple Quick Changes to Oasis

kjbranch:
I was working on Oasis, when we realized that a `reusableRoomwideNotable` needed to be added to one of the strats, so it's a very small PR with a bit of the runway migration included


blkerby:
Looks good, thanks!

PR #1131, opened 2023-11-02
Cross Room Migration: West Sand Hall

kjbranch:


osse101 (dismissed):
You have to be going so slow to get half of a useful tile for `comeInShinecharging` without Gravity.  14 tiles with `canSlowShortCharge` if you want to put that in, or as a fixme.  Any faster and its more likely you slide into the sand.

PR #1133, opened 2023-11-03
Runway migration: a couple Green Brin rooms

blkerby:


region/brinstar/green/Brinstar Pre-Map Room.json
```diff
@@ -229,26 +201,79 @@
...
+        {"canShineCharge": {
+          "usedTiles": 14,
+          "openEnd": 1
+        }},
+        "canSpeedball"
       ],
-      "clearsObstacles": ["A"]
+      "clearsObstacles": ["A"],
+      "note": "Run on the upper platform to perform the speedball.",
+      "devNote": "FIXME: This length implies 3 unusable tiles, yet the docs say 5 for a slow speedball in general?"
```
osse101, line 232: Other slow speedballs are through door so there's a bit of weirdness.
But moreso, this is a harder shortcharge at 14 so a smaller short hop is expected.  `canSlowShortCharge` is targetted more at 16 tiles.  Really the tech doesn't need to be here because you jump as soon as you turn blue so it would be covered by doing a 14 shortcharge.

region/brinstar/green/Brinstar Pre-Map Room.json
```diff
@@ -229,26 +201,79 @@
...
+      "exitCondition": {
+        "leaveWithRunway": {
+          "length": 15,
+          "openEnd": 0
+        }
+      }
+    },
+    {
+      "link": [2, 2],
+      "name": "Leave Shinecharged",
```
osse101, line 263: [2,1] is able to leave shinecharged using the same runway.  60-65 frames remaining.
Strange how that would be the only reason to have `A` and there is no strat which uses it.

region/brinstar/green/Brinstar Pre-Map Room.json
blkerby: Makes sense. Sounds like at some point there could be an opportunity to refine the requirements for other speedballs, to reflect that with a good enough shortcharge it can be doable with fewer than 5 tiles. The fact that going through a door transition can make a shortcharge harder is another thing we aren't modeling in general but maybe we could; not sure exactly what that would look like. Also shortcharges requiring heat damage -- maybe it could be useful to include a leniency multiplier (or require heatProof) to represent the need for many tries, for shortcharges close the minimum length for the selected difficulty, similar to what we are doing for other tricks like gate glitches and X-ray climb setups. On the other hand, there are some cases where shortcharges are not retriable, e.g. shinecharging out the top left door of Crocomire Escape with the gate open, so this could get tricky.

region/brinstar/green/Brinstar Pre-Map Room.json
blkerby: Added a 2->1 leaveShinecharged strat now. I'm guessing the "A" obstacle was mainly for the scenario where you go to the other side to kill the enemies to unlock the gray door, then go back. This can also be useful with ammo doors like in Map Rando (though right now the randomizer is not very smart in how it handles doors -- doesn't remember they're unlocked, but we'll fix that eventually).

region/brinstar/green/Brinstar Pre-Map Room.json
osse101: `canShortHopMockball` is a tech that is hard to define so it is assumed that you can do shorter hops as difficulty rises.

Missing a shortcharge is related to how close it is to your selected runway length.  So it would take many tries at every level if you are going for the shortest shortcharge in logic.  And adding a few tiles makes it much easier to get every time.  

One shot tricks are expected at extreme, VeryHard-Expert would opt-in to them with notables.  Hard we're trying to avoid them for race settings.  But its still very frustrating below expert.  And any skill level can blow the shinespark even if they hit the shortcharge.  Looping around to try again adding some more costs.  But its mostly on the randomizer to keep the situation from feeling too bad.

PR #1134, opened 2023-11-03
Runway migration: Etecoon ETank Room

blkerby:
This room had a bug (which has come up in at least a couple seeds reported in the Map Rando Discord) where there were "canLeaveCharged" strat on the bottom left door using the runway on the top, but without an "initiateRemotely". So that gets fixed in this PR.

### DIFF
--- a/region/brinstar/green/Etecoon Energy Tank Room.json
+++ b/region/brinstar/green/Etecoon Energy Tank Room.json
@@ -15,52 +15,6 @@
       "nodeSubType": "green",
       "nodeAddress": "0x0018f3a",
       "doorEnvironments": [{"physics": "air"}],
-      "runways": [
-        {
-          "name": "Base Runway - Etecoon E-Tank Top Left Door (to Etecoon Supers)",
-          "length": 5,
-          "strats": [
-            {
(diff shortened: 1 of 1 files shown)


osse101 (commented):
[1,2], [1,3]
[2,1], [2,3]
All are able to come in shinecharging and then leave shinecharged through a different door.

blkerby:
> [1,2], [1,3] [2,1], [2,3] All are able to come in shinecharging and then leave shinecharged through a different door.

It's true but would those be useful strats to add? My intuition was that three-room shinespark strats would need canShinechargeMovementComplex; and in Map Rando at least, at that difficulty level (Very Hard) the in-room 20-tile runway would be enough.

osse101:
I hadn't thought of it like that.  Should 3 room shinespark strats be a new tech then, or a component of `canShinechargeMovementComplex`?
Only looking at this room, the movement is `canShinechargeMovement`.  But then planning to use a shinespark, especially one through the save room door, is more complicated.  Jumping the secret pit isn't free at intermediate.  And piecing together a runway like that to use later is probably too much.

PR #1135, opened 2023-11-03
Runway migration: Early Supers

blkerby:


region/brinstar/green/Early Supers Room.json
```diff
@@ -253,8 +175,12 @@
     {
       "from": 4,
       "to": [
-        {"id": 1},
-        {"id": 2}
+        {
+          "id": 1,
+          "devNote": "FIXME: A leaveCharged might be possible here from above? If so, it would be very tight."
```
osse101, line 180: I'd need another 2 or 3 frames to leave shinecharged, with none remaining.  But I can leave in spark.
It technically possible to start running from within the leftmost shutter and leave shinecharged. Not reliable and takes exact tap timings.

region/brinstar/green/Early Supers Room.json
```diff
@@ -357,6 +295,42 @@
...
+          "openEnd": 2
+        }},
+        "canShinechargeMovementComplex"
+      ],
+      "exitCondition": {
+        "leaveShinecharged": {
+          "framesRemaining": 145
+        }
+      },
+      "note": "Gain the shinecharge near the edge of the runway so that you slide off quickly, while shooting the door open."
```
osse101, line 332: Slide off is more like an expert movement, then as a requirement a bit higher than that since this takes a room reset to retry.

region/brinstar/green/Early Supers Room.json
blkerby: I'm curious where/how you do the spark. Do you land on the floor and then jump up, or on the platform and then jump down? Or spin jump off the crumble and spark mid-air from there? How many shinespark frames?

region/brinstar/green/Early Supers Room.json
blkerby: I see, do we need a new tech then? Sliding off like this is going to be useful in a lot of places for quick shinecharge movement. To me this feels much less difficult than Expert tech like Moat CWJ or Baby Skip for example. I was thinking Very Hard ("canShinechargeMovementComplex") would be fine

osse101: Slide off doesn't actually feel that bad at mid run speeds so maybe it is ok in very hard.  I mostly remember it being difficult in speedruns because you also go full speed there.  In this case you have to run near full speed to beat the crumbles and shutters.  With the door opening shot, its pretty hard to do.

How about lowering frames remaining to either 140 for a middling number, or 135 for walking off the ledge while opening the door?

region/brinstar/green/Early Supers Room.json
blkerby: Thanks for the video, I added the strat now. I assume it should be notable because it seems super precise. I put a "canQuickCrumbleEscape" requirement on it, been thinking that we should rename this to "canQuickCrumble" or "canCrumbleQuickDrop" and define it to just be the quick drop; "crumble escape" seems overly niche to be a tech.

region/brinstar/green/Early Supers Room.json
blkerby: Sure, lowered it to 140. I've also been thinking that we could add some leniency on the randomizer side, a number of frames to be subtracted from the 180 frames that you're initially assumed to have, maybe something like this:

Basic/Medium: 60 frames
Hard: 30 frames
Very Hard: 20 frames
Expert: 10 frames
Extreme: 5 frames
Insane: 0 frames

On the other hand, to make that work for in-room shinecharges, we would need to add a `framesRemaining` field to `canShinecharge`.

osse101: It is feeling to me more important to add shinespark frames leniency.  It is another case where we would take off 10 or 20 frames as a way to add leniency, in the past.  And I take off closer to 5 now.

In room shinecharge strats should have the appropriate tech for the difficulty of the strat so wouldn't need leniency.

PR #1136, opened 2023-11-04
Runway migration: a few more Green Brin rooms

blkerby:
Green Brinstar Beetom Room had the runways as free, which was unsound. Added an obstacle to track the Beetoms being dead or frozen.

In Green Brinstar Fireflea Room, I migrated the left-side X-mode leaveShinecharged, but skipped the right-side one, because it seems like short-charging using the 17-tile runway connected to the door would be easier?

### DIFF
--- a/region/brinstar/green/Green Brinstar Beetom Room.json
+++ b/region/brinstar/green/Green Brinstar Beetom Room.json
@@ -104,7 +104,7 @@
         {"enemyDamage": {
           "enemy": "Beetom",
           "type": "contact",
-          "hits": 2
+          "hits": 4
         }}
       ],
       "exitCondition": {
--- a/region/brinstar/green/Green Brinstar Beetom Room.json
+++ b/region/brinstar/green/Green Brinstar Beetom Room.json
@@ -15,21 +15,6 @@
       "nodeSubType": "blue",
       "nodeAddress": "0x0018f22",
       "doorEnvironments": [{"physics": "air"}],
-      "runways": [
-        {
-          "name": "Base Runway - Green Brinstar Beetoms Left Door (to Etecoon E-Tank)",
-          "length": 2,
-          "strats": [
-            {
-              "name": "Base",
-              "notable": false,
-              "requires": []
-            }
-          ],
-          "usableComingIn": false,
-          "openEnd": 1
-        }
-      ],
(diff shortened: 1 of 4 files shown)


region/brinstar/green/Green Brinstar Beetom Room.json
```diff
@@ -99,11 +76,26 @@
...
       ]
     }
   ],
   "strats": [
+    {
+      "link": [1, 1],
+      "name": "Leave With Runway",
+      "notable": false,
+      "requires": [
+        {"obstaclesCleared": ["A"]}
```
osse101, line 90: Taking damage is a pretty reasonable way to use the runway with beetoms alive.

region/brinstar/green/Green Brinstar Beetom Room.json
```diff
@@ -168,7 +162,8 @@
       "notable": false,
       "requires": [
         "Ice"
-      ]
+      ],
+      "clearsObstacles": ["A"]
```
kjbranch, line 166:
```suggestion
      "clearsObstacles": ["A"],
      "devNote": "This doesn't kill the Beetoms, but clears them as an obstacle."
```

region/brinstar/green/Green Brinstar Beetom Room.json
```diff
@@ -234,7 +231,8 @@
...
```
kjbranch, line 235:
```suggestion
      "clearsObstacles": ["A"],
      "devNote": "This doesn't kill the Beetoms, but clears them as an obstacle."
```

region/brinstar/green/Green Brinstar Beetom Room.json
blkerby: Yeah, I hesitated to add it, because we have to think about all the ways that a runway can be used (e.g. not just for gaining a shinecharge, but also for door stuck setups, bomb boosts, etc.). In this case I think it's fine though; the various ways to use the runway will probably have a similar number of Beetom hits; using it to start a short-charge will probably be the worst case so we could just look at that. In the future, if we need to we could refine the `leaveWithRunway` to allow expressing that only certain ways of using the runway are allowed.

blkerby: Added now.

PR #1138, opened 2023-11-06
Pants Rooms Restructure, Runway Migration & Connection Cleanup

kjbranch:
Coming from either door in the "left leg" makes it so the door to Shaktool's room is not submerged. Coming from either door in the "right leg" will make it so the door is submerged.

This change in the definitions of the two rooms' nodes will split them based on this feature, and allow us to easily represent how the runway to Shaktool's room behaves.

Minimal connection cleanup:
- Removed `InRoomTransition` connection type from the connections, as this was only used in this room. 
- Removed the `Backward` connection type, as this can always be represented by a `Forward` connection type, and fixed its single random use.


osse101 (commented):
I like this room restructure.  It looks a lot cleaner.

I guess connections were listed in order by roomID, but it doesn't seem important to be done that way 🤷

PR #1147, opened 2023-11-15
East Sand Hole Sandfall Bounce Boulder Description

kjbranch:
I felt like this could be useful, as it's hard to avoid the Boulder. But maybe this description is not need at this difficulty?


osse101:
At this level, you are aware there are boulders up here, but not where exactly.  Like when I had to do this trick recently.  Because its rare to take this path to the item.  And if the rock falls on you, you'll remember.  So I think a note just needs to say jumping out of the water triggers the boulder, if that is how it works.

The Underwater walljump strat would have the same note.
The shinespark strat needs tips the most, I suppose.
And it doesn't seem important if you have gravity.

PR #1149, opened 2023-11-17
Add 10 PB CF and Ammo Refill Helpers

kjbranch:


blkerby:
The PR looks good to me. 

Strats using the new helper `h_can10PowerBombCrystalFlash` can later be added to other rooms, such as Kronic Boost Room, where farming energy may not be possible but where there are enemies that drops power bombs. 

Outside the scope of this PR -- general Crystal Flash strats (using 11 Power Bombs) could be handled implicitly in most cases, as another application of runways, since if a runway is clear for use in bomb boosts, X-ray climb setups, etc., then it should also be usable for a Crystal Flash. I can make a proposal for this soon.

PR #1153, opened 2023-11-18
Update Green Maridia Room Diagrams

### LINKED ISSUE #744: Room Diagram Colors
Should we add new colors or line types for different types of nodes and links?
- Nodes: 
  - Utility, 
  - Event
- Links: 
  - SpawnAt,
  - Possibly cross room strats that bypass junctions: x-ray climb, shinespark, cross room jump?
  - Are G-Mode links still worth neglecting?
  
What would we do if there are multiple types of links that go between two nodes (i.e. a SpawnAt and normal link)?

kjbranch:
With the cross room migration, room diagrams are drifting from the logic. I'm looking to start fixing them when I want a more simple task to work on. So, I'm looking for feedback on this style, which was slightly discussed in #1123 and looks to resolve #744.

The idea here is:
- 1 line will exist between a pair of nodes, prioritizing the type closer to the top of the next list. If the two directions need to be different colors, it will instead exist as two arrows.
  - Red: spawn at
  - Yellow: normal strats
  - Blue: cross room
  - Purple: g-mode
  
Unfortunately the purple doesn't work well in this room, but I think that's ok ... Few people will be looking for the g-mode arrows, and they will look fine in almost every room. I could redraw this room where 4 is smaller and move the purple line inward, but that's more work and doesn't feel very impactful.


osse101:
There would also be 1>2 and 2->1 cross room lines for entering and leaving in various states of Shinecharge.

kjbranch:
Oh, I forgot to mention those. I was thinking of leaving out strats that come in and leave. Do you think they should be included? I imagine some rooms it will be chaotic (i.e. Business Center). If we do keep them, do you think they should be a different color to not confuse them with other cross room strats (i.e. dark blue)?

osse101:
I haven't liked them since it can really clutter a room.  Its "proper" to include them until there are enough comeIn/Leave combinations that they lose meaning.  Such as a canChainTemporaryBlue from any door to any other door.  So yeah, they probably shouldn't be shown.

kjbranch:
Finished off Green Maridia, except for a single g-mode arrow that's going to need to move after the gmode migration

PR #1154, opened 2023-11-19
Schema change for migrating door-lock bypass strats

blkerby:
As part of the cross-room strat migration, this PR makes the necessary schema changes to allow converting "bypassStrats" (in of `node/locks`) into regular strats. This also allows more types of door-lock bypass strats to be expressed, since a bypass strat could now start at any node, not necessarily only at the door node that it exits through. A couple examples are included: one for migrating the Butterfly Room gray lock bypass strat, and another for a Very Deep Stuck X-Ray climb in Red Tower that wouldn't have been expressible before.

At some point we will also want to migrate "unlockStrats", e.g. to help us more accurately represent requirements to unlock doors where heat, lava, acid, and/or enemies are a factor; this is more complicated, because it can depend on the door color, meaning multiple possibilities might need to be considered since we want the logic to be able to work with randomized doors.

**Note**: Map Rando never made use of the `bypassStrats` in `node/locks`; as we migrate these to regular strats, it will start using them.

### DIFF
--- a/region/brinstar/red/Red Tower.json
+++ b/region/brinstar/red/Red Tower.json
@@ -604,6 +604,28 @@
       "note": "This is a short climb of only a few tiles."
     },
     {
+      "link": [2, 1],
+      "name": "Very Deep Stuck X-Ray Climb",
+      "notable": false,
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "direct",
+          "morphed": false
+        }
+      },
+      "requires": [
+        "canXRayClimb"
+      ],
+      "bypassesDoorShell": true,
+      "note": [
+        "Enter with G-mode direct, back up to between 1 and 6 pixels from the door transition, and activate X-ray to get very deep stuck in the door.",
+        "Climb up 2 screens.",
+        "When you reach the door, perform a turnaround buffered spin-jump away from the door, and the wall-jump check will trigger the transition, bypassing any lock on the door.",
+        "If you position 1 pixel from the door transition (not recommended), at the start of the climb you will need to use X-ray when turning around toward the door, to avoid triggering the transition back through the door you came in;",
+        "in this case when you reach the destination door, there is no need for a spin-jump, as simply turning around toward the door will trigger the transition."
+      ]
+    },
+    {
       "link": [2, 2],
       "name": "Leave With Runway",
       "notable": false,
--- a/strats.md
+++ b/strats.md
@@ -6,13 +6,15 @@ Strats are usually presented within an array of strats, where all individual str
 ## Structure
 
 A `strat` can have the following properties:
-  * _name:_ The name of the strat.
-  * _notable:_ Indicates whether the strat is notable.
-  * _reusableRoomwideNotable:_ The name of the reusable roomwide notable strat. This must share an identical name with an entry in the `reusableRoomwideNotable` array and the other strats that are connected to this one. This is only applicable for strats where `notable` is `true`.
+  * _name_: The name of the strat.
+  * _notable_: Indicates whether the strat is notable.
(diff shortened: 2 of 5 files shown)


kjbranch (approved):

strats.md
```diff
@@ -6,13 +6,15 @@ Strats are usually presented within an array of strats, where all individual str
...
   * _entranceCondition_: Indicates that this strat requires entering through the door transition in a special way, combining with a strat in the previous room.
-  * _requires:_ The [logical requirements](logicalRequirements.md) that must be fulfilled to execute that strat.
+  * _requires_: The [logical requirements](logicalRequirements.md) that must be fulfilled to execute that strat.
   * _exitCondition_: Indicates that this strat leaves through the door transition in a special way that combines with a strat in the next room. 
-  * _clearsObstacles:_ An array containing the ID of obstacles that will be cleared by executing this strat (if they are not already cleared).
+  * _clearsObstacles_: An array containing the ID of obstacles that will be cleared by executing this strat (if they are not already cleared).
+  * _gModeRegainMobility_: Indicates that this strat allows regaining mobility when entering with G-mode immobile.
```
osse101, line 16: sus

region/brinstar/red/Red Tower.json
```diff
@@ -603,6 +603,28 @@
...
+          "mode": "direct",
+          "morphed": false
+        }
+      },
+      "requires": [
+        "canXRayClimb"
+      ],
+      "bypassesDoorShell": true,
+      "note": [
+        "Enter with G-mode direct, back up to between 1 and 6 pixels from the door transition, and activate X-ray to get very deep stuck in the door.",
```
osse101, line 621: A long note for what's likely to be a common strat.

osse101 (commented):
Looks good.

Adding bypass strats for all the doors will be a ton of work.

Bypass in GMode can be used to bring GMode into another room and this structure helps realize that.  Depends on the GMode restructure for how to really continue, but I guess we're looking at a GMode versions of bypass strats that would say that they are bringing GMode indirect into the next room.

osse101:
So I forgot GMode schema migration already happened.

It should be called out that GMode doors don't open but a bypass would be possible on those doors still.

And it would just look something like
```
EntranceCondition: GMode: any
ExitCondition: GMode
bypass: true
```
Examples would be ice clips up through a door, long moonfalls or enemyStuckMoonfalls, the colliseum grapple clip seems barely possible.  Maybe there aren't very many.  Mostly ice clips.

blkerby:
> Bypass in GMode can be used to bring GMode into another room and this structure helps realize that.

Yeah, this is exactly what I had in mind too, with allowing that "bypassesDoorShell" can be combined with an exit condition. I'm also wondering if it might be possible with a Very Deep Stuck X-ray climb to bypass a locked door while setting up door-stuck for X-ray climbing in the next room; I tried for a bit and didn't get it, but I expect it could be done.

strats.md
blkerby: Haha, yeah just noticed that was missing in the overview here, so went ahead and added it

region/brinstar/red/Red Tower.json
blkerby: Yeah it would be fine to be shortened. Mostly I was thinking we might want to create a tech for it, and the description could be moved there. I wasn't sure because this would be a pretty easy tech for Extreme (aside from the dependence on G-mode), but there is something extra in just knowing that it's possible to do, so having a new tech would be a way to highlight that. On the other hand, G-mode in general involves a lot of knowledge about how it can be used, so that would be an argument for not needing another tech here. In any case, figured a new tech would be a separate issue from this PR

osse101: Yep, for this PR its fine as is.

GMode very deep stuck is mostly a knowledge check, that should be explained someplace.  And as a knowledge check it could be a tech.  But its really not hard or complex compared to GMode itself so its kinda weird.

PR #1157, opened 2023-11-19
Create script to mostly automate the G-mode cross-room migration

blkerby:
This script will not migrate strats with "comeInWithGMode" in a few cases:
- if there are multiple in the same strat
- if there are multiple fromNodes
- if it is not at the top level of "requires" (e.g. inside an "or")

Those cases will generally need to be manually migrated, splitting the strat into multiple strats.

Aside from that, everything should be covered: "leaveWithGModeSetup", "leaveWithGMode", "gModeImmobile", and occurrences of "comeInWithGMode" other than those exceptional cases.

This PR also makes "morphed" be a required property of "leaveWithGMode"; it was required in the old schema, and I think we want to keep it that way. Often "leaveWithGMode" strats should come in pairs, one with "morphed" false and one with "morphed" true, and it would be easy to make a mistake by forgetting about this property.

### DIFF
--- a/schema/m3-room.schema.json
+++ b/schema/m3-room.schema.json
@@ -500,7 +500,7 @@
               "type": "object",
               "title": "Leave With G-Mode",
               "description": "Represents that Samus can leave through this door while in G-mode, resulting in indirect G-mode in the next room.",
-              "required": [],
+              "required": ["morphed"],
               "additionalProperties": false,
               "properties": {
                 "morphed": {
--- a/scripts/gmode_migration.py
+++ b/scripts/gmode_migration.py
@@ -1,6 +1,7 @@
 import json
 from pathlib import Path
 import copy
+import argparse
 
 import format_json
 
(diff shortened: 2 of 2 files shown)


kjbranch:
I wouldn't mind taking some extra time on this and somewhat thoroughly review the output, while thinking on if there are other strats that were not included because of the limitations of the old schema, and manually migrating the strats that weren't able to be included with this script. Would it be possible for this to be run on a smaller region at a time because of this (like one of the 6 regions at a time) and for you to teach me how to run it?

blkerby:
Sounds good. Thoroughly reviewing the output seems wise. I added a "path" argument now so you could run it on on a region or subregion at a time, e.g.:

```
python gmode_migration.py  ../region/crateria/central
```

osse101 (approved):
Looks good

I like how it fixed some misordered links.

I like the plan of migrating with the script in smaller chunks.

PR #1162, opened 2023-11-21
Rest of Green Brin Runway migration

blkerby:
There is opportunity to add more leave charged/sparking strats, e.g. in Green Hill Zone for the top doors (and a ton of 3-room strats in Green Brinstar Main Shaft). This PR is mostly just trying to migrate the existing ones

### DIFF
--- a/region/brinstar/green/Noob Bridge aka A Bridge Too Far.json
+++ b/region/brinstar/green/Noob Bridge aka A Bridge Too Far.json
@@ -151,7 +151,7 @@
       ],
       "exitCondition": {
         "leaveShinecharged": {
-          "framesRemaining": 120
+          "framesRemaining": 150
         }
       }
     },
--- a/region/brinstar/green/Noob Bridge aka A Bridge Too Far.json
+++ b/region/brinstar/green/Noob Bridge aka A Bridge Too Far.json
@@ -15,20 +15,6 @@
       "nodeSubType": "blue",
       "nodeAddress": "0x0018efe",
       "doorEnvironments": [{"physics": "air"}],
-      "runways": [
-        {
-          "name": "Base Runway - Noob Bridge Left Door (to Green Hill Zone)",
-          "length": 0,
-          "strats": [
-            {
-              "name": "Base",
-              "notable": false,
-              "requires": []
-            }
-          ],
-          "openEnd": 1
-        }
-      ],
       "leaveWithGModeSetup": [
         {"strats": [
           {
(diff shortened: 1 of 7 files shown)


region/brinstar/green/Noob Bridge aka A Bridge Too Far.json
```diff
@@ -195,11 +120,41 @@
...
+      "requires": [
+        {"canShineCharge": {
+          "usedTiles": 33,
+          "openEnd": 2
+        }},
+        "canShinechargeMovement"
+      ],
+      "exitCondition": {
+        "leaveShinecharged": {
+          "framesRemaining": 120
```
osse101, line 154: At full run I was seeing ~140, but when shortcharging I could get 150.
So there is a full run vs shortcharge difference, and that would be something to cover with a `h_shinechargeLeniency` in the future.

region/brinstar/green/Spore Spawn Kihunter Room.json
```diff
@@ -166,11 +120,41 @@
...
+        {"id": 2}
       ]
     }
   ],
   "strats": [
+    {
+      "link": [1, 1],
+      "name": "Leave With Runway",
+      "notable": false,
+      "requires": [],
```
osse101, line 133: This requires killing the left kihunter so it should have some basic requirement.

blkerby: Done.

region/brinstar/green/Noob Bridge aka A Bridge Too Far.json
blkerby: There are a lot of other factors too that could justify shinecharge leniency:
- tendency for less experienced players to idle for a moment after getting the shinecharge, before beginning moving
- lack of knowledge/ability of how to ledge-grab, down-back, or other movement techniques
- lack of knowledge/ability to quickly hone in accurately on the place where the spark should be activated, e.g. by stopping on a dime or carefully controlling a jump.

I think we could treat this kind of like heat frames, where we put tight values into the strats, and based on settings, the randomizer assumes you need extra frames available for leniency, e.g. maybe something like this:

Basic: 60 frames
Medium: 30 frames
Hard: 20 frames
Very Hard: 10 frames
Expert: 5 frames
Extreme: 2 frames
Insane: 0 frames

In this case, the leniency wouldn't need to be explicitly written into the strats.

osse101 (approved):

PR #1163, opened 2023-11-21
Make notable false by default, create script to remove notable false

blkerby:
This changes the room schema to make "notable" an optional property which only needs to be specified if it is "true". A script is created which can then strip out `"notable": false` from all the strats.

### DIFF
--- a/schema/m3-room.schema.json
+++ b/schema/m3-room.schema.json
@@ -56,7 +56,6 @@
       "type": "object",
       "required": [
         "name",
-        "notable",
         "requires"
       ],
       "additionalProperties": false,
@@ -64,15 +63,11 @@
         {
           "if": {
             "properties": {
-              "notable": {"const": false}
-            }
+              "notable": {"const": true}
+            },
+            "required": ["notable"]
           },
           "then": {
-            "not": {
-              "required": ["reusableRoomwideNotable"]
-            }
-          },
-          "else": {
             "required": ["note"]
           }
         }
@@ -104,7 +99,8 @@
         },
         "notable": {
           "$id": "#/definitions/strat/properties/notable",
-          "type": "boolean"
+          "type": "boolean",
+          "default": false
         },
         "reusableRoomwideNotable": {
           "$id": "#/definitions/strat/properties/reusableRoomwideNotable",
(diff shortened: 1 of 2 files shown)


schema/m3-room.schema.json
```diff
@@ -56,23 +56,18 @@
...
       "additionalProperties": false,
       "allOf": [
         {
           "if": {
             "properties": {
-              "notable": {"const": false}
-            }
+              "notable": {"const": true}
+            },
+            "required": ["notable"]
```
osse101, line 68: Does the if/then/else not work with a default value notable?  This loses `reusableRoomwideNotable` requiring notable to be true.

And I don't think `required: notable` makes sense here.

blkerby: Ok, I was confused about what was going on with the `reusableRoomwideNotable`. Fixed it now. The `"required": ["notable"]` is necessary in order for it to work, otherwise it will treat cases with unspecified `notable` property as if they were true. It's very confusing, but basically the "const" condition will always evaluate as true if there is nothing there that it would apply to. See the docs here: https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse.

I think as it turns out, it could also have worked to have left this the way it was before. I was concerned about how it would handle the unspecified `notable` values, but it turns out it would work since `{"const": false}` evaluating to true for a missing notable is what you would want in this case. It wouldn't not taking the default value into account, it's just a coincidence that it would happen to do the right thing here. So I think it's more clear to rework it like this, basing it on a positive condition of `notable` rather than negative.

blkerby: And oops, I committed a version that reverted to the old form of the validation here (which would have also worked, but wasn't intended).

For me the key thing that made this make more sense is to realize that `required` does not really mean "required" here; it means to test for if a property is present or not, evaluating to true if it, and false if it is not.

osse101: Haha good to have that cleared up. 
It looks better this way.

osse101 (approved):
Looks good, works good.

It doesn't edit strats that are attached to nodes so when using the GMode migration script any `notable: false` will have to be removed, by the script or otherwise.

PR #1164, opened 2023-11-21
Runway Migration: Finishing off Pink Maridia

kjbranch:
This should be the last of the Pink Maridia migration, minus a few strats here and there which I couldn't replicate.

### DIFF
--- a/region/maridia/inner-pink/Halfie Climb Room.json
+++ b/region/maridia/inner-pink/Halfie Climb Room.json
@@ -477,10 +477,6 @@
       },
       "requires": [
         "canWaterShineCharge",
-        {"canShineCharge": {
-          "usedTiles": 33,
-          "openEnd": 2
-        }},
         {"shinespark": {"frames": 20, "excessFrames": 3}}
       ],
       "devNote": "This runway length is just an estimate of the worst case scenario for difficulties that can't use canStutterWaterShineCharge."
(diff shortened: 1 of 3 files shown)


region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -502,57 +453,50 @@
...
-          "fromNode": 2,
-          "usedTiles": 0.5,
-          "inRoomPath": [2, 1],
-          "physics": ["air"]
-        }}
+        "canWaterShineCharge",
+        {"canShineCharge": {
+          "usedTiles": 33,
+          "openEnd": 2
+        }},
```
osse101, line 483: What does this mean here? 33+2 is going to be always true.
Then there is `comeInShinecharging`  which does the real comparison with your usedTiles count.

kjbranch: Looks like it doesnt mean anything here - I'll delete it.

Sometimes i think it's used to show something like `canShinecharge`, which might be better to add. Or simply adding `SpeedBooster`. It's used all over the place, like in that X-Mode strat you were reviewing above.

blkerby: The `canShineCharge` requirement implicitly includes a SpeedBooster item requirement. So as long as the SpeedBooster requirement is covered (e.g. in this case by the `canWaterShineCharge`), I think it would be fine to start dropping these 33+2 canShineCharge requirements.

Issue #586, opened 2023-03-02
Underwater shinespark health cost


osse101:
Shinesparking underwater (or in acid) has a different health cost if done with or without gravity suit due to a difference in acceleration.
This only applies to horizontal movement or the horizontal component of diagonal shinesparks.
https://wiki.supermetroid.run/Horizontal_Speed

No gravity movement will be slower and therefore cost more shinesparkframes and health.

PR #1166, opened 2023-11-22
Runway Migration: Mt. Everest

kjbranch:
During the cross-room migration, nearly all of #586 has been completed, and the rest will be soon. This was the primary remaining room with impactful differences (long horizontal sparks), so I think it can be closed as complete after this PR is merged.


region/maridia/outer/Mt. Everest.json
```diff
@@ -1699,6 +1658,65 @@
...
+      "link": [5, 4],
+      "name": "Midair Shinespark",
+      "notable": false,
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 55
+        }
+      },
+      "requires": [
+        "canSuitlessMaridia",
```
osse101, line 1698: I think Gravity is missing here.

And there could be shinesparks which do not need to reach the grapple blocks first, in both directions.

kjbranch: Good catch. 

I'll look to add those strats next I have time. I also should also add strats coming in shinecharging from the right with gravity, and sparking to the grapple blocks with a down grab from both sides (Probably just suitless sparks as it's easier to gravity jump up there if you have it).

PR #1169, opened 2023-11-23
Brinstar Crystal Flash strats

blkerby:


region/brinstar/green/Green Brinstar Beetom Room.json
```diff
@@ -140,6 +140,36 @@
...
+        ]},
+        "h_canCrystalFlash"
+      ]
+    },
+    {
+      "link": [1, 1],
+      "name": "10 Power Bomb Crystal Flash",
+      "requires": [
+        {"or": [
+          {"obstaclesCleared": ["A"]},
```
osse101, line 163: Parts of clearing A can work, like using Ice.  But you need the Beetoms to drop PBs.

`can10PowerBombCrystalFlash` is important to not be in health bomb so we'll need a way to check for that.

region/brinstar/kraid/Warehouse Energy Tank Room.json
```diff
@@ -124,6 +124,23 @@
         }
       }
     },
+    {
+      "link": [1, 1],
+      "name": "Crystal Flash",
+      "requires": ["h_canCrystalFlash"]
```
osse101, line 130: There is a `enter in ball` way to avoid damage, but otherwise a Beetom attaches instantly.

region/brinstar/pink/Big Pink.json
```diff
@@ -1195,6 +1202,11 @@
       ],
       "note": "If using Bombs, lure and kill the Reo. If using Spring Ball, it's important to release jump before reaching to peak of each jump to outrun the Reo."
     },
+    {
+      "link": [7, 7],
+      "name": "Crystal Flash",
+      "requires": ["h_canCrystalFlash"]
```
osse101, line 1208: 7 can farm at 13 pretty quick.
CFing at 9, and also 1,  are the same situation as 7, but I think this strat can just go.

region/brinstar/pink/Dachora Room.json
```diff
@@ -231,6 +233,11 @@
         {"refill": ["Energy", "Missile", "Super"]}
       ]
     },
+    {
+      "link": [1, 1],
+      "name": "Crystal Flash",
+      "requires": ["h_canCrystalFlash"]
```
osse101, line 239:
```suggestion
      "requires": ["h_canCrystalFlash"],
      "clearsObstacles": ["A"]
```

region/brinstar/pink/Dachora Room.json
```diff
@@ -493,6 +505,11 @@
         }}
       ],
       "note": "Use The camera scroll blocks above the speed blocks and by the bomb wall to overload PLMs."
+    },
+    {
+      "link": [3, 3],
+      "name": "Crystal Flash",
+      "requires": ["h_canCrystalFlash"]
```
osse101, line 512:
```suggestion
      "requires": ["h_canCrystalFlash"],
      "clearsObstacles": ["A"]
```

region/brinstar/pink/Pink Brinstar Hopper Room.json
```diff
@@ -522,6 +530,11 @@
       ],
       "clearsObstacles": ["A"],
       "note": "With access to the right side of the room, it's possible to kill the enemies safely with Wave or a PB or screw attack once on the other side."
+    },
+    {
+      "link": [2, 2],
+      "name": "Crystal Flash",
+      "requires": ["h_canCrystalFlash"]
```
osse101, line 537: This gets a partial clear of A

region/brinstar/red/Beta Power Bomb Room.json
```diff
@@ -134,6 +134,14 @@
         "If any Sidehoppers remain after using a Power Bomb, safely kill them from the hole below."
       ]
     },
+    {
+      "link": [1, 1],
+      "name": "Crystal Flash",
+      "requires": [
+        {"obstaclesCleared": ["A"]},
+        "h_canCrystalFlash"
+      ]
```
osse101, line 143:
```suggestion
      ],
      "clearsObstacles": ["B"]
```

region/brinstar/red/Red Tower.json
```diff
@@ -637,6 +638,11 @@
         }
       }
     },
+    {
+      "link": [2, 2],
+      "name": "Crystal Flash",
+      "requires": ["h_canCrystalFlash"]
```
osse101, line 644:
```suggestion
      "requires": ["h_canCrystalFlash"],
      "clearsObstacles": ["A"]
```

osse101 (commented):
I don't think damaging down is supported yet.  Is the thinking that you can probably find a room to lower health below 50?  Low damage enemies can be used to set health to some low number, and heat rooms as well.  CF logic seems complicated.

kjbranch:
Yeah, that is the assumption. It's basically impossible to get to CF ammo before finding a low hitting enemy or a heated room. 

The main problem I see is when you get cornered and have to use a CF to refill on the escape, but there are only high-hitting enemies. It still seems pretty unlikely to be a problem, as you can probably lower your energy a bit before getting cornered (unless you're grabbing a tank in there), and you can probably slow-farm or something if necessary.

blkerby:
Yeah, in theory we would want to model places where you can damage down; ever if we did, though, it would not be straightforward to use that information in the randomizer. In practice (from the CF clips already in logic) I don't think we've ever seen a case yet where you want to CF but have no way to damage down. I think it may be ok to postpone worrying about this until we see it becoming a problem (if that ever happens).

region/brinstar/green/Green Brinstar Beetom Room.json
blkerby: If you're in health bomb range, it should be possible to farm a few times to get above 50 energy. You'll have 10 PBs which should be more than enough. I'm assuming that you can reset the room here. I can add that as a requirement.

region/brinstar/kraid/Warehouse Energy Tank Room.json
blkerby: Jumping into the room is also an option. So I think we can say either `canPrepareForNextRoom` or take 2 hits.

osse101 (approved):

PR #1173, opened 2023-11-24
Tourian Crystal Flash strats

blkerby:
For Tourian Blue Hopper Room, I was experimenting for a while with a strat to lure both Hoppers to the right when entering from the top, to be able to do a Crystal Flash safely on the left. It works a lot of the time but I wasn't able to find a consistent setup yet so I didn't add that one. The one entering from the left though seems very reliable.

### DIFF
--- a/region/tourian/main/Blue Hopper Room.json
+++ b/region/tourian/main/Blue Hopper Room.json
@@ -328,6 +328,10 @@
           "nodes": [1],
           "mustStayPut": false
         }},
+        {"resetRoom": {
+          "nodes": [2],
+          "mustStayPut": false
+        }},
         "canTrickyJump",
         "h_canCrystalFlash"
       ],
@@ -336,7 +340,8 @@
         "After 6 hops, the bottom Hopper will do three big hops in a row.",
         "Roll under the Hopper while it does the third of those three big hops.",
         "Quickly unmorph and run to the right side of the room.",
-        "If successful, both Hoppers will remain off-camera, so you can safely Crystal Flash."
+        "If successful, both Hoppers will remain off-camera, so you can safely Crystal Flash.",
+        "Reset the room through the top door before returning to the left."
       ]
     }
   ]
--- a/region/tourian/main/Blue Hopper Room.json
+++ b/region/tourian/main/Blue Hopper Room.json
@@ -319,6 +319,25 @@
         }
       },
       "devNote": "FIXME: Add requirements to be able to use this runway."
+    },
+    {
+      "link": [2, 2],
+      "name": "Crystal Flash",
+      "requires": [
+        {"resetRoom": {
+          "nodes": [1],
+          "mustStayPut": false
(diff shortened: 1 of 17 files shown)


region/tourian/main/Blue Hopper Room.json
```diff
@@ -319,6 +319,25 @@
...
+      "requires": [
+        {"resetRoom": {
+          "nodes": [1],
+          "mustStayPut": false
+        }},
+        "canTrickyJump",
+        "h_canCrystalFlash"
+      ],
+      "note": [
+        "Enter the room from the left door, morph, and wait at the left wall.",
```
osse101, line 335: This is "Blue Hopper Room Roll Under Hoppers", and I'd never been able to reliably get to the right side without one of the hoppers chasing me.  Especially the ceiling one because I'm timing the movement against the bottom one.

Another problem is that you can't reach the door at 2 after CFing.

There are some kill strats that would be safe, if you can kill the bottom guy mostly.  But that's not saved as an obstacle.

region/tourian/main/Metroid Room 1.json
```diff
@@ -422,6 +427,15 @@
...
       }
     },
+    {
+      "link": [2, 2],
+      "name": "Crystal Flash",
+      "requires": [
+        {"obstaclesCleared": ["A"]},
+        "h_canCrystalFlash"
+      ],
+      "note": ["Be sure to be at a safe distance from Rinkas before performing the Crystal Flash."]
```
osse101, line 437: Node 2 looks like the free one.
Node 1 has a rinka at the door.

region/tourian/main/Tourian Escape Room 4.json
```diff
@@ -431,6 +437,12 @@
       ],
       "clearsObstacles": ["A"],
       "devNote": "IBJ can be used to avoid wall jumps before the acid starts rising, but a crouch jump down grab is needed near the right door."
+    },
+    {
+      "link": [3, 3],
+      "name": "Crystal Flash",
+      "requires": ["h_canCrystalFlash"],
+      "devNote": "The Crystal Flash should be performed on a platform away from the acid trigger or any Pirates."
```
osse101, line 445: This should work.  CF before triggering the acid and entering the box that is Node 3.
[3, 2] adds a trivial movement requirement whereas [2, 2] would slowly kill the pirates.

region/tourian/main/Metroid Room 1.json
blkerby: Right, fixed now.

region/tourian/main/Blue Hopper Room.json
blkerby: Added a resetRoom requirement to be able to reset the room using the top door after the CF, because you're right that going back left immediately doesn't seem to work, at least not consistently (or without some additional requirements).

PR #1175, opened 2023-11-25
Rest of Upper Norfair Crystal Flash strats

blkerby:
It looks like I slightly underestimated the heatFrames needed in the `h_canHeatedCrystalFlash` before, probably had a lucky RNG on the heat cycle. Bumping it up from 180 to 185.

### DIFF
--- a/region/norfair/east/Rising Tide.json
+++ b/region/norfair/east/Rising Tide.json
@@ -150,6 +150,12 @@
       "link": [1, 2],
       "name": "Roll Under",
       "notable": false,
+      "entranceCondition": {
+        "comeInRunning": {
+          "minTiles": 0,
+          "speedBooster": "any"
+        }
+      },
       "requires": [
         "h_canNavigateHeatRooms",
         "Morph",
@@ -347,6 +353,12 @@
       "link": [2, 1],
       "name": "Roll Under",
       "notable": false,
+      "entranceCondition": {
+        "comeInRunning": {
+          "minTiles": 0,
+          "speedBooster": "any"
+        }
+      },
       "requires": [
         "h_canNavigateHeatRooms",
         "Gravity",
@@ -368,6 +380,12 @@
       "link": [2, 1],
       "name": "Tricky Roll Under",
       "notable": false,
+      "entranceCondition": {
+        "comeInRunning": {
+          "minTiles": 0,
+          "speedBooster": "any"
+        }
+      },
       "requires": [
         "h_canNavigateHeatRooms",
         "Gravity",
--- a/region/norfair/east/Rising Tide.json
+++ b/region/norfair/east/Rising Tide.json
@@ -101,6 +101,11 @@
       ]
     },
     {
+      "link": [1, 1],
+      "name": "Crystal Flash",
+      "requires": ["h_canHeatedCrystalFlash"]
+    },
+    {
       "link": [1, 2],
       "name": "SpaceJump",
       "notable": false,
@@ -389,6 +394,11 @@
           "openEnd": 1
         }
       }
+    },
+    {
+      "link": [2, 2],
+      "name": "Crystal Flash",
+      "requires": ["h_canHeatedCrystalFlash"]
     }
   ]
 }
--- a/region/norfair/west/Ice Beam Gate Room.json
+++ b/region/norfair/west/Ice Beam Gate Room.json
@@ -457,7 +457,7 @@
     {
       "link": [4, 4],
       "name": "Crystal Flash",
-      "requires": ["h_canHeatedCrystalFlash"]
+      "requires": ["h_canCrystalFlash"]
     },
     {
       "link": [5, 2],
--- a/region/norfair/west/Ice Beam Gate Room.json
+++ b/region/norfair/west/Ice Beam Gate Room.json
@@ -157,6 +157,11 @@
       }
     },
     {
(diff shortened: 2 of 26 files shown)


region/norfair/east/Rising Tide.json
```diff
@@ -100,6 +100,11 @@
         {"refill": ["Energy", "Missile", "Super"]}
       ]
     },
+    {
+      "link": [1, 1],
+      "name": "Crystal Flash",
+      "requires": ["h_canHeatedCrystalFlash"]
```
osse101, line 106: This breaks the roll under strat 🤔

region/norfair/west/Ice Beam Gate Room.json
```diff
@@ -392,6 +403,12 @@
         {"refill": ["Energy", "Missile", "Super", "PowerBomb"]}
       ]
     },
+    {
+      "link": [3, 3],
+      "name": "Crystal Flash",
+      "requires": ["h_canCrystalFlash"],
```
osse101, line 409: 3 is safe only if you spend some ammo then farm it back

region/norfair/west/Ice Beam Snake Room.json
```diff
@@ -542,6 +564,11 @@
         "h_canNavigateHeatRooms",
         {"heatFrames": 50}
       ]
+    },
+    {
+      "link": [5, 5],
```
osse101, line 569: I don't think 5 adds anything

region/norfair/west/Ice Beam Snake Room.json
```diff
@@ -284,6 +296,11 @@
         }
       }
     },
+    {
+      "link": [3, 3],
+      "name": "Crystal Flash",
+      "requires": ["h_canHeatedCrystalFlash"]
```
osse101, line 302: 3 has to avoid the Fune shot

osse101 (commented):
Some rooms could have a node added to CF for being long and heated.

Single Chamber between 5 and 6
Spiky Platforms Tunnel between 1 and 2

blkerby:
> Some rooms could have a node added to CF for being long and heated.
> 
> Single Chamber between 5 and 6 Spiky Platforms Tunnel between 1 and 2

For sure, there are a lot of cases where more nodes could help: Rising Tide, Kronic Boost, and Lava Dive seem like other examples. For this first pass now I'm just focusing on adding strats at the existing nodes, and figure it can be refined later with additional nodes

region/norfair/east/Rising Tide.json
blkerby: Ok, added entrance conditions to the "Roll Under" strats, which prevents a CF strat from happening in the middle. I used 
a comeInRunning with 0 tiles. Is that accurate, or do these actually need some speed from the previous room?

region/norfair/west/Ice Beam Gate Room.json
blkerby: True, added an enemyKill requirement to this now.

region/norfair/west/Ice Beam Snake Room.json
blkerby: So I just realized the CF at node 1 doesn't work, because of the shot blocks. So I removed that one, and in this case I think it makes the one at node 5 more useful.

region/norfair/west/Ice Beam Snake Room.json
blkerby: True, added requirements for that now.

region/norfair/east/Rising Tide.json
osse101: It was only a problem for this strat, because the lava is rising.  `comeInRunning: 0` is a good way to write it, there is a different strat that would come in with speed for a mockball.

region/norfair/west/Ice Beam Snake Room.json
osse101: I was just pretending that you jump off the shot blocks.

The Fune fireball  can hit you as you rise up with the CF starting to heal, and that is deadly suitless.

region/norfair/east/Rising Tide.json
blkerby: Ok, `comeInRunning` will also require the other room to have normal physics (air or Gravity). So we may want to refine this at some point, e.g. by adding a new entrance condition that doesn't care about the neighboring room's physics, or add a property to the existing entrance conditions to override implicit physics requirements.

region/norfair/west/Ice Beam Snake Room.json
blkerby: Jumping off the shot blocks would need some more heat frames before and after the CF (when entering through node 1). Seems like using 5 is more practical?

Added requirements to kill the Fune before doing the CF.

osse101: There are small pros and cons to using 1 or 5.  I don't know how likely it is that you want to enter a heated room to CF and then leave through the same door.

But there is a Fune across from 5.  And the fireball reaches me during the CF.  If I don't try to delay it.

blkerby: Oh, missed the Fune there, and forgot those shots go through walls. In that case, seems more useful to scrap the node 5 strat in favor of node 1. Added extra heat frames to position at node 1.

PR #1180, opened 2023-11-27
Runway Migration: Fish Tank

kjbranch:


region/maridia/outer/Fish Tank.json
```diff
@@ -349,6 +329,12 @@
       "link": [1, 5],
       "name": "Cross Room Jump to the Frozen Fish",
       "notable": false,
+      "entranceCondition": {
+        "comeInJumping": {
+          "speedBooster": true,
+          "minTiles": 3.4375
```
kjbranch, line 335: This is an extra ~1 tile compared to what was in there before, but I couldn't get past the ledge with anything less. If it's possible, it may need to be notable. (same for the spring ball variant)
Most of the other cross room jumps from this door I was able to tighten a bit.

kjbranch: @blkerby mentioned to me that it is helpful to be in a downward pose before hitting the turnaround. This isn't how I've been testing them, so there are probably some that should be changed (fortunately most of the runway lengths were based on bobbob's videos, so those should be correct).

region/maridia/outer/Fish Tank.json
```diff
@@ -693,6 +672,22 @@
       "notable": false,
       "requires": []
     },
+    {
+      "link": [4, 5],
+      "name": "Shinespark",
+      "notable": false,
+      "entranceCondition": {
```
osse101, line 679: Is it a probalem that this is a spawnAt?

region/maridia/outer/Fish Tank.json
kjbranch: I went back through and it looks like this shouldnt change the number of tiles for any of the other strats that I migrated

region/maridia/outer/Fish Tank.json
```diff
@@ -360,15 +346,10 @@
...
-        {"adjacentRunway": {
-          "fromNode": 1,
-          "usedTiles": 1.5,
-          "inRoomPath": [1, 5],
-          "physics": ["air"]
-        }}
+        ]}
       ],
       "note": [
+        "Requires 4 tiles of run speed (with no open end) to make it past the overhang above the door.",
```
osse101, line 352: The note wasn't updated, here and below.  
I needed 3 tiles no open end.  Any less wasn't enough height or distance.

region/maridia/outer/Fish Tank.json
kjbranch: Talked to @blkerby about this. Looks like this should work fine .

Issue #1185, opened 2023-11-30
Sparking through the door vertical restrictions


kjbranch:
Sparking through door assumes you can spark at either the bottom of the door or the top (with `canMidairShinespark`). However there are some restrictions which are not accounted for, which can cause problems. For example, sparking out of the right of the Gauntlet Entrance can only be done through the top of the door, and sparking through Mt. Everest can only be done through the bottom of the door.

PR #1188, opened 2023-12-01
Schema changes to represent cross-room spark position

blkerby:
This addresses issue #1185 by adding a new property `position` to the exit/entrance conditions `leaveWithSpark`/`comeInWithSpark`. Only horizontal sparks are considered for now, and the new `position` property indicates the vertical position in the doorway during the transition. Just having two positions `top` and `bottom` appears to be precise enough to model everything that we need (for now).

The new property is applied in Gauntlet Entrance and Mt. Everest, fixing the example of unsound logic mentioned in the issue.

### DIFF
--- a/strats.md
+++ b/strats.md
@@ -466,7 +466,7 @@ A strat with a `comeInWithSpark` condition should include a `shinespark` require
 
 A `comeInWithSpark` condition must match with either a `leaveWithSpark`, `leaveShinecharged`, or `leaveWithRunway` condition on the other side of the door:
 
-- A match with `leaveWithSpark` is valid as long as the `position` properties are compatible. The `position`` properties of a `leaveWithSpark` and `comeInWithSpark` are compatible if they are equal or if at least one of them are unspecified.
+- A match with `leaveWithSpark` is valid as long as the `position` properties are compatible. The `position` properties of a `leaveWithSpark` and `comeInWithSpark` are compatible if they are equal or if at least one of them are unspecified.
 - A match with `leaveShinecharged` is always valid and does not come with any implicit requirements.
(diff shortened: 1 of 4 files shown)


strats.md
```diff
@@ -452,13 +455,19 @@ A `comeInShinecharged` must match with either a `leaveShinecharged` condition or
...
+
+The direction of the spark is assumed to be horizontal when sparking through horizontal door transitions, or vertical when sparking through vertical door transitions.
 
 A strat with a `comeInWithSpark` condition should include a `shinespark` requirement in its `requires`.
 
 A `comeInWithSpark` condition must match with either a `leaveWithSpark`, `leaveShinecharged`, or `leaveWithRunway` condition on the other side of the door:
 
-- A match with `leaveWithSpark` or `leaveShinecharged` is always valid and does not come with any implicit requirements.
+- A match with `leaveWithSpark` is valid as long as the `position` properties are compatible. The `position`` properties of a `leaveWithSpark` and `comeInWithSpark` are compatible if they are equal or if at least one of them are unspecified.
```
kjbranch, line 469:
```suggestion
- A match with `leaveWithSpark` is valid as long as the `position` properties are compatible. The `position` properties of a `leaveWithSpark` and `comeInWithSpark` are compatible if they are equal or if at least one of them are unspecified.
```

kjbranch (approved):
That was fast. I like it.

This _could_ be used to model sparking through the top of the door without `canMidairShinespark`, if there is a raised place to stand nearby. I dont think it's necessary to do this though, since `canMidairShinespark` is a pretty low-difficulty tech and this would probably be just as difficult from being obscure.

PR #1190, opened 2023-12-01
Runway Migration: Most of 

kjbranch:


region/maridia/inner-yellow/Maridia Elevator Room.json
```diff
@@ -309,11 +291,12 @@
       "link": [1, 2],
       "name": "Shinespark",
       "notable": false,
+      "entranceCondition": {
```
osse101, line 294: 1<->2 can enter shinecharging and leave shinecharged.  Not many frames remaining on 1->2 however.

osse101 (approved):

region/maridia/inner-yellow/Maridia Elevator Room.json
kjbranch: I'm definitely not skilled enough to time those types of strats. Could you or @blkerby help with them?

PR #1195, opened 2023-12-02
Maridia ice vertical door lock skips (+ one more WS)

blkerby:
Skipping Mt. Everest (just adding a FIXME) because it looks horrible.

### DIFF
--- a/region/maridia/inner-pink/Crab Shaft.json
+++ b/region/maridia/inner-pink/Crab Shaft.json
@@ -571,6 +571,27 @@
       "requires": []
     },
     {
+      "link": [1, 3],
+      "name": "Sciser Ice Clip Door Lock Skip",
+      "notable": true,
+      "requires": [
+        {"ammo": {"type": "Super", "count": 1}},
+        "canTrickyUseFrozenEnemies",
+        {"or": [
+          "Gravity",
+          "HiJump"
+        ]},
+        "h_canNonTrivialCeilingClip"
+      ],
+      "bypassesDoorShell": true,
+      "note": [
+        "Shoot the shot block in the middle of the room, to allow a global Sciser (crab) to pass to the top part of the room.",
+        "Use Ice to climb the Scisers to the top of the room.",
+        "Shoot a Super as a Sciser crawls near the top door, then freeze it mid-air and use it to clip up past the door shell.",
+        "If you don't have Morph and X-Ray, you must shoot the Super while the Sciser is on a block next to the door, not on the door itself."
+      ]
+    },
+    {
       "link": [3, 3],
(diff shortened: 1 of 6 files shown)


region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -438,6 +438,27 @@
...
+    {
+      "link": [1, 3],
+      "name": "Sciser Ice Clip Door Lock Skip",
+      "notable": true,
+      "requires": [
+        {"ammo": {"type": "Super", "count": 1}},
+        "canTrickyUseFrozenEnemies",
+        {"or": [
+          "Gravity",
+          "HiJump"
```
osse101, line 450: Can it be done with HiJump?  Seems like a snail clip situation.  When trying the puyo clip method, I was pushed through the crab instead of the door.

region/maridia/inner-yellow/Pseudo Plasma Spark Room.json
```diff
@@ -300,6 +300,22 @@
         ]}
       ]
     },
+    {
+      "link": [2, 2],
+      "name": "Skulltera Ice Moonfall Door Lock Skip",
```
osse101, line 305: So I did this one and fell into the center box of Crab Shaft and was softlocked.

osse101 (commented):
The ice clip was the worst of the batch.
Mochtroids just took a bit of practice.  I used the walls to help position them but positioning isn't that important.

I suppose falling with precise timing doesn't feel that much harder than setting up some of the other enemyStucks,

And I had a much better time killing the lower enemy instead of sliding off.

region/maridia/inner-yellow/Pseudo Plasma Spark Room.json
blkerby: Yeah I found that for a lot of these moonfalls, holding an angle button (or up) through the transition is helpful to expand your hitbox, too avoid falling too deep in the next room. In the examples I tested, that seemed to be enough to prevent getting stuck.

In principle the way to model this accurately would be to put an exitCondition on these strats, with a leavesWithStoredFallSpeed or something similar. Then strats on the bottom room would have a corresponding entrance condition, to take into account the effects of the fall speed (either positively or negatively). But then we'd need a way to put specific speed values in those exit/entrance conditions. Also the precise position through the transition (horizontal and vertical) could matter, and that could get messy fast. And I'm not sure if we have any examples yet where it would actually matter logically.

region/maridia/inner-pink/Crab Shaft.json
osse101: I still can't get it

PR #1200, opened 2023-12-03
New strat property resetsObstacles, and application in Bat Cave

blkerby:
This new strat property `resetsObstacles` can probably be useful in a number of situations. One application is here in Bat Cave, where an X-Ray climb is possible (e.g. as an alternative to walljumping) but where we have to be careful with the farm logic since the farm can't be used in the normal way while off-camera.

### DIFF
--- a/region/norfair/east/Bat Cave.json
+++ b/region/norfair/east/Bat Cave.json
@@ -184,7 +184,7 @@
         {"heatFrames": 1600}
       ],
       "clearsObstacles": ["A"],
-      "note": "Climb up 1 screens."
+      "note": "Climb up 1 screen."
     },
     {
       "link": [2, 1],


region/norfair/east/Bat Cave.json
```diff
@@ -163,6 +172,20 @@
...
+      "notable": false,
+      "entranceCondition": {
+        "comeInWithDoorStuckSetup": {}
+      },
+      "requires": [
+        "canXRayClimb",
+        {"heatFrames": 1600}
+      ],
+      "clearsObstacles": ["A"],
+      "note": "Climb up 1 screens."
```
kjbranch, line 187:
```suggestion
      "note": "Climb up 1 screen."
```

kjbranch (dismissed):
Looks good. I think this will have uses elsewhere - I remember needing something like it before, but I can't remember where or if I figured out a workaround.

PR #1212, opened 2023-12-05
Remove Notable: false

kjbranch:
run auto formatting and remove notable scripts.


blkerby (approved):
Looks good. That was my bad with doing all the Crystal Flash strats with wrong formatting. We were thinking about having the formatting shorten the `requires` to one line when it is simple like that, so I wrote them with that in mind. But sounds like we probably want to stick with just always having requirements on separate lines even in simple cases.

PR #1215, opened 2023-12-05
G-Mode Migration: Crateria Central Pt. 1

kjbranch:
This PR was based off of #1212. It contains the output of the g-mode migration script with a thorough review and a few small edits. Only errors in the migration itself are included, not errors in the strats (unless severe), new strats, nor strats from non-door nodes. Those will be included in a follow up PR.

Edits included in this PR:
- Strats with a leave with g-mode have problems in that their `requires` is before the `entrance condition` instead of after it. (Crateria Tube)
- Places with multiple strats in a `leave with gmode setup` with a devnote are split and the note is duplicated. This is expected, but not ideal. (Parlor)
- There was a major error in one strat which included a `{"not": "f_ZebesAwake"}`, but neglected `canRiskPermanentLossOfAccess`.

Edits that will be included in a follow-up PR:
- Strats that started at a junction: Landing Site, Parlor, Blue Brin Elevator
- Crateria Supers
  - This room is a bit of a mess and could greatly benefit from another node at 4 or 6.
  - 1->3->1, 1->3->5 remote acquire
- Pit Room 2->3 with artificial morph and PBs works around the pirates, however they won't spawn without morph
- Parlor
  - Doors 2-7 setups and immobile need to be timed to see if they need `canBePatient` (they were encoded before the lower-level tech existed).
  - Setups leaving through 1 can now be encoded without starting at 1.

### DIFF
--- a/region/crateria/central/Climb.json
+++ b/region/crateria/central/Climb.json
@@ -262,13 +262,13 @@
     {
       "link": [1, 2],
       "name": "G-Mode through Bomb Blocks",
-      "requires": [
-        {"comeInWithGMode": {
-          "fromNodes": [1],
+      "entranceCondition": {
+        "comeInWithGMode": {
           "mode": "any",
-          "artificialMorph": false
-        }}
-      ],
+          "morphed": false
+        }
+      },
+      "requires": [],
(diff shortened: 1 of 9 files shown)


blkerby (approved):
Looks good.

PR #1217, opened 2023-12-06
G-Mode Migration: Crateria Central Pt. 2

kjbranch:
Applied the needed changes mentioned in #1215. 

It turns out the Crateria Supers room looks to be working properly, its just very complex. I'm not sure if it could be significantly improved with the addition of another node or two or not, but I'm planning on leaving it alone for now.


region/crateria/central/Crateria Super Room.json
```diff
@@ -925,6 +925,14 @@
       ],
       "devNote": "The strat specific requirements are included when clearing the abstract obstacle."
     },
+    {
+      "link": [3, 5],
+      "name": "Remote Acquire Item",
+      "requires": [
+        {"obstaclesCleared": ["C"]}
+      ],
+      "note": "Walk through the item to overload PLMs, fall through the speed blocks, then exit G-Mode to remote acquire the item."
```
osse101, line 934: Using the item to overload PLMs would require the item to be there, but this strat can be run multiple times.  Clearing C requires having Bombs and being in morph, so its still safe.

osse101 (dismissed):
RIP pit room dboosts.

Single IBJing past the rippers in parlor looks ok.  I had to end with a double bomb jump, but that's different from double IBJing.

PR #1218, opened 2023-12-06
Update Climb Room Diagram

kjbranch:
This is what I have for the updated Climb room diagram. It's a mess of a room - probably the worst room in the game, since it's so tall and there are bomb blocks by so many doors ... 

I still think including all of the arrows is helpful. In this example, I could see people not knowing you can cross the room with Temp Blue (2->3/4) or the Screw + X-Ray Climb (5->3/4), and the arrows could help there. But this room is a mess and I'm open to changes (to this room or the rework in general).


osse101:
My initial reaction to different coloured arrowheads was that it looked like a mistake.  And then I thought about how it made sense.

Very organized lines 🧮

kjbranch:
Thanks, I think this is a big improvement. 
The line from 2->3/4 is now blue, so it's less obtrusive, and it lets me combine it with the blue 5->3/4 lines.
There were a couple duplicate strats that existed in both 2->3/4 and 6->3/4, so I combined them.

PR #1230, opened 2023-12-12
Update Beta Power Bomb Room

kjbranch:


region/brinstar/red/Beta Power Bomb Room.json
```diff
@@ -264,16 +264,45 @@
       ],
       "clearsObstacles": ["A"]
     },
+    {
+      "link": [2, 3],
```
osse101, line 268: [2, 2]?

kjbranch: yep. Whoops

PR #1231, opened 2023-12-13
Update Kraid Fight

kjbranch:
Without being able to wall jump, the fight is much harder. Using the platforms can be particularly annoying without being able to dodge while shooting. A crouch jump + down grab is possible, but at low difficulties it doesnt really seem easier than using the platforms (at higher difficulties it seems easier).

### DIFF
--- a/region/brinstar/kraid/Kraid Room.json
+++ b/region/brinstar/kraid/Kraid Room.json
@@ -81,8 +81,11 @@
                   "SpaceJump",
                   "canSpringBallJumpMidAir",
                   {"and": [
-                    "canCarefulJump",
-                    "canDodgeWhileShooting"
+                    "canDodgeWhileShooting",
+                    {"or": [
+                      "canCarefulJump",
+                      "h_canCrouchJumpDownGrab"
+                    ]}
                   ]}
                 ]}
               ],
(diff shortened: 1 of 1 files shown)


osse101:
The platforms have a bit of trick to them where you want to jump over the spikes so that you have more time with the platforms.  If you know that, then its about the same as crouchjump+downgrab.

Crouchjump+downgrab isn't listed as an option, and its as hard as a normal crouchjump+downgrab.

Looking at the rest of the fight:
The Charge only fight took me 110seconds and that was while knowing about the hand stuck trick with charge, but not playing amazingly.  An option for using a stronger beam could be more friendly.
The Missile fight is fine.  Hunting missile drops could be annoying.  And maybe you don't know a beam shot opens the mouth in Phase1.
The super fight; it's not that unlikely to miss a super at low difficulty, but you can still farm it back eventually.

osse101:
Still wondering about Crouchjump+downgrab.

I like the other changes.

kjbranch:
What about something like this?
I think the cj+dg is a lot tighter in dry areas, and this one requires moving rightward a tile as well as focusing on a lot of other things going on. I was thinking it should be in Hard.

PR #1233, opened 2023-12-13
Runway Migration - Stragglers

kjbranch:
This finishes off the majority of the remaining strats that aren't in LN

### DIFF
--- a/region/brinstar/green/Green Brinstar Beetom Room.json
+++ b/region/brinstar/green/Green Brinstar Beetom Room.json
@@ -287,16 +287,31 @@
     {
       "link": [1, 2],
       "name": "Jump Through the Room",
+      "entranceCondition": {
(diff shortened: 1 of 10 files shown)


region/brinstar/green/Green Brinstar Beetom Room.json
```diff
@@ -287,17 +287,32 @@
     {
       "link": [1, 2],
       "name": "Jump Through the Room",
+      "entranceCondition": {
+        "comeInJumping": {
```
osse101, line 291: This strat kinda depends on Beetom RNG tbh but I think just landing near the opposite door is enough to get through damageless.  
Jumping after the transition, with this tile count, lands on the opposite runway.
Jumping before the transition lands in the pit but the Beetoms avoid Samus anyway.

PR #1234, opened 2023-12-14
Runway Migration - Remove Remaining canLeaveCharged

kjbranch:


region/lowernorfair/east/Lower Norfair Fireflea Room.json
```diff
@@ -286,6 +265,30 @@
...
+          {"obstaclesCleared": ["A"]}
+        ]},
+        "canShinechargeMovement"
+      ],
+      "exitCondition": {
+        "leaveShinecharged": {
+          "framesRemaining": 120
+        }
+      },
+      "note": "FIXME: There may be an X-Mode strat that could work here without killing the Fune (possibly with Ice)."
```
osse101, line 290: I couldn't get an xmode shinespark in the spike pit.  The entry angle doesn't work for it, plus it is very risky towards causing a crash.

osse101 (dismissed):

PR #1235, opened 2023-12-14
Simplify Node Names

kjbranch:
This only renames nodes, not enemy groups, door locks, etc.

### DIFF
--- a/region/lowernorfair/east/Red Kihunter Shaft Save Room.json
+++ b/region/lowernorfair/east/Red Kihunter Shaft Save Room.json
@@ -11,7 +11,7 @@
   "nodes": [
     {
       "id": 1,
-      "name": "Red Kihunter Shaft Save Room Door (to Red Kihunters)",
+      "name": "Left Door",
       "nodeType": "door",
       "nodeSubType": "blue",
       "nodeAddress": "0x0019ab6",
@@ -34,7 +34,7 @@
     },
     {
       "id": 2,
-      "name": "Red Kihunters Save Station",
+      "name": "Save Station",
       "utility": ["save"],
       "nodeType": "utility",
       "nodeSubType": "save"
--- a/region/lowernorfair/east/Red Kihunter Shaft.json
+++ b/region/lowernorfair/east/Red Kihunter Shaft.json
@@ -11,7 +11,7 @@
   "nodes": [
     {
       "id": 1,
-      "name": "Red Kihunter Shaft Left Door (to Amphitheatre)",
+      "name": "Top Left Door",
       "nodeType": "door",
       "nodeSubType": "blue",
       "nodeAddress": "0x00199f6",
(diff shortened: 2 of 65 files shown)


region/brinstar/blue/Blue Brinstar Energy Tank Room.json
```diff
@@ -30,23 +30,23 @@
...
+      "name": "Hidden Ceiling Item",
       "nodeType": "item",
       "nodeSubType": "hidden",
       "nodeItem": "ETank",
       "nodeAddress": "0x7879E"
     },
     {
       "id": 4,
-      "name": "Missile (blue Brinstar middle)",
+      "name": "Right Pedestal Item",
```
blkerby, line 49: Could just be "Pedestal Item" since there's only one pedestal item?

region/brinstar/blue/Morph Ball Room.json
```diff
@@ -90,7 +90,7 @@
     },
     {
       "id": 6,
-      "name": "Morph Ball Room Junction (Right of Sidehoppers)",
+      "name": "Junction (Right of Sidehoppers)",
```
osse101, line 93: `Right of Sidehoppers` isn't clear to me.
I'd like `Left of Morph Tunnel` more.

region/brinstar/blue/Blue Brinstar Energy Tank Room.json
kjbranch: I couldnt decide between `Right Item` and `Pedestal Item`. I think if I were to choose one it would be `Right Item` in this case.

blkerby: Sounds good

kjbranch:
I renamed the nodes in every section except Upper/Lower Norfair. I should be able to get to them by tomorrow.

region/crateria/east/West Ocean.json
```diff
@@ -11,15 +11,15 @@
...
+      "name": "Left Door",
       "nodeType": "door",
       "nodeSubType": "blue",
       "nodeAddress": "0x00189ca",
       "doorEnvironments": [{"physics": "air"}]
     },
     {
       "id": 2,
-      "name": "West Ocean Top Right Door (to Attic)",
+      "name": "Upper Right Section Top Right Door",
```
osse101, line 22: Green Brin Main Shaft does sections like
`Top Right Door (Upper Right Section)`
and that is confusing at first read, but I think this should copy the format to be consistent.  And I like that way of writing it more.

kjbranch: I was moving more towards this method, because I think it will be nicer to read. For `Bottom Right Door (Upper Right Section)`, if you read it quickly you might think it's the `Bottom Right Door`. Maybe `Upper Right Section - Bottom Right Door` would work the best?

osse101: Right, `Bottom Left Door (Above Power Bomb Blocks)` is hard to understand at first, like here.  
`Above Power Bomb Blocks - Bottom Left Door` works.  Its just different from the majority of node names since only two or three rooms have sections.

kjbranch:
Finished the rest of the rooms (Upper/Lower Norfair), and reworked the node names in West Ocean, Green Brinstar Main Shaft, and Big Pink

osse101 (approved):

PR #1236, opened 2023-12-14
East Sand Hole - Right item with bomb escape

kjbranch:


region/maridia/inner-pink/East Sand Hole.json
```diff
@@ -155,9 +146,29 @@
               "canSunkenTileWideWallClimb"
             ]}
           ]},
+          "canSpringBallJumpMidAir",
+          {"and": [
+            "h_canCrouchJumpDownGrab",
+            "canBombJumpWaterEscape",
+            "h_canIBJ"
```
osse101, line 153:
```suggestion
            "h_canJumpIntoIBJ"
```

region/maridia/inner-pink/East Sand Hole.json
```diff
@@ -155,9 +146,29 @@
...
+        "canPlayInSand",
+        {"or": [
           {"and": [
-            "HiJump",
-            "canSpringBallJumpMidAir"
+            "h_canMaxHeightSpringBallJump",
+            {"or": [
+              "canWalljump",
+              "can3HighMidAirMorph",
+              "h_canJumpIntoIBJ"
```
osse101, line 170:
```suggestion
              {"and": [
                "canBombJumpWaterEscape",
                "h_canJumpIntoIBJ"
              ]}
```

region/maridia/inner-pink/East Sand Hole.json
kjbranch: Is this possible to do with a single jump? I can't get it

region/maridia/inner-pink/East Sand Hole.json
kjbranch: This is done with spring ball, so it's a lot easier than what I think `canBombJumpWaterEscape` calls for. You can hold jump constantly and just place a well-timed bomb at the peak of any of the jumps

region/maridia/inner-pink/East Sand Hole.json
osse101: This is a funny way to look at it, but I agree with your reasoning.

With HiJump it looks like your jumping out of the water, and without HJB you are bomb jumping at the water line so they look like they should be switched.

PR #1242, opened 2023-12-17
Entrance/exit condition schemas for jumpways

blkerby:
This PR introduces the schema changes needed to migrate jumpways to the new style of cross-room strat, continuing the work of #1059.

The mechanics of the matching is mostly left unchanged; i.e., as in the old jumpway schema, the strat in the upper room specifies constraints on the `height`, `leftPosition`, and `rightPosition` properties of the matching strat in the lower room. The most significant change is that rather than having a single generic leaveWithJumpway/comeInWithJumpway with multiple possible jumpwayTypes, I split it up a bit more into finer-grained conditions, eliminating the need for a `jumpwayType`:

- Exit conditions: `leaveWithDoorFrameBelow`, `leaveWithPlatformBelow`
- Entrance conditions: `comeInWithWallJumpBelow`, `comeInWithSpaceJumpBelow`, `comeInWithPlatformBelow`

The main reason for splitting up `comeInWithWallJumpBelow` and `comeInWithSpaceJumpBelow` as entrance conditions is because of the somewhat substantial difference in heat frames between these two, if the lower room is heated. Strats which do a Spring Ball jump in the lower room can be expressed with `leaveWithPlatformBelow`/`comeInWithPlatformBelow`; I don't currently see a need for it to have its own entrance condition.

### DIFF
--- a/strats.md
+++ b/strats.md
@@ -86,6 +86,8 @@ In all strats with an `exitCondition`, the `to` node of the strat must be a door
 - _leaveWithStoredFallSpeed_: This indicates that is is possible to walk through the door with the stored velocity to clip through floor tiles using a Moonfall.
 - _leaveWithGModeSetup_: This indicates that Samus can take enemy damage through the door transition, to set up R-mode or direct G-mode in the next room.
 - _leaveWithGMode_: This indicates that Samus can carry G-mode into the next room (where it will become indirect G-mode).
+- _leaveWithDoorFrameBelow_: This indicates that Samus can go up through this door with momentum by jumping in the door frame, e.g. using a wall-jump or Space Jump.
+- _leaveWithPlatformBelow_: This indicates that Samus can go up through this door with momentum by jumping from a platform below, possibly with run speed.
 
 Each of these properties is described in more detail below.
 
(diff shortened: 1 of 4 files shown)


strats.md
```diff
@@ -264,6 +266,56 @@ and another where the `false` values for `morphed` are replaced with `true`. Her
 
 Aside from the implicit strats, there are a limited amount of `leaveWithGMode` strats possible. Normally entering a room with G-mode (or a G-mode setup) and then leaving with G-mode through a different door is not possible, since door shells cannot be opened while in G-mode. However, some door transitions do not have door shells (e.g. in Crateria Tube, Glass Tunnel, Crab Hole, Big Pink; also elevators and sand transitions), and some door shells are possible to bypass using glitches, so `leaveWithGMode` can be used in these situations.
 
+### Leave With Door Frame Below
+
+A `leaveWithDoorFrameBelow` exit condition represents that Samus can go up through this door with momentum by jumping in the door frame, e.g. using a wall-jump or Space Jump. A `leaveWithDoorFrameBelow` exit condition can satisfy `comeInWithWallJumpBelow` and `comeInWithSpaceJumpBelow` entrance conditions in the room above.
+
+A `leaveWithDoorFrameBelow` object has the following property:
+
+- _height_: The number of tiles beneath the door transition (not including the transition tiles) usable for wall-jumping.
```
osse101, line 275: Does this value have a use?  Some kind of low walljump that needs HiJump but also has to dodge a ledge after the transition?

osse101:
Looks like a good migration

- comeInWithSpaceJumpBelow could be a fixed height room entry, but does have different heat frames, yeah.
- door frame walljump could do fractional walljump position values, but putting the specific requirements on the comeIn side cuts out unneeded leave conditions
- comeInWithBlueFromBelow would be a separate condition with an entry direction.  The low speed entries which have a direction can be done from either side from existing jumpways.

strats.md
blkerby: Yeah it's used for Pseudo Plasma Spark Room. The extra tile of height for a lower walljump makes it much easier, which affects the logical requirements.

osse101 (approved):

PR #1245, opened 2023-12-17
Add R-Mode and G-Mode strats to Gauntlet E-Tank Room

kjbranch:


region/crateria/west/Gauntlet Energy Tank Room.json
```diff
@@ -203,6 +233,151 @@
...
+        ]}
+      ],
+      "clearsObstacles": ["A"],
+      "note": [
+        "Carefully walk through the room. It is possible but tricky to do so damageless, but it is always possible to go back and farm if needed.",
+        "While in the morph tunnel, simply unmorph to break the bomb blocks overhead."
+      ]
+    },
+    {
+      "link": [1, 4],
```
osse101, line 339: I think this is supposed to be [1,1] and then uses a obstaclesCleared[B] strat to reach 4

PR #1247, opened 2023-12-18
Refine leaveShinecharged in Spore Spawn Kihunter Room

blkerby:
Using the full 32 tiles of runway to get a shinecharge makes it not really practical to get up the steps with any frames remaining (technically possible, but hard and pointless). So this PR rewrites the strat to require a 25-tile charge, to leave room to gain run speed to jump up the steps. 

I think the movement here goes beyond expectations of "canShinechargeMovement", so this PR changes it to "canShinechargeMovementComplex". And the frames remaining are increased significantly, which for example will put it in logic to be used with Reverse Croc Speedway.

### DIFF
--- a/region/brinstar/green/Spore Spawn Kihunter Room.json
+++ b/region/brinstar/green/Spore Spawn Kihunter Room.json
@@ -137,16 +137,18 @@
           "h_hasBeamUpgrade"
         ]},
         {"canShineCharge": {
-          "usedTiles": 32,
+          "usedTiles": 25,
           "openEnd": 0
         }},
-        "canShinechargeMovement"
+        "canShinechargeMovementComplex"
       ],
       "exitCondition": {
         "leaveShinecharged": {
-          "framesRemaining": 30
+          "framesRemaining": 75
         }
-      }
+      },
+      "note": "Store a shinecharge with enough space remaining to run and jump over all the steps with a single jump.",
+      "devNote": "The runway is 32 tiles, but about 7 of those should be used to run after getting the shinecharge."
     },
     {
       "link": [1, 1],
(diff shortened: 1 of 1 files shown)


osse101:
It would be good if we could keep a 32 tile or `canShinechargeMovement` option.  Such as by shinesparking out the left door from the middle of the room.  Aiming a midair spark that far is a challenge itself :eyes:
But making breakpoints for movement ability->framesremaining is an issue.  And so is using a full runway and then moving afterwards.

blkerby:
> It would be good if we could keep a 32 tile or `canShinechargeMovement` option. Such as by shinesparking out the left door from the middle of the room. Aiming a midair spark that far is a challenge itself 👀 But making breakpoints for movement ability->framesremaining is an issue. And so is using a full runway and then moving afterwards.

Yeah, my thinking was that at the difficulty level where you can't shortcharge at all (Medium), there's not really anything that you could reasonably be expected to do after getting the shinecharge right up against the steps. A mid-air horizontal spark (much less one without an on-camera target to aim for) is fairly challenging for that level. @kjbranch has talked about wanting to move mid-air horizontal shinesparks out of Medium and into Hard (and maybe also move grounded horizontal shinesparks from Basic into Medium?), and that makes some sense to me.

The 25-tile charge (which seems close to an optimal distance?) can be done comfortably with just a 1-tap (Hard), so in this case I'm not sure there's a need for additional breakpoints for other difficulty levels. I still would like to add leniency frames across the board for cross-room shinecharges, but that's something that could just be handled on the randomizer side (similar to heat frames) without needing to be explicit in the strats.

osse101:
> wanting to move mid-air horizontal shinesparks out of Medium and into Hard (and maybe also move grounded horizontal shinesparks from Basic into Medium?), and that makes some sense to me.

Both of those suggestions make sense to me.  There is a new stepping stone of shinesparking when learning horizontal and midair horizontal sparks since they take finer control to do.

PR #1248, opened 2023-12-19
Add G-Mode to Green Pirate Shaft

kjbranch:
Also added the energy refill to the artificial morph CF helper, and reformatted the ammo lines in this file, since I copy/paste them a lot and the formatter doesn't touch this file.


region/crateria/west/Green Pirates Shaft.json
```diff
@@ -267,6 +313,46 @@
...
+          }},
+          {"enemyDamage": {
+            "enemy": "Green Space Pirate (standing)",
+            "type": "laser",
+            "hits": 3
+          }}
+        ]}
+      ],
+      "note": [
+        "Crystal Flash then damage down on the Pirates to Reserve trigger to exit G-Mode.",
```
osse101, line 340: The goal of the strat is to fill your reserves and gain up to 400 health?

kjbranch: Yes, in order to do Beetom shenanigans. Although I guess this needs to model the PB blocks being broken, otherwise everything requires Morph to get back down from 9 except the Beetom x-ray climb

PR #1253, opened 2023-12-19
West Ocean - Use Bridge to Bypass Door Locks

kjbranch:
I think this is the best way to represent these strats. With this method, the tests will need to be changed to allow nodes with the same `nodeAddress`.


connection/inter.json
```diff
@@ -152,6 +152,56 @@
...
+          "area": "Wrecked Ship",
+          "subarea": "Main",
+          "roomid": 154,
+          "roomName": "Wrecked Ship Entrance",
+          "nodeid": 1,
+          "nodeName": "Wrecked Ship Entrance Left Door (to West Ocean)",
+          "position": "right"
+        }
+      ]
+    },
```
blkerby, line 179:
```suggestion
```
This appears to be a copy-paste of the previous connection.

region/crateria/east/West Ocean.json
```diff
@@ -1687,4 +1850,4 @@
     "FIXME: The area inside the bridge should be modeled, including adding two additional door nodes.",
     "It is possible to reach it in various ways, e.g. using deep-stuck X-ray climb from below, or a high-speed moonfall from above."
```
blkerby, line 1851: This can be deleted now.

blkerby (commented):
The "Stored Moonfall Clip" from 3 to 5 can also have a version from 3 to 8.

region/crateria/east/West Ocean.json
```diff
@@ -534,6 +574,24 @@
...
+      "entranceCondition": {
+        "comeInWithStoredFallSpeed": {
+          "fallSpeedInTiles": 2
+        }
+      },
+      "requires": [
+        "canSkipDoorLock"
+      ],
+      "note": [
+        "Moonfall onto the Homing Geemer walkway twice to clip all the way down into the ocean below.",
```
blkerby, line 589:
```suggestion
        "Moonfall onto the Homing Geemer walkway twice to clip down into the bridge below.",
```

region/crateria/east/West Ocean.json
```diff
@@ -534,6 +574,24 @@
...
+        "The Moonfall must be performed at the very edge of the slope so that Samus fully falls to the lower area.",
+        "The first Moonfall will clip into the slope above the walkway, and the second Moonfall will clip all the way through.",
+        "Pose changing will allow Samus to remain inside of the walkway, where the right door connects to the WS main entrance door, and the left door connects to the Moat door."
```
blkerby, line 592:
```suggestion
        "Pose changing will allow Samus to remain inside of the bridge, where the right door connects to the WS main entrance door, and the left door connects to the Moat door."
```

blkerby (dismissed):

PR #1254, opened 2023-12-20
Add support for carrying a shinecharge between 3 rooms

blkerby:
This PR updates the schema and docs to support representing strats that enter a room with a shinecharge and leave with a shinecharge. This is done by putting both a `comeInShinecharged` entrance condition and a `leaveShinecharged` exit condition on the strat, where a new "auto" value is used on the `framesRemaining` of the `leaveShinecharged` to indicate that the final frames remaining should be inferred by subtracting the amount of frames required in the `comeInShinecharged` from the amount of frames remaining on entry.

I also added some details on expectations for Samus' position and momentum during the door transition in strats with `comeInShinecharged` and `leaveShinecharged`. Hopefully it aligns reasonably well with what we have already been doing; if not it could probably be adjusted. Just trying to iron out potential ambiguities, to avoid the possibility that an entrance strat and exit strat could make conflicting assumptions about the transition position/momentum which could result in the combination being impossible to execute.

Crateria Tube is included as an example.

### DIFF
--- a/region/crateria/central/Crateria Tube.json
+++ b/region/crateria/central/Crateria Tube.json
@@ -85,6 +85,21 @@
     },
     {
       "link": [1, 2],
+      "name": "Carry Shinecharge",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 50
+        }
+      },
+      "requires": [],
+      "exitCondition": {
+        "leaveShinecharged": {
+          "framesRemaining": "auto"
+        }
+      }
+    },
+    {
+      "link": [1, 2],
       "name": "Spark Through Tunnel",
       "entranceCondition": {
         "comeInWithSpark": {}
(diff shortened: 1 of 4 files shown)


osse101 (dismissed):
That's some nice tech.

I thought `details on expectations` would include auto implying a `canShinechargeMovementComplex` requirement since we want to fit that into 3 room shinecharges somehow.

blkerby:
> That's some nice tech.
> 
> I thought `details on expectations` would include auto implying a `canShinechargeMovementComplex` requirement since we want to fit that into 3 room shinecharges somehow.

Yeah, I wanted that at one point but now I'm not sure. Running through Crateria Tube for example might be simple enough to not need `canShinechargeMovementComplex`. It should at least need `canShinechargeMovement` (added that now). I think it may be ok to keep things where we just add those tech requirements explicitly on the strats where they apply?

osse101:
It's the combination of multiple shinecharge conservation strats which makes it complex.  And that situation shows up the clearest with `framesRemaining: auto`.  But trivial rooms are still trivial, and rooms with movement are still complex.  So maybe there's not an issue here.

PR #1256, opened 2023-12-20
Add Plasma Spark Fake Kassiuz Junction

kjbranch:


blkerby (approved):
It's a little awkward how the 3 -> 6 strat mentions going through the door and then 6 -> 2 mentions it again. But I understand why it's like this, since the 3 -> 6 is notable and going the door is currently the only useful thing to do up there

kjbranch:
Yeah, I wanted the notable variant to have the full description. I gave a few more details in the second description which werent overly necessary, but still felt like the strat should have some description.

PR #1269, opened 2023-12-23
Red Tower top: easier shinespark strats

blkerby:
This adds alternative, easier shinespark strats for getting up the top of Red Tower without the hero shot. These need Ice in order to get the rest of the way up, but they have lower requirements in terms of tech, energy, and shinecharge frames. In particular they allow getting up without a tank, whereas the hero shot method requires a tank.

### DIFF
--- a/region/brinstar/red/Red Tower.json
+++ b/region/brinstar/red/Red Tower.json
@@ -780,7 +780,7 @@
     },
     {
       "link": [1, 5],
-      "name": "Red Tower Hero Shot Spark (Come In Shinecharged)",
+      "name": "Red Tower Hero Shot Shinespark (Come In Shinecharged)",
       "notable": true,
       "reusableRoomwideNotable": "Red Tower Hero Shot Shinespark",
       "entranceCondition": {
--- a/region/brinstar/red/Red Tower.json
+++ b/region/brinstar/red/Red Tower.json
@@ -340,7 +340,7 @@
       ]
     },
     {
-      "name": "Red Tower Spark",
+      "name": "Red Tower Hero Shot Spark",
       "note": [
         "Come in shinecharged (or shinecharging) from the top left door. With missiles selected, position Samus roughly in the horizontal center of the room.",
         "Crouch, aim up, then in very quick succession, shoot a missile upwards then spark up.",
@@ -780,9 +780,9 @@
     },
     {
       "link": [1, 5],
-      "name": "Red Tower Spark (Come In Shinecharged)",
+      "name": "Red Tower Hero Shot Spark (Come In Shinecharged)",
       "notable": true,
-      "reusableRoomwideNotable": "Red Tower Spark",
+      "reusableRoomwideNotable": "Red Tower Hero Shot Spark",
       "entranceCondition": {
         "comeInShinecharged": {
           "framesRequired": 100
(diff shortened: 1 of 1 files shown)


region/brinstar/red/Red Tower.json
```diff
@@ -340,7 +340,7 @@
       ]
     },
     {
-      "name": "Red Tower Spark",
+      "name": "Red Tower Hero Shot Spark",
```
kjbranch, line 343: I'm preferring moving towards full long-hand names, but I'm fine either way.
```suggestion
      "name": "Red Tower Hero Shot Shinespark",
```

region/brinstar/red/Red Tower.json
```diff
@@ -780,9 +780,9 @@
     },
     {
       "link": [1, 5],
-      "name": "Red Tower Spark (Come In Shinecharged)",
+      "name": "Red Tower Hero Shot Spark (Come In Shinecharged)",
```
kjbranch, line 783:
```suggestion
      "name": "Red Tower Hero Shot Shinespark (Come In Shinecharged)",
```

region/brinstar/red/Red Tower.json
```diff
@@ -780,9 +780,9 @@
...
       "notable": true,
-      "reusableRoomwideNotable": "Red Tower Spark",
+      "reusableRoomwideNotable": "Red Tower Hero Shot Spark",
```
kjbranch, line 785:
```suggestion
      "reusableRoomwideNotable": "Red Tower Hero Shot Shinespark",
```

region/brinstar/red/Red Tower.json
```diff
@@ -801,9 +801,9 @@
     },
     {
       "link": [1, 5],
-      "name": "Red Tower Spark (Come In Shinecharging)",
+      "name": "Red Tower Hero Shot Spark (Come In Shinecharging)",
```
kjbranch, line 804:
```suggestion
      "name": "Red Tower Hero Shot Shinespark (Come In Shinecharging)",
```

region/brinstar/red/Red Tower.json
```diff
@@ -801,9 +801,9 @@
...
       "notable": true,
-      "reusableRoomwideNotable": "Red Tower Spark",
+      "reusableRoomwideNotable": "Red Tower Hero Shot Spark",
```
kjbranch, line 806:
```suggestion
      "reusableRoomwideNotable": "Red Tower Hero Shot Shinespark",
```

region/brinstar/red/Red Tower.json
```diff
@@ -837,6 +837,43 @@
         ]}
       ]
     },
+    {
+      "link": [1, 9],
+      "name": "Shinespark (Come In Shinecharged)",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 50
```
kjbranch, line 845: sparking diagonally immediately on room entry can be done with a single frame. Probably worth modeling separately from the jump option ... but it may have different frames remaining, as youll be on the right side of the shaft ...

kjbranch (approved):

region/brinstar/red/Red Tower.json
blkerby: True, added a strat for that now. Also realized that the framesRequired for the vertical spark could also be reduced by sparking closer to the door, though this breaks the platform so you have to land on the frozen enemies. I changed all the new strats to go directly to the door node (5) instead of the platform junction (9), as there doesn't appear to be anything useful you could do here starting from node 9 other than freezing the enemies and continuing to node 5 (e.g., if you could kill the enemies and use walljump, IBJ, or SpaceJump then there would be no need for the shinespark).

PR #1271, opened 2023-12-23
Tighten Reverse Croc Speedway shinecharge frames

blkerby:
Tightening the shinecharge frames here since they were fairly loose, and this seemed like an important strat. 

Increased the shinespark frames by a couple since we're no longer assuming you have to get all the way next to the speed blocks before sparking.

I'm planning on implementing leniency for cross-room shinecharge frames probably in the next release of Map Rando, so this is part of why I'm thinking it can a good time to start moving toward tighter shinecharge frames where possible.

### DIFF
--- a/region/norfair/west/Crocomire Speedway.json
+++ b/region/norfair/west/Crocomire Speedway.json
@@ -253,14 +253,14 @@
       "notable": true,
       "entranceCondition": {
         "comeInShinecharged": {
-          "framesRequired": 60
+          "framesRequired": 35
         }
       },
       "requires": [
         "h_canNavigateHeatRooms",
         "canShinechargeMovement",
-        {"shinespark": {"frames": 84, "excessFrames": 10}},
-        {"heatFrames": 760}
+        {"shinespark": {"frames": 86, "excessFrames": 10}},
+        {"heatFrames": 700}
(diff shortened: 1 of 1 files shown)


region/norfair/west/Crocomire Speedway.json
```diff
@@ -253,14 +253,14 @@
...
+          "framesRequired": 35
         }
       },
       "requires": [
         "h_canNavigateHeatRooms",
         "canShinechargeMovement",
-        {"shinespark": {"frames": 84, "excessFrames": 10}},
-        {"heatFrames": 760}
+        {"shinespark": {"frames": 86, "excessFrames": 10}},
+        {"heatFrames": 700}
```
kjbranch, line 263: After the spark you right right then back while blue in order to break the rest of the speed blocks. Is using spring ball, to get over the speed blocks, better if you have it?

kjbranch (approved):

region/norfair/west/Crocomire Speedway.json
blkerby: I gave it a try but it looks like no, even in best case scenario it takes almost a second longer to try to get through with Spring Ball.

PR #1275, opened 2023-12-24
G-Mode Migration: Aqueduct

kjbranch:
This one needs to be reviewed and not just merged, as there was a lot more than just running the scripts.

The first commit was to run the script, the second was the rest of the room, including:
- Migrated strats with multiple from nodes
- Added a setup strat from door 1, starting at 9 and breaking the bomb blocks
- Move some [X, 1] strats which used a PB to go up from the bottom to instead be [X, 2] and break A

Things to still complete in the Sand section:
- Strats going down into the sand in Aqueduct and Botwoon E-Tank Rooms
  - The Aqueduct ones will need to moonfall, using 2 snails
- Edit: There are also some aftificial morph + spring ball snail climbs to look into. It's barely possible to get onto the snail with spring ball alone by pressing against the snail on the ascent to get a little more jump height


region/maridia/inner-pink/Aqueduct.json
```diff
@@ -726,6 +683,41 @@
...
+      "name": "G-Mode Morph Break the Power Bomb Blocks",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "any",
+          "morphed": true
+        }
+      },
+      "requires": [
+        {"ammo": {"type": "PowerBomb", "count": 1}}
+      ],
```
blkerby, line 718:
```suggestion
      ],
      "clearsObstacles": ["A"],
```

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -1206,12 +1249,6 @@
...
             ]}
-          ]},
-          {"and": [
-            "Bombs",
-            "HiJump",
-            "SpringBall",
-            "canSnailClimb"
           ]}
         ]}
       ]
```
blkerby, line 1254:
```suggestion
      ],
      "clearsObstacles": ["A"]
```

blkerby (approved):
I'm finding it hard to check everything in detail because of the amount of duplication. Some G-mode junctions could help simplify things? But that's just an idea, doesn't have to block the merge

kjbranch:
Another node or two could probably help. Likely one at node 2 with PLMs overloaded. It's not a huge help though, since each of the doors have different ways to do things. I'll put it in my to-do list, I just dont think it's a priority right now.

PR #1277, opened 2023-12-26
G-mode Migration: Botwoon E-Tank Room, Go Down the Sand

kjbranch:


region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -220,13 +222,154 @@
...
+      "notable": false,
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "any",
+          "morphed": true
+        }
+      },
+      "requires": [
+        {"or": [
+          "Bombs",
```
osse101, line 329: I think you need Grav or Spring to reach 3.  Maybe you could damage boost across with the Puyos.

PR #1279, opened 2023-12-26
Schema changes to represent unlocking doors

blkerby:
This PR adds a strat property `unlocksDoors` for representing that a strat can unlock a door while being executed, possibly subject to additional requirements, depending on the door type/color.
 
This will be useful in many situations, for example:

- For strats that gain a shinecharge, go down a one-way, and then spark out a door, there may be no opportunity to unlock the door beforehand. But if it is a red or green door and Supers are available, then it could be unlocked while executing the strat.
- In heated rooms, it may be possible to unlock the door while on the way to it, reducing the heat frames required compared to reaching the door and then unlocking it.
- It could be possible to unlock a door remotely, i.e. without being able to reach the door but being able to fire ammo at it. This could then make a later strat possible to go through the door from the other direction (e.g. Power Bomb door at left of  Amphitheatre). 
- It can be possible to unlock a Power Bomb door using G-Mode artificial morph.

The PR also adds a node property `useImplicitDoorUnlocks` which makes it possible to override default strats for unlocking door nodes which normally would be assumed to be possible. This will be useful for certain doors that can be particularly difficult to unlock, e.g. the left door in Tourian Blue Hopper Room or the left door of Amphitheatre. In these cases, `unlocksDoors` can be used to explicitly encode requirements for unlocking the doors. This PR also spells out in the docs what the implicit door unlock requirements are (which is non-obvious in heated rooms); previously this was something that Map Rando did but was never specified in this project. 

This PR also sets up tests to output warnings if a strat has a requirement for a door to be unlocked but does not specify requirements in `unlocksDoors` for how it can be unlocked. The plan is that once these are fixed for all existing strats, the warning can be promoted to an error. In theory, it could be possible to take another approach where if `unlocksDoors` requirements are unspecified, the logic would just assume the door has to be unlocked previously; however, that would make it necessary for randomizers and other tools to track door lock state within a room similar to obstacles, and the sheer amount of doors in some rooms would seem to make that impractical as the amount of states to track can blow up exponentially. Also, without these tests it would be easy to forget to include `unlocksDoors` in applicable strats. So even though it will be a bit tedious to explicitly include `unlocksDoors` requirements, doing it this way should lead to better accuracy while being easier for randomizers to use. The schema is also designed so that common cases can be handled with a one-liner `unlocksDoors`.

The choice to classify door types by their unlock method ("missiles", "super", "powerbomb") rather than by color ("red", "green", "yellow") was deliberate, to simplify writing `unlocksDoors` in strats; it avoids having to duplicate Super unlock requirements across both "red" and "green" doors.

As an example, this PR adds `unlocksDoors` strats across Green Brinstar (only the mandatory ones that would otherwise generate warnings with the new tests), and `"useImplicitDoorUnlocks": false` is added for the two bad doors in Blue Hopper Room and Amphitheatre; additional work would still be needed to flesh out door unlock strats in those rooms (and throughout the game), but it should all be possible with the new schema. I'm also planning to eventually migrate gray locks to also use `unlocksDoors` (replacing the `unlockStrats` embedded in the `lock` node property), consistent with the cross-room migration goal of moving all strats into the top-level room "strats" array.


region/brinstar/green/Early Supers Room.json
```diff
@@ -202,6 +202,34 @@
       },
       "requires": []
     },
+    {
+      "link": [1, 1],
```
blkerby, line 206: This strat was 4->1. I converted it to be 1->1 to make it so unlocking the door can be automatically handled without needing to add `unlocksDoors`.

kjbranch: But what if you can only come in at 2?

region/brinstar/green/Early Supers Room.json
```diff
@@ -457,7 +501,10 @@
         "leaveShinecharged": {
           "framesRemaining": 40
         }
-      }
+      },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
+      "note": "Use the runway on the screen above to gain a shinecharge near the edge of the runway.",
+      "devNote": "This starts at node 2 to be able to unlock the door if needed."
```
kjbranch, line 507: This, and the one below, needs to be moved to start at 2 and have reqs added to get up to 4.

region/brinstar/green/Early Supers Room.json
blkerby: If you can come in at 2, you can get to 1 (via 4), unlock the door at 1, exit and re-enter the room to clear the obstacle. This is assuming that resetting the room through an unlocked door is always possible, which is true in Map Rando (though it may have an energy cost for heat damage which currently isn't taken into account). If that assumption is wrong (e.g. in a randomizer with asymmetric door connections), the strat will still be sound, just won't be applicable in a case where it could be.

We could have a second strat [4, 1] to cover those other possibilities, and it just wouldn't be able to unlock the door in some cases. Do you think it's worth having a second strat? Or any idea how we could represent it using just one strat?

region/brinstar/green/Early Supers Room.json
blkerby: Done. Yeah, I had it like that before and then was thinking it would be ok, since if you could get to 4 in the first place then you could get up there again. But there could be scenarios where that might not be true (e.g. if 4 were your starting location) so it's safer to do it this way like you suggest.

region/brinstar/green/Early Supers Room.json
```diff
@@ -202,6 +202,34 @@
       },
       "requires": []
     },
+    {
+      "link": [1, 1],
+      "name": "Early Supers Leave With Spark To The Left",
+      "notable": true,
+      "requires": [
+        {"obstaclesNotCleared": ["A"]},
+        "canCrouchJump",
```
osse101, line 211:
```suggestion
        "h_canCrouchJumpDownGrab",
```

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -350,6 +363,19 @@
...
         "canShinechargeMovement"
       ],
+      "unlocksDoors": [
+        {"types": ["missiles"], "requires": ["never"]},
+        {"types": ["super"], "requires": []},
+        {
+          "types": ["powerbomb"],
+          "requires": [
+            {"or": [
+              "canCrumbleJump",
```
osse101, line 373: Little bit harder than a normal crumblejump.  A walljump instant morph works too, I think.

osse101 (commented):
Reminds me of the old obstacle system.

What we want to allow is performing a strat from node to door lets you open a lock on the door as you go through. Usually with supers, but the etecoon etank powerbomb strat is cool.  Missiles/PowerBombs would actually be able to continue through with fewer frames remaining in some cases.

The other case is opening the door and going back to your start node.  So some door unlocks will look just like door opens, but with the possibility of reloading the room.  Tracking lock state seems like it'd be more punishing to play while being complex to randomize.

strats.md
```diff
@@ -878,3 +881,55 @@ A strat with `"bypassesDoorShell": true` has an implicit tech requirement of `ca
...
+
+If an `unlocksDoors` property is not specified, then it is assumed to be an empty array. If a strat has any `doorUnlockedAtNode` requirements (including an implicit one based on having an exit condition without a `bypassesDoorShell`), then the `unlocksDoors` property should be specified explicitly and include items for each of the three possible types "missiles", "super", and "powerbomb" (or the catch-all "any") for each applicable node. The only exception is if the strat has no entrance condition then the starting node of the strat does not need to be included in the `unlocksDoors` property; in this case, the door could be unlocked immediately prior to the strat being executed (e.g. by an implicit unlock strat; see below), so generally it would not be necessary to describe how to unlock it as part of the strat. Where applicable, cases should be included for all three types of door unlock methods, "missiles", "super", and "powerbomb" (or using "any" as a catch-all), in order to support randomizers which may modify the door colors. The case of gray doors only needs to be included in places where the vanilla game has gray doors.
```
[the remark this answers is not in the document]
osse101: `"any"` -> `"ammo"`

region/tourian/main/Blue Hopper Room.json
```diff
@@ -124,6 +125,7 @@
           "enemies": [["Blue Sidehopper", "Blue Sidehopper"]]
         }}
       ],
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
```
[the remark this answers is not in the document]
blkerby: Makes sense. Added a simple `unlocksDoors` on the other [1, 2] strats that can open with Supers for free, or take an extra enemy hit to open with Missiles or PB. More advanced dodging strat(s) like you mentioned would be good to add too; I imagine they'll need a detailed description, since it sounds very tricky. If I remember it's hard to control exactly how the left Hopper behaves.

We'll need to go over all the rooms in the game to refine the `unlocksDoors` strats. This PR is just intended as a conservative starting point. At the moment, the randomizer doesn't even allow putting a lock on the left door of Blue Hooper Room or Amphitheatre.

region/brinstar/green/Etecoon Energy Tank Room.json
blkerby: Makes sense. Added a walljump instant morph option, with a "canTrickyJump" requirement on it and on the crumblejump option.

osse101 (approved):
424 warnings makes it harder to see the errors so we will want to work on these right away.

Or the `print:Warnings` can be taken out until after there has been an effort to add the majority of the `unlockDoors`.

PR #1280, opened 2023-12-26
Misc Strats: Some Speed Jumps and IBJing Above Spikes

kjbranch:


region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -361,66 +395,6 @@
...
-        ]},
-        {"or": [
-          "canIframeSpikeJump",
-          {"spikeHits": 2}
-        ]}
-      ]
-    },
     {
       "link": [2, 1],
       "name": "X-Ray Access SpikeJumps with HiJump",
```
osse101, line 400: This strat can be split to use 3 and 4.
`spikeHits:3` = [2,3] spikehits:1 + [3,4] spikehits:2
`or#1` = [2,3]
`or#2` = [4,1]
And the [4,1] part would be notable.  The reset isn't as important.

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -557,6 +518,133 @@
       "requires": [
         "h_canCrystalFlash"
       ]
+    },
+    {
+      "link": [2, 3],
```
osse101, line 523: [2,3] could use a damageboost strat since it's not as hard as the full room waver ride.  But it would still be notable and not that much easier, so maybe don't bother 🤷

region/norfair/east/Double Chamber.json
```diff
@@ -639,12 +639,9 @@
...
     {
       "link": [3, 4],
-      "name": "Double Chamber Spike IBJ",
+      "name": "Spike IBJ",
       "requires": [
-        "h_canNavigateHeatRooms",
-        "canIframeSpikeJump",
-        {"spikeHits": 1},
-        "h_canIBJ",
+        "h_IBJFromSpikes",
```
osse101, line 644: heated version of the helper?  Since it has that `or` which I think means number of attempts.

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -573,6 +535,36 @@
...
+          "type": "contact",
+          "hits": 1
+        }},
+        "canHorizontalDamageBoost",
+        {"or": [
+          "canIframeSpikeJump",
+          {"spikeHits": 1}
+        ]}
+      ],
+      "note": "Damage boost off of the Fireflea in order to get onto the upper spikes."
```
osse101, line 566: This jump felt a little easier than the hijumpless version, but still harder than the fast Waver boost.
I'd rather wait for the waver to come back than set this up.  But I don't the timing for when the waver comes back to a good spot yet.

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -573,6 +535,36 @@
...
       "note": "Use the Fireflea when it moves to the right to cross the first gap."
     },
+    {
+      "link": [2, 3],
+      "name": "HiJump and Wall Jump",
+      "requires": [
+        "HiJump",
+        "canPreciseWalljump",
+        "canIframeSpikeJump",
+        {"spikeHits": 2}
```
osse101, line 545: Jumping on the fire flea after the walljump can save some health.  Not really well known/ intuitive though.

region/brinstar/red/Red Brinstar Fireflea Room.json
osse101: The [4, 1] or has
- A pixel precise positioning with moonwalk/xray/morph turnaround + insane jump +  insanewalljump
- A spike damage boost + iframeSpikeRun + TrickyJump + ~precise walljump

kjbranch:
Added a helper `h_canBackIntoCorner`, which I can apply to other strats later if you like it. I can also convert IBJ above spikes into a tech if we want to move it from `H`->`VH`.

osse101:
I was just thinking we should bundle canBackIntoCorner together as a set.  But I was thinking it would be a tech.  Morph turnaround, and XRay turnaround aren't commonly used for pixel precise positioning.  But then I also haven't seen anyone have trouble with it, or just always run with moonwalk.  And a tech for this would be awkward with how xrayturnaround and moonwalk already exist to do the same thing.  So it would make more sense for morphTurnaround to be the new tech, if needed.

As a widespread helper, there should be another version `h_canBackOrJumpIntoCorner` which wouldn't have any item requirements because you spinjump into the wall to get the pixel alignment.  So a tech requirement.  And the main use is Pink Brin Hopper room (wave gate).  Or the notable strat is selected while the `backIn` Items are not acquired.  I've basically never seen anyone do the spinjump align setup for this trick; they'd rather not align and take multiple shots at the trick.  And I think it would also broadly be used for `canTrickyDashJump` and `Back to the wall shortcharges`.  But those situations being so precise, you could just assume it to be known.

osse101:
IBJ above spikes sounds pretty good.  JumpIntoIBJ isn't a first try success trick at hard.  So it is a good tech for falling floors, springball into ibj, and maybe the water escape ibj.  SandIBJ is it's own tech.  Heat room jumpIntoIBJ is very fuzzy, and it actually feels kinda bad to have to do it on a timer.  JumpIntoIBJ over spikes messes with your positioning due to the knockback, and short iframe timer.  I'd like it in `VH`.  Then `RollAroundOnSpikesAndIBJ` only makes sense as something you would do with artificial morph to me, if those cases are still around in logic.

blkerby:
The helpers here are requiring either "canTrickyJump" or 3 extra leniency hits, so that puts it in Hard only if you have energy to take 4 hits, which seems reasonable? You could put the whole thing in Very Hard, but I feel like it's kind of nice to introduce tech like this gradually (possible on a lower tier but with leniency)?

Heated IBJs I think could use greater lenience on lower settings. We had the idea a while ago to represent them using a new logical requirement that would encode the number of tiles climbed; then heat frames could be automatically determined based on whether canDoubleBombJump is enabled or not, and extra lenience could be included that way too. But that needs some schema changes, probably out of scope for this PR?

kjbranch:
If there's nothing pressing on this PR, I would prefer having this one merged and then adding and using new tech in a separate PR.

PR #1282, opened 2023-12-27
G-Mode Aqueduct Cleanup

kjbranch:
Clean up Aqueduct by adding new nodes, as suggested. This helped fix a couple small errors, and enabled an easy addition of a couple more strats - namely, starting a snail climb while morphed with only spring ball.


blkerby (approved):
Looks good.

PR #1285, opened 2023-12-28
Add Missing G-Mode Strats in Wrecked Ship

kjbranch:


region/wreckedship/main/Basement.json
```diff
@@ -350,11 +386,37 @@
...
+          "morphed": true
+        }
+      },
+      "requires": [
+        "f_DefeatedPhantoon",
+        {"ammo": {"type": "PowerBomb", "count": 2}}
+      ],
+      "note": [
+        "On room entry, dodge the atomic and quickly place a Power Bomb on the left half of the platform, then a second in order to kill both Atomics.",
+        "Placing it on the right half will hit the bomb block and it will no longer be passable.",
```
osse101, line 404: You are safe until about halfway between the platform and the wall.  Where the large foreground structure is.

region/wreckedship/main/Wrecked Ship East Super Room.json
```diff
@@ -166,6 +174,31 @@
...
+      "clearsObstacles": ["A"]
+    },
+    {
+      "link": [1, 1],
+      "name": "Coverns and Atomics Farm (Power Off)",
+      "requires": [
+        {"not": "f_DefeatedPhantoon"},
+        "canRiskPermanentLossOfAccess",
+        {"or": [
+          "Ice",
```
osse101, line 193: Ice isn't really strong enough to make the covern farm good, and the atomic farm is still tricky.
Screw only helps with the Coverns.

As an R-Mode farm spot, only the coverns respawn so you would want items which kill them faster.

region/wreckedship/main/Wrecked Ship East Super Room.json
```diff
@@ -276,6 +309,26 @@
...
+      "name": "WS East Supers R-Mode Standup Clip (Right to Left)",
+      "notable": true,
+      "requires": [
+        {"obstaclesCleared": ["A"]},
+        "canCeilingClip",
+        "canBePatient",
+        "Morph"
+      ],
+      "note":  [
+        "In R-Mode, kill the Coverns until there is Energy in Samus's Reserves. Get into the Morph tunnel and go to the far left.",
```
osse101, line 324: S'pose the Player would refill on the coverns before and after the clip.  So canBePatient fits.

region/wreckedship/main/Basement.json
kjbranch: Weird. I tried this 2-3 times and couldnt get them overloaded. Not sure why, but it's working now

PR #1286, opened 2023-12-29
More Crateria shinecharges and sparks

blkerby:


blkerby:
A lot of these strats probably should go in "Expert" but we might need a new tech, maybe a higher-level version of "canShinechargeMovementComplex" or of "canTrickyJump"?

osse101:
There could be another tier of `ShinechargeMovement` for things like waiting in shinespark windup, morph-unmorph to avoid activating the shinespark, and maybe there are more quirky techniques.  It would come up a lot with spike suit movement.  Like how RBO does maridia tube with an unmorph, and underwater crouch jumps, and spinjump grapple-jumps.  West SandHole is another room with particular movements to keep the shinespark alive.

`canTrickyJump` needs another tier to separate hard jumps from frame perfect jumps.

osse101 (commented):
Moat ocean fly would be expert, but with some more frames you could jump to the pedestal and from there spark through the door, and it wouldn't be too hard.

Bowling path hero spark right to left with plasma is probably very hard.
Left to right hero spark is more like expert.
Damage boosting through the room shouldn't be expert, but it is something you practice a lot for any% so it's hard to judge.  Probably not reliable enough for VHard.

Final Bombway feels like Extreme.  Jumping into the morph tunnel without wasting frames is a pain and you potentially don't have a way back.

blkerby:
I added a new tech "canShinechargeMovementTricky" to differentiate the Expert-level strats here.

With the hero shots to open the door, I think the main thing that distinguishes Expert-level strats from Very Hard is whether you can bring the camera close enough to the door to eliminate the risk of the shot despawning. Added several alternative strats now in Moat and Bowling Alley Path where this is possible.

I've been thinking it may make sense to move "canHeroShot" tech from Hard to Very Hard; I can't think of an example that uses it and actually seems to belong in Hard. For these shinecharge hero shot strats, they all have "canShinechargeMovementComplex" putting them in at least Very Hard so it doesn't matter much anyway.

I hadn't noticed the [12, 2] strat in West Ocean, removed that now. Starting a cross-room strat at the door (when logically sound to do so) has an advantage that it's simpler to infer that the door can be unlocked before doing the strat, so we won't have to put "unlockDoors" on it. On the other hand, [12, 2] is more descriptive of where the strat actually begins, so I'm not sure which is better. I guess I'll leave it [2, 2] for now.

I made the Final Missile Bombway Shinespark into a notable so we can put it in Extreme.

I couldn't get the damage boost strat in Bowling Alley Path; I'm not familiar with how it's supposed to work exactly. Maybe you could add it?

tech.json
```diff
@@ -1089,6 +1089,16 @@
...
                     "The ability to make multiple movement actions with a shinespark charge timer running to get into position to shinespark.",
                     "This reposition may be necessary to avoid obstacles, conserve energy, etc."
+                  ],
+                  "extensionTechs": [
+                    {
+                      "name": "canShinechargeMovementTricky",
+                      "requires": [ "canShinechargeMovementComplex" ],
+                      "note": [
+                        "The ability to make precise and potentially unintuitive movement actions with a shinespark charge timer running to get into position to shinespark.",
+                        "A common example is using a well-timed hero shot to spark through an off-camera door."
```
osse101, line 1099:
```suggestion
                        "A common example is using the Shinespark Wind-Up animation to wait for a hero shot to open an off-camera door."
```

PR #1293, opened 2024-01-03
Blue Brinstar shinecharges/sparks

blkerby:


region/brinstar/blue/Construction Zone.json
```diff
@@ -177,6 +214,23 @@
       "gModeRegainMobility": {},
       "note": "Shoot a Super at the wall, while the Geemer is moving vertically. If it is on the bottom of the shot blocks, it will not fall."
     },
+    {
+      "link": [2, 3],
```
blkerby, line 218: A [1, 3] version of this strat might be possible, by shooting the blocks while down-backing like people do in speedruns, but I'm bad and couldn't get it.

osse101: The downback is exactly as fast as shooting through the shot blocks from above.  I could get a `leaveWithSpark` using 160 frames.

region/brinstar/blue/Construction Zone.json
```diff
@@ -177,6 +214,23 @@
...
+      "name": "Carry Shinecharge",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 160
```
osse101, line 222: Best I could get with room grinding was 165

osse101: 159 new best 💪Had to morph immediately on entry.
Also Morph is missing from requirements.

region/brinstar/blue/Construction Zone.json
blkerby: Ah, I forgot about trying to only leaveWithSpark, was able to get the spark out 160 frames too now. Seems like only leaving in "bottom" position is feasible. 

One thing that I'm just noticing is that if you spark out with a crouch jump, it puts your vertical position in a bad place where Samus' hitbox intersects all four doorway tiles, and so doesn't satisfy the requirements of either "bottom" or "top". But you can hold angle down while initiating the spark from crouch to prevent getting the crouch jump height, making it a valid "bottom" position.

region/brinstar/blue/Construction Zone.json
blkerby: Good catch, added Morph requirement now

region/brinstar/blue/Construction Zone.json
osse101: I noticed that crouch-jump-sparks put you a little higher.  I forgot that `canRJump` cancelled that out.

PR #1307, opened 2024-01-10
G-Mode Migration: Boyon Gate, Crab Hole

kjbranch:
Manually migrating the remaining strats which had multiple fromNodes, and adding artificial morph strats from the bottom to top of Crab Hole

### DIFF
--- a/region/maridia/outer/Crab Hole.json
+++ b/region/maridia/outer/Crab Hole.json
@@ -272,7 +272,13 @@
             "h_canArtificialMorphBombHorizontally",
             {"or": [
               "Bombs",
-              {"ammo": {"type": "PowerBomb", "count": 1}}
+              {"ammo": {"type": "PowerBomb", "count": 1}},
+              "canTrickyJump",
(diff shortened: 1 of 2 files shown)


region/maridia/outer/Crab Hole.json
```diff
@@ -385,6 +232,57 @@
...
+        {"or": [
+          "Morph",
+          "SpringBall",
+          {"and": [
+            "Gravity",
+            "h_canArtificialMorphIBJ"
+          ]},
+          {"and": [
+            "Gravity",
+            "h_canArtificialMorphBombHorizontally",
```
blkerby, line 272: Similar to the strat going the other direction, It's possible to save a PB by rolling on the platform with Gravity turned off. This direction is trickier because you need to wait a moment for the crab to move, and since you're stopping I found you have to move back toward the transition a bit in order to have space to gain enough momentum to make it.

blkerby (approved):

osse101 (approved):

PR #1308, opened 2024-01-10
Tourian shinecharges

blkerby:


region/tourian/main/Tourian Escape Room 2.json
```diff
@@ -100,11 +100,92 @@
...
+      "link": [1, 2],
+      "name": "Come in Shinecharged, Leave With Spark",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 95
+        }
+      },
+      "requires": [
+        {"or": [
+          "ScrewAttack",
```
osse101, line 113: I don't think screw attack works without another jump... maybe sometimes.
A super to kill the pirate would work.

strats.md
```diff
@@ -349,6 +350,22 @@ In all strats with an `entranceCondition`, the `from` node of the strat must be
 
 Each of these properties is described in more detail below.
 
+### Come In Normally
+
+A `comeInNormally` entrance condition represents the need for Samus to enter the room through this door, with no other particular requirements. It has no properties.
```
osse101, line 355: This looks like a way to ensure that you just entered the room.
I thought it was a condition for checking that Samus is in some generic state that is able to move normally.
Because `canBabySkip` leaves you in a bouncing Morph ball state, except for the insane version, and does need a condition to model that.

region/tourian/main/Seaweed Room.json
```diff
@@ -94,11 +95,110 @@
...
+      "link": [1, 3],
+      "name": "Come in Shinecharging, Leave Shinecharged (Wave)",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 4,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "Wave",
```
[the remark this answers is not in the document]
blkerby: Interesting. Looking at this again, seems like using Spazer should be "canShinechargeMovementTricky" while Wave can just be "canShinechargeMovementComplex". Updated the strat.

region/tourian/main/Seaweed Room.json
```diff
@@ -94,11 +95,110 @@
...
+          "length": 4,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canShinechargeMovementTricky",
+        {"shinespark": {"frames": 2}}
+      ],
+      "exitCondition": {
+        "leaveWithSpark": {}
```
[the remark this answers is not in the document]
blkerby: Sure enough. Added a strat for it.

region/tourian/main/Tourian Escape Room 2.json
blkerby: Good catch.

strats.md
blkerby: Makes sense. At some point I would like to propose we make it so that entrance/exit conditions are the only way to cross between rooms. At the moment, there are implicit strats that cross rooms (i.e. the normal way to cross rooms, in cases where no cross-room strat is being used), and it potentially creates a lot of little issues, due to hidden assumptions which may not always be satisfied. By making the entrances/exits all explicit, it can make it easier to get everything right.

For example, except in cases with a spawnAt, logically you could go through a door into a heated room then back into the original room with no requirements; this could be done to reset obstacles for instance, which would be unsound, since in reality you take some heat damage to open the door and go back through.

region/tourian/main/Blue Hopper Room.json
```diff
@@ -213,6 +213,31 @@
         "leaveWithGModeSetup": {}
       }
     },
+    {
+      "link": [1, 2],
+      "name": "Come in Shinecharged, Leave With Spark (Tank the Damage)",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 95
```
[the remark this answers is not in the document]
osse101: You have to land before you can turn around.

region/tourian/main/Seaweed Room.json
```diff
@@ -94,11 +95,139 @@
...
+      "name": "Come in Shinecharging, Leave Shinecharged (Missiles)",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 4,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        {"ammo": {"type": "Missile", "count": 7}}
+      ],
```
osse101, line 163:
```suggestion
      "requires": [
        "canShinechargeMovementTricky",
        {"ammo": {"type": "Missile", "count": 7}}
      ],
```

osse101 (dismissed):

PR #1311, opened 2024-01-11
Brinstar unlocksDoors

blkerby:
In addition to adding `unlocksDoors` there are a couple more little things:
- A new "Speedy Airball" strat in "Blue Brinstar Boulder Room" (pointed out by insomniaspeed).
- Fix to Red Tower 1 -> 5 "Leave with Moondance" strats; it seemed to be missing requirements for how to get to the top of the room; freezing the Rippers alone is not enough.

### DIFF
--- a/region/brinstar/blue/Blue Brinstar Boulder Room.json
+++ b/region/brinstar/blue/Blue Brinstar Boulder Room.json
@@ -134,7 +134,8 @@
         "leaveShinecharged": {
           "framesRemaining": 100
         }
-      }
+      },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}]
     },
     {
       "link": [1, 2],
@@ -152,6 +153,7 @@
           "framesRemaining": 60
         }
       },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
       "note": [
         "Shoot the bridge to see where it ends.",
         "Opening the door requires a shot that is fired while Samus has some momentum from running to the right.",
@@ -235,6 +237,16 @@
       "note": [
         "It's a delayed walljump while the water is low, followed by a tight walljump off the bridge.",
         "This strat is easily bypassed by jumping through the door, or if the door can be opened."
+      ]
+    },
+    {
+      "link": [2, 1],
+      "name": "Speedy Airball",
+      "requires": [
+        {"doorUnlockedAtNode": 2},
+        "SpeedBooster",
+        "canLateralMidAirMorph",
+        "canCarefulJump"
       ]
     },
     {
--- a/region/brinstar/blue/Blue Brinstar Boulder Room.json
+++ b/region/brinstar/blue/Blue Brinstar Boulder Room.json
@@ -153,7 +153,10 @@
           "framesRemaining": 60
         }
       },
-      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
+      "unlocksDoors": [
+        {"types": ["powerbomb"], "requires": []},
+        {"types": ["missiles", "super"], "requires": ["canTrickyJump"]}
+      ],
       "note": [
         "Shoot the bridge to see where it ends.",
         "Opening the door requires a shot that is fired while Samus has some momentum from running to the right.",
--- a/region/brinstar/kraid/Warehouse Kihunter Room.json
+++ b/region/brinstar/kraid/Warehouse Kihunter Room.json
@@ -558,6 +558,10 @@
       "exitCondition": {
         "leaveWithSpark": {}
       },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],
       "note": "Charge a spark, then break the shot blocks, drop through, and spark out the bottom right door."
     },
     {
--- a/region/brinstar/pink/Dachora Room.json
+++ b/region/brinstar/pink/Dachora Room.json
@@ -304,56 +304,6 @@
     },
     {
       "link": [1, 3],
-      "name": "Blockade Leave with Moondance",
-      "requires": [
-        "h_canUseMorphBombs",
-        "canMoondance",
-        "SpeedBooster"
-      ],
-      "exitCondition": {
-        "leaveWithStoredFallSpeed": {
-          "fallSpeedInTiles": 1
-        }
-      },
-      "note": [
-        "Break exactly the lower-middle-right and top-right Bomb Blocks, leaving the upper-middle-right and bottommost Blocks intact.",
-        "Clear all enemies before starting.",
-        "Unmorph while on the top block to begin the Moondance.",
-        "Wiggle out to the left and use SpeedBooster to run through to the right side door."
-      ]
-    },
-    {
-      "link": [1, 3],
-      "name": "Blockade Leave with Extended Moondance",
-      "notable": true,
-      "requires": [
-        "h_canUseMorphBombs",
-        {"or": [
-          "SpeedBooster",
-          {"and": [
-            "ScrewAttack",
-            "canDisableEquipment"
-          ]}
-        ]},
-        "canExtendedMoondance"
-      ],
-      "exitCondition": {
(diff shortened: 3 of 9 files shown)


region/brinstar/blue/Blue Brinstar Boulder Room.json
```diff
@@ -237,6 +239,16 @@
...
     },
+    {
+      "link": [2, 1],
+      "name": "Speedy Airball",
+      "requires": [
+        {"doorUnlockedAtNode": 2},
+        "SpeedBooster",
+        "canLateralMidAirMorph",
+        "canCarefulJump"
+      ]
```
kjbranch, line 250: This should be fine because at this difficulty it can be assumed you can cj+dg. I want to add a spawn at node here sometime soon too.

region/brinstar/kraid/Warehouse Kihunter Room.json
```diff
@@ -558,6 +558,10 @@
       "exitCondition": {
         "leaveWithSpark": {}
       },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
```
kjbranch, line 563: might be possible to lay a PB by the save door, but would need to be its own strat which has access to 3

region/brinstar/pink/Dachora Room.json
```diff
@@ -752,6 +719,56 @@
...
+      "exitCondition": {
+        "leaveWithStoredFallSpeed": {
+          "fallSpeedInTiles": 1
+        }
+      },
+      "note": [
+        "Break exactly the lower-middle-right and top-right Bomb Blocks, leaving the upper-middle-right and bottommost Blocks intact.",
+        "Clear all enemies before starting.",
+        "Unmorph while on the top block to begin the Moondance.",
+        "Wiggle out to the left and use SpeedBooster to run through to the right side door."
```
kjbranch, line 739: right/left are backwards now. Should mention that this can start on either side of the blockade and still work (i think?). Could be added to strats going out the left as well.

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -289,7 +290,9 @@
       "exitCondition": {
         "leaveWithSpark": {}
       },
-      "note": "Use the really small runway in the pit in the middle of the room. The door must be already opened."
+      "unlocksDoors": [{"types": ["ammo"], "requires": [], "useImplicitRequires": false}],
+      "note": "Use the really small runway in the pit in the middle of the room. The door must be already opened.",
+      "devNote": "Unlocking the door is free since obstacle A being cleared implies it was already unlocked earlier if needed."
```
kjbranch, line 295: maybe we can add a [4, 4] strat that opens the door as well. Not sure what all would be required, but it's something to think about after the rework to this room is merged

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -499,7 +506,8 @@
           "mustStayPut": false
         }},
         {"refill": ["Energy", "PowerBomb"]}
-      ]
+      ],
+      "resetsObstacles": ["A"]
```
kjbranch, line 510: good catch. I wonder if we might want to make `resetRoom` implicitly reset all obstacles?

region/brinstar/red/Red Tower.json
```diff
@@ -761,6 +859,11 @@
       "name": "Leave with Extended Moondance",
       "requires": [
         {"obstaclesNotCleared": ["A"]},
+        {"or": [
+          "HiJump",
```
kjbranch, line 863: curious on this. I had a very hard time getting a Beetom up here with just wall jumps - it would knock the Beetom off. I'm assuming you might want A not cleared. It may be nice to have another node at the top of the room with a Beetom. It could be helpful to reduce the duplication and inconsistencies here and with g-mode setups, and allow for a more detailed description.

region/brinstar/blue/Blue Brinstar Boulder Room.json
```diff
@@ -152,6 +153,7 @@
           "framesRemaining": 60
         }
       },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
```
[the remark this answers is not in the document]
blkerby: Good point, added a "canTrickyJump" req in those cases now. It really does seem pretty tricky because you need to know basically exactly where the end of the invisible ledge is.

region/brinstar/kraid/Warehouse Kihunter Room.json
blkerby: Makes sense, I was thinking we could do another pass through all the rooms later to add strats like that

region/brinstar/pink/Dachora Room.json
blkerby: Hmm I'm not seeing what is backwards? This strat starts and ends at the right door, so naturally it would start from the right side; with Morph Bombs you can freely pass between the two sides anyway (before gaining the high fall speed). There's a separate [1, 1] strat for the left door, which doesn't need Speedbooster since the wiggle out at the end already puts you on the left side.

region/brinstar/red/Red Brinstar Fireflea Room.json
blkerby: Makes sense

region/brinstar/red/Red Brinstar Fireflea Room.json
blkerby: This could be awkward to make work properly in the randomizer. How about an alternative where we add a test to check that any strat with `resetRoom` has `resetsObstacles` covering all the obstacles? Either way, at the moment there's not a way to make the `resetsObstacles` conditional, so if `resetRoom` appears in an `or` then we could need to split those into separate strats, which will take some manual work.

region/brinstar/red/Red Tower.json
blkerby: In this scenario we also have Ice to freeze the Rippers, so the wall jumps are only needed to get up the big gaps before the Rippers. I tested it now and while the Beetom can be annoying and fall off, it is possible to get it to the top.

region/brinstar/pink/Dachora Room.json
kjbranch: I dont see how you can wiggle to the left. Do you wiggle to the right, or can you start by fully breaking all but the left column before starting?

region/brinstar/red/Red Brinstar Fireflea Room.json
kjbranch: looks like tricky jump + wave can open a blue door, or PBs can open a yellow one (probably tricky for obscurity)

It's also possible to open the door with a super/normal shot when getting to 4, but thats pretty tricky too

kjbranch:
I'd say this is good to merge after checking on the comment in the Dachora room. The fireflea stuff can be added now or later

region/brinstar/pink/Dachora Room.json
blkerby: Oh I see what you mean. I don't see a way to break the blocks in the correct pattern from the right side, so I think it needs to be started from the left side. I converted these [3, 3] strats to [1, 3] now and added an "obstaclesNotCleared" requirement on the bomb blocks.

region/brinstar/pink/Dachora Room.json
```diff
@@ -302,6 +305,60 @@
...
+          ]}
+        ]},
+        "canExtendedMoondance"
+      ],
+      "exitCondition": {
+        "leaveWithStoredFallSpeed": {
+          "fallSpeedInTiles": 2
+        }
+      },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
```
kjbranch, line 351: Thought these couldn't use PBs to unlock doors?

blkerby:
As long as the blocks weren't broken before, you can break open the correct
ones from the left side and then freely pass back and forth with Morph. So
you can unlock the door before doing the moondance

PR #1312, opened 2024-01-12
Migrate remaining adjacentRunway and canComeInCharged

blkerby:
This eliminates all the remaining `adjacentRunway` and `canComeInCharged` logical requirements, migrating the strats to use the new entrance conditions instead. It also removes them from the schema and docs.

Migrating the remaining runways (to `leaveWithRunway` exit conditions) will be done in a separate PR.

Handling the temporary blue strats in Mickey Mouse Room required adding a new entrance condition "comeInWithTemporaryBlue", which hopefully will be useful in other situations too.

The main strat I wasn't sure about was "SpeedKeep for Temporary Blue, Bounce on Runway" in Crateria Super Room. It had 13 unusable tiles, extending into the previous room, which we can't represent with the existing entrance conditions. I may be mistaken since I'm not really familiar with how speed keep works, but it looked like this strat was only an alternative to "canSlowShortCharge", and at the difficulty level expected of canSpeedKeep this seemed unnecessary, so I thought it may be ok to just delete it. Eventually I would like to add an entrance condition for jumping into a room while blue, with constraints on speed and/or position through the transition, but it may take some time to figure out exactly what this should look like.


blkerby:
Also, before this could be merged, we need some solution for the strats in Botwoon ETank Room that expect coming into the room shinecharged with a cross-room jump. The comeInShinecharged entrance condition currently doesn't have a way to require that the neighboring room is air, which is something we'd need to check.

osse101:
> he main strat I wasn't sure about was "SpeedKeep for Temporary Blue, Bounce on Runway" in Crateria Super Room. It had 13 unusable tiles, extending into the previous room, which we can't represent with the existing entrance conditions. I may be mistaken since I'm not really familiar with how speed keep works, but it looked like this strat was only an alternative to "canSlowShortCharge", and at the difficulty level expected of canSpeedKeep this seemed unnecessary, so I thought it may be ok to just delete it.

Right, the purpose was just for cases where you are barely shortcharging, and need to bounce on the platform instead of the spikes.  So its pretty safe to delete.

region/norfair/east/Lava Dive Room.json
```diff
@@ -351,6 +351,12 @@
       "link": [2, 1],
       "name": "Hjless Lava Dive BounceBall",
       "notable": true,
+      "entranceCondition": {
+        "comeInRunning": {
+          "minTiles": 4,
+          "speedBooster": "any"
```
osse101, line 357:
```suggestion
          "speedBooster": "false"
```

region/norfair/east/Lava Dive Room.json
```diff
@@ -311,19 +311,19 @@
     {
       "link": [2, 1],
       "name": "Lava Dive BounceBall",
+      "entranceCondition": {
+        "comeInRunning": {
+          "minTiles": 4,
+          "speedBooster": "any"
```
osse101, line 317:
```suggestion
          "speedBooster": "false"
```

osse101 (commented):
A `comeInWithTemporaryBlue` can follow a `leaveWithTemporaryBlue`, by chaining it from some runway.  But, future condition to add.

`comeInShinechargedJumping` surely needs a runway so you have some runspeed, if needed.  Right door aqueduct jumping to the speedblocks has come up a lot recently watching streams.

blkerby:
> A `comeInWithTemporaryBlue` can follow a `leaveWithTemporaryBlue`, by chaining it from some runway. But, future condition to add.
> 
> `comeInShinechargedJumping` surely needs a runway so you have some runspeed, if needed. Right door aqueduct jumping to the speedblocks has come up a lot recently watching streams.

For this application in Botwoon ETank, no runway is needed so that's why I didn't add it yet. In general for sure it would be useful to be able to specify a "minTiles" runway length. This would be non-trivial though to make work, because it needs more information on the exit condition side:
- we would need to add runway data to each `leaveShinecharged` strat.
- we would need to define how to match with `leaveWithRunway`. Basically the part of the runway used to gain the required speed would need to be subtracted from the part of the runway used to gain the shinecharge; or you could gain the shinecharge by running in the opposite direction, and use the full runway length both for the shinecharge and for gaining speed (but at the cost of shinecharge frames for the return run).

PR #1313, opened 2024-01-12
G-Mode Migration: Glass Tunnel, Main St., Mt. Everest pt.1

kjbranch:
This is primarily just running the scripts and fixing a few small errors and I think it's fine to merge:
- Reordering the entrance/requires/exit conditions
- Reducing the requirements to get leave a doorway with no doorcap (before it had to get to the node that comes in with g-mode)
- Removing some useless strats in Mt. Everest
  - [7, 1] and [7, 3] are free with Gravity alone. so converted strats to instead go [6, 7] and [2, 7] to reduce duplication and requirements.

There are still many things to do, including migrating strats with multiple fromNodes, reducing the requirements from [X, 6] in Mt. Everest, and adding some missing strats and options within strats, and more.


region/maridia/outer/Mt. Everest.json
```diff
@@ -906,6 +844,35 @@
         "In the likely event of failure, the room will need to be reset and the crabs repositioned for the next attempt."
       ]
     },
+    {
+      "link": [2, 7],
```
blkerby, line 846: I think these [2, 7] G-mode strats may be pointless, since with Gravity alone you can just Gravity jump 2->7 (though that seems to be missing from the logic).

region/maridia/outer/Mt. Everest.json
```diff
@@ -2123,6 +2016,41 @@
       ],
       "gModeRegainMobility": {}
     },
+    {
+      "link": [6, 7],
```
blkerby, line 2018: I think these two [6, 7] strats could be simplified to a [6, 2] strat that just falls down, since you can Gravity jump from 2 to 7.

blkerby (approved):

PR #1316, opened 2024-01-13
Add G-Mode Strats to Everest and Glass Tunnel, Update Glass Tunnel

kjbranch:
Everest
- Reduce unnecessary requirements from the old schema which were used to leave at 6
- Reduce duplication by adding a g-mode node at 6
- Add more options for setting up g-mode out the top 3 doors
- Fix some errors and add some strats

Glass Tunnel
- Migrate strats with multiple fromNodes
  - This was difficult, as g-mode strats must start at a door, but flag yielding strats must start at event nodes
- Add an obstacle to fix these and remove a now unnecessary node (still need to update the room diagram)

### DIFF
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -568,6 +568,7 @@
       },
       "requires": [
         {"or": [
+          "Morph",
           "SpringBall",
           {"and": [
             "Gravity",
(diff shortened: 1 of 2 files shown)


region/maridia/outer/Mt. Everest.json
```diff
@@ -562,6 +553,34 @@
...
+            "h_canArtificialMorphIBJ"
+          ]},
+          {"and": [
+            "Gravity",
+            "h_canArtificialMorphBombHorizontally"
+          ]},
+          {"enemyDamage": {
+            "enemy": "Sciser",
+            "type": "contact",
+            "hits": 1
```
blkerby, line 583: I think this needs 2 crab hits unless the morph tunnel is expanded?

region/maridia/outer/Mt. Everest.json
```diff
@@ -1889,62 +1867,38 @@
...
-        }
-      }
-    },
-    {
-      "link": [5, 9],
-      "name": "Base",
-      "requires": []
+      "note": [
+        "Bounce from the right peak into the morph tunnel.",
+        "Samus can't bounce twice in a single fall, so pause on the top platform or avoid it completely."
```
blkerby, line 1900:
```suggestion
        "Samus can't bounce twice in a single fall, so come to a stop on the top platform or avoid it completely."
```

blkerby (dismissed):

PR #1319, opened 2024-01-16
G-Mode Migration: Run Scripts for the Few Strats in Norfair

kjbranch:
Ran the script, then moved two strats in Post Croc Shaft that climb with a frozen Viola to start at the bottom of the room to save a Super


osse101 (dismissed):
Hope I didn't forget to add too many `canUpwardGModeSetup` as requirements.

PR #1322, opened 2024-01-18
G-Mode Migration: Remaining Strats

kjbranch:
- Migrate strats with multiple `fromNodes` in Crab Hole, Main Shaft, Etecoon E-Tank
- Migrate strats that now start at a different node to have fewer requirements in Dachora, Red Tower
- Migrate R-mode strats and other fixed in Red Tower

I believe this is the end of the migration. I can work on removing the schema elements after this is merged.


region/maridia/outer/Crab Hole.json
```diff
@@ -469,12 +490,13 @@
     {
       "link": [2, 1],
       "name": "G-Mode Morph IBJ",
-      "requires": [
-        {"comeInWithGMode": {
-          "fromNodes": [2, 3],
+      "entranceCondition": {
```
osse101, line 493: A springball grav jump could be added

osse101 (approved):

region/maridia/outer/Crab Hole.json
kjbranch: Good call. I think I've missed several of these. I should look through Maridia again and see if there are more sometime

kjbranch:
Combined duplicate strats that were migrated in an earlier PR but weren't deleted.

PR #1327, opened 2024-01-23
Frozen Enemy Runways in Brinstar

kjbranch:


osse101 (commented):
Think about adding some damage to the Beetom strats.  There is probably a need to add "Random Damage" throughout the project.

These felt pretty reasonable overall.  Strats with setups for the correct height are ~VeryHard.  But the Beetom bridge may be too much hassle for `VH` and potentially `expert` too.
Wallcrawlers are a lot like `canPreciseCeilingClip` where the only difference is that there is more to the strat than freezing the enemy in the right spot.
No-setup farm bugs weren't that bad when next to the spawner, but were really tedious when they had to fly a longer distance.

kjbranch:
Are you thinking these should be in `VH`? I was originally thinking `Expert`, since they can occur all over and only give you one attempt for the shinecharge/cross room jump

osse101:
I'm between Expert and Extreme.  With an easy setup, they are `VH` difficult to do, but you aren't thinking about frozen runways at that level and its hard to get good value out of the extra tile.

Setting up the enemies for what is presumably a really tight shortcharge could get tedious for expert.  I'm also not fond of `Just freeze it on the right pixel` as a strat, like `canPreciseCeilingClip` has you do.

osse101 (approved):

PR #1331, opened 2024-01-27
Add Horizontal Shinespark Tech

kjbranch:
- Add horizontal shinespark tech and use throughout.
- Add a more detailed description for how to do it on the ground and mid-air.
- Fix various errors noticed in shinespark strats.

### DIFF
--- a/tech.json
+++ b/tech.json
@@ -1069,27 +1069,29 @@
                 "This can be achieved by starting without horizontal momentum and pressing jump and then forward while still holding jump (or repressing jump).",
                 "These inputs can be done in rapid succession or with a short delay.",
                 "Shinesparking in water can be done the same way, but requires a slightly longer delay between the jump and forward presses, and it is particularly important to ensure Samus has no horizontal momentum."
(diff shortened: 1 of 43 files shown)


tech.json
```diff
@@ -1056,13 +1056,32 @@
...
+              "note": [
+                "The ability to shinespark horizontally from the ground.",
+                "This can be achieved by starting without horizontal momentum and pressing jump and then forward while still holding jump (or repressing jump).",
+                "These inputs can be done in rapid succession or with a short delay.",
+                "Shinesparking in water can be done the same way, but requires a slightly longer delay between the jump and forward presses, and it is particularly important to ensure Samus has no horizontal momentum."
+              ]
+            },
             {
               "name": "canMidairShinespark",
               "requires": [ "canShinespark" ],
```
osse101, line 1076: midairShinespark works better as an extension of horizontal.

kjbranch: Oh whoops. That's what I intended it to be.

PR #1338, opened 2024-01-31
Add spring ball jump and IBJ options for Tourian Escape 4 L->R

blkerby:
This also adds item requirements (Morph + Bombs) that appear to have been missing in the right-to-left strat "Tourian Escape Room 4 IBJ and Acid Bath".

Also adds a shinespark strat (coming in shinecharged) for getting up left->right walljumpless.

### DIFF
--- a/region/tourian/main/Tourian Escape Room 4.json
+++ b/region/tourian/main/Tourian Escape Room 4.json
@@ -154,6 +154,9 @@
         }
       },
       "requires": [
+        "canPrepareForNextRoom",
+        "canShinechargeMovement",
+        "canMidairShinespark",
         {"shinespark": {"frames": 53, "excessFrames": 13}}
       ]
     },
--- a/region/tourian/main/Tourian Escape Room 4.json
+++ b/region/tourian/main/Tourian Escape Room 4.json
@@ -139,7 +139,8 @@
           "HiJump",
           "SpaceJump",
           "canWalljump",
-          "h_canIBJ"
+          "h_canIBJ",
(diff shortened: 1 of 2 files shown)


region/tourian/main/Tourian Escape Room 4.json
```diff
@@ -139,11 +139,24 @@
...
         ]}
       ],
       "note": "The pirates are free to kill, although they take some time."
     },
+    {
+      "link": [1, 3],
+      "name": "Shinespark",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 15
```
osse101, line 153: 50 frames if just running in.  But entranceConditions have an implicit `canPrepareForNextRoom`?  Or at least you are aware of the next room since you know to try this strat.
Still should have `canShinechargeMovement` and `canMidairShinespark`

blkerby: Ah yeah, makes sense for this strat to have a `canPrepareForNextRoom` requirement since if you fall down without getting the spark off in time then you may be unable to get back to the door to retry. And I always forget about those other requirements.

Spin-jumping through the door seems like the reasonable thing to assume, though if the previous room is water (e.g. from doing a water shinecharge involving two rooms to the left) then the shinespark frames need to be a bit higher since the spin jump in will be low. Adjusted this and also updated the docs for `comeInShinecharged`. An alternative is we could use `comeInShinechargedJumping` for this and add a `physics` field to it, which could be "any" in this case.

PR #1342, opened 2024-02-01
Tighten heat frames for GT crumble item

blkerby:


blkerby:
I noticed this because I rolled an Insane Challenge seed for testing (with Ship start), which had GT at top-right of Landing Site, and it gave Space Jump to reach the crumble item instead of Speed Booster, which was surprising. With the somewhat lenient heat frames here, there wasn't enough energy to logically grab the item after shinesparking up Landing Site.

blkerby:
Also, at @kjbranch 's suggestion, just added a couple cross-room running jump strats which have even lower heat frames. With short runways this makes for a very tight jump (much harder than the crumble jump), so I added a "canInsaneJump" tech for that; I figure we've been wanting to add that for a while anyway. I'm not sure if it would be at Expert or Extreme level, probably Extreme?

At some point I'm also interested in adding an entrance condition for cross-room running jumps in heated rooms, where you could specify the in-room runway and the horizontal distance to travel, and it could compute the heat frames based on the achievable speed, depending on the runway length in the other room (and whether or not you have SpeedBooster). It wouldn't make that much difference here, but in long heated rooms like Rising Tide it could make a big difference especially for Space Jump strats where the higher run speed can be carried all the way across the room. So it could give us a way to refine heat frames in situations like that without having to add variant strats for every possible runway length in the neighboring room (and SpeedBooster vs. no-Speedbooster).

osse101:
Numbers are pretty tight but seem fine.
`comeInRunning` strats are missing from most heated rooms; just the ones where spacejump would save time are in already, and some special cases.  So they could begin to be added.
Calculating the health save from entering with more speed sounds tricky but would make a better model.  Controlling Samus tends to get harder when you are moving faster too.

PR #1350, opened 2024-02-07
Crateria Enemy Extend Runways

kjbranch:
I skipped some tough ones which may be added later:
- Flyway + Pre Map Flyway - The bugs are fast and don't add much. It doesn't look like the Reo can be used.
- West Ocean - Top right two doors, added a devNote explaining it.
- Gauntlet Entrance - A waver can be used relatively easily depending on which blocks are broken and when, but it only allows a single try.


region/crateria/east/Bowling Alley Path.json
```diff
@@ -437,6 +598,148 @@
...
+    {
+      "link": [2, 2],
+      "name": "Bowling Alley Path Double Frozen Choot Runway, Suitless (Right)",
+      "notable": true,
+      "requires": [
+        "h_canFrozenEnemyRunway"
+      ],
+      "exitCondition": {
+        "leaveWithRunway": {
+          "length": 14,
```
osse101, line 666: It feels like a 16 runway where a stutter 3 gets an in room shortcharge. Going in and out of water is a unique thing, so its hard to actually know.
The double freeze is fine for extreme, single freeze is probably ok on expert.  I don't know how to go about a triple freeze without the first ones thawing.

osse101: I used a crouch+angle down shot while standing between the two of them.
You can also use XRay to buffer the shot but then you need to avoid getting hit from the other choot.

kjbranch:
- Added `canTrickyEnemyExtendRunway` and used it throughout
- Timed the suitless runways in Bowling Pathway using Samus' horizontal speed - using the runway for a shinecharge will likely add a small amount of runway, as Samus gains some shinecharge frames but no speed while submerged. I took a good but not great runway length, as it is highly dependent on the water cycle.

region/brinstar/kraid/Kraid Eye Door Room.json
```diff
@@ -129,6 +129,7 @@
         {"or": [
           "Morph",
           {"and": [
+            "canTrickyEnemyExtendRunway",
```
osse101, line 132: Door 2 too

region/brinstar/red/Hellway.json
```diff
@@ -125,7 +125,7 @@
       "notable": true,
       "requires": [
         "canTrickyJump",
-        "h_canFrozenEnemyRunway",
+        "h_canTrickyFrozenEnemyRunway",
```
osse101, line 128: Should there be an option for using a Zeela now?  Here or at 2.

region/brinstar/kraid/Kraid Eye Door Room.json
kjbranch: I was only including it on door 1/3 when you need to run in the spikes to align the Zeb (no Morph). The top right door seemed much less tricky to me - I also didnt include it using a Zeb on the right door in Big Pink.

region/brinstar/red/Hellway.json
kjbranch: Definitely. Good catch.

PR #1351, opened 2024-02-08
Wrecked Ship Enemy Extend Runways

kjbranch:
I was using `canInsaneJump` as a proxy for annoying positioning of Coverns. I wonder if we should instead have a second tech `canTrickyEnemyExtendRunway`?

### DIFF
--- a/region/wreckedship/main/Wrecked Ship Main Shaft.json
+++ b/region/wreckedship/main/Wrecked Ship Main Shaft.json
@@ -335,6 +335,22 @@
     },
     {
       "link": [2, 2],
+      "name": "Leave With Runway - Frozen Atomic",
+      "requires": [
+        "f_DefeatedPhantoon",
+        "h_canFrozenEnemyRunway"
+      ],
+      "exitCondition": {
+        "leaveWithRunway": {
+          "length": 8,
+          "openEnd": 0
+        }
+      },
+      "note": "Position and freeze an Atomic to extend the runway. Maintain a half-tile gap between the runway to extend it as far as possible.",
+      "devNote": "It could be possible to freeze a Covern in its place, but it seems very difficult to position."
+    },
+    {
+      "link": [2, 2],
       "name": "G-Mode Setup - Get Hit By Covern or Atomic",
       "notable": false,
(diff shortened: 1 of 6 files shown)


region/wreckedship/main/Wrecked Ship Main Shaft.json
```diff
@@ -743,6 +760,36 @@
...
+          ]}
+        ]},
+        "canSlowShortCharge",
+        {"canShineCharge": {
+          "usedTiles": 14,
+          "openEnd": 0,
+          "steepUpTiles": 3
+        }},
+        "canTemporaryBlue",
+        "canSpringBallBounce"
```
osse101, line 783:
```suggestion
        "canTemporaryBlue"
```

osse101 (commented):
There is a set of easy extendRunways, and tedious ones.  A `canTrickyEnemyExtendRunways` could be made, and reduce the number of notables too.

kjbranch:
Thinking a bit more about `canTrickyEnemyExtendRunways`, I'm not quite sure what would be best. It seems like it should be put in `Extreme`, but we have very few strats that are currently above `Expert`. So there are a few paths we could take if we add this tech:
- Move some of the more difficult strats from `Expert` to `Extreme` by adding this new tech to them.
- Keep this tech in `Expert` and add it to some of the more difficult strats, so they are toggleable, but remain where they are.
- Add this new tech and only have it apply to a small number of strats (Would we add more of the most annoying runways, such as moving the Zebs to the top right doors of West Ocean?)

What do you think would be best?

osse101:
The way I see it; Beetoms, wallcrawlers, easy setup pipe bugs, atomics, choots don't feel too bad at expert.  And there could be more like mochtroids and snails who are cooperative.  I liked wallcrawlers the least, and choots should be worse.  Next-to-spawner freezes like hellway and red fish could go either way.

And the tricky set are the ones where we need to get vertical and horizontal right with different steps.  Zebs, coverns, fall off a ledge and freeze with the right timing strats.  Things that would take more time or attempts to setup.

osse101 (approved):

PR #1358, opened 2024-02-09
Add Morph req to Spooky Missile Temp Blue strat

blkerby:
This recently added strat seemed to be missing a Morph requirement. Also the note referenced Spring Ball which I don't think is required


kjbranch:
Whoops

PR #1360, opened 2024-02-10
Deep transition tech: X-Ray climbs in Big Pink and Everest

blkerby:
This felt like it should be a tech even though there's only these two places where we (currently) know of an application.

### DIFF
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -307,7 +307,8 @@
         {"id": 6},
         {"id": 7},
         {"id": 8},
-        {"id": 9}
+        {"id": 9},
+        {"id": 11}
       ]
     },
     {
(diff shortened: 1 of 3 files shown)


osse101 (approved):
I like it as a new tech.

These set-ups look good; work well.  `comeInWithSpark` sometimes cannot be done at the door, so it becomes RNG.  An at door spark would be more like `comeInShinecharged`.

Running start is nice to have documented, but turnaround start would always be used.

These exactly measured tile counts to me are still unfun because its hard to tell where to stand for whichever pair of rooms you are given.  I'd like an in game tool to help with tile counting and run speeds.  Samus hitbox outline would help and is typically banned in tournaments for it.  I think I'd open practice hack on the side to figure out which pixels to stand on before trying the strat.
I couldn't eyeball 6 tiles or 3.75 tiles but it does seem like you should be able to learn Samus' hitbox as it relates to the sprite, and same for floor tiles.  With enough practice.  And then look up the strat numbers on the logic pages.

region/maridia/outer/Mt. Everest.json
```diff
@@ -2113,6 +2114,28 @@
...
+    {
+      "link": [6, 11],
+      "name": "Shinespark Deep Transition X-Ray Climb",
+      "entranceCondition": {
+        "comeInWithSpark": {
+          "position": "top"
+        }
+      },
+      "requires": [
+        "canDeepTransition",
```
kjbranch, line 2126: I don't have time to review this PR right now, but does this strat need `h_EverestMorphTunnelExpanded`?

blkerby: I think it should be ok without it. With the tunnel not expanded, the strat still works, and in fact you don't even need a deep transition; any top-position shinespark would work. If a different project did something different, like expanding the height of the entire tunnel instead of just the entrance, then the strat wouldn't work.

PR #1362, opened 2024-02-10
Everest HiJump grapple bomb jump

blkerby:
We already had the crazy strat that does two consecutive grapple bomb jumps here. But with HiJump you can do it with a single grapple bomb jump.


blkerby:
Whoops, accidentally branched this off of the Mission Impossible branch. I guess review that one first?

PR #1365, opened 2024-02-11
Beetom 10 PB CF strats

blkerby:


region/norfair/east/Frog Speedway.json
```diff
@@ -101,6 +101,41 @@
...
+            }}
+          ]}
+        ]},
+        {"or": [
+          {"enemyDamage": {
+            "enemy": "Beetom",
+            "type": "contact",
+            "hits": 2
+          }},
+          "canInsaneJump"
```
osse101, line 127: Killing them with PBs is rather safe.
Lay PB when the first Beetom comes on screen. Run through the drops before they appear. PB in the corner, pick up drops, run back.

blkerby: Nice. Added it to the description and adjusted the requirements. I guess there's not a comparable strat if you enter from the right and cannot reset the room on the left?

Thinking about the possibility of one-hit-kill mode (or high damage scaling) has been changing how I think about enemy damage requirements, since it means small enemy damage can't be ignored even if there is an available way to farm.

osse101: You can kill the right 3 beetoms with speed but not break all the speed blocks.  Waste a PB first so you get 3 drops to work with.  More 🤡 than the left door strat.

blkerby: Oh yeah, that works nicely. I think if someone is playing in a "one-hit kill" mode with "can10PowerBombCrystalFlash" then this sort of stuff should be fair game, so probably no need for extra requirements for the safe strats.

PR #1366, opened 2024-02-11
Botwoon ETank speedball left to right

blkerby:


region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -447,6 +447,22 @@
...
+      "link": [1, 4],
+      "name": "Speedball",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 0,
+          "openEnd": 1
+        }
+      },
+      "requires": [
+        "canSpeedball"
```
kjbranch, line 460: Needs some sort of suitless maridia tech - it's not a perfect fit, but maybe `canWaterShineCharge` since it's obscure or `canShinechargeMovementComplex`? 

Do you have to jump in the previous room? if so, should this have an `openEnd` of 0?

blkerby: Yeah, `canWaterShineCharge` seems like a reasonable fit. There's technically no shinecharge, but it's same idea of charging from an air room into a water room. I don't think `canShinechargeMovementComplex` would work since that's more specific to moving with a shinecharge, which doesn't apply here.

You don't have to jump in the previous room, though it could often be easier to do so. Doing it in-room is probably a `canTrickyJump`, so I added that as a requirement too.

PR #1370, opened 2024-02-11
East Sand Hall Spring Ball jump right-to-left

blkerby:
This adds a spring ball jump option (in combination with Ice) to get from right to left. Also adds an option to avoid Evir damage and Power Bomb usage going from the center to left. See insomniaspeed's video: https://discord.com/channels/1053421401354285129/1053436236628496454/1189600996469514340. I also tested that avoiding Evir damage and PB usage is possible when entering from the sandfall above, by entering the room with a jump, freezing the first Evir on the way down, and avoiding bringing the second Evir on camera until after landing.

Similar to how other 2->4 strats are done in this room, the new Spring Ball strat mainly only includes the requirements to get to the first pillar, since getting from there to the center is considered to be free relative to the requirements to get from the center to the left.

### DIFF
--- a/region/maridia/inner-green/East Sand Hall.json
+++ b/region/maridia/inner-green/East Sand Hall.json
@@ -847,9 +847,7 @@
         "alternatively, jump slightly to the left and freeze the Evir, then return back to the center pillar before jumping across the sand.",
         "Freeze the right Evir and jump onto it.",
         "Quickly move left and fire an angle-down shot to freeze the left Evir before it descends too far.",
-        "Jump onto the left Evir.",
-        "Back up to the right side of it, to maximize space to build momentum, and jump to the left.",
-        "Break spin before landing on the sand, then spin jump onto the ledge.",
+        "Jump onto the left Evir, back up to the right side of it to maximize space to build momentum, and jump to the left onto the sand and then onto the ledge.",
         "If Samus does not quite get enough height to make it onto the ledge, then break spin, turn around to the right, land on the sand again, and do a turnaround spin-jump to the left.",
         "If the Evirs descend too far, it is possible to use a Power Bomb to bring them back up again, but caution is needed to avoid killing the first Evir, and it is difficult to avoid getting shot."
       ]
--- a/region/maridia/inner-green/East Sand Hall.json
+++ b/region/maridia/inner-green/East Sand Hall.json
@@ -676,7 +676,7 @@
       "name": "Spring Ball Jump",
       "requires": [
         "canTrickyUseFrozenEnemies",
-        "canSpringBallJumpMidAir"
+        "canTrickySpringBallJump"
       ],
       "note": [
         "Freeze the Evir soon after entering the room.",
--- a/region/maridia/inner-green/East Sand Hall.json
+++ b/region/maridia/inner-green/East Sand Hall.json
@@ -821,7 +821,7 @@
     },
     {
       "link": [4, 1],
-      "name": "East Sand Hall Suitless Bootless Evir Freeze (Right to Left)",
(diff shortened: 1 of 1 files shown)


osse101:
You can only jump into the room if you have walkable sand and not all sand exits have that.  But we can now add comeInJumping conditions to these sandfalls to represent that.  Without it I couldn't drop in and freeze the evirs correctly.
And the Power bomb placement is very picky from what I recall.

region/maridia/inner-green/East Sand Hall.json
```diff
@@ -815,17 +827,29 @@
...
-          "type": "particle",
-          "hits": 1
-        }}
+        {"or": [
+          {"and": [
+            {"ammo": {"type": "PowerBomb", "count": 1}},
+            {"enemyDamage": {
+              "enemy": "Evir",
+              "type": "particle",
+              "hits": 1
```
osse101, line 836: I would want to add another particle hit to this.  Very annoying.

blkerby:
So I looked into this some more and found the situation fortunately seems to be simpler than we were thinking:
- You can get a spin jump through any sand exit. Even the quicksand rooms have ledges at the bottom that can be used to get a spin jump out of the room, giving you a sandfall bounce coming into East Sand Hall to avoid the Evir shot. It's fairly forgiving in that if a Bull hits you it will break your spin but in that case you instead bring i-frames into the next room which also works (except in rare scenarios where there is not energy available to take a Bull hit). For the "Suitless Bootless Evir Freeze" notable strat (which is at Extreme), I think it's fine to assume you can evade the Bulls and exit with a spin jump; with a little practice it doesn't seem too hard.
- The camera manipulation strats seem much safer and reliable than using a Power Bomb. Coming from the sand transition above with a sandfall bounce or i-frames, all you have to do is land on the right side of the pillar, being careful not to scroll too far left, and then you can just freeze the Evirs and go. So that simplifies the requirements, as there's no need for Power Bombs, no risk of damage, and no need for a crouch jump. Coming from the right, you use moonwalk to scroll the camera to the right before jumping across the sand to the left (insomniaspeed video: https://discord.com/channels/1053421401354285129/1053436236628496454/1189600996469514340); or if moonwalk isn't available, you can jump out and freeze the Evir and return back to the center pillar. Again, if you do that and just go, then there's no need for Power Bombs or risk of damage. So I'm thinking Power Bombs only really seem useful as a backup, in case the player messes up and lets the Evirs descend too far.

region/maridia/inner-green/East Sand Hall.json
```diff
@@ -814,18 +828,28 @@
...
-        "Freeze the first Evir such that its bottom is only slightly lower than the top of the pillar.",
-        "Crouch jump onto the Evir, then freeze the second about 1.5 to 2.5 tiles lower. Jump off of the sand to get to the left door."
+        "If entering from the sandfall above, come in at the far left of the transition, with either a spin-jump or i-frames, to avoid the Evir shot;",
+        "land on the right side of the pillar to the left, being careful to not bring left Evir on camera.",
+        "If coming from the right side of the room, then moonwalk on the center pillar to scroll the camera as far right as possible, to avoid getting hit by an Evir shot;",
+        "alternatively, jump slightly to the left and freeze the Evir, then return back to the center pillar before jumping across the sand.",
+        "Freeze the right Evir and jump onto it.",
+        "Quickly move left and fire an angle-down shot to freeze the left Evir before it descends too far.",
+        "Jump onto the left Evir, back up to the right side of it to maximize space to build momentum, and jump to the left onto the sand and then onto the ledge.",
```
osse101, line 850: Momentum doesn't transfer when you jump while underwater.
But I have seen signs of gaining ~a pixel like you get pushed forward slightly on the jump frame.

osse101 (approved):

PR #1373, opened 2024-02-13
Green Brinstar Main Shaft Bottom Ice Clip

blkerby:
We unintentionally ended up with this strat now being in Map Rando Hard logic. It is missing requirements and needed to be refined:
- Added a "canTrickyJump" requirement. Even in the best case scenario (XRay + Spring Ball available) this is tricky for several reasons: you can't see Samus or the crumble block, you have to use i-frames from a hit, and even after successfully getting the clip it's nontrivial to get into the morph tunnel (which again, you can't see) since the crumble respawns.
- Made the strat notable. It's a very unique and obscure one, so we might want to bump it up to Expert.
- Added an alternative that uses a mid-air morph instead of Spring Ball; this one has "canInsaneJump" because it's tight and you may only get one chance; if the Zeela is frozen high enough, it can be more feasible to make a few tries, though in this case the mid-air morph is also tighter.
- Likewise, the version without X-ray gets a "canInsaneJump" requirement because it is pixel perfect.

So this ends up requiring "canInsaneJump" unless both X-Ray and Spring Ball are available.


osse101 (commented):
I think you need to quickly aimdown when jumping onto the frozen zeela, but couldn't confirm why I was missing that sometimes.

The clip needs to delay the down press a bit until you are above the crumble block, and that killed most of my early attempts.

Bombing the crumble helps a ton with visibility if you have Bombs.

Springball was much easier, jump morph was a little bit harder than a normal 4high.

And the no-XRay standup was just unreliable.

So expert on execution.  Springball+Xray is still maaaaybe extreme for being slow to setup and still easy to fail.  And no-XRay goes into insane.

blkerby:
> So expert on execution. Springball+Xray is still maaaaybe extreme for being slow to setup and still easy to fail. And no-XRay goes into insane.

Ok, added a "canBeVeryPatient" for the no-X-Ray version; maybe not quite the right requirement, but I'm not sure how else to put that variant into Insane unless we made a separate notable for it

PR #1374, opened 2024-02-13
East Sand Hall speedjumps

blkerby:


region/maridia/inner-green/East Sand Hall.json
```diff
@@ -203,6 +211,29 @@
...
+        "comeInRunning": {
+          "minTiles": 19.3625,
+          "maxTiles": 21.25,
+          "speedBooster": true
+        }
+      },
+      "requires": [
+        "canPlayInSand",
+        "canCrossRoomJumpIntoWater",
+        "canTrickyJump"
```
osse101, line 228:
```suggestion
        "canTrickyDashJump"
```

region/maridia/inner-green/East Sand Hall.json
```diff
@@ -453,6 +484,32 @@
...
+        "comeInRunning": {
+          "minTiles": 19.3625,
+          "maxTiles": 20.3,
+          "speedBooster": true
+        }
+      },
+      "requires": [
+        "canPlayInSand",
+        "canCrossRoomJumpIntoWater",
+        "canTrickyJump"
```
osse101, line 501:
```suggestion
        "canTrickyDashJump"
```

osse101 (commented):
Not as big a window of pixels as I'd like, but it works really well at 20 tiles.

Not technically a trickyDashJump, but you are targetting a specific pixel range to get a particular jump height.

I need a system for counting 20 tiles without prachack tools.  Trying out counting the number of occurances of repating patterns in floor tiles for now.

PR #1375, opened 2024-02-13
Add resourceAvailable to schema

blkerby:
This adds a new logical requirement `resourceAvailable` analagous to `resourceCapacity` but based on the current amount of resources not capacity.

I was thinking this could be useful to refine the logic in Metal Pirates Room, i.e. to add strats for killing both pirates with only 5 Supers, based on the assumption that you get at least one Super dropped (given enough energy to be out of health bomb range), or tankless with shinespark echoes, based on the assumption that you will get enough health drops to survive. I don't think I'm experienced enough with the room to be able to add the strats, but @osse101 maybe sometime you could? It seems like this room is currently one of the most common sources of "easy" sequence breaks on Challenge seeds.

### DIFF
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -291,10 +291,30 @@ A `resourceCapacity` object represents the need for Samus to be capable of holdi
 __Example:__
 ```json
 {"resourceCapacity": [
-    { "type": "Missile", "count": 10},
-    { "type": "Super", "count":10},
-    { "type": "PowerBomb", "count": 11},
-    { "type": "RegularEnergy", "count":899}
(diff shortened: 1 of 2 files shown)


osse101 (approved):
I think this pairs with a `resourceFull` style requirement since we want to increase the odds of certain drops.  And if we guarantee health drops, then more heat runs open up with lower health.

Super Metal Pirates if played extremely aggresively :eyes: (cancel the kick and first throw animations) you can do it 1 super pack, no etanks.  Or you could take a contact hit and double hit with supers.  We're assuming the first drops 1 super, but then there are 9 more drops that can be all health or all ammo.

PR #1376, opened 2024-02-13
Frozen Enemy Runways in Pink & Green Maridia

kjbranch:
Skipped a few here, which can be added on a second pass:
- East Pants with a Puyo - its bad but not terrible ... but it has no use if East Pants connects to Pants.
- Botwoon E-Tank - it doesnt look like the Puyos align easily. The Zoa from across the room also doesnt align with any easy setup.
- The Oums in Halfie Climb - they are a lot of fun to play with, but I couldn't find anything too reliable and it wasnt clear if a short charge is possible on them.

### DIFF
--- a/region/maridia/inner-pink/Aqueduct.json
+++ b/region/maridia/inner-pink/Aqueduct.json
@@ -1160,16 +1160,26 @@
       "name": "Leave Shinecharged",
       "requires": [
         "Gravity",
-        {"canShineCharge": {
-          "usedTiles": 20,
-          "openEnd": 2
-        }}
+        {"or": [
+          {"canShineCharge": {
+            "usedTiles": 20,
+            "openEnd": 2
+          }},
+          {"and": [
+            "canEnemyExtendRunway",
+            {"canShineCharge": {
+              "usedTiles": 20,
+              "openEnd": 2
+            }}
+          ]}
+        ]}
       ],
       "exitCondition": {
         "leaveShinecharged": {
           "framesRemaining": 60
         }
-      }
+      },
+      "devNote": "It is possible, although not overly useful, to extend the runway here using a snail."
     },
     {
       "link": [5, 5],
@@ -1221,11 +1231,25 @@
       "name": "Shinespark",
       "requires": [
         "Gravity",
-        {"canShineCharge": {
-          "usedTiles": 20,
-          "openEnd": 2
-        }},
+        {"or": [
+          {"canShineCharge": {
+            "usedTiles": 20,
+            "openEnd": 2
+          }},
+          {"and": [
+            "canEnemyExtendRunway",
+            {"canShineCharge": {
+              "usedTiles": 20,
+              "openEnd": 2
+            }}
+          ]}
+        ]},
         {"shinespark": {"frames": 25}}
+      ],
+      "devNote": [
+        "It is possible, although not overly useful, to extend the runway here using a snail.",
+        "FIXME: It is possible to use a snail here and in Shaktool's room to extend the runway at the door.",
+        "This can be useful when short charging, but can not be used to leave running, as running on the snail does not work."
       ]
     },
     {
--- a/region/maridia/inner-pink/Aqueduct.json
+++ b/region/maridia/inner-pink/Aqueduct.json
@@ -1168,7 +1168,7 @@
           {"and": [
             "canEnemyExtendRunway",
             {"canShineCharge": {
-              "usedTiles": 20,
+              "usedTiles": 21,
               "openEnd": 2
             }}
           ]}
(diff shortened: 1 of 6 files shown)


osse101:
Oums do align with the door runway which is good.  I could position one by standing in a certain spot in the doorway, where it would stop and attack.  And you could time a jump on it's back just right, or kago the damage to get run speed out the door.  But I couldn't get any reliable shortcharge on it's back.  And didn't look at using more than one.

Shortcharging by running on+against 1 Oum fails once you get a little bit of run speed, it looks like.

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -1160,16 +1160,26 @@
...
-        }}
+        {"or": [
+          {"canShineCharge": {
+            "usedTiles": 20,
+            "openEnd": 2
+          }},
+          {"and": [
+            "canEnemyExtendRunway",
+            {"canShineCharge": {
+              "usedTiles": 20,
```
osse101, line 1171: numbers are the same

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -1221,11 +1231,25 @@
...
-        }},
+        {"or": [
+          {"canShineCharge": {
+            "usedTiles": 20,
+            "openEnd": 2
+          }},
+          {"and": [
+            "canEnemyExtendRunway",
+            {"canShineCharge": {
+              "usedTiles": 20,
```
osse101, line 1242: here too

region/maridia/inner-pink/Below Botwoon Energy Tank.json
```diff
@@ -126,6 +126,24 @@
...
+      ],
+      "exitCondition": {
+        "leaveWithRunway": {
+          "length": 5,
+          "openEnd": 0
+        }
+      },
+      "note": [
+        "The Zoa needs to be at least 3 pixels higher than when it is frozen at standing height.",
+        "It may help to freeze one while standing to use as a reference, then jump and freeze a second while, noting that it will move up 1 more pixel once it thaws.",
```
osse101, line 143: Freeze one facing left and still rising, then stand in a particular spot on the slope behind it, on the right.

region/maridia/inner-pink/Colosseum.json
```diff
@@ -135,6 +135,35 @@
...
       }
     },
+    {
+      "link": [1, 1],
+      "name": "Leave With Runway - Frozen Mochtroid",
+      "requires": [
+        "h_canNavigateUnderwater",
+        "h_canFrozenEnemyRunway",
+        {"or": [
+          "canMochtroidIceClimb",
```
osse101, line 145: Pretty easy to take damage.  Should I just try to be extra careful?

region/maridia/inner-pink/Colosseum.json
```diff
@@ -776,6 +805,35 @@
         }
       }
     },
+    {
+      "link": [3, 3],
```
osse101, line 809: [2, 2] isn't hard with an angle down shot while walking off the platform.
I liked doing walk off shots for 1 and 3 too tbh.

region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -975,6 +1004,35 @@
...
       }
     },
+    {
+      "link": [4, 4],
+      "name": "Leave With Runway - Frozen Mochtroid",
+      "requires": [
+        "h_canNavigateUnderwater",
+        "h_canFrozenEnemyRunway",
+        {"or": [
+          "canMochtroidIceClimb",
```
osse101, line 1014: Little bit obnoxious to set up.  I find it more likely to take damage too, especially when positioning the runway freeze.  It is also a bit of a choice to stand with Mochtroid attached at the correct height to freeze vs timing it as it floats up somewhat fast.

region/maridia/inner-pink/Colosseum.json
kjbranch: 2 seems good, but the drop down method seems much worse to me. I added it as a note.

region/maridia/inner-pink/Below Botwoon Energy Tank.json
```diff
@@ -126,6 +126,25 @@
...
+      "exitCondition": {
+        "leaveWithRunway": {
+          "length": 5,
+          "openEnd": 0
+        }
+      },
+      "note": [
+        "The Zoa needs to be at least 3 pixels higher than when it is frozen at standing height.",
+        "It may help to freeze one while standing to the left of it before it starts moving horizontally,",
+        "then move to the slope to the right of it to position it. Note that it will move up 1 more pixel once it thaws.",
```
osse101, line 144:
```suggestion
        "It may help to freeze one while standing to the left of it before it starts moving horizontally,",
        "then move to a specific point on the slope to the right of it to position it vertically.",
```

osse101 (approved):

PR #1377, opened 2024-02-13
resourceMissingAtMost requirement

blkerby:
This is following up on your idea @osse101 about a `resourceFull` requirement. I think this generalization `resourceMissingAtMost` can be useful in a other situations too. For example we could use it to refine some of the farming strats, e.g. their patience requirements based on how long the refill would take. It's still limited by the fact that we're still only representing full refills, but it could allow quite a bit of improvement by putting limits on it, e.g. so you don't end up needing to spend 10 minutes refilling energy in LN Firefleas on vanilla map.

### DIFF
--- a/schema/m3-requirements.schema.json
+++ b/schema/m3-requirements.schema.json
@@ -469,7 +469,7 @@
                   "type": "integer",
                   "minimum": 0,
                   "title": "Resource Count",
-                  "description": "The amount of the resource that Samus must not be missing too much of, relative to being full."
+                  "description": "The amount of the resource that Samus must not be missing, relative to being full."
                 }
               }
             }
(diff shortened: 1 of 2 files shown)


schema/m3-requirements.schema.json
```diff
@@ -435,6 +435,45 @@
...
+                  ],
+                  "title": "Resource Type",
+                  "description": "The type of resource held."
+                },
+                "count": {
+                  "$id": "#/definitions/logicalRequirement/items/properties/resourceMissingAtMost/properties/items/count",
+                  "type": "integer",
+                  "minimum": 0,
+                  "title": "Resource Count",
+                  "description": "The amount of the resource that Samus must not be missing too much of, relative to being full."
```
osse101, line 472:
```suggestion
                  "description": "The amount of the resource that Samus must not be missing, relative to being full."
```

osse101 (approved):
Very powerful addition for adding small farms everywhere.

PR #1378, opened 2024-02-14
Red Tower Frozen Beetom Climb

blkerby:
This is a recent strat from Bobbob: https://discord.com/channels/1053421401354285129/1053436236628496454/1207096037836193802

I wasn't really sure how many Beetom hits to put in for leniency because it's hard for me to tell how viable it is to do these jumps consistently.

### DIFF
--- a/region/brinstar/red/Red Tower.json
+++ b/region/brinstar/red/Red Tower.json
@@ -1037,6 +1037,26 @@
       ]
     },
     {
+      "link": [1, 9],
+      "name": "Red Tower Frozen Beetom Ice Climb",
+      "notable": true,
+      "requires": [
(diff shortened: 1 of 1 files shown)


osse101:
Trying it I could get the jump about 1/5, taking damage every other setup attempt.  I also frequently missed the freeze when placing the Beetom for the second and third flatley jump.
With a pause buffer it was more like 2/5 for the jump and taking 1.5 hits per setup.

blkerby:
Yeah the pause buffer helps a lot. Using the practice hack it's hard for me not to cheat by noticing if I got the turnaround input before the pause, but it feels much more doable.

PR #1379, opened 2024-02-14
Remove walljump req from Indy Jones Long Mella

blkerby:
The walljump requirement here isn't needed; somerando figured this out a while ago: https://youtu.be/9t2AWPOJNYQ?list=PLMfL5HDn1Tf6wtTlVNGcwCUMdUYDWUbLd&t=3818.

### DIFF
--- a/region/norfair/crocomire/Post Crocomire Jump Room.json
+++ b/region/norfair/crocomire/Post Crocomire Jump Room.json
@@ -799,7 +799,6 @@
         "canUseFrozenEnemies",
         "canManipulateMellas",
         "canBeVeryPatient",
-        "canConsecutiveWalljump",
         {"or": [
           {"obstaclesCleared": ["B"]},
           "Morph"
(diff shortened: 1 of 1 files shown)


region/norfair/crocomire/Post Crocomire Jump Room.json
```diff
@@ -799,15 +799,15 @@
         "canUseFrozenEnemies",
         "canManipulateMellas",
         "canBeVeryPatient",
```
osse101, line 801:
```suggestion
        {"or": [
          {"and": [
            "canBePatient",
            "canWalljump"
          ]},
          "canBeVeryPatient"
        ]},
```

osse101: 120frames door to item with 1 walljump.
Watch it climb from the ground and once it's lowest point is around jump height, climb aboard and MellaIceClimb and that puts you close enough to walljump to the ledge.  A Springball jump probably works _quickly_ if done this way too.

PR #1380, opened 2024-02-14
Update X-Ray Exit Damage Boost waver hits

blkerby:
Noticed this on Eddie's stream: https://clips.twitch.tv/TardyBoredBaboonPJSalt-4DmlZ-_BxOgdpA2Y. Tried it out and it seems reasonable. If there were a different tech requirement that could bump up the 2-hit version to Expert, that may be ideal? But I don't know what we would use?


osse101:
I think I like expert more too, for 2 hit only.  But its fine in `Very Hard`.  You just need to be aware that the waver will be there to catch you. And trust your jumps.  The movement is fine at very hard.

PR #1381, opened 2024-02-15
canManageReserves tech

blkerby:


tech.json
```diff
@@ -133,6 +133,42 @@
...
+            "1) If Samus takes a hit that exceeds her energy capacity, while having any reserve energy, it should logically result in 1 Energy and 0 Reserve Energy afterward, under an assumption that reserve energy was previously drained to allow i-frames to be conserved",
+            "2) Samus should not be required to refill twice (nor refill once particularly precisely) during a shinespark if she has no Energy Tanks, roughly if the frames minus excess frames exceeds 120."
+          ],
+          "extensionTechs": [
+            {
+              "name": "canPauseAbuse",
+              "requires": [ "canManageReserves" ],
+              "note": [
+                "The ability to pause in order to avoid death while reaching 0 Energy.",
+                "Energy must be obtained before the unpause fade-in finishes, either by picking up an Energy drop or refilling with Reserve Energy."
```
osse101, line 162: Regarding trickling in health while dying to heat damage: In my experience I would lose as much health through movement inefficiency as I would gain from pause abuse.  Situationally, perhaps you could hold easy inputs and cheat health with pause abuse.  But I'm not a fan of modelling this.  Except for when you are waiting for drops to appear, like after Ridley.

And then, related, I think pausing to manually refill in a heat room should have a cost to it.  Tied to selected difficulty, but `Heat Leniency Multiplier`'s may cover that.

blkerby: Yeah, we weren't planning on having pause abuse implicitly modeled where it would save on total energy requirements like that. Strats could be manually added in cases where that may make sense (e.g. tanking Three Musketeers). The main thing to be implicitly modeled would be the ability to convert from reserve energy to regular energy as needed (even in the middle of an enemy hit or shinespark), which is what the randomizer already assumes but which we want to stop assuming at lower difficulty levels.

We had been thinking of supplementing this with a reserve energy multiplier that would assume some fraction of reserve energy is unusable (e.g., due to autofilling in heated rooms, to taking a a random double hit, or to worse movement in heated rooms due to a manual or auto refill). But I'm not sure it's necessary, since the heat damage multiplier could probably already cover it well enough.

PR #1382, opened 2024-02-15
Pink Hoppers more walljumpless strats

blkerby:
These are new right-side X-Ray climb strats by bobbob:

HiJump: https://www.youtube.com/watch?v=_hblD6uxx58
G-Mode: https://www.youtube.com/watch?v=hYkD5BnNjW8
Spring Ball: https://www.youtube.com/watch?v=Nj6_FBbgFsY

I tested them to work out details on positioning. Also added more cross-room Speed jump strats on both sides.

### DIFF
--- a/region/brinstar/pink/Pink Brinstar Hopper Room.json
+++ b/region/brinstar/pink/Pink Brinstar Hopper Room.json
@@ -884,6 +884,7 @@
       "requires": [
         "HiJump",
         "canXRayClimb",
+        "canOffScreenMovement",
         "canInsaneJump"
       ],
       "note": [
@@ -906,6 +907,7 @@
       },
       "requires": [
         "canXRayClimb",
+        "canOffScreenMovement",
         "canInsaneJump"
       ],
       "note": [
--- a/region/brinstar/pink/Pink Brinstar Hopper Room.json
+++ b/region/brinstar/pink/Pink Brinstar Hopper Room.json
@@ -432,6 +432,36 @@
       "entranceCondition": {
         "comeInJumping": {
           "speedBooster": true,
+          "minTiles": 34.4375
+        }
+      },
+      "requires": [
+        "canPrepareForNextRoom",
+        "canTrickyJump"
+      ],
+      "note": "Jump into the room and onto the Grapple block."
+    },
+    {
+      "link": [1, 3],
+      "name": "Tricky Speedy Jump",
+      "entranceCondition": {
+        "comeInJumping": {
+          "speedBooster": true,
+          "minTiles": 25
+        }
+      },
+      "requires": [
+        "canPrepareForNextRoom",
+        "canTrickyDashJump"
+      ],
+      "note": "Jump into the room and onto the Grapple block."
+    },
+    {
+      "link": [1, 3],
+      "name": "Speedy HiJump",
+      "entranceCondition": {
+        "comeInJumping": {
+          "speedBooster": true,
           "minTiles": 6.4375
         }
       },
@@ -445,7 +475,7 @@
     },
     {
       "link": [1, 3],
-      "name": "Tricky Speedy Jump",
+      "name": "Tricky Speedy HiJump",
       "entranceCondition": {
         "comeInJumping": {
           "speedBooster": true,
@@ -663,6 +693,18 @@
     },
     {
       "link": [2, 3],
+      "name": "Speedy Jump",
+      "entranceCondition": {
+        "comeInRunning": {
+          "speedBooster": true,
+          "minTiles": 29
+        }
+      },
+      "requires": [],
+      "note": "Jump at the end of the runway in-room. Assumes there is a runway in the adjacent room of at least 29 tiles."
+    },
+    {
+      "link": [2, 3],
       "name": "Tricky Speedy HiJump",
       "entranceCondition": {
         "comeInRunning": {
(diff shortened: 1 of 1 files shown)


region/brinstar/pink/Pink Brinstar Hopper Room.json
```diff
@@ -812,6 +854,71 @@
...
+      "entranceCondition": {
+        "comeInWithDoorStuckSetup": {}
+      },
+      "requires": [
+        "HiJump",
+        "canXRayClimb",
+        "canInsaneJump"
+      ],
+      "note": [
+        "Get stuck 3 pixels inside the door, and X-ray climb to a precise height:",
```
osse101, line 890: > 3 pixels inside the door

Is it not a normal doorstuck?

blkerby: It's a "normal" doorstuck setup as far as I know, but that setup can end up getting you stuck either 1, 2, or 3 pixels in the door, and this strat needs 3. I assume there must be some subtlety to the exact movement to get 3 pixels vs. other values but I don't understand it.

osse101: Different run speeds and transition heights can work... so I'd say it's a combination of the two, as well as fall speed, to land on the correct pixel position.  So RNG to me.

Slightly easier than a `canRightSideDoorStuckFromWater`.  The jump from the wall is a textbook `canInsaneJump`.  Running for 2 pixels to get the run speed takes a bit of practice to get the feel for.

region/brinstar/pink/Pink Brinstar Hopper Room.json
```diff
@@ -812,6 +854,71 @@
...
+        "after just one pixel of Samus' feet are visible while crouching facing left, perform exactly 7 more X-Ray stand-ups.",
+        "Use X-ray to face left, then run and jump to barely make it onto the ledge."
+      ],
+      "devNote": [
+        "3 pixels stuck in the door is X position 478 (0x1DE)."
```
osse101, line 895: This information would be useful to show on the strat page but also doesn't fit as part of a normal note.

region/brinstar/pink/Pink Brinstar Hopper Room.json
osse101: There is some concern over blind xray standups since it is often hard to tell if you missed one.  Sometimes you can tell if the crouch feels wrong, but we do want a level of comfort with Xray Climbing when blind counting.

region/brinstar/pink/Pink Brinstar Hopper Room.json
```diff
@@ -812,6 +854,71 @@
...
+        "canInsaneJump"
+      ],
+      "note": [
+        "Enter with G-Mode and use X-Ray to get stuck 9 pixels inside the door.",
+        "X-Ray climb to a relatively specific height:",
+        "after just one pixel of Samus' feet are visible while crouching facing left, perform between 20 and 22 more X-Ray stand-ups (any of these work).",
+        "Use X-ray to face left, then run and jump to barely make it onto the ledge."
+      ],
+      "devNote": [
+        "9 pixels stuck in the door is X position 484 (0x1E4).",
```
osse101, line 918: 482 and 483 looked to work.  485 lets you run out so should work too.

Harder than the HiJump strat because the jump timing is really weird.  More blind standups could be a problem but weren't in my test run.

osse101 (approved):

region/brinstar/pink/Pink Brinstar Hopper Room.json
blkerby: Do you think we need a new tech for the specific doorstuck setup? The "canInsaneJump" already gates these strats to a high difficulty level. I also just added a "canOffScreenMovement" requirement for the parts of counting your climbs (and knowing if you missing any)? The longer blind climb at least has some lenience, where you could shoot for the high number of climbs (22), and if you miss a couple (e.g. only get 20 climbs) it will still work.

region/brinstar/pink/Pink Brinstar Hopper Room.json
blkerby: Yeah I've thought about adding a button or something that could reveal any devNotes on a strat, or maybe the full JSON for the strat.

region/brinstar/pink/Pink Brinstar Hopper Room.json
blkerby: I tried 485 and couldn't get it to work: it did let you run out but it would always bonk when I tried to jump. I saw that some of the positions to the left could also work but I didn't thoroughly test to see if they worked at all the same vertical positions or not.

region/brinstar/pink/Pink Brinstar Hopper Room.json
osse101: I think there does need to be a tech for describing getting stuck with a particular pixel value.  `doorStuckFromWater` I'd like to leave as it's own thing.  GMode doorstuck is part of GMode, but knowledge of which pixels are good for climbing could be under the new tech.

I'd leave this as it is for this case, but I'd guess you can run from a wall in more rooms like East Cac Alley.  GT's room would be 🤡.  Maybe not too many places.  I'm guessing you need `Normal` physics and right side door stuck for the run to work.  

Turnaround spinjumps like the springball strat would use basic door stuck tech.

PR #1383, opened 2024-02-15
Frozen Enemy Runways in Yellow & Outer Maridia

kjbranch:
I didn't skip many in this section:
- NW Maridia Bug Room
- Thread the Needle


osse101 (commented):
Boyon Gate Hall can use a Zeb for runway overtop the Boyon hole.  Wave helps a lot.
And there is a possible double freeze that I didn't want to attempt.

The strats here all appear ok.

kjbranch:
Oh, cool. I was wondering about that and then moved on. I'll add it to my to-do list and tackle it in a future PR.

Issue #1189, opened 2023-12-01
Add Grapple Warping strats


blkerby:
Triggering a door transition while grappling results in Samus' position being overwritten after the transition, allowing the player to potentially warp to a distant part of the next room. Red Brinstar Fireflea Room is an example where you can exit the left door in this way. We could add a new pair of exit/entrance conditions to be able to model strats like this.

osse101:
I think Red Brinstar Firefleas is the only room where you can do this.
Entrance/Exit conditions make sense.  The tech itself is very similar to a grappleClip but with a frame perfect moonwalk.
And the effect puts Samus on the leftmost room tile of the room.  And I forget if the room tile is on the top row, or in line with the door.  But there certainly are skips possible with a grapple warp.

blkerby:
Bobbob created a playlist showing all the grapple teleport applications from Red Brinstar Firefleas: https://www.youtube.com/watch?v=g3jEO-WxvK0&list=PLUMF22jq7du0mfVSKbEZZhA1OaKH1Wtzh.

He also gave these notes about other setup locations (https://discord.com/channels/1053421401354285129/1053436236628496454/1208621382393532526):

> 
>     red brinstar firefleas left:  leftward transition, top left screen, 6 blocks from the left and 4 blocks from the top. we'll call it (6,4).
>     moat left:  leftward transition, top left screen, (8,3).
>     colosseum left:  leftward transition, top left screen, (4,13) or (4,14) [yes these can have different behavior where sometimes things work from one but not the other]
>     halfie climb top left:  leftward transition, one screen down from top left, (3,3) or (3,13) or (3,14).
>     draygon left:  leftward transition, one screen down from top left, (3,3) or (3,4).
>     halfie climb bottom left:  leftward transition, two screens down from top left, (3,3).
>     halfie climb top right:  rightward transition, top left screen, (13,13) or (13,14).
>     colosseum top right:  rightward transition, seven screens right of top left, (13,13) or (13,14).
>     colosseum bottom right:  rightward transition, seven screens right and one screen down of top left, (13,3).
>

PR #1389, opened 2024-02-20
Grapple teleports from Red Brin Firefleas

blkerby:
This is based on bobbob's video playlist: https://www.youtube.com/playlist?list=PLUMF22jq7du0mfVSKbEZZhA1OaKH1Wtzh.

This makes progress on issue #1189.


region/maridia/outer/Glass Tunnel.json
```diff
@@ -586,6 +586,38 @@
...
+      "entranceCondition": {
+        "comeInWithGrappleTeleport": {
+          "blockPositions": [[5, 3]]
+        }
+      },
+      "requires": [
+        {"doorUnlockedAtNode": 4},
+        {"or": [
+          {"and": [
+            "canGrappleJump",
```
osse101, line 601: Couldn't do it.
It's a harder grapple than the fish tank one because of the platform under the door.
You have to retract the grapple a bit before you hit it.

region/norfair/west/Crumble Shaft.json
```diff
@@ -354,6 +354,26 @@
...
+    {
+      "link": [2, 1],
+      "name": "Grapple Teleport",
+      "entranceCondition": {
+        "comeInWithGrappleTeleport": {
+          "blockPositions": [[5, 3]]
+        }
+      },
+      "requires": [
+        {"heatFrames": 100},
```
osse101, line 366:
```suggestion
        {"heatFrames": 120},
```

osse101: And there should be a version without offscreenmovement

region/wreckedship/main/Attic.json
```diff
@@ -139,6 +139,7 @@
     {
       "from": 3,
       "to": [
+        {"id": 1},
```
osse101, line 142: 💭

osse101:
I think I do like a `canTrickyGrappleJump` tech.
- very precise grapple releases and morph/unmorphes
- being active during the grapple jump
- wiggling to position the grapple jump after release

So Pants room grapple jump through the grapple block would be included, but draygon escape would not be.

region/norfair/west/Crumble Shaft.json
blkerby: Done.

region/maridia/outer/Fish Tank.json
```diff
@@ -510,6 +535,20 @@
...
+      "name": "Grapple Teleport",
+      "entranceCondition": {
+        "comeInWithGrappleTeleport": {
+          "blockPositions": [[5, 3]]
+        }
+      },
+      "requires": [],
+      "note": [
+        "Swing over the right to make it to the ledge.",
+        "Be cautious of extending Grapple, to prevent being damaged by the Skulltera."
```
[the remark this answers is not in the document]
blkerby: I think it's not so bad; you can fix the camera while avoiding damage by just continuing to swing. Added this to the note.

blkerby:
I think `canTrickyGrappleJump` sounds good. Mind adding that in a follow-up PR? I figure I'm probably not the one to do it since I barely know how to grapple jump at all. In the meantime, all these strats involving tricky grapple jumps should be notable now, so we can put them in Insane or Beyond as appropriate.

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -616,6 +628,26 @@
...
+      },
+      "requires": [
+        {"heatFrames": 250},
+        {"or": [
+          "canWalljump",
+          "h_canCrouchJumpDownGrab"
+        ]}
+      ],
+      "note": [
+        "After teleporting, extend Grapple to position Samus on the floor one tile to the right of the Grapple attachment point.",
```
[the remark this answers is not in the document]
osse101: My method doesn't work everytime so I'll leave it as a backup.
Bob's way has a tiny frame window.  It's fair as an expert strat but means you need a feel for the timing.  A more consistent method would just be nice.

region/maridia/outer/Fish Tank.json
```diff
@@ -489,6 +490,30 @@
...
+      "name": "Grapple Teleport into Grapple Jump",
+      "entranceCondition": {
+        "comeInWithGrappleTeleport": {
+          "blockPositions": [[5, 3]]
+        }
+      },
+      "requires": [
+        {"doorUnlockedAtNode": 4},
+        "canGrappleJump",
+        "canInsaneJump"
```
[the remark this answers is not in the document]
osse101: - You can't avoid the fish by dead hanging, but you can learn a simple movement pattern to avoid it
- I didn't think of turning off grapple while leaving the beam out

PR #1390, opened 2024-02-20
Frozen Enemy Runways in Tourian

kjbranch:
Skipped a few strats:
- Metroid room 3 right door
- Rinka Shaft top door
- Both MB doors - the left door could be added with a helper, but the right door could cause some problems for Map Rando. I'll add it on my next pass through.

### DIFF
--- a/region/tourian/main/Metroid Room 2.json
+++ b/region/tourian/main/Metroid Room 2.json
@@ -108,7 +108,7 @@
       "note": [
         "Kill or lure and freeze the Metroids at the bottom of the room.",
         "It may be easiest to stand on a frozen Rinka from the top left spawner to position a freeze of the other Rinka.",
-        "It is possible maintain a half-tile gap between the Rinka and the runway in order to extend it as far as possible."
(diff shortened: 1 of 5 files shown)


region/tourian/main/Metroid Room 2.json
```diff
@@ -93,6 +93,24 @@
...
+      "exitCondition": {
+        "leaveWithRunway": {
+          "length": 5,
+          "openEnd": 0
+        }
+      },
+      "note": [
+        "Kill or lure and freeze the Metroids at the bottom of the room.",
+        "It may be easiest to stand on a frozen Rinka from the top left spawner to position a freeze of the other Rinka.",
+        "It is possible maintain a half-tile gap between the Rinka and the runway in order to extend it as far as possible."
```
osse101, line 111:
```suggestion
        "Maintain a half-tile gap between the Rinka and the runway in order to extend it as far as possible."
```
?

osse101 (approved):

PR #1391, opened 2024-02-21
Warehouse Entrance arm pump jump

blkerby:


blkerby:
Yeah it feels like it could probably go in Expert, since it would only be required if walljump is disabled, in which case the player is kind of asking for obscure things like this. I'm just not sure what tech would go in place of "canInsaneJump". Maybe an arm pump tech could do it

PR #1392, opened 2024-02-21
Fish Tank tricky frozen fish

blkerby:
This strat is currently in Medium. Skilled players regularly seem unaware that this is possible. I figure it should at least need "canTrickyUseFrozenEnemies", which would put it in Hard. @osse101 Do you think Hard is ok, or should it need something to bump it to Very Hard, either a "canTrickyJump" or notable? The jump doesn't seem very precise as long as the fish is frozen in a good spot.

### DIFF
--- a/region/maridia/outer/Fish Tank.json
+++ b/region/maridia/outer/Fish Tank.json
@@ -927,7 +927,11 @@
         "canTrickyJump",
         "canTrickyUseFrozenEnemies"
       ],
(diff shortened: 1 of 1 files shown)


blkerby:
Sounds good, added the `canTrickyJump` requirement. I found that freezing the fish while it's moving right also works fine, so I removed the part in the note about freezing it when it's moving left.

PR #1397, opened 2024-02-27
Main Hall Elevator Reserve Trigger

kjbranch:


blkerby:
I think the reserve trigger option shouldn't also be requiring the 438 heat frames. 

The heat damage during the refill I think will need to be modeled as part of the autoReserveTrigger, since it depends on the amount in reserves; might be worth a mention in the docs for autoReserveTrigger

kjbranch:
I was just trying to be conservative here. If we are wanting to use the elevator with low energy and lots of reserve to use fewer heat frames, that will need some tech. 

The heat frames during the reserve trigger is important, but I'm unclear how to implement that. In this example, it actually depends on if the next room is heated ...

blkerby (approved):
Looks good. I think later we'll probably want to use a new entrance/exit condition (e.g. comeInWithElevator, leaveWithElevator) to fully accurately encode the pending heat damage and how it gets resolved across rooms.

PR #1399, opened 2024-02-29
West Sand Hall Spring Fling with Momentum

blkerby:


osse101:
Standing on the farthest pixel out felt like the most important factor.
Then springball jumping around max height.
Then jumping from the right half of last tile.
Springfling was sometimes helpful and the pauses are awkward.  It gives leeway for doing the other parts suboptimally.

blkerby:
> Standing on the farthest pixel out felt like the most important factor. Then springball jumping around max height. Then jumping from the right half of last tile. Springfling was sometimes helpful and the pauses are awkward. It gives leeway for doing the other parts suboptimally.

For sure standing on the farthest pixel would be important. I wasn't sure about putting that in the note since it may not apply if the runway is longer? I guess it may already go without saying for a near-max-height spring ball jump, but for me pressing pause a little before the morph seemed important; it helps with giving a comfortable window for the repause. Also it didn't seem to work if I morphed while still in the first sandfall. Springfling isn't strictly necessary but for me the amount that it helps seems huge, turning it from a very precise trick into a fairly lenient one.

PR #1400, opened 2024-03-01
Refine Bowling Alley X-Ray Climb

blkerby:


region/wreckedship/main/Bowling Alley.json
```diff
@@ -442,17 +442,33 @@
...
       ],
       "note": [
-        "Climb up 1 screen.",
-        "Climb quickly once you reach the spikes to minimize damage (if Phantoon is dead).",
-        "I-frames will run out while using X-Ray, so you don't want to hold X-Ray for long."
+        "Climb up about half a screen.",
+        "If Phantoon is dead, the last part of the climb should be done carefully to avoid triggering collision with the spikes:",
+        "Do not press forward or turn-around except while X-Ray is active.",
+        "For each X-ray stand-up, release the turn-around input quickly to make sure that it is not still held after X-Ray deactivates.",
+        "Once is Samus is high enough, unequip X-Ray, jump up (from either standing or crouch) to clip past the spikes, and land, taking a single spike hit.",
```
osse101, line 464:
```suggestion
        "Once Samus is high enough, unequip X-Ray, jump up (from either standing or crouch) to clip past the spikes, and land, taking a single spike hit.",
```

osse101 (commented):
These xray turnarounds are ~extreme level; maybe expert if you're being real careful, much harder than basic xray climbing.  canXRayClimb is more like a very hard tech which has missed xray standups and doesn't jump through the top tiles of the climb.  So a base option of `Number of spike hits to pop out, then jump to the door` would be best.  But, there is still some offscreen movement so limiting the whole strat to expert with another tech could be good.  For other randomizers.

PR #1404, opened 2024-03-04
Tricky Heated Power Bombs with Reserves Only

kjbranch:
This is just a quick attempt to address some of the hardest uses of Power Bombs in heated rooms with tight Reserve management. This would move these strats from `VH` to `Expert` in Map Rando if the player has no E-Tanks.

There are other heated rooms that require using Power Bombs, but I don't think they are on the same level of difficulty with precise movement while placing the PBs and triggering reserves:
Right side of Single Chamber, LN Springball Maze, Three Musketeers, Screw Attack, Mickey Mouse, LN Kihunter, Wasteland, Pillar Setup, GT right item

@osse101, what are your thoughts on something like this: is it useful, should it be included in more rooms, should there be a `canInsaneJump` alternative?

### DIFF
--- a/region/lowernorfair/east/Pillar Room.json
+++ b/region/lowernorfair/east/Pillar Room.json
@@ -232,13 +232,19 @@
             {"heatFrames": 240},
             {"acidFrames": 96}
           ]}
+        ]},
+        {"or": [
+          "h_heatResistant",
+          "canPauseAbuse",
+          {"resourceCapacity": [{"type": "RegularEnergy", "count": 149}]}
         ]}
       ],
       "reusableRoomwideNotable": "Lower Norfair Pillar Room with Two Power Bombs",
       "note": [
         "Place the PBs next to the pillars in order to only use 2.",
         "Minimize acid by unmorphing high to land back on the jump spot or walljumping before placing the bomb."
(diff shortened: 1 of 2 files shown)


osse101:
Looking at these examples, needing to pause mid strat makes it a full difficulty tier harder.  You pause just before laying the PB, which actually isn't intuitive until expert, where `canPauseAbuse` lives.
Pausing will throw off your next movement, and the low% ice strats are very technical.  So it is harder to do the existing inputs, plus there are additional inputs.
The diagonal bomb jump seems the most reasonable for just adding `canPauseAbuse`, but the pause is at the hardest part of the strat.

kjbranch:
What do you think of this? This will move all strats up to `Expert`, except the 3 strats that were already there+: Both `WRITG Low Ice` are moved up from `Expert` to `Extreme`. `WRITG Diagonal Bomb Jump` remains in `Extreme`, but now also includes `canInsaneJump`.

If we need to address `WRITG Diagonal Bomb Jump` further, we could just require no Reserve only strats or we could split off a Reserve only into another `Insane` notable. Thoughts?

osse101 (approved):

PR #1418, opened 2024-03-18
Screw Attack Room spring ball jump

blkerby:
This adds a strat for spring ball jumping from the doorway to the top of Screw Attack Room, with HiJump (where blocks are already broken). This was something that Kewlan recently noticed was missing. I also saw that the heat frames on the "Doorway Speedjump" strat were very lenient, so I tightened them up to some extent. 

Probably other strats in the room could also use another look at the heat frames; I also notice there's no canDoubleBombJump strat, but I'm not skilled enough to work that out.

### DIFF
--- a/region/lowernorfair/west/Screw Attack Room.json
+++ b/region/lowernorfair/west/Screw Attack Room.json
@@ -869,7 +869,7 @@
     },
     {
       "link": [5, 3],
-      "name": "Space",
+      "name": "Space Jump",
       "requires": [
         "h_canNavigateHeatRooms",
         "SpaceJump",
@@ -878,7 +878,7 @@
     },
     {
       "link": [5, 3],
-      "name": "Screw Attack Room Springwall",
+      "name": "Springwall",
       "requires": [
         "h_canNavigateHeatRooms",
         "canSpringwall",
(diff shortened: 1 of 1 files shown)


region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -904,6 +903,21 @@
       ],
       "note": "Position yourself in the door way, then run and jump. Makes it possible to walljump up."
     },
+    {
+      "link": [5, 3],
+      "name": "Spring Ball Jump",
```
osse101, line 908:
```suggestion
      "name": "Doorway Spring Ball Jump",
```

osse101 (approved):
How exactly node 5 works is a little undefined.  2->5 with PBs would try to land back in the door frame.  But if the door didn't need to be opened, like with spacejump and bombs, you fall to the lower bomb blocks.  5 is drawn on the left but would be more accurate to be on the right. 

There's also a hijump+speed+springball that runs on the lower bomb blocks.  It works with partially broken lower blocks, so its slightly complicated.

blkerby:
Makes sense, I was wondering about that too. I didn't check all the strats that end at node 5 so I wasn't sure if it was safe to assume you could land back in the door. I based the heat frames on assuming you're below the door but close enough to jump directly up to it. I was curious if it might make more sense to use an obstacle to represent the broken blocks, and then the position where you end up could be less ambiguous since we would no longer need a separate node to represent that state? Just a thought for a possible follow-up later.

osse101:
Top blocks already are an obstacle but it doesn't look like it is being used much.  5 Could probably be removed.

Lower blocks also have an obstacle, but may need another for partially broken.  Where partially broken is used for jumpIntoIBJ & run+jumpIntoSpringBallJump.

Then doorframe as a new node appears to duplicate some strats but maybe makes sense that way.

PR #1420, opened 2024-03-20
Upper Norfair Farming Room Gate Glitch

kjbranch:
This gate requires significantly less leniency due to the farm and has been a reoccurring problem.

@osse101 , @blkerby - What do you think about this solution?

### DIFF
--- a/region/norfair/east/Upper Norfair Farming Room.json
+++ b/region/norfair/east/Upper Norfair Farming Room.json
@@ -639,10 +639,10 @@
     },
     {
       "link": [5, 5],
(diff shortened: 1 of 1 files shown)


osse101:
I like having the second notable strat, more for sm-json than map rando.  Since it does treat the GGG differently here.
I'd like some more heat frames for below Very Hard still.  It's important for `Hard` to have one Etank still.

blkerby:
If the heat frames were increased, say, to 300, that would be a simple way to make it require a tank on Hard but not on Very Hard, based on the heat frame multiplier. The precise value of the heat frames seems unimportant because of the fact that you can farm before and after.

PR #1421, opened 2024-03-21
Add comesThroughToilet strat property

blkerby:
This PR sets up the schema to represent vertical cross-room strats (e.g. jumps, G-mode setups, etc.) that pass through the Toilet. For now it just adds one example strat, the Mt. Everest one that was discovered/shared this morning by 123456789you123456789 on the Discord (https://discord.com/channels/1053421401354285129/1053436236628496454/1220378509046841468).

### DIFF
--- a/strats.md
+++ b/strats.md
@@ -354,13 +354,14 @@ A `leaveWithGrappleTeleport` comes with an implicit tech requirement `canGrapple
 
 ## Entrance conditions
 
-In all strats with an `entranceCondition`, the `from` node of the strat must be a door node or entrance node. An `entranceCondition` object must contain exactly one property:
+In all strats with an `entranceCondition`, the `from` node of the strat must be a door node or entrance node. An `entranceCondition` object must contain exactly one of the following properties:
 
 - _comeInNormally_: This indicates that Samus must come into the room through the specified door, with no other particular requirements.
 - _comeInRunning_: This indicates that Samus must run into the room, with speed in a certain range.
 - _comeInJumping_: This indicates that Samus must run and jump just before hitting the transition, with speed in a certain range.
 - _comeInShinecharging_: This indicates that Samus must run into the room with enough space to complete a shinecharge.
 - _comeInShinecharged_: This indicates that Samus must enter the room with a shinecharge with a certain amount of frames remaining.
+- _comeInShinechargedJumping_: This indicates that Samus must jump into the the room with a shinecharge with a certain amount of frames remaining.
 - _comeInWithSpark_: This indicates that Samus must shinespark into the room.
 - _comeInWithBombBoost_: This indicates that Samus must come into the room with a horizontal bomb boost.
 - _comeInStutterShinecharging_: This indicates that Samus must run into the room with a stutter immediately before the transition.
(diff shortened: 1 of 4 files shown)


blkerby:
Once we add these everywhere, I think we'll probably want to add a test to enforce that this `comesThroughToilet` property is always present on any strat with a vertical entranceCondition; otherwise it would be very easy to overlook and have mistakes. The default value of "no" should keep things conservative in the meantime.

blkerby:
The other thing I was thinking about is that "comesThroughToilet" could be an entrance condition, instead of a strat property as in this PR. The trouble is that then we would have to allow multiple entrance conditions on the same strat, and I'm not yet convinced that we want to go down that road, because of how complicated and error-prone it could become to define in general how multiple entrance conditions should resolve.

An example scenario (unrelated to the toilet condition) where we might want to be able to combine multiple entrance conditions would be "comeInRunning" with "comeInShinecharged". In that case the way that this combination of conditions would resolve with a "leaveWithRunway" would be different from just checking the two entrance conditions separately: you'd have to take into account that the part of the runway used to gain run speed through the transition must be subtracted from the available runway for short-charging (or alternatively, run in the reverse direction to use the full runway, but leave with fewer shinecharge frames remaining). That's why I think it could be better to just have an entrance condition like "comeInShinechargedRunning" to represent this combination, because it really has to be resolved in its own special way.

The toilet condition is different in that it's something that makes sense to apply in combination with any vertical entrance condition, and the way that it gets resolved is independent of the entrance condition. So that's what motivates defining it as a strat property rather than an entrance condition.

osse101:
I was thinking more along the lines of making `comesThroughToilet` a property of EntranceConditions which can travel through vertical doors, with a common referenced schema.  Or as part of the EntranceCondition container.

But then on `Base` type strats that just enter the room, you are hit by `comesThroughToilet` automatically so you do want a passive `comesThroughToilet : no` on all strats.  In a case like Doorless / Open Doors you could fall back through the door if you were centered, like what happens with Sporespawn.

Looking at strats that work or don't work depending on if you passed through toilet, I'd rather it not be a top level property.

strats.md
```diff
@@ -929,6 +930,32 @@ A `comeInWithGrappleTeleport` comes with an implicit tech requirement `canGrappl
...
+  "name": "Cross Room Jump - Toilet HiJump Wall Jump",
+  "entranceCondition": {
+    "comeInWithWallJumpBelow": {
+      "minHeight": 2
+    }
+  },
+  "comesThroughToilet": "yes",
+  "requires": [
+    "canCrossRoomJumpIntoWater",
+    "HiJump",
```
kjbranch, line 953:
```suggestion
    "HiJump",
    "canPreciseWalljump",
```

blkerby:
So the amount of entrance conditions that (in theory at least) could go through vertical doors is quite a few:
- comeInShinecharged
- comeInWithSpark
- comeInWithTemporaryBlue
- comeInWithRMode
- comeInWithGMode
- comeInWithWallJumpBelow
- comeInWithSpaceJumpBelow
- comeInWithPlatformBelow
- comeInWithGrappleTeleport

It would seem messy to have to add a new property in each of these places, and by having it defined in so many places it would also make it more awkward to consume, e.g. in the randomizer. 

But the point makes sense that the meaning of the property is tied to the entranceCondition, and it doesn't make sense for it to exist (even implicitly, with a default value of "no") on other kinds of strats.

So yeah, maybe it would make sense to put it in the "entranceCondition" container, having it be treated as an independent property which could co-exist with exactly one other property (one of the already existing entrance condition properties). This would be a step toward relaxing the "exactly one entrance condition" rule, which I think can be ok as long as we are clear about what kinds of combinations are allowed

blkerby:
Ok, I moved `comesThroughToilet` into `entranceCondition` now. 

A consequence was that `"minProperties": 1` and `"maxProperties": 1` could no longer be used to enforce the "exactly one" requirement on the remaining properties in the `entranceCondition` schema, so I used a `oneOf` instead.

kjbranch (approved):

osse101:
If we forget to add a new entranceCondition to the `oneOf` list, it will error, and then we can add it.

PR #1423, opened 2024-03-22
Add Artificial Morph Helpers - A Step Towards Reducing Errors

kjbranch:
Many of our errors are accidentally using items or tech that don't have item requirements instead of their helpers that do. This PR is designed to be a first step towards addressing these problems by creating and utilizing more artificial morph helpers to help reduce the total number of items/tech in the project that are common offenders. With this we can more easily search for and eventually prevent the use of error prone tech and items such as `canIBJ` and `SpringBall`. 

This also fixes a few small errors found in the project in Maridia where `Power Bombs` were accidentally used without `Morph`. It looks like there is also the reverse problem in several places in Main Street, where `enemyKill: Skultera` is used with `Power Bombs` that shouldn't require `Morph`; this will be addressed in a future PR.

### DIFF
--- a/region/crateria/west/Green Pirates Shaft.json
+++ b/region/crateria/west/Green Pirates Shaft.json
@@ -570,15 +570,18 @@
             "h_additionalBomb",
             {"or": [
               "HiJump",
-              "h_additionalBomb",
+              "h_canArtificialMorphSpringBallBombJump",
               "canJumpIntoIBJ"
             ]}
           ]}
         ]}
       ],
       "note": [
-        "With Spring Ball, it is possible to save a Power Bomb by placing it high on the first jump and bouncing on it on the second,",
-        "but this doesn't save anything if breaking the Power Bomb blocks."
+        "With Spring Ball, it is possible to save a Power Bomb by placing it on the descent of the first jump by the bottom corner of the overhang,",
+        "then bouncing on it on the ascent of the second.",
+        "It is also possible but tighter to get high enough from bouncing on a Power Bomb with a single jump, similar to jumping into an IBJ.",
+        "This doesn't save anything if breaking the Power Bomb blocks above.",
+        "With an extra Power Bomb to spare, simply Spring Ball Bomb Jump."
       ]
     },
     {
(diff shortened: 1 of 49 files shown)


kjbranch:
I ended up skipping some of the more difficult edits as well, such as those with many PBs. They will come in a future PR.

region/crateria/west/Green Pirates Shaft.json
```diff
@@ -565,11 +565,12 @@
...
-            {"ammo": { "type": "PowerBomb", "count": 2}},
+            "h_canArtificialMorphSpringBall",
+            "h_canArtificialMorphPowerBomb",
+            "h_additionalBomb",
             {"or": [
               "HiJump",
-              {"ammo": { "type": "PowerBomb", "count": 1}},
+              "h_additionalBomb",
               "canJumpIntoIBJ"
             ]}
```
osse101, line 575: Potentially a separate issue, I notice this PB doesn't help you jump high enough, but combining it  with canSpringBallBombJump would work.  The [4,10] version too.
The `canJumpIntoIBJ` seems hard at first but has a good visual cue with the side of the platform.

osse101 (approved):
Just tiny things I don't care about
- h_canArtificialMorphPowerBomb, Bombs, and SpringBall aren't "equivalent wordage" to h_canUsePowerBombs
- h_canArtificialMorphBombHorizontally was still confusing because I was comparing it to canBombHorizontally where its unclear whether a bomb is used or not, so I had to count.
- There was a mix of
```
h_canAritificialMorphPowerBomb, h_canAritificialMorphPowerBomb, h_canAritificialMorphPowerBomb

h_canAritificialMorphPowerBomb, h_additionalBomb, h_additionalBomb

{"ammo": "type": "PowerBomb", "count": 3}
```

kjbranch:
I was trying to be consistent, but was probably shifting how I was doing it as I went. But those 3 options aren't typically the same.
The first option is used if `Bombs` dont help - i.e. need Bombs + 3PB or 3PB and Bombs dont contribute.
The second option is used if `Bombs` or `3 PBs`

I was sticking to only replacing `{"ammo": "type": "PowerBomb", "count": 2}` or lower, then started including 3. I think we need a better solution for more PBs, as there are quite a few with 6+ which I am thinking we dont want to paste as 6+ lines.

The naming can definitely be changed, and it wouldnt be difficult. I do prefer not using the term "morph bombs" though, like in `h_canUseMorphBombs` ... I think that should be renamed.

kjbranch:
Currently we have:
`h_canBombThings` / `h_canArtificialMorphBombThings`
`h_canUseSpringBall` / `h_canArtificialMorphSpringBall`
`h_canUsePowerBombs` / `h_canArtificialMorphPowerBomb`
`h_canUseMorphBombs` / `h_canArtificialMorphBombs`

My vote would be to change:
`h_canUsePowerBombs` -> `h_canUsePowerBomb`
`h_canUseMorphBombs` -> `h_canUseBombs`

osse101:
> My vote would be to change: 
> `h_canUsePowerBombs` -> `h_canUsePowerBomb`
> `h_canUseMorphBombs` -> `h_canUseBombs`

I like it.  `h_canUse<Item>`

PR #1426, opened 2024-03-23
Add comesThroughToilet data on strats

blkerby:
This adds "comesThroughToilet" data to all strats having an "entranceCondition" through a vertical transition, to specify whether or not the strat is valid when passing through the Toilet. This is necessary because of how the Toilet affects Samus' horizontal and vertical spawn position in the next room.

A few other things:
- Added some new strats in cases where the Toilet makes an important difference.
- For some strats where the Toilet has a tiny effect, e.g. on shinecharge frames, I just adjusted the strat slightly to adapt to this, rather than creating variant strats.
- Set up tests to verify that "comesThroughToilet" is always (and only) specified on strats having an "entranceCondition" through a vertical transition.

Note that although Map Rando does not allow the Toilet to connect to sand transitions, the logic here still handles that possibility, and it is possible to test these cases in the Practice Hack. For elevators, neither Map Rando nor the Practice Hack allow a connection with the Toilet (or other vertical doors), but for completeness we still include these cases, always with "comesThroughToilet": "any", under the assumption that passing through the Toilet would have no effect when spawning into an elevator ride.


tests/asserts/keywords.py
```diff
@@ -370,6 +370,26 @@ def process_strats(src, paramData):
...
+connectionPath = os.path.join(".","connection")
+for root, dirs, files in os.walk(os.path.join(".", "connection")):
+    for filename in files:
+        if not filename.endswith(".json"):
+            continue
+        with open(os.path.join(root, filename), "r", encoding="utf-8") as connectionFile:
+            connections_json = json.load(connectionFile)
+            for connection in connections_json["connections"]:
+                for i, node in enumerate(connection["nodes"]):
+                    if node["position"] in ["top", "bottom"]:
```
osse101, line 389: `connectionType` in tears 😢

osse101 (commented):
- SandHalls entry would work differently... without hijump you get stuck in the sand.  West sand hall needs to actively dodge the evir shot.  And I'm just noticing most rooms let you enter Sandhalls with high fall speed so the evir dodge is easy.
- Vertical doors perhaps could use a `Position` property too.  Not sure that it matters enough yet.
- Blue hopper dodge should get additional requirements.
- Botwoon quicksand room may need new strats cause I miss the morph through the crossover tunnel all the time.  But I've still never needed to do it in maprando.
- `Fish Tank Entry Reset Fall Speed`, Freeze the fish, and fish damage boost do not work.  But the walljump entry does work.  They should have EntranceConditions added then, maybe ignoring that they do require entering in a particular way.   

Nothing stood out as a problem, and I can do a follow up to work on the strat changes I mentioned here.

tests/asserts/keywords.py
blkerby: True, `connectionType` is there, and with that we could easily exclude the elevators, which might be better.

blkerby: Ok, I switched it over to use `connectionType` now and excluded the elevators since the data isn't meaningful for them.

I found one case where the `connectionType` was wrong, so this also fixes that.

PR #1428, opened 2024-03-24
Refine some Three Musketeers strats

blkerby:
Some of the strats that killed Kihunters from top to bottom assumed the top would be left alive. This was a problem because these were clearing obstacle "A", which meant the strats going back from bottom to top would assume they were all three dead.

If we want to correctly model leaving the top one alive, we would probably need to add another obstacle and take it into account in the strats that go back up. But it wasn't clear if this would be useful. So I simplified the strats to assume you kill all 3 or none at all.

Also checked the heat frames for the top-to-bottom strats and tightened some of them. Removed an unnecessary "Morph" requirement from the dodging strat going down; Morph will be needed anyway to do anything at the bottom of the room but it's not used as part of this strat.

Also removed the "canCarefulJump" requirement from the safe version of the Super kill strat. I assume it was for firing 2 Supers per cycle, but with the heat frame leniency on low settings it wouldn't be expected to do this.

### DIFF
--- a/region/lowernorfair/east/Three Musketeers' Room.json
+++ b/region/lowernorfair/east/Three Musketeers' Room.json
@@ -272,7 +272,10 @@
         {"heatFrames": 800},
         {"or": [
           "canDodgeWhileShooting",
-          {"heatFrames": 800}
+          {"and": [
+            "canCarefulJump",
+            {"heatFrames": 800}
+          ]}
         ]}
       ],
       "clearsObstacles": ["A"],
--- a/region/lowernorfair/east/Three Musketeers' Room.json
+++ b/region/lowernorfair/east/Three Musketeers' Room.json
@@ -184,8 +184,7 @@
       "requires": [
         "h_canNavigateHeatRooms",
         "canTrickyJump",
-        "Morph",
-        {"heatFrames": 540}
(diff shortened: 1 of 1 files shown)


blkerby:
On second thought, "canCarefulJump" makes sense just for positioning close enough to the edge to be able to jump down and fire Supers, while under pressure of heat damage; you have to get fairly close and it's easy to fall off. So I put that back now.

osse101:
My thought was to make new strats for kill all 3 in addition to the only kill 2, since you could save ammo and heatframes.  But when playing it, its scary to leave the top one alive, and I'd rather kill all 3 in a kill strat.

blkerby:
> My thought was to make new strats for kill all 3 in addition to the only kill 2, since you could save ammo and heatframes. But when playing it, its scary to leave the top one alive, and I'd rather kill all 3 in a kill strat.

Yeah, it seems like if you're not going to be coming back up, then it would make more sense to just dodge them all rather than killing 2? I'm not sure about strats that come back up (after getting the item); I assume you'd have to kill that top one anyway, but I guess dodging it or tanking a hit could be a possibility. I guess there are probably a lot of variations of mixed killing/dodging/tanking combinations; just not sure which are useful/important to model.

PR #1432, opened 2024-03-27
Add leaveNormally exit condition to downward door & sand transitions

blkerby:
This is just something simple to allow #1429 to go through.

I'm thinking that at some point we may want to add more detail such as describing precisely which types of states may be assumed by this condition, e.g. leaving on a specific side of the door, leaving with a walljump or an unmorph, and we may want to add additional properties or conditions to express those things more concretely in the strats that assume coming into the room in one of these ways.

Eventually I'd still be interested in moving to a setup where exit/entrance conditions are the only logical way to pass to another room. That means we'd need to add `comeInNormally` strats everywhere, or at least define where they are implicit and allow them to be overridden. This could replace the need for the `spawnAt` property (which has always been a somewhat annoying, error-prone thing to have to deal with), as you could instead just have a `comeInNormally` strat that leads to a different node from the entrance node.


osse101 (dismissed):
Leaving through sand rooms is a little more complicated for leaving in a state that isn't just falling.  It's something that could be expanded on later.  More of the issue is that my Fish Tank `comeInNormally` aren't normal, but really have a `position:left` or `comeInJumping` condition.  Sand lets you be on either side easily enough but jumping through the transition can be kinda hard, if the ledge spinjump trick even works.

PR #1436, opened 2024-03-28
Add canBeExtremelyPatient, canLongCeilingBombJump, G-Mode in Red Firefleas

kjbranch:
I'm adding this in hopes that it will be used as a `Beyond` tech in Map Rando. 

Alternatively/additionally, I wouldnt be opposed to a `canLongCeilingBombJump`, which could be `Beyond` and have a lower threshold (i.e. 2 screens instead of 4).

### DIFF
--- a/region/brinstar/red/Red Brinstar Fireflea Room.json
+++ b/region/brinstar/red/Red Brinstar Fireflea Room.json
@@ -326,7 +326,8 @@
       "notable": true,
       "requires": [
         "h_canCeilingBombJump",
-        "canBeExtremelyPatient"
+        "canLongCeilingBombJump",
+        "canBeVeryPatient"
       ],
       "reusableRoomwideNotable": "Red Brinstar Firefleas Ceiling Bomb Jump",
       "note": [
@@ -346,7 +347,8 @@
       },
       "requires": [
         "h_canArtificialMorphCeilingBombJump",
-        "canBeExtremelyPatient",
+        "canLongCeilingBombJump",
+        "canBeVeryPatient",
         "canInsaneJump"
       ],
       "reusableRoomwideNotable": "Red Brinstar Firefleas Ceiling Bomb Jump",
@@ -534,7 +536,8 @@
       "notable": true,
       "requires": [
         "h_canCeilingBombJump",
-        "canBeExtremelyPatient"
+        "canLongCeilingBombJump",
+        "canBeVeryPatient"
       ],
       "reusableRoomwideNotable": "Red Brinstar Firefleas Ceiling Bomb Jump",
       "note": "This is a very long ceiling bomb jump. Kill the Wavers before crossing the room."
@@ -553,7 +556,8 @@
         "h_canArtificialMorphBombHorizontally",
         "h_canArtificialMorphIBJ",
         "h_canArtificialMorphCeilingBombJump",
-        "canBeExtremelyPatient",
+        "canLongCeilingBombJump",
+        "canBeVeryPatient",
         "canInsaneJump"
       ],
       "reusableRoomwideNotable": "Red Brinstar Firefleas Ceiling Bomb Jump",
--- a/region/brinstar/red/Red Brinstar Fireflea Room.json
+++ b/region/brinstar/red/Red Brinstar Fireflea Room.json
@@ -352,7 +352,7 @@
       "reusableRoomwideNotable": "Red Brinstar Firefleas Ceiling Bomb Jump",
       "note": [
         "This is a very long ceiling bomb jump.",
-        "Crossing the room with artificial morph is particularly diffuclt without a good way to kill the Wavers."
+        "Crossing the room with artificial morph is particularly difficult without a good way to kill the Wavers."
       ]
     },
     {
@@ -559,7 +559,7 @@
       "reusableRoomwideNotable": "Red Brinstar Firefleas Ceiling Bomb Jump",
       "note": [
         "This is a very long ceiling bomb jump.",
-        "Crossing the room with artificial morph is particularly diffuclt without a good way to kill the Wavers.",
+        "Crossing the room with artificial morph is particularly difficult without a good way to kill the Wavers.",
         "It is recommended to avoid killing the Firefleas, as the room gets dark fast."
       ]
     },
--- a/region/brinstar/red/Red Brinstar Fireflea Room.json
+++ b/region/brinstar/red/Red Brinstar Fireflea Room.json
@@ -90,6 +90,12 @@
       "obstacleType": "inanimate"
     }
   ],
+  "reusableRoomwideNotable": [
+    {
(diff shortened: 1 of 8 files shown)


osse101:
I think Moat has a long but reasonable ceiling that I've seen people go for in challenge runs.  And I already can't do that one.  So that's similar to 1 screen | more than 1 screen.  (CIBJ window widener Max QoL patch 😸, super zero mission has a wider window and its still bad.)

Main street Ice only is my only video over 5 minutes, but there are a few other strats that would fit; mostly GMode IBJs.
There could be room to change the meaning of canBeXXXXPatient to include waiting for less long but also doing a difficult trick after waiting.  But I don't think it's been an issue in map rando since the patient placements are so high.

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -314,6 +320,41 @@
...
+      },
+      "requires": [
+        "h_canArtificialMorphCeilingBombJump",
+        "canBeExtremelyPatient",
+        "canInsaneJump"
+      ],
+      "reusableRoomwideNotable": "Red Brinstar Firefleas Ceiling Bomb Jump",
+      "note": [
+        "This is a very long ceiling bomb jump.",
+        "Crossing the room with artificial morph is particularly diffuclt without a good way to kill the Wavers."
```
blkerby, line 355:
```suggestion
        "Crossing the room with artificial morph is particularly difficult without a good way to kill the Wavers."
```

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -487,6 +528,41 @@
...
+        "h_canArtificialMorphBombHorizontally",
+        "h_canArtificialMorphIBJ",
+        "h_canArtificialMorphCeilingBombJump",
+        "canBeExtremelyPatient",
+        "canInsaneJump"
+      ],
+      "reusableRoomwideNotable": "Red Brinstar Firefleas Ceiling Bomb Jump",
+      "note": [
+        "This is a very long ceiling bomb jump.",
+        "Crossing the room with artificial morph is particularly diffuclt without a good way to kill the Wavers.",
```
blkerby, line 562:
```suggestion
        "Crossing the room with artificial morph is particularly difficult without a good way to kill the Wavers.",
```

kjbranch:
The main changes:
- Redefined `extremely patient` as 6 minutes and removed it from the long ceiling IBJs (which were both ~4 minutes and 4 screens).
- Made a new tech `canLongCeilingBombJump`, which is defined as 2+ screens (but could be changed), which can also be `Beyond` in Map Rando. Applied it to all the current ceiling IBJs that should have it (there are many missing ceiling IBJs).

The most notable strats that didn't make the cut for `extremely patient`, but it could be added if you think it's worthwhile: 
- GT charge+varia is about 5.5 minutes. But it's easy and seems much nicer than the 4 minute missile only fight, so I thought it could remain in `Insane`.
- Main St full climb with ice and limited supers was about 7:45, (7 min from one door up). This one is long, but it feels like youre constantly making progress. Though there are two sections where you have to wait 2 minutes each. 
- Crateria supers G-mode ceiling bomb jump takes about 4:30-6:30 with bombs alone. This consists of 2 minutes to kill the Boyons, which could be cut with PBs if available, 1 minute to double bomb jump or 2 to single, then ceiling IBJ for 20-30s then climb again. Worst case, this could have `extremely patient` or `double IBJ` or `a few PBs` added.

region/tourian/main/Metroid Room 1.json
```diff
@@ -281,7 +281,9 @@
       },
       "requires": [
         "f_KilledMetroidRoom1",
-        "h_canArtificialMorphCeilingBombJump"
+        "h_canArtificialMorphCeilingBombJump",
```
osse101, line 284: This strat was already here so perhaps a comment for later, but I don't think you can get past the center rinka.  
With the Morph item strat you jump to the floating platforms first.
A diagonal bombjump looks like you could make it with some luck.

[2,1] could HBJ to the floating platform and ceiling/diagonal IBJ from there. And HBJ from the left platform to end.  A Purely ceiling IBJ wouldn't get past the double Rinkas.

region/wreckedship/main/Bowling Alley.json
```diff
@@ -428,9 +422,9 @@
       "notable": true,
       "requires": [
         "h_canCeilingBombJump",
+        "canLongCeilingBombJump",
         "canBeVeryPatient"
```
osse101, line 426:
```suggestion
        "canBeExtremelyPatient"
```

region/wreckedship/main/Wrecked Ship Energy Tank Room.json
```diff
@@ -380,17 +384,22 @@
...
       "name": "G-mode Morph Ceiling Bomb Jump",
       "requires": [
         "h_canArtificialMorphCeilingBombJump",
+        "canLongCeilingBombJump",
+        "canBeVeryPatient",
         {"obstaclesCleared": ["A"]}
       ],
-      "note": "Touch the item while remaining in artificial morph. Ceiling bomb jump back to the right, then use x-ray to cancel g-mode and obtain the item."
+      "note": "Touch the item while remaining in artificial morph. Ceiling bomb jump back to the right, then use x-ray to cancel g-mode and obtain the item.",
+      "devNote": "This strat alone would only require canBePatient, but it is only possible after Ceiling Bomb Jumping there, so it would be a combined 4 minues."
```
osse101, line 402:
```suggestion
      "devNote": "This strat alone would only require canBePatient, but it is only possible after Ceiling Bomb Jumping there, so it would be a combined 4 minutes."
```

region/wreckedship/main/Bowling Alley.json
osse101: Looks longer than the fire flea CIBJ

region/tourian/main/Metroid Room 1.json
kjbranch: The Rinkas shouldn't spawn once killed once with Bombs

region/wreckedship/main/Bowling Alley.json
kjbranch: I'm on the fence about this one. It will be in `Beyond` either way, but this one should only take 4 minutes (according to your notable video).

region/tourian/main/Metroid Room 1.json
osse101: You are right, pseudo GMode treated them incorrectly.

region/wreckedship/main/Bowling Alley.json
osse101: I didn't think I ran that CIBJ to the end.  I like very patient then.  My setup could have been done a little faster whereas in firefleas you have to figure out how to kill the Wavers.

PR #1437, opened 2024-03-29
Crossing Crocomire's Room acid

blkerby:
We can refine the ceiling bomb jump requirements here once #1436 is merged.


kjbranch (commented):
Would an air ball + wall jump be useful?
the way I've input the ceiling bomb jump strats in #1436, this would be 2 screens (2 minutes), so `canBePatient` and `canLongCeilingBombJump`

osse101:
Airball saves a good 16frames in the acid, gravityless + left to right.

PR #1438, opened 2024-03-29
Add partialRefill logical requirement, apply to CF helpers

blkerby:


osse101 (commented):
The content looks good.
I think I like "Limit" more than `Level`.  `Amount` is normally a good word, but isn't quite right here.

helpers.json
```diff
@@ -609,22 +673,29 @@
           ]
         },
         {
-          "name": "h_canCrystalFlash",
+          "name": "h_CrystalFlashBase",
```
kjbranch, line 676: Not sure I like `Base`. Maybe `Requirements`?

kjbranch (commented):
I think `Limit` sounds good

PR #1441, opened 2024-03-30
Some grapple teleports from Halfie Climb

blkerby:
This introduces a new tech "canGrappleBombHang", with a grapple teleport setup from the top right of Halfie Climb using it. It wasn't as bad as I was expecting. It's different but seems not necessarily harder than the Moat setup.

I'm planning on going through the rest of the rooms to add all the applications (from this door) but figured it could make sense to start with a smaller PR to get feedback.

This setup has a little more nuance because it looks like that there are multiple ways to enter the transition (e.g. by varying your momentum) which can slightly affect your position in the next room, so we'll have to be careful with that. Also this setup has the ability to bypass door shells and even chain the grapple teleport into subsequent rooms.

### DIFF
--- a/region/brinstar/blue/Construction Zone.json
+++ b/region/brinstar/blue/Construction Zone.json
@@ -87,6 +87,7 @@
       "from": 3,
       "to": [
         {"id": 1},
+        {"id": 2},
         {"id": 3},
         {"id": 4}
       ]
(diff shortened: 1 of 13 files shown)


osse101 (approved):
What's here looks to work.

And the tech isn't as bad as it sounds, like you said.  Grappling the wall after a miss lets you retry pretty fast.  
I didn't look at momentum into the transition.  The lag is a pain when just trying to reach the door.  So I idled at the transition and slowly went through.

PR #1443, opened 2024-03-30
More Halfie Climb top-right grapple teleports

blkerby:
This should be all of them, except I'll go back and add all the "Carry Grapple Teleport" strats after this.


helpers.json
```diff
@@ -222,6 +222,14 @@
           "name": "h_SupersDoubleDamageMotherBrain",
           "requires": [ "never" ],
           "devNote": "In Vanilla, Supers do 300 damage to Mother Brain. Randomizers may want to double this to help prevent Charge being required so frequently."
+        },
+        {
+          "name": "h_BypassMotherBrainRoom",
```
osse101, line 227: Is Morphless a reality 🪄 

This helper is meant to skip the zebs+map rando blockade+be in position to hit MB.
There was mention of a teleport to the left of MB and you could think the wall would be open since it is a door when entering from the left, but room state won't have it there I don't think.  As a way to travel through the room without fighting anything.
And I want to say Colosseum lets you teleport to the right door but I don't know what room state looks like.  And I find right side teleports harder to imagine.

region/lowernorfair/east/Lower Norfair Escape Power Bomb Room.json
```diff
@@ -93,6 +94,32 @@
...
+      "note": [
+        "In the previous room, avoid triggering the transition too deeply, otherwise Samus can end up stuck too far right in the wall.",
+        "After teleporting, Samus should be standing inside the wall.",
+        "Wait a moment for the acid to lower, then retract Grapple by pressing up, which will pull Samus left and down.",
+        "Hold left, and release Grapple while still holding left.",
+        "Perform a turn-around spin jump (to the right).",
+        "Then morph and roll out to the left."
+      ],
+      "devNote": [
+        "Coming in from position [12, 12] can also work but is more difficult:",
```
osse101, line 119: Looks like my bottom position save was too deep but my top position save had no problem with this.

region/norfair/west/Hi Jump Energy Tank Room.json
```diff
@@ -150,6 +151,26 @@
...
+      "requires": [
+        "canXRayClimb",
+        "Morph",
+        "canOffScreenMovement"
+      ],
+      "note": [
+        "After teleporting, hold left and release Grapple to remain in a standing pose (and not be force crouched).",
+        "X-ray climb a specific distance, performing exactly 15 X-Ray stand-ups.",
+        "Then morph and roll out to the right.",
+        "If Bombs or Power Bombs are available, this can allow more lenience in how far to climb."
```
osse101, line 171: Screw Attack too

osse101 (commented):
I'm suspecting you take a frame of heat damage when door bypassing heat rooms.  And that would only matter for OHKO

region/lowernorfair/east/Lower Norfair Escape Power Bomb Room.json
osse101: It works if I hold right while hitting the transition, as opposed to coming to a stop inside the transition and then pressing right. 👍

helpers.json
blkerby: Yeah the intent is to use this helper in Map Rando to add requirements for the objectives to be logically complete. Getting to the MB in this way is fine, but we wouldn't want it to result in objectives not being completable. Being able to skip the objectives may be too powerful, so we might eventually also want to enforce them in some other way (e.g., not being able to hurt MB until they are complete).

The left door being already blown open is a thing that Map Rando adds to avoid getting instantly softlocked, and it only does it when entering through that door. So in Map Rando, teleporting to the left side of the room from the right side (e.g. from Moat or Red Brin Firefleas) doesn't seem to be useful, as you would just be stuck, unless the glass is already broken, but in that case you would have had to have gone through the room normally earlier.

blkerby:
> I'm suspecting you take a frame of heat damage when door bypassing heat rooms. And that would only matter for OHKO

I didn't think about this, but trying it out now, it looks like if you hold forward through the transition, it triggers the next transition on the first frame and you never take heat damage. I tried it like 15 times or so, at different times, to try to rule out the possibility of getting lucky 3/4 chance every time

PR #1446, opened 2024-03-31
Refine WS ETank grapple fling

blkerby:
This came up in Oats' seed tonight where you had to use Grapple to get this item and get out without taking a hit. It was an Extreme seed but it could have happened in Hard.

I'm not sure if I completely understand this strat. For some reason, to me going left-to-right seems way easier than going right-to-left, even though the geometry is symmetric so I would have expected them to be basically the same. A lateral mid-air morph seems to help a lot with going right-to-left but not really with going left-to-right.

### DIFF
--- a/region/wreckedship/main/Wrecked Ship Energy Tank Room.json
+++ b/region/wreckedship/main/Wrecked Ship Energy Tank Room.json
@@ -275,9 +275,20 @@
       "link": [1, 2],
       "name": "Precise Grapple",
       "requires": [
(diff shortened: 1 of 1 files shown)


osse101 (approved):
Can confirm it took me a very long time to grapple across right-to-left when it came up in one of my seeds.  And when I was trying this when it was first added I did left-to-right, and it felt like you could get consistent with some practice.  Which makes Hard not a good place for it.

PR #1454, opened 2024-04-04
Colosseum top-left door grapple teleports

blkerby:
I think this door is the one with the most applications, especially in terms of how many ways there are to chain it into subsequent rooms. All it takes is a door on the left of the top-left screen of the room, which is very common.

Also, for cases where there's a wall instead of door there, the further-out position of the grapple block (3 tiles vs. 2 tiles for the left side of Halfie Climb) means in some cases you can retract Grapple to pull Samus out to the right, to avoid being stuck inside the wall; it's like the mirror image of the strat that teleports into Jail from the top-right of Halfie Climb.

### DIFF
--- a/region/maridia/inner-pink/Crab Shaft.json
+++ b/region/maridia/inner-pink/Crab Shaft.json
@@ -658,7 +658,7 @@
       "name": "Grapple Teleport",
       "entranceCondition": {
         "comeInWithGrappleTeleport": {
-          "blockPositions": [[5, 3], [7, 2]]
+          "blockPositions": [[5, 3], [7, 2], [3, 12], [3, 13]]
         }
       },
       "requires": []
--- a/region/maridia/inner-pink/Crab Shaft.json
+++ b/region/maridia/inner-pink/Crab Shaft.json
@@ -784,6 +784,26 @@
       ]
     },
     {
+      "link": [2, 3],
+      "name": "Grapple Teleport",
+      "entranceCondition": {
+        "comeInWithGrappleTeleport": {
+          "blockPositions": [[3, 12], [3, 13]]
+        }
+      },
+      "requires": [
+        {"or": [
+          "canUseFrozenEnemies",
+          "HiJump",
+          "canSpringBallJumpMidAir"
+        ]}
+      ],
+      "devNote": [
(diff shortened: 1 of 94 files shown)


region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -810,6 +810,26 @@
         "FIXME: Add a variation with canRiskPermanentLossOfAccess if the door is not blue."
       ]
     },
+    {
+      "link": [2, 3],
```
osse101, line 814: [2, 1] version that just drops down

region/norfair/east/Single Chamber.json
```diff
@@ -243,6 +243,33 @@
         {"heatFrames": 45}
       ]
     },
+    {
+      "link": [2, 1],
```
osse101, line 247: 3, 4, & 5

region/wreckedship/main/Wrecked Ship Energy Tank Room.json
```diff
@@ -373,6 +373,26 @@
...
+          "blockPositions": [[3, 12], [3, 13]]
+        }
+      },
+      "requires": [
+        "canOffScreenMovement",
+        "Morph"
+      ],
+      "note": [
+        "After teleporting, Samus should be standing inside the wall.",
+        "Retract Grapple by pressing up, which will pull Samus down and right.",
```
osse101, line 390: From top position I had to either end the grapple retract early and do a turn and roll out of the wall, or CF and xray climb.  I couldn't end the grapple in a way that let me spinjump.

osse101 (commented):
+Gauntlet Entrance
+Draygon Save
+Plasma Tutorial Room

region/wreckedship/main/Wrecked Ship Energy Tank Room.json
blkerby: Interesting, tried it again and found that even from a save state after the teleport, sometimes after retracting and turning right it leaves me standing where I can spin jump, and other times it leaves me in a force crouch.

Looks like if after retracting I turn around left before turning back right, then it always leaves me standing. Added that to the note.

PR #1455, opened 2024-04-05
A Few G-Mode Setups in Unheated Norfair

kjbranch:
If the Frog Speedway [1, 2] leave with G-Mode is too hard for no requirements, I can convert the speed blocks to an obstacle and split the strat.

### DIFF
--- a/region/norfair/east/Frog Speedway.json
+++ b/region/norfair/east/Frog Speedway.json
@@ -147,6 +147,22 @@
       ]
     },
     {
+      "link": [1, 1],
+      "name": "G-Mode Setup - Get Hit By Beetom",
+      "notable": false,
+      "requires": [],
+      "exitCondition": {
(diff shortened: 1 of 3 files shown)


osse101:
Frog Speedway:
[11] I run at walking speed and tap dash softly so that I'm running the same speed as the frog.
[12] Jumping over the beetom, and staying on its left for its next jump is tricky but fair.  There should be another tech to describe that, but feels like an ok assumption of something a GMode player could do.
[21] Jumping the frogs should be RNG, but isn't hard and works most of the time.
[22] I don't know if there is a completely safe way to lure a beetom to the door without getting early grabbed but there are a few ways that work most of the time.  I do a slow run to break just enough speedblocks to free a frog.  The most in control way seems to be doing the [21] followed by the [12] strats.

PR #1457, opened 2024-04-07
Draygon's Room grapple teleports

blkerby:


osse101 (commented):
`canUnmorphGrappleHang` with Gravity felt frame perfect.
But suitless with HiJump its kinda reasonable.  Jump+angle down for lower block, but grapple+Jump+angle down for top block.  And some unmorph locations wouldn't allow for lower block.

The setups with bombs feel reliable except for reaching the transition.  Maybe you can learn the inputs to just go through, but I bonk the wall most of the time and then making it through takes twice as long.  And still, you may need to exit in a certain way.  So I'd like more damage leniency added.

blkerby:
If it sounds all right, I think I'll leave the `canUnmorphGrappleHang` strats out of this PR (leaving them as a FIXME). It seems like there's a lot that's possible using that and also grapple jumps to reach the higher position, but it will be a fair bit of work to figure it all out and I've got some other things I should probably be turning my attention to for now. Feel free to add these in a follow-up PR if you're interested.

PR #1461, opened 2024-04-12
Colosseum bomb jump water escape by Spiffy

blkerby:
Spiffy came up with this strat the other day in one of his seeds:
https://clips.twitch.tv/RepleteGeniusRabbitHassanChop-ng1dgYjhS-MVH-Il

Compared to the Insane "Colosseum with no Equipment" strat, this version uses one less spike hit, and it avoids the final spike jump which was the insanely difficult part. Escaping the water with the spring ball jump into IBJ is precise but can be retried as many times as needed. I put "canInsaneJump" on it since the frame-perfect bomb placement (2 frames before the spring ball jump) makes it probably take a while to get, given that you have to pause/unpause twice with every attempt.

### DIFF
--- a/region/maridia/inner-pink/Colosseum.json
+++ b/region/maridia/inner-pink/Colosseum.json
@@ -1060,6 +1060,30 @@
     },
     {
       "link": [3, 1],
+      "name": "Colosseum Bomb Jump Water Escape",
+      "notable": true,
+      "requires": [
+        "HiJump",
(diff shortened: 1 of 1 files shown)


osse101:
A SpringFling lets you place the bomb later, probably with a larger window, but I don't think it makes it any easier.

kjbranch (approved):

osse101 (approved):

PR #1464, opened 2024-04-14
Blue Hopper left-to-right dodging and Screw Attack

blkerby:
This is the Blue Hopper dodge that Eddie did in his seed yesterday: https://clips.twitch.tv/DistinctSwissFiddleheadsLeeroyJenkins-pYlLF0RMYe6yet2H.

Also a new, easy way to get through with Screw Attack, obsoleting the other Screw Attack strats. I put "canTrickyJump" on it only because the movement is potentially unintuitive; it's very lenient, with no specific timings on the jumps.

### DIFF
--- a/region/tourian/main/Blue Hopper Room.json
+++ b/region/tourian/main/Blue Hopper Room.json
@@ -340,45 +340,18 @@
     },
     {
       "link": [2, 1],
-      "name": "Screw Attack with Space Jump",
-      "requires": [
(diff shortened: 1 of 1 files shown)


osse101:
New screw strat does look like trickyJump movement.
The weird part is spacejump and screw attack being in very hard when they are so intuitive.
But that's about where I'd trust them to avoid the single hopper hit, so it must be right.

PR #1466, opened 2024-04-15
Spring fling bomb water escapes

blkerby:
This is based on insomniaspeed's technique:
https://youtu.be/hf-dM3JpZbk?si=Le3jZ-xWsN_LkQz4
https://youtu.be/1du4nXXcD8c?si=c007eDdU5cQOgFy1

After learning the timing and practicing for a bit, the spring fling method in Colosseum does seem much easier.

### DIFF
--- a/region/maridia/inner-green/Pants Room.json
+++ b/region/maridia/inner-green/Pants Room.json
@@ -638,6 +638,7 @@
         "h_canMaxHeightSpringBallJump",
         "canSpringFling",
         "canBombJumpWaterEscape",
+        "canDoubleBombJump",
         "h_canJumpIntoIBJ"
       ],
       "note": [
--- a/region/maridia/inner-green/Pants Room.json
+++ b/region/maridia/inner-green/Pants Room.json
@@ -630,7 +630,7 @@
       ]
     },
     {
-      "link": [3, 1],
+      "link": [5, 2],
       "name": "Pants Room Bomb Jump Water Escape",
       "notable": true,
       "requires": [
--- a/region/maridia/inner-green/Pants Room.json
+++ b/region/maridia/inner-green/Pants Room.json
@@ -630,6 +630,30 @@
       ]
     },
     {
+      "link": [3, 1],
+      "name": "Pants Room Bomb Jump Water Escape",
+      "notable": true,
+      "requires": [
+        "HiJump",
+        "h_canMaxHeightSpringBallJump",
+        "canSpringFling",
+        "canBombJumpWaterEscape",
+        "h_canJumpIntoIBJ"
+      ],
+      "note": [
+        "Wait the water tide to reach its peak, then crouch jump into a spring ball jump into an IBJ.",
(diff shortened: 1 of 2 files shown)


region/maridia/inner-pink/Colosseum.json
```diff
@@ -1073,16 +1073,28 @@
...
         "h_canJumpIntoIBJ"
       ],
       "note": [
         "Stay out of the water, and by extension the sand, of Colosseum by using the spikes as platforms.",
         "Morphing before landing on the spikes helps to be able to control the knockback.",
         "The final spike jump (which would be the most difficult) is avoided by doing a crouch jump into spring ball jump into IBJ to reach the left door.",
-        "The spring ball jump should be performed just before max height, during the 4-frame window between 2 and 5 frames before the last possible frame to jump (there are also 1-frame windows at 0 and 8 frames before the last frame).",
-        "In every case the first bomb must be placed exactly two frames before the spring ball jump."
+        "Perform the spring ball jump near max height.",
+        "Place the first bomb between about 4 and 6 frames after the spring ball jump; ideally it should be just above the water line.",
```
osse101, line 1088: I pause before placing the bomb.

region/maridia/inner-green/Pants Room.json
```diff
@@ -629,6 +629,30 @@
...
+    {
+      "link": [5, 2],
+      "name": "Pants Room Bomb Jump Water Escape",
+      "notable": true,
+      "requires": [
+        "HiJump",
+        "h_canMaxHeightSpringBallJump",
+        "canSpringFling",
+        "canBombJumpWaterEscape",
+        "h_canJumpIntoIBJ"
```
osse101, line 641: Sorta want to add canDoubleBombJump since the water catches up to me half the time.
Bombing into a springballjump is halfway to being a new tech.

kjbranch: It's a really weird mechanic that I thought wasn't possible. It could be useable in walljumpless hell runs, but seems a bit rough. It's also possible to kill them with a PB, but it's kinda tight.

region/maridia/inner-pink/Colosseum.json
blkerby: It can work, yes. But from my testing, doing this only gives a 1-frame or 2-frame window for placing the bomb, compared to a 5-frame window with pausing later. It makes sense because when pausing earlier the pause end ups hitting near the peak of the jump, where the fall speed reset doesn't do as much; by pausing later you get the fall speed reset after you've already accumulated a significant amount of fall speed, which buys more time.

region/maridia/inner-green/Pants Room.json
blkerby: Yeah it looks like you can keep getting Spring ball jumps indefinitely in this way to speed up a long IBJ (at least in terms of in-game time, e.g. for heat damage)?

PR #1467, opened 2024-04-16
Ice beam gate room temporary blue

blkerby:


blkerby:
Nito found this one, did it on stream a while ago.

region/norfair/west/Ice Beam Gate Room.json
```diff
@@ -359,6 +359,70 @@
...
+          "length": 4,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canChainTemporaryBlue"
+      ],
+      "clearsObstacles": ["A", "B"],
+      "note": [
+        "Use a temporary blue chain to jump through the bomb blocks and land on the center of the ledge above.",
```
osse101, line 376: I'd rather not land on the ledge above and go directly for the bomb block.
If you do land above, this needs can4Highmidairmorph.

region/norfair/west/Ice Beam Gate Room.json
```diff
@@ -359,6 +359,70 @@
...
+        {"or": [
+          {"canShineCharge": {
+            "usedTiles": 14,
+            "openEnd": 0
+          }},
+          {"and": [
+            {"canShineCharge": {
+              "usedTiles": 15,
+              "openEnd": 0
+            }},
```
osse101, line 412:
```suggestion
            }},
            "can4HighMidAirMorph",
```

osse101 (commented):
[4, 5] has a use in setting up GMode.

The jump into the 1 tile bomb block is pretty hard.  I don't like it.

region/norfair/west/Ice Beam Gate Room.json
blkerby: I found going directly for the bomb block to be much more difficult. Just checked with Bizhawk and confirmed that doing it that way requires a frame-perfect morph (at least from the position I tested), whereas from the ledge there is a 2-frame window. Adding can4Highmidairmorph sounds good.

PR #1472, opened 2024-04-17
Add G-Mode in Ice Beam Gate Room

kjbranch:


region/norfair/west/Ice Beam Gate Room.json
```diff
@@ -639,6 +720,49 @@
         "canMockball"
       ]
     },
+    {
+      "link": [4, 2],
+      "name": "G-Mode Morph Overload PLMs",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "any",
```
osse101, line 728: Indirect shouldn't need to overload PLMs.

Also, its a way to move [2, 4]

osse101 (commented):
I thought it'd be cool to climb the crumbles to 1, but then you can't end GMode.

The Mellas aren't as much of a problem as I thought for [37] so trickyjump is fine.

PR #1482, opened 2024-04-22
Add collectsItems and setsFlags

### LINKED ISSUE #1086: Collect Items During a Strat? (for Big Pink remote acquire g-mode)
Remote acquiring items while in g-mode has been regularly handled by setting an obstacle cleared and then using that to return. One thing to be careful with, is that the requirements to get back can't use the item itself, so all of the requirements are included on the strat getting there, which sets the obstacle cleared.

Big Pink is particularly difficult, because there are two items that can be obtained (10, 11) and many doors that can be started from (2, 3, 6, 8) and locations to go after obtaining one or both items (1, 4, 8, 12). Setting the requirements on the obstacle before obtaining the item is not possible, because it depends where you go, so this can only be remedied with making a ton of obstacles. However, because this room already has many obstacles and links, the room is also already very slow to compute, and adding more obstacles will make this much worse. Even if this were possible, it would still have major problems: go to item 1 setting the remote acquire obstacle, unmorph and get item 2, return to 1, and the obstacle will still be set for remote acquiring with artificial morph.

The only way I can think of solving this is by adding a way to obtain an item while performing a strat, similar to how obstacles are cleared. This is likely a pain to implement, but it could be used in other locations to simplify the logic as well. Reducing the number of obstacles and possibly nodes in some instances. If this is worth doing, it may also be worth considering adding a way to clear flags the same way.

blkerby:
This will resolve #1086.

Earlier I was resistant to add these because it requires adding extra nodes on the randomizer side to deal with them. But we're going to have to do that kind of thing anyway in order to handle door locks properly, so I no longer see a reason not to do this. Given that we have `unlocksDoors` on strats (and `clearsObstacles`, etc.), it only seems natural to have `collectsItems` and `setsFlags`. Eventually I would like if all the flags could be set using `setsFlags` in this way, and then we could remove flag nodes and their associated locks/yields (the way that flags are currently getting set), as part of the long-term goal of migrating strats away from being embedded inside "node" properties and into the room "strats" list instead.

### DIFF
--- a/strats.md
+++ b/strats.md
@@ -1140,7 +1140,6 @@ A `setsFlags` array lists the names of game flags that become set (if not alread
   "link": [5, 6],
   "name": "Break the Tube",
   "requires": [
-    {"not": "f_MaridiaTubeBroken"},
     "h_canUsePowerBombs"
   ],
   "setsFlags": ["f_MaridiaTubeBroken"]
--- a/strats.md
+++ b/strats.md
@@ -18,7 +18,8 @@ A `strat` can have the following properties:
   * _gModeRegainMobility_: Indicates that this strat allows regaining mobility when entering with G-mode immobile.
   * _bypassesDoorShell_: Indicates that this strat allows exiting without opening the door.
   * _unlocksDoors_: An array describing possible doors that can be unlocked as part of this strat.
-  
+  * _collectsItems_: An array listing items that are collected as part of this strat (e.g. for G-mode remote item acquire).
+  * _setsFlags_: An array listing game flags that are set as part of this strat.
 These properties are described below in more detail.
 ### Example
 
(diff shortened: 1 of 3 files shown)


osse101 (dismissed):
This seems to fix some problems that we were running into with item collection and flags.

A test should be added to ensure the collectsItems node exists.
And that setFlags is in gameFlags.

This could be used to never visit an item node.  But we would want to model physically going to the location still.  There's probably some kaizo hacks where the only way to get an item is with remote acquire though.

osse101 (commented):
looks good to me

strats.md
```diff
@@ -1114,3 +1115,34 @@ In a heated room, it instead has the form:
 ```
 
 The implicit unlock strats can be disabled by setting the node property `useImplicitDoorUnlocks` to false.
+
+## Collects Items
+
+An `collectsItems` array lists the node IDs (within the room) of items that become collected (if not already collected) as part of executing this strat. This can be useful, for example, for strats that use G-mode to "remote acquire" items at a location different from the item node.
```
kjbranch, line 1121:
```suggestion
A `collectsItems` array lists the node IDs (within the room) of items that become collected (if not already collected) as part of executing this strat. This can be useful, for example, for strats that use G-mode to "remote acquire" items at a location different from the item node.
```

strats.md
```diff
@@ -1114,3 +1115,34 @@ In a heated room, it instead has the form:
...
+## Sets Flags
+
+A `setsFlags` array lists the names of game flags that become set (if not already set) as part of executing this strat.
+
+```json
+{
+  "link": [5, 6],
+  "name": "Break the Tube",
+  "requires": [
+    {"not": "f_MaridiaTubeBroken"},
```
kjbranch, line 1143:
```suggestion
  "requires": [
```

kjbranch (approved):

kjbranch:
I believe that currently going to an item node it is optional to collect the item (Even though Map Rando doesn't use that). Should `collectsItems` force the collection, and should we use it for that purpose?

blkerby:
> I believe that currently going to an item node it is optional to collect the item (Even though Map Rando doesn't use that). Should `collectsItems` force the collection, and should we use it for that purpose?

I think `collectsItems` should force collection (if not already collected), since it may need to be this way for remote acquire strats where collecting the item is also needed for overloading PLMs. And then we can also use `collectsItems` for other situations where collecting an item is unavoidable.

PR #1486, opened 2024-04-24
 Remove Diagonal IBJ from Crateria Supers

kjbranch:
I believe this is too long too low of an angle for `canDiagonalBombJump`. This was expected on Spiffy's recent seed, which caused some confusion. 

I also checked all other uses of `canDiagonalBombJump`, which looked fine. And added `canLongCeilingBombJump` here, as it is 2 screens wide.

### DIFF
--- a/region/crateria/central/Crateria Super Room.json
+++ b/region/crateria/central/Crateria Super Room.json
@@ -200,7 +200,7 @@
     },
     {
       "link": [1, 3],
-      "name": "Direct G-Mode Morph",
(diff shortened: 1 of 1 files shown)


osse101:
There is room to add a new tech for infinite horizontal bomb jumps, or sinking diagonal bomb jumps.  Its sort of a `canLongCeilingBombJump` equivalent.  I wouldn't be able to make it across either way, but I'd have a chance if it were infinite horizontal bomb jumps.

PR #1487, opened 2024-04-27
Add doorOrientation node property

blkerby:
This adds a `doorOrientation` node property to indicate which direction a door leads. This information could already be inferred from the connection files but it was inconvenient to use that way. I included a test to check that the new data is consistent with the connection files.


osse101 (commented):
It's not new to this change, but `doorOrientation` is a property dependant on `nodeType` being a form of door.  Item has it's own special properties.

And that is something that should be organized in a better way, or tested for.

osse101 (approved):

PR #1489, opened 2024-04-27
Mama Turtle Speedy HiJump

kjbranch:


osse101 (commented):
A 1 frame jump or need to reset the room is like a `canInsaneJump`
Reacting to the camera shift so you can catch yourself on the lower ledge is expert mechanics.
But the jump is a lot like moat CWJ to me and moat without a softlock (or second frame precise jump) is very hard.

PR #1493, opened 2024-04-29
G-Mode Bottom of Vertical Doors - Maridia

kjbranch:


osse101 (commented):
GravityJump with bombs with enough momentum to reach the door is closer to insane than extreme when using power bombs.

kjbranch:
the bomb method is quite precise. I dont think itll work with a PB because of not being able to use pause when a PB has been laid.

osse101:
I thought I remembered trying it with PBs, but it does look pretty impossible.
Without it the bomb gravity jump is mostly pointless.  Want to take them out?

kjbranch:
Oh, didnt think of that. I'll remove them

PR #1497, opened 2024-04-30
Schemas to support flash suit logic

blkerby:
This lays down the schema support needed to encode flash suit logic, along with some examples:
- New logical requirements `gainFlashSuit`, `useFlashSuit`, and `noFlashSuit` for describing how to obtain, apply, and carry around (or rather, not be able to carry around) a flash suit, respectively.
- New logical requirement `getBlueSpeed`, similar to `canShineCharge` but without the shinecharge. This is an important distinction since a shinecharge will cause a flash suit to be lost while blue speed won't. In the future, these could also possibly have slightly different shortcharge difficulty requirements from each other, since the crouching animation to gain a shinecharge may require a little more space than just getting blue speed.
- Similarly, a new entrance condition `comeInGettingBlueSpeed` similar to `comeInShinecharging`.
- A new tech `canSpikesuit`.
- A new strat property `flashSuitChecked`, to help us (and the randomizer) keep track of which strats have been checked for logical soundness, as far as their requirements for carrying around a flash suit.
- A new helper `h_canMidAirShootUp`. Just one example application is included. Other applications and more helpers like this will be needed later.
- `noFlashSuit` requirements added to tech/helpers that imply doing actions incompatible with preserving a flash suit. Some of these may end up having exceptions, which we could deal with later.

The idea is that adding flash suit logic will be a long-term project, but thanks to the property `flashSuitChecked` it can start safely being used before it is complete. @kjbranch is planning on starting to go through room-by-room and check strats one-by-one. In the near term, I plan to treat `canSpikesuit` as an ignored tech in Map Rando, until there are enough strats in logic that it would have a significant chance of being used.

Ceiling clips are an example where we'll need to be careful. In most cases they seem incompatible with preserving a flash suit. The only exception I'm aware of is the type of clip used with Mochtroids in Botwoon Hallway, where the enemy must be in a pixel-perfect very high position (compared to other ice clip setups), you jump + aim-down to get on it, and then stand with up or forward; angle-down jump can then clip through. I think we'll want to define a helper specifically for that type of ice clip, as it can be applied in many places but is more difficult than clipping either with X-Ray or a regular `canPreciseCeilingClip` (with the exception of Botwoon Hallway where the setup happens to be trivial).

If I remember, there is some way to do gate glitching while preserving a flash suit, but I assume this will need a new tech, and this could be handled later, so I put `{"noFlashSuit": {}}` on it for now.


osse101:
Gate glitching with flash suit would most easily be done with springball.  Jump to the ceiling and unmorph, then the shot isn't too different on the way down.  Without springball, you do a super hard jump morph.

tech.json
```diff
@@ -1479,7 +1488,7 @@
         },
         {
           "name": "canXMode",
-          "requires": ["XRayScope"],
+          "requires": ["XRayScope", {"noFlashSuit": {}}],
```
osse101, line 1491: One of the ways to activate canSuperJump is with a FlashSuit in XMode.

region/brinstar/green/Green Brinstar Fireflea Room.json
```diff
@@ -146,6 +146,28 @@
...
+            {"doorUnlockedAtNode": 2},
+            {"canShineCharge": {
+              "usedTiles": 17,
+              "openEnd": 0
+            }}
+          ]}
+        ]},
+        "canSpikeSuit",
+        {"shinespark": {"frames": 5, "excessFrames": 5}}
+      ],
```
kjbranch, line 168: needs a thorn hit

schema/m3-requirements.schema.json
```diff
@@ -640,6 +704,30 @@
...
+            "title": "Use Flash Suit",
+            "description": "Fulfilled if Samus has a flash suit, with an effect of losing the flash suit.",
+            "additionalProperties": false,
+            "properties": {}
+          },
+          "noFlashSuit": {
+            "$id": "#/definitions/logicalRequirement/items/properties/noFlashSuit",
+            "type": "object",
+            "title": "No Flash Suit",
+            "description": "Fulfilled if Samus does not have a flash suit.",
```
kjbranch, line 728: does it assume Samus can always get rid of a flash suit for free? (assuming its probably fine)

tech.json
```diff
@@ -1318,6 +1318,15 @@
             "Being able to 4 tap is expected as that brings out the most value from this tech."
           ]
         },
+        {
+          "name": "canSpikeSuit",
+          "requires": [{"gainFlashSuit": {}}],
+          "note": [
+            "The ability to gain a flash suit using spikes or similar damage source.",
+            "Gain a shinecharge, take a spike hit while morphed, unmorphing either 1 or 2 frames after taking damage.",
```
kjbranch, line 1326: spike, thorn, electricity etc

I was reading the wiki and was under the impression you unmorph on frame 0 or 1, and then the spark is activated on frame 10 after the spike, independent of when you unmorph (so 9-10 after the unmorph)

tech.json
```diff
@@ -1630,7 +1639,7 @@
         },
         {
           "name": "canCrystalFlash",
-          "requires": [],
+          "requires": [{"noFlashSuit": {}}],
```
kjbranch, line 1642: is using a cf to gain a suit going to be a problem?
i.e. a single strat for elevator CF having `{"noFlashSuit": {}}` and `{"gainFlashSuit": {}}`

tech.json
blkerby: The Practice Hack [docs](https://smpractice.speedga.me/infohudmode) give more detail about this so I based the description off of that:

> E1 indicates you unmorphed the same frame you took damage, which is one frame early.

I suppose this could still be consistent with the Wiki if damage happens one frame after contact? I'm not really sure.

tech.json
blkerby: In a strat like that, it will be fine as long as the `gainFlashSuit` requirement comes after the `noFlashSuit` requirement, which we can be sure to do

schema/m3-requirements.schema.json
blkerby: The way it's worded here wouldn't necessarily assume that, but I think we will assume that in the randomizer implementation, as it should make things more efficient while being logically equivalent as far as I can tell. It's not that you can necessarily always get rid of a flash suit for free; it's that you could have chosen to have not gotten it in the first place. This should be logically sound as long as we don't add strats that unavoidably produce a flash suit as a side effect of something else (and without a free way to get rid of it); I'm not aware of any strats like that, but if it comes up then we can think about it some more. 

When we get to blue suits, I think superjumps are an example where you can get a blue suit as a side effect of something else (whether you wanted it or not); but a blue suit is trivial to get rid of by just pressing dash so I think it should also be fine in that case.

region/brinstar/green/Green Brinstar Fireflea Room.json
blkerby: Done. 

I think in general we might want some kind of leniency helper(s) for spikesuit attempts. But in this spot it seems probably unnecessary since there's a convenient farm in the room and it's only thorns.

tech.json
blkerby: Ah, right. I think it might still be good to have the "noFlashSuit" here, since any flash suit will be lost after releasing X-Ray (when X-mode is canceled). That kind of super jump strat could be represented by having a "useFlashSuit" requirement before the "canXMode" requirement; since the flash suit usage happens during the X-mode, it doesn't seem any more inaccurate to put it before "canXMode" rather than after? We could also make a helper for that combination to avoid risk of mixing it up?

osse101: A helper would be nice.  I mostly need a note somewhere since I'll forget.

region/crateria/central/Crateria Super Room.json
```diff
@@ -1140,6 +1141,19 @@
       ],
       "devNote": "It's not really insane jump difficult, but that is where you think about needing to avoid the damage.  And failing the jump is still very punishing due to falling into the acid."
     },
+    {
+      "link": [5, 3],
+      "name": "Use Flash Suit",
+      "requires": [
+        {"or": [
+          "canConsecutiveWalljump",
+          "h_canFly"
```
osse101, line 1150: You can't IBJ into using a flash suit

tech.json
blkerby: Done.

region/crateria/central/Crateria Super Room.json
blkerby: Added a `canUnmorphBombBoost` requirement now, since that's the "trick" to make this work: I press up to unmorph while starting to descend toward the last bomb; I keep holding up the whole time after that, which buffers the spark to activate when the last bomb boost is finished. I was just doing single bomb jumping and for me this is just what happened anyway, without specifically going for it; so there's no precise timing to it or anything. But yeah, if you don't get a bomb boost after unmorphing then it won't let you activate the spark.

osse101: I didn't know it worked like that, but it makes sense.

osse101 (approved):
Gonna keep on eye on `h_canMidAirShootUp`.  It seems like there are some trick shots you could do with a flash suit and morph.

PR #1501, opened 2024-05-03
Some clean-up in rooms with useImplicitDoorUnlocks false

blkerby:
- Add tests so we don't forget to add `unlocksDoors` on strats ending at door nodes with `"useImplicitDoorUnlocks": false`.
- Make these tests pass by adding at least rudimentary `unlocksDoors` in the places where it was missing.
- In Amphitheatre:
  - Remove obsolete "h_canNavigateHeatRooms"
  - Add `obstaclesCleared` or `obstaclesNotCleared` to more cleanly separate the acid vs. no-acid strats.
  - Split off `gravitylessHeatFrames`/`gravitylessAcidFrames` in the Gravity jump strat. The total frame count seemed loose (even with Screw Attack, the slowest method of killing the Pirate), but I didn't try to tighten it because I wasn't confident about the intended starting position of the strat.
  - Strats for unlocking the left door with Missiles or Power Bombs could be possible, but they are difficult to test without actually changing the door color; leaving those out for now.
- In Tourian Blue Hoppers:
  - Add a strat to use the left doorway runway by killing the enemies from right to left.
  - Simplify the "Tourian Blue Hopper Crystal Flash" strat to end at node 1 and remove the reset-room requirements. Previously we were concerned about this strat being able to be followed by a dodge strat without first resetting the room, but the dodge strats are now guarded by a `comeInNormally` entrance condition, so it should be fine.

### DIFF
--- a/region/lowernorfair/east/Amphitheatre.json
+++ b/region/lowernorfair/east/Amphitheatre.json
@@ -200,13 +200,20 @@
     {
       "link": [1, 1],
       "name": "Leave With Runway",
-      "requires": [],
+      "requires": [
+        {"obstaclesCleared": ["A"]}
+      ],
       "exitCondition": {
         "leaveWithRunway": {
           "length": 1,
           "openEnd": 1
         }
-      }
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],
+      "devNote": "FIXME: Add acid-filled version of the runway, and requirements to unlock the door with Missiles and Power Bombs."
     },
     {
       "link": [1, 1],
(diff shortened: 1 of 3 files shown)


region/lowernorfair/east/Amphitheatre.json
```diff
@@ -200,13 +200,20 @@
...
+        {"obstaclesCleared": ["A"]}
+      ],
       "exitCondition": {
         "leaveWithRunway": {
           "length": 1,
           "openEnd": 1
         }
-      }
+      },
+      "unlocksDoors": [
```
osse101, line 212: At this point the door at 1 is already unlocked.
Here, we're standing on the door frame.
There could be a scenario where you door stuck to unlock the door, and that wouldn't leave 1.
As it is, the unlock requirements are the same everywhere so it's not causing an issue.

PR #1502, opened 2024-05-05
Crateria remote runway strats

blkerby:
- Setups and applications of remote runways in Crateria
- One application in East Sand Hall
- Update `comeInWithSpringBallBounce` docs to clarify it can also match with `leaveWithRunway`
- New entrance condition `comeInWithBlueSpringBallBounce` (like `comeInWithSpringBallJump` but with blue speed).
- Add `startingDownTiles` in (remote) runway schema.

### DIFF
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -418,6 +418,25 @@ __Example:__
 }},
 ```
 
+
+#### speedBall object
+
+A `speedBall` object represents the need for Samus to be able to gain blue speed and jump into a speedball using a specified runway. It includes a `canSpeedball` tech requirement. Whether a speedball can performed within a given runway length depends on the player's ability to shortcharge as well as to perform a short-hop mockball. The [blue run speed table](strats.md#blue-run-speed-table) describes several tiers of shortcharging skill, their associated runway lengths over which blue speed can be attained, and the resulting run speed. For determining whether a speedball is logically possible, the lowest run speed should be used within the range of what is possible for that skill level; this corresponds to using the "Ideal speed" from the table. The total runway length required can then be computed as follows:
+
+$$\text{minimum speedball runway} = \text{minimal blue speed runway} + \text{short-hop mockball frames} * \text{mid-air speed}$$
+
+Here the "mid-air speed" is obtained by adding the extra run speed (from the "Ideal speed" at the end of getting blue speed) to a base speed of $1.5, which is the average of the spin-jump base speed of $1.6 and the aim-down base speed of $1.4, assuming that roughly half of the jump will be spent in each of those two states. The "short-hop mockball frames" is the amount of frames between the start of the jump and when the morph-sized hitbox is achieved; it is a parameter that can be varied based on assumed player skill: a value of 6 is at or near TAS-level, while 40 would be a lenient value where Samus can jump several tiles into the air.
+
+__Example:__
+```json
+{"speedBall": {
+  "length": 25,
+  "steepUpTiles": 3,
+  "steepDownTiles": 3,
+  "openEnd": 1
+}},
+```
+
 #### itemNotCollectedAtNode object
 An `itemNotCollectedAtNode` object represents the need to have not yet collected the item at a given node in the same room. For example, such
 an item could be used to overload PLMs in G-mode assuming the item has spawned. Note that any conditions for the item to spawn (e.g. for
--- a/region/crateria/central/Landing Site.json
+++ b/region/crateria/central/Landing Site.json
@@ -467,7 +467,6 @@
       "requires": [
         {"or": [
           "canBlueSpaceJump",
-          "canSpringBallBounce",
           {"and": [
             "canSpringBallBounce",
             "canTrickyJump"
--- a/region/crateria/central/Landing Site.json
+++ b/region/crateria/central/Landing Site.json
@@ -315,14 +315,14 @@
       "exitCondition": {
         "leaveSpinning": {
           "remoteRunway": {
-            "length": 28,
+            "length": 31,
             "openEnd": 1,
-            "steepUpTiles": 8
-          }
+            "steepUpTiles": 9
+          },
+          "maxExtraRunSpeed": "$5.A"
         }
       },
-      "note": "Gain speed using the long runway at the top-right of room, and use Space Jump to carry it into the next room.",
-      "devNote": "This runway has some unusable tiles because the Space Jump descent will not work with too high speed."
+      "note": "Gain speed using the long runway at the top-right of room, and use Space Jump to carry it into the next room."
     },
     {
       "link": [1, 1],
@@ -334,18 +334,18 @@
       "exitCondition": {
         "leaveWithMockball": {
           "remoteRunway": {
-            "length": 28,
+            "length": 31,
             "openEnd": 1,
-            "steepUpTiles": 8
+            "steepUpTiles": 9
           },
           "landingRunway": {
             "length": 3,
             "openEnd": 1
-          }
+          },
+          "maxExtraRunSpeed": "$5.A"
         }
       },
-      "note": "Gain speed using the long runway at the top-right of room, and use Space Jump to carry it across the room into a mockball (or speedball) at the doorway.",
-      "devNote": "This runway has some unusable tiles because the Space Jump descent will not work with too high speed."
+      "note": "Gain speed using the long runway at the top-right of room, and use Space Jump to carry it across the room into a mockball (or speedball) at the doorway."
     },
     {
       "link": [1, 1],
@@ -357,22 +357,19 @@
       "exitCondition": {
         "leaveWithSpringBallBounce": {
           "remoteRunway": {
-            "length": 28,
+            "length": 31,
             "openEnd": 1,
-            "steepUpTiles": 8
+            "steepUpTiles": 9
           },
           "landingRunway": {
             "length": 3,
             "openEnd": 1
           },
-          "movementType": "uncontrolled"
+          "movementType": "uncontrolled",
+          "maxExtraRunSpeed": "$5.A"
         }
       },
-      "note": "Gain speed using the long runway at the top-right of room, and use Space Jump to carry it across the room into a spring ball bounce into the other room.",
-      "devNote": [
-        "This runway has some unusable tiles because the Space Jump descent will not work with too high speed.",
-        "FIXME: This could be done without Space Jump, but it needs some way to encode a minimum speed that we will have to exit with. "
-      ]
+      "note": "Gain speed using the long runway at the top-right of room, and use Space Jump to carry it across the room into a spring ball bounce into the other room."
     },
     {
       "link": [1, 1],
@@ -384,10 +381,11 @@
       "exitCondition": {
         "leaveSpaceJumping": {
           "remoteRunway": {
-            "length": 28,
+            "length": 31,
             "openEnd": 1,
-            "steepUpTiles": 8
-          }
+            "steepUpTiles": 9
+          },
+          "maxExtraRunSpeed": "$5.A"
         }
       },
       "note": "Gain speed using the long runway at the top-right of room, and use Space Jump to carry into the other room.",
@@ -741,6 +739,34 @@
         "The jump can be set up by using the full runway with a one-tap shortcharge, with the tap being at the top of the lowest slope."
       ],
       "unlocksDoors": [{"types": ["ammo"], "requires": []}]
+    },
+    {
+      "link": [3, 1],
+      "name": "Leave With Controlled Spring Ball Bounce (Long Runway)",
+      "requires": [
+        {"obstaclesCleared": ["A", "C"]},
+        "SpeedBooster"
+      ],
+      "exitCondition": {
+        "leaveWithSpringBallBounce": {
+          "remoteRunway": {
+            "length": 31,
+            "openEnd": 1,
+            "steepUpTiles": 9
+          },
+          "landingRunway": {
+            "length": 3,
+            "openEnd": 1
+          },
+          "movementType": "controlled",
+          "minExtraRunSpeed": "$4.9"
+        }
+      },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
+      "note": [
+        "Gain speed using the long runway at the top-right of room, do a big jump across the room into a speedball, then use controlled bounces to make it through the top-left door."
+      ],
+      "devNote": "Unlocking doors would be done ahead-of-time when the door is opened, i.e. when obstacle 'C' is set."
     },
     {
       "link": [3, 1],
--- a/region/crateria/central/Landing Site.json
+++ b/region/crateria/central/Landing Site.json
@@ -748,7 +748,8 @@
       "name": "Leave With Controlled Spring Ball Bounce (Long Runway)",
       "requires": [
         {"obstaclesCleared": ["A", "C"]},
-        "SpeedBooster"
+        "SpeedBooster",
+        "canInsaneJump"
       ],
       "exitCondition": {
         "leaveWithSpringBallBounce": {
--- a/region/crateria/central/Landing Site.json
+++ b/region/crateria/central/Landing Site.json
@@ -259,6 +259,139 @@
     },
     {
       "link": [1, 1],
+      "name": "Leave With Mockball (Short Runway)",
+      "requires": [],
+      "exitCondition": {
+        "leaveWithMockball": {
+          "remoteRunway": {
+            "length": 8,
+            "openEnd": 1
+          },
+          "landingRunway": {
+            "length": 3,
+            "openEnd": 1
+          }
+        }
+      }
+    },
+    {
+      "link": [1, 1],
+      "name": "Leave With Spring Ball Bounce (Short Runway)",
+      "requires": [],
+      "exitCondition": {
+        "leaveWithSpringBallBounce": {
+          "remoteRunway": {
+            "length": 8,
+            "openEnd": 1
+          },
+          "landingRunway": {
+            "length": 3,
+            "openEnd": 1
+          },
+          "movementType": "uncontrolled"
+        }
+      }
+    },
+    {
+      "link": [1, 1],
+      "name": "Leave Space Jumping (Short Runway)",
+      "requires": [],
+      "exitCondition": {
+        "leaveSpaceJumping": {
+          "remoteRunway": {
+            "length": 8,
+            "openEnd": 1
+          }
+        }
+      }
+    },
+    {
+      "link": [1, 1],
+      "name": "Leave Spinning (Long Runway)",
+      "requires": [
+        "canPreciseSpaceJump",
+        {"obstaclesCleared": ["A", "C"]}
+      ],
+      "exitCondition": {
+        "leaveSpinning": {
+          "remoteRunway": {
+            "length": 28,
+            "openEnd": 1,
+            "steepUpTiles": 8
+          }
+        }
+      },
+      "note": "Gain speed using the long runway at the top-right of room, and use Space Jump to carry it into the next room.",
+      "devNote": "This runway has some unusable tiles because the Space Jump descent will not work with too high speed."
+    },
+    {
+      "link": [1, 1],
+      "name": "Leave With Mockball (Long Runway)",
+      "requires": [
+        "canPreciseSpaceJump",
+        {"obstaclesCleared": ["A", "C"]}
+      ],
+      "exitCondition": {
+        "leaveWithMockball": {
+          "remoteRunway": {
+            "length": 28,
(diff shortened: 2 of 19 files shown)


blkerby:
I made a little table to help with measuring effective runway lengths: https://docs.google.com/spreadsheets/d/1b2hP3lt-AOWfas38jbwZ1ySehpKwSjDp7GmGBeUTYAg/edit?usp=sharing.

The idea is to use the "Horizontal Speed" Infohud mode in the Practice Hack and then convert the speed value into a runway length. The table only gives integer-valued runway lengths; this could be refined, but it's not clear yet if more precision would actually be needed for the applications. To be conservative, I've been rounding down the setup runway lengths and rounding up the applications.

For the mockball setups with little horizontal distance between the end of the remote runway and the start of the landing, there can be a trade-off between how much remote runway is used and how much of the landing is used. For mockballs, for now I've been using a small enough amount of runway that the full landing can be used (or enough of it to ensure a mockball reaches its max speed before the transition, in cases where this is possible). For uncontrolled spring ball bounces, I've been doing the opposite, using as much of the runway as possible, leaving just 1 tile of landing, since I'm not yet aware of applications where you would need more landing space. If it ends up mattering, we could refine this to give multiple possible combinations of lengths for these exit conditions, similar to what we did for the entrance conditions with `remoteAndLandingMinTiles`.

region/crateria/east/East Ocean.json
```diff
@@ -159,25 +159,63 @@
...
+            "canXRayTurnaround"  
+          ]},
+          {"and": [
+            {"canShineCharge": {
+              "usedTiles": 22,
+              "openEnd": 0,
+              "steepUpTiles": 3,
+              "steepDownTiles": 3
+            }},
+            "HiJump"
```
osse101, line 209: Not sure where HiJump is supposed to be used.  You can gravity jump the water sections, and barely jump morph to reach the above water platforms.

Turning gravity off helped with the underwater jump morphs, but the ones by the door are still hard.

You can go under the structure facing left and not need any xray turnarounds either.

region/crateria/east/East Ocean.json
```diff
@@ -159,25 +159,63 @@
...
+    {
+      "link": [1, 1],
+      "name": "Leave With Mockball (Space Jump)",
+      "requires": [
+        "canPreciseSpaceJump"
+      ],
+      "exitCondition": {
+        "leaveWithMockball": {
+          "remoteRunway": {
+            "length": 18,
```
osse101, line 171: This looks like missing Gravity?

osse101: I can't figure out how to reach the door platform.
Landing on the first platform and springballing to the door is much easier.

region/crateria/east/West Ocean.json
```diff
@@ -592,6 +663,75 @@
...
+      "name": "Leave With Mockball",
+      "requires": [],
+      "exitCondition": {
+        "leaveWithMockball": {
+          "remoteRunway": {
+            "length": 45,
+            "openEnd": 1
+          },
+          "landingRunway": {
+            "length": 7,
```
osse101, line 692: More like 11

region/crateria/east/West Ocean.json
```diff
@@ -592,6 +663,75 @@
...
+            "openEnd": 1,
+            "steepUpTiles": 1
+          },
+          "movementType": "uncontrolled"
```
osse101, line 716: I missed that spacejump was here but did see that you can neutral bounce into an uncontrolled springball bounce.  Hard to control, but was fun.

region/crateria/west/Gauntlet Entrance.json
```diff
@@ -394,6 +394,35 @@
...
+      "entranceCondition": {
+        "comeInGettingBlueSpeed": {
+          "length": 6,
+          "openEnd": 1,
+          "steepUpTiles": 1,
+          "steepDownTiles": 1
+        }
+      },
+      "requires": [
+        "canSlowShortCharge"
```
osse101, line 422: - Springball and speedball tech
- trickyJump to each, but the `canSlowShortCharge` places it better at expert.

region/crateria/west/Gauntlet Entrance.json
```diff
@@ -552,6 +581,35 @@
...
```
osse101, line 609: - Springball and speedball tech

region/maridia/inner-green/East Sand Hall.json
```diff
@@ -507,6 +507,30 @@
...
       ]
     },
+    {
+      "link": [2, 1],
+      "name": "East Sand Hall Cross-Room Blue Spring Ball Bounce (Right to Left)",
+      "notable": true,
+      "entranceCondition": {
+        "comeInWithBlueSpringBallBounce": {
+          "adjacentMinTiles": 18,
+          "adjacentBlueSpeedTiles": 15,
```
osse101, line 517: At the lower speeds, I sometimes get stuck in the middle or don't make it out of the sand at the end.

strats.md
```diff
@@ -1075,6 +1084,55 @@ There are additional requirements depending on the exit condition:
...
+  - A `getBlueSpeed` requirement based on remote runway length.
+- A match with `leaveWithMockball` has the following requirements:
+  - The `movementType` of the entrance condition must be "controlled" or "any".
+  - The `canMockball` tech is required.
+  - `remoteAndLandingMinTiles` must contain at least one pair `(minRemoteLength, minLandingLength)` such that the effective length of the `remoteRunway` is at least `minRemoteLength` and the effective length of the `landingRunway` is at least `minLandingLength`.
+  - The `blue` property of the exit condition must be "yes" or "any".
+  - A `getBlueSpeed` requirement based on remote runway length.
+- A match with `leaveWithRunway` has the following requirements:
+  - The effective runway length must be at least `adjacentMinTiles`.
+  - If the previous room is heated, then `heatFrames` are included based on the time spent running in that room. The minimally required heat frames are calculated the same way as for `comeInRunning` (and `comeInJumping`).
```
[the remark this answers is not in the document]
blkerby: Yeah, I think a `minTiles` on the exiting side (to ensure being able to reach the door) would need to be matched with something like a `maxTiles` on the entrance side (to ensure low enough speed to be able to adequately control the bounces). The thing I've been struggling with is how to capture the relationship between tiles, momentum values, and shortcharge requirements when getting blue speed.

When using a runway purely for momentum, you hold dash for the whole segment of runway, and there's a straightforward conversion between tiles of runway used and the final horizontal speed attained.

When using a runway for a combination of gaining blue speed and gaining momentum, it's more complicated: you're only holding dash part of the time, and even over the same length of runway the final momentum can vary depending on how the shortcharge is executed. To minimize final momentum, you would do a multi-stutter and a 4-tap with as close to perfect taps as possible. To maximize final momentum (assuming a fixed-length runway of, say, less than 27 tiles), I think you still want a multi-stutter and perfect taps but as few taps as necessary, pressing and holding dash for the final stretch of the run, ideally timed so that you get blue just at the end of the runway. So on both the low and high end, the range of momentum values that a player can attain while getting blue on a fixed length of runway depends on shortcharge ability. So this creates a problem that the difficulty level of the shortcharge can no longer be measured by just the length of runway, because it depends on the required momentum too (e.g. on how close the required momentum is to either the low or high extreme of what is possible). We could try to translate the shortcharge difficulty into a standard shortcharge runway length, based on the crispness of taps/stutters required. But I'm fairly fuzzy on exactly what this should look like or if this is the right approach. 

The tech "canSlowShortCharge" was created to solve this on the low-momentum side, but for what we're trying to do now it seems like we need a more granular solution. For these kinds of strats, I'm wondering if it might be better to encode constraints on momentum directly instead of using runway length as a proxy. When we get to setups and applications in heated rooms, the momentum constraints could also have a significant impact on heat frames.

strats.md
```diff
@@ -1075,6 +1084,55 @@ There are additional requirements depending on the exit condition:
 }
 ```
 
+### Come In With Blue Spring Ball Bounce
+
+A `comeInWithBlueSpringBallBounce` entrance condition indicates that Samus must enter the room by spring ball bouncing in from the previous room, while having blue speed. This is applicable to horizontal transitions. It has the following properties:
+
+- _adjacentMinTiles_: This is the minimum effective runway length in case an adjacent runway (connected to the door) is used to gain speed, jump, and bounce. This includes the entire length of runway from where the run begins, up to the door transition.
```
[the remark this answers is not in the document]
blkerby: Yeah, I was thinking about this, and in theory it would be cleaner that way. It's just that coming up with a formula for how to compute those unusable tiles seems non-trivial. I think we could have some kind of "short-hop mockball proficiency" that factors into it. So I guess there's a trade-off between the amount of extra detail needed in the applications/entrances, vs. how much hidden complexity is stuffed into behind-the-scenes math for how the cross-room conditions get resolved. I think we could always make a change later if we figure out that math and decide we want it to be automated like that.

region/crateria/central/Landing Site.json
```diff
@@ -257,6 +257,139 @@
...
+            "openEnd": 1
+          }
+        }
+      },
+      "note": "Gain speed using the long runway at the top-right of room, and use Space Jump to carry it across the room into a mockball (or speedball) at the doorway.",
+      "devNote": "This runway has some unusable tiles because the Space Jump descent will not work with too high speed."
+    },
+    {
+      "link": [1, 1],
+      "name": "Leave With Spring Ball Bounce (Long Runway)",
```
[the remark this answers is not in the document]
blkerby: This is true. The problem is that without SpaceJump you would only be able to exit with high momentum, which might not work in the next room. I think we do need a way to encode these momentum constraints. I added a FIXME for the moment but will think about it more.

region/crateria/east/East Ocean.json
blkerby: I was using HiJump to get onto the first ocean platform. Wasn't thinking of Gravity jumping. Good catch. Added "canGravityJump" option and added a little more detail to the note. Added a devNote for the tight morph option.

blkerby: Oh, took me a minute to realize what you meant, that you can also use the left runway right-to-left without X-Ray. Knowing that, I don't think there would be any reason to use the right runway, unless Gravity jumps were not possible for some reason. Based on that, simplified the strat & note more now.

region/crateria/east/East Ocean.json
blkerby: Good catch, added Gravity requirement.

Landing on the first platform is much easier, yeah. I split off a separate strat for that now, since it requires Spring Ball to make it to the door from there. After some more practice, found how to make it all the way to the door while still spinning, which seems very precise. It will take some more work to figure out momentum constraints for all these strats

region/crateria/east/West Ocean.json
blkerby: Yeah, that works. I went ahead now and added a "canInsaneJump" option for that version, with a note. I'm not sure yet if these high-speed uncontrolled bounces have any applications that wouldn't already be covered by controlled bounces, but I guess we'll find out.

region/maridia/inner-green/East Sand Hall.json
blkerby: Yeah I think we wmay ant a way to express the minimum speed requirement directly, since the tiles requirements don't really make it clear.

region/crateria/west/Gauntlet Entrance.json
```diff
@@ -394,6 +394,38 @@
...
+    {
+      "link": [1, 2],
+      "name": "Come In With Blue Spring Ball Bounce",
+      "entranceCondition": {
+        "comeInWithBlueSpringBallBounce": {
+          "movementType": "controlled"
+        }
+      },
+      "requires": [
+        "canSlowShortCharge"
```
osse101, line 406:
```suggestion
        "canSlowShortCharge",
        "canTrickyJump"
```

region/crateria/west/Gauntlet Entrance.json
```diff
@@ -552,6 +584,38 @@
...
+    {
+      "link": [2, 1],
+      "name": "Come In With Blue Spring Ball Bounce",
+      "entranceCondition": {
+        "comeInWithBlueSpringBallBounce": {
+          "movementType": "controlled"
+        }
+      },
+      "requires": [
+        "canSlowShortCharge"
```
osse101, line 596:
```suggestion
        "canSlowShortCharge",
        "canTrickyJump"```

**Comment by blkerby** [comment]
- Added `minExtraRunSpeed` and `maxExtraRunSpeed` properties so that we can now properly express speed constraints in exit/entrance conditions. At least for now, this is only being applied to entrance conditions that involve blue speed, since without blue speed I think the existing method is still working fine, of using tile counts as a proxy for speed.
- Worked out docs/tables describing essentials of how run speed works and what ranges can be achieved with different combinations of runway lengths and short-charging techniques.
- Added a "speedBall" logical requirement, which can be used to represent in-room speedballs (instead of using "canShineCharge"); this makes it possible to properly account for the interaction between the two skills of short-charging and short-hop mockball. I'm planning on adding "short-hop mockball frames" as a new difficulty parameter in Map Rando.
- Updated the docs for "comeInSpeedballing" to describe how it gets resolved in the same way as "speedBall" on the combined runway. This should be quite a bit more accurate than what we were doing before here.

One thing that came up with the blue run speed table is that there are several types of actions which could be required after gaining blue speed, and I don't think we've yet worked out how to account for them completely accurately:
- running through speed blocks (e.g. like in PCJR)
- jumping before the end of the runway (e.g. to go into a speedball, spring ball bounce, blue space jump, etc.)
- gaining a shinecharge without falling off
- gaining a shinecharge while falling off
So that's stuff that should maybe be sorted out later as part of a larger issue. The blue speed run table that I made is based on an assumption of jumping before the end of the runway, but it assumes a last-frame jump, not directly accounting for lenience that should maybe be there. So it could probably be refined.

File: logicalRequirements.md [hunk 13]
```diff
@@ -418,6 +418,25 @@ __Example:__
 }},
 ```
 
+
+#### speedBall object
+
+A `speedBall` object represents the need for Samus to be able to gain blue speed and jump into a speedball using a specified runway. It includes a `canSpeedball` tech requirement. Whether a speedball can performed within a given runway length depends on the player's ability to shortcharge as well as to perform a short-hop mockball. The [blue run speed table](strats.md#blue-run-speed-table) describes several tiers of shortcharging skill, their associated runway lengths over which blue speed can be attained, and the resulting run speed. For determining whether a speedball is logically possible, the lowest run speed should be used within the range of what is possible for that skill level; this corresponds to using the "Ideal speed" from the table. The total runway length required can then be computed as follows:
+
+$$\text{minimum speedball runway} = \text{minimal blue speed runway} + \text{short-hop mockball frames} * \text{mid-air speed}$$
```
osse101, line 426: Markdown not showing right

osse101 (commented):
- Is the `MaxExtraRunSpeed` the `Extra Run Speed` part of the table?

- Speed values usually are measured in tiles and to use frame counts makes them harder to read and is too much detail in most cases.  It is better for `canTrickyDashJump` and a few notables that use an exact speed amount.  But those strats are also not built for counting an exact subtile position ingame.  I can see stutters and taps making tile counts not helpful so there could just not be a friendlier way to show the required speed.

- The `blue run speed table` looks like a suggestion, but seems pretty good.

- The run speed table looks like it could go in a reference file as well as the wiki

logicalRequirements.md
osse101, line 426: Markdown not showing right

osse101 (commented):
The main hangup is that hex speed values are hard to parse.  As a strat writer and someone reading the logic, you can't look at the strat and then execute it.  It's looking up the table and doing hex math first.

logicalRequirements.md
blkerby: It's looking fine on GitHub in my browser. Mind sharing the problem you're seeing?

blkerby:
> The main hangup is that hex speed values are hard to parse. As a strat writer and someone reading the logic, you can't look at the strat and then execute it. It's looking up the table and doing hex math first.

I don't disagree, but for these strats that involve getting blue speed while also getting a certain amount of momentum, I don't really see a better alternative? A tile count also wouldn't tell you how to execute it and would seem even more unclear. At least with the speed values you have something to compare against, by using the Practice Hack Infohud to see if your speed is high or low. I wish the Infohud had a built-in mode to show the extra run speed; I normally use either the RAM Watch Infohud mode or a debugger to see it. Using hexadecimal here I think makes sense, because that is what you're going to see in the Infohud or a debugger; also it makes it so the subspeed can be represented with 1 digit whereas if we converted to decimal then it would need 4 digits. For testing strats that requiring coming in with blue speed, I usually make a save state in the transition and then use either a debugger or the Customize RAM Watch to write and test different speed values

region/crateria/central/Landing Site.json
```diff
@@ -554,6 +740,34 @@
       ],
       "unlocksDoors": [{"types": ["ammo"], "requires": []}]
     },
+    {
+      "link": [3, 1],
+      "name": "Leave With Controlled Spring Ball Bounce (Long Runway)",
```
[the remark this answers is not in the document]
blkerby: For the name, I was trying to leave it neutral as to whether you get blue or not, because the application in the next room may or may not need it.

I was leaning toward removing the notables, because I think there are going to be a lot of strats that do this kind of thing and it could be hard to keep up with adding so many notables if we want to be consistent.

logicalRequirements.md
osse101: My mistake, github skill issue

region/crateria/central/Landing Site.json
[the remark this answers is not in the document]
osse101: Ah yes, blue vs not blue is a difference.  If its a common situation then it can be described with tech.  Like I think the blue space jump strats in this room are `canPreciseSpaceJump` with slight difference in the difficulty of the jumps if you are rising or descending or if the wall is broken.

This is a 
- very careful speed amount and jump timing from the higher runway
- a bit of a tricky mockball
- controlled springball jumps that depend on your speed
- not necessarily in-room free retries

So its missing `canInsaneJump`.
The note says `speedball` but that's only technically inaccurate.

osse101:
I think the main requirement is to be able to broadly tell if the strat is possible or not.  
For a tricky dash jump, you tell someone to stand on a certain pixel.

To exit with some speed value, you point to a runway, and say you need at least some level of shortcharging.  The tables may fill that role, and remote runways have a runway length listed.  Knowing if a strat is in logic takes a table look up to find how much speed I can leave with at my skill.  Hex values can be directly compared between entrance and exit conditions, and maybe that's trivial but its also more confusing to general readers.

I think the number of strats using a shortcharge and have limits on your speed are few and you could maybe ignore the problem for a while.  Strats with a max/min speed are a little harder to read but not really that different.

And maintainers can work out a system for writing strats since I don't see a better way either.

region/crateria/central/Landing Site.json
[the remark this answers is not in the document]
blkerby: You have to go over to the door ahead of time to open it, which means you must have some way back up to the top right. Since the Ship is there for refilling I think it means you do get in-room retries, but it takes a little time. Added the `canInsaneJump` now and updated the wording in the note to say "mockball or speedball"

blkerby:
> I think the main requirement is to be able to broadly tell if the strat is possible or not. For a tricky dash jump, you tell someone to stand on a certain pixel.
> 
> To exit with some speed value, you point to a runway, and say you need at least some level of shortcharging. The tables may fill that role, and remote runways have a runway length listed. Knowing if a strat is in logic takes a table look up to find how much speed I can leave with at my skill. Hex values can be directly compared between entrance and exit conditions, and maybe that's trivial but its also more confusing to general readers.
> 
> I think the number of strats using a shortcharge and have limits on your speed are few and you could maybe ignore the problem for a while. Strats with a max/min speed are a little harder to read but not really that different.
> 
> And maintainers can work out a system for writing strats since I don't see a better way either.

Yeah the logic for these strats is not going to be easy to interpret, as far as telling if something is in logic or not with a specific room configuration. I just don't see a way around it because it does depend on an interaction between the runway lengths and the shortcharging ability in a way that just is fundamentally not straightforward. I think maybe the answer is that eventually we need to have "logic calculator" tools that could help players answer questions like this. There are other things like adding up heat frames across a room path (to find total energy needed) that are more conceptually simple but in practice also not easy for players to do.

blkerby:
By the way, I also added a couple more columns to the run speed table, to show the initial vertical speeds for jumping (with/without HiJump), since that's something that's fairly important for tricky dash jumps. I never realized before that it's only 2-frame windows to get the extra tricky dash jump height without HiJump.

osse101 (dismissed):

PR #1504, opened 2024-05-08
Flash Suit Strat Compatibility - Green Brinstar Pt. 1

kjbranch:
This is mostly just the easier strats which were easy to test.  More will come later.

### DIFF
--- a/region/brinstar/green/Brinstar Reserve Tank Room.json
+++ b/region/brinstar/green/Brinstar Reserve Tank Room.json
@@ -100,19 +100,22 @@
           "length": 3,
           "openEnd": 1
         }
-      }
+      },
+      "flashSuitChecked": true
     },
     {
       "link": [1, 1],
       "name": "Crystal Flash",
       "requires": [
         "h_canCrystalFlash"
-      ]
+      ],
+      "flashSuitChecked": true
     },
     {
       "link": [1, 2],
       "name": "Base",
-      "requires": []
+      "requires": [],
+      "flashSuitChecked": true
     },
     {
       "link": [1, 3],
@@ -123,7 +126,8 @@
           "morphed": true
         }
       },
-      "requires": []
+      "requires": [],
+      "flashSuitChecked": true
     },
     {
       "link": [1, 3],
@@ -144,6 +148,7 @@
         ]}
       ],
       "clearsObstacles": ["A"],
+      "flashSuitChecked": true,
       "note": [
         "Blindly grab the item before PLMs are overloaded, then return and exit g-mode to remotely acquire the item.",
         "Starting a completely blind IBJ can be somewhat tricky, even if for such a short distance."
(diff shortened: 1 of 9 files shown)


region/brinstar/green/Brinstar Reserve Tank Room.json
```diff
@@ -232,35 +237,41 @@
...
         "Morph"
-      ]
+      ],
+      "flashSuitChecked": true
     },
     {
       "link": [3, 2],
       "name": "G-mode Morph Remote Acquire Return",
       "requires": [
+        "canEnterGMode",
```
kjbranch, line 263: I was thinking of adding this tech to all g-mode strats that don't use a door node, so they no longer show in `Basic`. I think it could help make things clearer, but maybe the tech would need to be renamed as well?

kjbranch:
I think I will do this differently going forward; almost all of the strats are trivial to think about, but they're occupying a lot of space. So I'm planning one PR for all the CF, one for leave/enter shinecharged, etc. so that it can be easier to review.

region/brinstar/green/Brinstar Reserve Tank Room.json
osse101: That makes sense.
The tech name still sounds like it fits, to me.

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -307,6 +311,7 @@
         ]}
       ],
       "clearsObstacles": ["B"],
+      "flashSuitChecked": true,
```
osse101, line 314: It's going to be so much harder to not lose it.

osse101 (commented):
It's the `canShinespark` that I don't know if are covered because they don't have a `noFlashSuit`.  `comeInShinecharging` says it doesn't preserve flashSuit, and `canShinecharge` is an implicit use of `noFlashSuit`, but those notes are written in distant places.

I think the `canShinecharge` with tiles over 33 were being omitted more recently.  But checking the strats is how we're going to find them. 

Looks like gaining and using Flash Suits will be done later too.

region/brinstar/green/Etecoon Energy Tank Room.json
kjbranch: Currently I'm counting all G-Mode strats as not allowing a flash suit. You will lose the suit when entering G-Mode, so it should be fine as long as we dont add ways to gain flash suits while already in G-Mode. 

I think any combinations of flash/blue/gmode are probably a very low priority for now (except generating suits with R-Mode).

blkerby:
> It's the `canShinespark` that I don't know if are covered because they don't have a `noFlashSuit`.

When gaining a spikesuit, there's a shinespark that happens and you have the flash suit afterward, so I was thinking that a shinespark itself might not take away the flash suit. That was my thinking for not including `noFlashSuit` on it. Though maybe there's something special going on here with spikesuit, and it could just be modeled by having the `gainFlashSuit` come after that shinespark.

Either way, I think `shinespark` requirements should never appear on their own; they should always be preceded by either a `canShinecharge`, a `useFlashSuit`, or some entrance condition that implies carrying in a shinecharge or shinespark. That's maybe something that would be worth adding as a test. And as long as that is true, I guess it would be redundant to include a `noFlashSuit` into `shinespark` because it should already be covered by the shinecharge or other preceding action? 

I imagine there could be scenarios where you use the initial shinespark from a spikesuit to spark out a room while having the flash suit afterward. So by not including `noFlashSuit` in `shinespark`, those strats can be represented naturally with the current schema. Otherwise we would have to do something special to be able to represent them.

There are cases where you want to get a shinecharge without needing to spark, e.g. for the purposes of getting temporary blue while falling off a ledge, and this will cause a flash suit to be lost.

region/brinstar/green/Etecoon Energy Tank Room.json
kjbranch: Someday we can change this if we model gaining a flash suit in G-Mode, but currently `canEnterGMode` requires `canEnterRMode` which requires `{"noFlashSuit": {}}`.

PR #1507, opened 2024-05-09
Split tech requirements into two fields: tech and other

kjbranch:
This is the first step towards adding items back to tech (i.e. morph to canIBJ); then `h_canArtificialMorphIBJ` can be defined as `{"tech": "canIBJ"}`.

### DIFF
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -535,3 +535,14 @@ __Example:__
 ```json
 {"noFlashSuit": {}}
 ```
+
+### Other requirements
+
+#### tech object
+
+A `tech` object indicates the need to be able to perform a `tech`, but without its `otherRequires`. This can be used to override item, ammo, or other requirements that are normally associated with the tech. The removal of `otherRequires` requirements applies recursively to any tech in the `techRequires` of the referenced tech.
+
+__Example:__
+```json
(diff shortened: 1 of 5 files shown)


osse101 (approved):
I think with this, we add more items to the techs in tech.json, then use helpers to build copycat techs with fewer item requirements.  And only a couple uses of the new `{tech:name}` object makes it to strats. Using the wrong version `e.g. canIBJ` would probably be the version with the most requirements.

`canGrappleJump` has a funny case of `or:[item, tech]`
`canUnderwaterBombIntoSpringBallJump` has a different kind of mixed tech in `otherRequirements`
I'd expect any techs inside otherRequirements get thrown out when using {tech}.

Seems like you can override `noFlashSuit` by using the tech object.

All looks syntactically fine and sorted properly.

kjbranch:
`canGrappleJump` is an odd one that I don't like, but I think it should work ok. It could be moved into a helper, which won't improve anything really, but it will help us not have "or"s, if that is something we want in the future.

The tech with helpers that include both tech and items were split in an awkward way. I plan to fix those and hopefully remove them in the next step.

I'll look through this PR one more time before I merge it - it's a bit of a scary one if there is an error somewhere.

PR #1509, opened 2024-05-11
Mother Brain R-mode strats

blkerby:


kjbranch (dismissed):
Weird stuff. I hadn't ever seen that before.

osse101 (commented):
Using `Energy` would be more technically accurate, but the main point is to show the individual tank count.  
It matters more for the light pillar where, from how it looked, you need a combined 301 instead of 399.  MB2 is between the refill and health check, so you need a damageless fight when you could actually take a little damage.

blkerby:
> Using `Energy` would be more technically accurate, but the main point is to show the individual tank count. It matters more for the light pillar where, from how it looked, you need a combined 301 instead of 399. MB2 is between the refill and health check, so you need a damageless fight when you could actually take a little damage.

Right, for the light pillar start, "Energy" requirement of 301 is better. Changed that now. It can also work to do the Crystal Flash right before rainbow beam, though I don't think there's any benefit, since either way you have the same amount of energy (98) that can be lost during the fight, and damaging down for the Crystal Flash is easier to do before MB2. 

For the non-light-pillar version, there is more to gain from doing the CF right before rainbow beam, since it allows the fight to use between 199 and 299 energy (minus whatever was lost before MB1), compared to having to be damageless. CF during ketchup phase also seems possible. But damaging down to CF range seems potentially tricky unless you have Gravity. Not sure if it's worth modeling or not. I figure in Map Rando these will probably be Extreme notables and damageless Mother Brain would be assumed at that level anyway?

osse101:
Setting up the pre rainbow CF is a tricky one; you kind of set it up in advance then dodge everything through ketchup phase, without suits there may not be a way to damage down correctly, you need to track when the phase ends.  Its an emergency option in randos.
These fit at extreme notable with reduced item pool; challenge type seeds.  If we compare to low ice, you should basically be able to do the fight damageless.  They do like to take <4 damage from the center of bombs, since its 0 with gravity.  And weaker beam fights would be straight to insane.

PR #1511, opened 2024-05-12
Add Items to Tech Pt.1 - Artificial Morph

kjbranch:
Added item requirements to tech that had artificial morph variants. This should improve clarity and reduce the likelihood of future errors, and became possible after #1507.
- Converted all remaining instances of `canIBJ`, `canJumpIntoIBJ`, etc. to `h_canIBJ` or `{"tech": "canIBJ"}`, etc.
- Added items, such as `Morph` to tech and used `{"tech": "canIBJ"}`, etc. in the artificial morph helpers, and simplified the regular helpers.
- Removed all helpers from the tech file.

This is only part of adding all items to their tech, but this is likely the majority of it. In a future PR, all helpers, such as `h_canIBJ` can be removed and `canIBJ` used instead.

The only error found was escaping Beta PBs with a springball bomb jump which didn't include a bomb, as well as 7 instances of `canResetFallSpeed` that didn't require `Morph`. These 7 still need to be verified and are all included in the final commit `Add morph to canResetFallSpeed`.

### DIFF
--- a/tech.json
+++ b/tech.json
@@ -728,7 +728,7 @@
                     "canTrickySpringBallJump",
                     "canJumpIntoIBJ"
                   ],
-                  "otherRequires": ["h_canJumpIntoIBJ"],
+                  "otherRequires": [],
                   "note": [
                     "Use the tiny amount of height gain from an underwater Bomb explosion to re-equip Springball and perform a Springball jump.",
                     "This is most often used to get a second SpringBall jump when underwater without HiJump.",
@@ -753,7 +753,10 @@
           {
             "name": "canSpringFling",
             "techRequires": ["canDisableEquipment"],
-            "otherRequires": ["SpringBall"],
+            "otherRequires": [
+              "Morph",
+              "SpringBall"
+            ],
             "note": [
               "The ability to gain extra distance meerly by turning Springball on or off.",
               "When equipping or unequipping Springball while Morphed, Samus' speed is reset.",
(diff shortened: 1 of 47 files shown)


tech.json
```diff
@@ -1808,13 +1852,21 @@
...
+            {"ammo": {"type": "PowerBomb", "count": 10}},
+            {"noFlashSuit": {}}
+          ],
           "note": [
             "Performing a Crystal Flash (CF).",
             "Initiating the Crystal Flash requires energy at 50 or lower, empty Reserve energy, 1 Power Bomb to initiate, and holding L+R+Shoot+Down with no other inputs.",
             "A Crystal Flash will refill 1500 energy at the cost of 10 Missiles, 10 Supers, and 10 Power Bombs.",
             "The Crystal Flash will also put Samus in a standing position even if she is in a Morph tunnel."
           ],
+          "devNote": "Note that this does not include the refill, as the amount is dependent on if Samus' environment.",
```
osse101, line 1869:
```suggestion
          "devNote": "Note that this does not include the refill, as the amount is dependent on Samus' environment.",
```

tech.json
```diff
@@ -1837,7 +1889,12 @@
                 "canCrystalFlash",
                 "canCeilingClip"
               ],
-              "otherRequires": [],
+              "otherRequires": [
+                {"or": [
+                  "SpringBall",
```
osse101, line 1894: Hopefully if you are doing `canJumpIntoCrystalFlashClip`, you understand the nuance of doing it with Springball.

osse101 (approved):
Looks at lot nicer already.

I didn't notice anything missing, but which direction is the use of helpers going for techs with all the items in the base tech?

tech.json
kjbranch: You'll still have Morph in the `canCrystalFlash` tech, and presumably can midair morph.

The CF tech and helpers could use a bit more cleanup after this PR as well.

kjbranch:
If you're referring to helpers like `h_canIBJ`, I think they should probably all be deleted. I could see some minimal use in keeping them, but I mostly just didnt delete them yet, because I thought it would make this PR too difficult to review.

PR #1517, opened 2024-05-13
Add Items to Tech Pt.2 and Cleanup

kjbranch:
This is very similar to previous PRs on the topic, but also with a bit of general helper cleanup.

Note that I removed the extension/requirement of `canXRayCeilingClip` in `canPreciseCeilingClip`. Although they are progressive in difficulty and accomplish the same thing, they didn't really feel like extensions to me. Let me know if you think it should be added back.

### DIFF
--- a/region/brinstar/red/Beta Power Bomb Room.json
+++ b/region/brinstar/red/Beta Power Bomb Room.json
@@ -401,11 +401,9 @@
         "canTrickyJump",
         {"or": [
           "canXRayCeilingClip",
-          {"and": [
-            "canTunnelCrawl",
-            "canPartialFloorClip"
-          ]}
+          "canTunnelCrawl"
         ]},
+        "canPartialFloorClip",
         "canCeilingClip"
       ],
       "note": [
@@ -413,8 +411,7 @@
         "The thorns do not cover the entire blocks above. Jump barely only onto the corner to avoid the damage.",
         "Carefully jump around the thorns, tunnel crawl slightly to avoid touching the thorns when partial floor clipping.",
         "Morph and x-ray can be used instead to setup the clip much easier."
-      ],
-      "devNote": "FIXME: Technically this canXRayCeilingClip does not require canUseEnemies."
+      ]
     },
     {
       "link": [3, 1],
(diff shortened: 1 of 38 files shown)


tech.json
```diff
@@ -1808,33 +1809,32 @@
               "name": "canXRayCeilingClip",
               "techRequires": [
                 "canCeilingClip",
-                "canUseEnemies"
+                "canUseEnemies",
```
osse101, line 1812: There shouldn't be a `canUseEnemies` requirement here since there are a number of places where you can unmorph in a two tile gap ... was it brought over with the trickyIceClip refactor?

osse101: Or maybe this tech is more enemy oriented than the name suggests

region/brinstar/red/Beta Power Bomb Room.json
```diff
@@ -400,7 +400,7 @@
       "requires": [
         "canTrickyJump",
         {"or": [
-          "h_canXRayCeilingClip",
+          "canXRayCeilingClip",
```
osse101, line 403: The usage here is
```
"canXRayStandUp",
"canPartialFloorClip",
"canCeilingClip"
```

tech.json
kjbranch: I was talking to Maddo about this, and we could move `canUseEnemies` into the `otherRequires` and still be able to reference it without. Alternatively I was thinking we could keep `canUseEnemies` off of it and add it on with a helper, such as `h_canXRayMorphIceClip` or something?

osse101: `canXRayCeilingClip` has 2 uses: on top of enemies and between solid tiles like in Beta PBs.  Then another one is at the top of an xray climb as seen in Bowling to avoid taking spike damage, but that is less the focus of the strat.

Helpers for both sounds good to me.  And this is used for snail clips as well as ice clips.

kjbranch:
Removed `canUseEnemies` from `canXRayCeilingClip` and `canPreciseCeilingClip`.
Created two new helpers: `h_canXRayMorphIceClip`, `h_canPreciseIceClip`
Renamed `h_canNonTrivialCeilingClip` to `h_canIceClip`

PR #1523, opened 2024-05-17
Add flashSuitChecked to all Strats with comeInShinecharged or leaveShinecharged

kjbranch:
None of these should be possible with a flash suit.


osse101:
I thought it was covering all shinesparks and missed a bunch 🐱

PR #1526, opened 2024-05-20
Another Pass at Ice Clips, Add canHighPixelCeilingClip for Preserving Flash Suits

kjbranch:


region/brinstar/pink/Big Pink.json
```diff
@@ -1930,7 +1932,8 @@
         "Once it is up there, the bee will become stuck and the room will need to be reset.",
         "But if the Reo does not fit in the hole, it is possible to just wait near the ledge for another slow hover to try again.",
         "With this setup, there is a one-frame window in which the Reo can be frozen. Without a precise setup, there is typically no way to clip through without Morph and X-Ray."
-      ]
+      ],
+      "devNote": "This may be possible with a high pixel ice clip."
```
osse101, line 1936: I don't think there is a way to position for this.  You would need to bring a Zeb over?
Get inside the bee's hitbox and you will fall, and there is no room to approach it from the side.

region/brinstar/pink/Big Pink.json
```diff
@@ -1955,6 +1959,28 @@
...
+        "canOffScreenMovement",
+        "Wave",
+        "canInsaneJump",
+        "h_canHighPixelIceClip"
+      ],
+      "flashSuitChecked": true,
+      "note": [
+        "Raise a Zeb to be just below the crumble blocks and blindly freeze it to set up an ice clip to reach the Power Bomb room exit door.",
+        "Get a bug from the pipe to spawn facing left and freeze it while moving to the ledge below the Hopper room door.",
+        "Go above, then blindly freeze it with a diagonal downward shot at a pixel perfect height.",
```
osse101, line 1978: A standing shot followed by a jump? from where the normal setup position ( below wave gate door) is should work too.  Just a smaller frame window.  Refreezing by the crumbles would work the same.  but I haven't tried it.

region/brinstar/pink/Big Pink.json
```diff
@@ -1955,6 +1959,28 @@
...
+        "Go back down and shoot it again to freeze it with a diagonal upward shot while on the ground to avoid losing the flash suit.",
+        "If the Zeb is frozen again, but not in a usable spot, it is possible to wall jump in the shaft below the crumble and refreeze it with a diagonal downward shot.",
```
osse101, line 1980: This part was confusing.  Is it at the correct height but inside of a wall or the middle of the shaft?

region/brinstar/pink/Big Pink.json
kjbranch: This felt harder to me, but I didnt try for too long. If you shoot before you jump, itll be frozen too low. So you might either need a second blind freeze or a high jump with diagonal shot (probably wont work).

osse101: I haven't gotten it yet, but the height looks possible from a crouch shot into (step forward) spin jump.  Then you could do it with spazer, and keep the flash suit with an unmorph upshot on the refreeze.

PR #1527, opened 2024-05-23
Add more Blue Boulder cross-room jumps

blkerby:
This adds some cross-room jump strats that were missing (as pointed out by zornor in the Discord). There are also cross-room spring ball bounce strats that should be considered; I figure we can come back to that after (or as part of) the pending Crateria remote runway PR. (**Edit**: On second thought, I guess spring ball bounce wouldn't necessarily be useful since if you have Morph then you can just do the airball instead, as that requires just the 1 tile of runway which is always available, assuming there's an air environment at all)

This room is currently a bit awkward because of not having a node to represent being in the water below the right door. I'm not sure if everything is logically sound if you were down there and not able to unlock the door (e.g. which could happen in Map Rando with Random start). So it could all still use another look later.

### DIFF
--- a/region/brinstar/blue/Blue Brinstar Boulder Room.json
+++ b/region/brinstar/blue/Blue Brinstar Boulder Room.json
@@ -249,7 +249,7 @@
       },
       "requires": [
         "canCarefulJump",
-        "canWalljump"
+        "canPreciseWalljump"
       ],
       "note": "A doorsill with an open end is really all the room that's needed on the other side."
     },
@@ -259,7 +259,7 @@
       "entranceCondition": {
         "comeInJumping": {
           "speedBooster": "any",
-          "minTiles": 1
+          "minTiles": 2
         }
       },
       "requires": [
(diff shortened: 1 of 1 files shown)


region/brinstar/blue/Blue Brinstar Boulder Room.json
```diff
@@ -242,7 +255,35 @@
     },
     {
       "link": [2, 1],
-      "name": "Boulder Room Doorsill Jump",
+      "name": "Jump Through the Door Air Ball",
+      "entranceCondition": {
+        "comeInJumping": {
+          "speedBooster": "any",
+          "minTiles": 1
```
osse101, line 262: At 1 tile this is a last frame jump and that is a tricky jump.

osse101 (commented):
I would also try to change walljumping off of an invisible bridge to `canPreciseWalljump`.

region/brinstar/blue/Blue Brinstar Boulder Room.json
blkerby: Good catch. Split it now into two strats, a 2-tile runway version with canCarefulJump and a 1-tile version with canTrickyJump.

PR #1533, opened 2024-06-04
Add and use canPowerBombMidIBJ

kjbranch:
For the Screw Attack Room, I just used the IBJ heat frames that were used in [5, 3]


osse101:
I'll comment that switching to PBs mid IBJ is easier than using reserves mid IBJ, manual or auto, and there is a specific problem to fix there.

PR #1538, opened 2024-06-08
Add flashSuitChecked to strats with tech that prevent it

kjbranch:
To strats with the following tech: (excluding ones in nontrivial `OR`s, those will need more work to check)
- canRJump
- canGateGlitch
- Temporary Blue (added a FIXME describing that it is possible and some need to be rechecked)
- canXRayWaitForIFrames
- canXRayTurnaround
- canXRayStandUp
- canXMode
- canPartialFloorClip
- canSuperJump

Still need to go through canCrouchJump strats, but likely many of them need to be checked if the crouch can be avoided.


region/lowernorfair/east/Wasteland.json
```diff
@@ -1512,6 +1516,7 @@
         {"heatFrames": 240}
       ],
       "clearsObstacles": ["B"],
+      "flashSuitChecked": true,
```
osse101, line 1519: This strat can break the block with blue without losing flash suit, and its not really much harder.

region/lowernorfair/west/Acid Statue Room.json
```diff
@@ -289,6 +289,7 @@
         ]},
         {"heatFrames": 285}
       ],
+      "flashSuitChecked": true,
```
osse101, line 292: The note suggests `canPartialFloorClip` is here for the tech component, and you actually do a spinjump

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -902,6 +906,7 @@
       ],
       "clearsObstacles": ["A", "B", "C"],
       "reusableRoomwideNotable": "Screw Attack Room Descent and Shinespark Escape",
+      "flashSuitChecked": true,
```
osse101, line 909: The ability to use the flash suit to escape might be easier.  You would jump with blue and unmorph to preserve it.  (Entering through middle door with very low speed probably works too).  But that would be written as new strats.

osse101 (commented):
I think this would be the time to put in the flash suit alternatives for `canGateGlitch`.  And that is jumping with springball and unmorphing, or a `canInsaneJump` spinjump+morph+unmorph which may itself be a tech.  The spinjump would be less important to add.

Blue brin etank, mickey mouse, and screw attack have variations of the strat that would let you keep the flash suit.  They could be added later.

Wasteland was using a chainTemporaryblue for simplicity but the more intuitive way of doing the strat is to jump with blue and morph+unmorph above the bomb block, and not lose flash suit.

PR #1539, opened 2024-06-08
Add flashSuitChecked to transitioning with stored fall speed

kjbranch:
Didn't add it to strats with screw to break a block while still having stored fall speed, because I wasn't able to test it.

### DIFF
--- a/region/norfair/east/Frog Speedway.json
+++ b/region/norfair/east/Frog Speedway.json
@@ -729,6 +729,7 @@
         {"types": ["missiles", "super"], "requires": []},
         {"types": ["powerbomb"], "requires": ["never"]}
       ],
+      "flashSuitChecked": true,
       "note": [
         "This strat is only usable right to left.",
         "This room has many offscreen shot blocks. Shooting enough of them with wave + spazer or wave + plasma allows you to pass through the speed blocks.",
@@ -785,6 +786,7 @@
         {"types": ["missiles", "super"], "requires": []},
         {"types": ["powerbomb"], "requires": ["never"]}
       ],
+      "flashSuitChecked": true,
       "note": [
         "This strat is only usable right to left.",
         "This room has many offscreen shot blocks. Shooting enough of them with wave + spazer or wave + plasma allows you to pass through the speed blocks.",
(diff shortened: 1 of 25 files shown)


osse101 (dismissed):
Moonfalling with Screw Attack let me keep the flash suit.  There's nothing about it that would trigger the flash suit other than accidently buffering a jump, but holding angle and jump prevent that from happening.

The did the dachora moondance with a flash suit then broke the blocks to reach the right side with Screw.  The only thing I had to do differently was bombjumping to reach the top of the bomb wall since you cannot jump morph.  Croc speedway and Bowling moondances won't work because of the Crystal Flash, but a Beetom moondance should be fine as long as you can carry the beetom to the door you need it for without losing the flash suit.

When using the stored moonfall with 1 tile of stored speed you need to be careful about buffering the use of the flash suit, by releasing jump asap.  I tested this with Gauntlet Etank right->left stored moonfall+flash suit to cross the room.

region/norfair/east/Frog Speedway.json
```diff
@@ -725,6 +729,7 @@
         {"types": ["missiles", "super"], "requires": []},
         {"types": ["powerbomb"], "requires": ["never"]}
       ],
+      "flashSuitChecked": true,
```
kjbranch, line 732: Added flashSuitChecked to the strats that transition and use screw. I figure the setups and uses can be added later. I could attempt to work on it next, or start to do the easier strats in rooms. 

One thing I noticed is in this strat (speedless speedway) it says you can transition with stored fall speed and deal with the Beetoms with bombs/PBs - I'm pretty sure that youll lose the stored fall speed?

osse101: Correct, you cannot use bombs or Pbs.  And Screw depends on your amount of stored fall speed.

kjbranch: I removed the bomb/PB options. I wasn't sure how to update the screw options if they needed it

region/norfair/east/Frog Speedway.json
```diff
@@ -700,14 +704,6 @@
           {"ammo": {"type": "Missile", "count": 1}},
           {"ammo": {"type": "Super", "count": 1}},
           "ScrewAttack",
```
osse101, line 706:
```suggestion
```

region/norfair/east/Frog Speedway.json
```diff
@@ -756,14 +753,6 @@
...
```
osse101, line 755:
```suggestion
```

osse101: Tried it, and Screw doesn't kill them at all when you do the moonfall.

PR #1546, opened 2024-06-12
Convert some comeInShinecharging to comeInGettingBlueSpeed

kjbranch:
I looked through all of the strats, but didn't touch ones with temporary blue. I'm thinking most of those need to be split based on if they gain the temp blue from a speedball or shinecharge.

### DIFF
--- a/region/crateria/central/Parlor and Alcatraz.json
+++ b/region/crateria/central/Parlor and Alcatraz.json
@@ -343,7 +343,7 @@
       "link": [1, 1],
       "name": "Leave with Runway, Come in with Speed to Break the Wall",
       "entranceCondition": {
-        "comeInShinecharging": {
+        "comeInGettingBlueSpeed": {
           "length": 13,
           "openEnd": 0,
           "steepUpTiles": 2,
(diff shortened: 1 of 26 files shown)


region/crateria/central/Parlor and Alcatraz.json
```diff
@@ -391,7 +391,7 @@
       "name": "G-Mode Setup - Speed Through the Blocks then Get Hit By Geemer",
```
osse101, line 391: Seems weird to have these strat components combined, to me.
strat 1 - comeInGettingBlueSpeed to reach the hub.
strat 2- G-Mode setup using speed from the hub.

region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -881,7 +881,7 @@
       "link": [4, 1],
       "name": "Suitless Water Shinecharge",
```
osse101, line 882: Tempting to call this Suitless Run Through.  The tech name being in the title is more important but its a tad confusing for those of us looking at the entrance condition

PR #1548, opened 2024-06-13
Add G-Mode Return Strats - Horizontal SpawnAt, Doorways

kjbranch:
This will fix some strats, but more spawn ats should be added.

### DIFF
--- a/region/lowernorfair/east/Amphitheatre.json
+++ b/region/lowernorfair/east/Amphitheatre.json
@@ -231,6 +231,51 @@
       "devNote": "FIXME: Add acid-filled version of the Crystal Flash, and requirements to unlock the door with Missiles and Power Bombs."
     },
     {
+      "link": [1, 1],
+      "name": "Carry G-Mode Back Through",
+      "entranceCondition": {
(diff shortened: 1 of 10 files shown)


kjbranch:
I also added strats to return through doors that have no cap. These will mostly be useful in rooms that can also setup g-mode, but I thought the others may have a use I'm not thinking of.

region/lowernorfair/east/Amphitheatre.json
```diff
@@ -230,6 +230,51 @@
...
+        ]}
+      ],
+      "exitCondition": {
+        "leaveWithGMode": {
+          "morphed": false
+        }
+      },
+      "bypassesDoorShell": true,
+      "flashSuitChecked": true,
+      "devNote": "This should be possible with a precise and very long canDoubleBombJump, or strats with lots of acid damage."
```
osse101, line 255: You can barely make it up in time without a jumpIntoIBJ if you do the DoubleIBJ very well.  With a jump start just need to not drop the IBJ.

osse101 (dismissed):

PR #1552, opened 2024-06-15
Add Spawn At Node in Screw Attack Room

kjbranch:
This removes node 5, which was not useful since there was an obstacle. It then adds a spawn at node and node on the bomb blocks.

It looked like the heat frames throughout were a bit inconsistently determined (prepared through the door vs fall first). I did my best to separate them, but I'm sure the heat frames will need to be adjusted throughout.


region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -955,55 +914,77 @@
...
         ]},
         "h_canUseMorphBombs",
-        {"heatFrames": 300}
+        {"heatFrames": 220}
       ],
       "clearsObstacles": ["B"]
     },
     {
-      "link": [4, 2],
+      "link": [4, 5],
```
osse101, line 981: This would be able to land at 2

osse101 (commented):
There is a [3, 2] canResetFallSpeed strat

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -727,97 +705,34 @@
...
+        {"doorUnlockedAtNode": 2},
+        "canResetFallSpeed",
+        "canTrickyJump",
+        {"heatFrames": 145}
       ],
-      "note": [
-        "Moonfall to clip through the Bomb blocks",
-        "Break spin with Shot or Angle Aim to avoid clipping all the way to the bottom of the room."
+      "unlocksDoors": [
+        {"types": ["super"], "requires": ["canPrepareForNextRoom"]},
```
osse101, line 734: ?  Seems unrelated to anything

kjbranch: Ah, I put this on as you'd need to know that the door is green beforehand. If you've never been in the room or have the map and it was green, I don't think you could have enough time to react and would instead fall to 5

osse101: That makes sense 🧠

PR #1555, opened 2024-06-17
G-Mode Regain Mobility Strats in Norfair and LN

kjbranch:


osse101 (approved):
I'm wondering about Ridley's room as a `spawnAt`.  If the comeInWithGmodeSetup keeps you at 2, and regainMobility keeps you at 2, then can you bypassDoorShell at 2 back through the door?
But that may be part of the general gmodeImmobile handling.

kjbranch:
I was wondering about this as well. I'm planning on going through the new Spawn Ats and adding G-mode return strats to them next.

PR #1556, opened 2024-06-19
Morph Ball Room add left spawn at

kjbranch:


osse101 (commented):
The hoppers are Labelled `A` on the diagram.

My preferred way of enterring the room and killing hoppers is to missile+Super the first two and doesn't get hit.

There is a way to enter in a mockball and cross the room without getting hit.  Pretty hard though.  Placing a PB in the middle of the hoppers is the safest way but then it doesn't reach the far bomb block.  So there are some strat tweaks I'd want to look more closely into.

region/brinstar/blue/Morph Ball Room.json
```diff
@@ -250,6 +246,239 @@
...
+    },
+    {
+      "link": [1, 1],
+      "name": "Prepared Missile and Super Kill",
+      "entranceCondition": {
+        "comeInNormally": {}
+      },
+      "requires": [
+        "canPrepareForNextRoom",
+        "canFarmWhileShooting",
```
kjbranch, line 334: This is not a great tech here, but I was looking for something Expert-level that had some sort of enemy shooting/dodging.

PR #1557, opened 2024-06-20
Mickey Mouse Room add left spawn at

kjbranch:


region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -1363,58 +1358,36 @@
...
       "requires": [
-        "h_canNavigateHeatRooms",
         "ScrewAttack",
         {"heatFrames": 320}
       ],
-      "clearsObstacles": ["C", "E"]
-    },
-    {
-      "link": [8, 1],
-      "name": "Base",
```
kjbranch, line 1380: Removed this strat and added this to all the [6, 1] and [7, 1] strats:
```
"unlocksDoors": [
  {"types": ["missiles"], "requires": [{"heatFrames": 30}]}
]
```

region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -1423,137 +1396,117 @@
...
+        {"or": [
+          {"and": [
+            {"enemyKill": {
+              "enemies": [["Multiviola"]],
+              "explicitWeapons": ["PseudoScrew", "ScrewAttack"]
+            }},
+            {"heatFrames": 30}
+          ]},
+          {"and": [
+            "canTrickyJump",
```
kjbranch, line 1481: Unfortunately this PR is very hard to read because of moving strats. I didn't change much besides updating a few dev/notes. I did add this tricky jump requirement for missiles/supers

osse101:
Seems ok
- `Get Hit on Entry` could use a tiny amount of heat frames.  It would be impactful when modelling a room peek and then leave, and that's not really a thing.  Otherwise there exist larger pools of heatFrames to traverse the room. 
- `Leave With Runway (Short Runway)` doesn't need an obstacle check now that it has an entrance condition.

And from the unchanged parts of the strats
- A number of medium strength weapon kill strats should should expect running away and that is `canDodgeWhileShooting`
- `Plasma Hitbox Run-Through` I recall being kinda hard to do; but I don't really remember for which direction it was.

kjbranch:
I think I addressed everything except the hitboxing. I found that I got hit a lot both directions, but I'm not really sure what I'm doing. I did notice that [1, 6] hitbox required plasma and [6, 1] didn't (?)

PR #1558, opened 2024-06-20
Blue Hopper Room add left spawn at

kjbranch:
This should be the end of the spawn ats. 
There were two that were skipped:
- Electric Death Room
  - Can be escaped with just a down grab
- Pink Hopper Room
  - Of the 4 rooms with hits on entry, this is the easiest to dodge and the least damage


region/tourian/main/Blue Hopper Room.json
```diff
@@ -603,64 +627,89 @@
...
       ],
-      "exitCondition": {
-        "leaveWithGModeSetup": {}
-      },
       "unlocksDoors": [
         {"types": ["super"], "requires": []},
         {"types": ["missiles", "powerbomb"], "requires": ["never"]}
       ],
-      "flashSuitChecked": true
+      "note": "Enter the room morphed to avoid a hopper hit on entry."
```
osse101, line 695:
```suggestion
      "note": [
        "Enter the room morphed to avoid a hopper hit on entry.",
        "Roll under the bottom hopper to lead it to the right, and then roll back to reach the door safely.
      ]
```

region/tourian/main/Blue Hopper Room.json
```diff
@@ -603,64 +627,89 @@
...
-      "gModeRegainMobility": {},
-      "flashSuitChecked": true
+      "unlocksDoors": [
+        {"types": ["ammo"], "requires": []}
+      ],
+      "note": "Enter the room morphed to avoid a hopper hit on entry."
     }
-  ]
+  ],
+  "devNote": "FIXME: This room could use obstacles indicating if each door is free of hoppers, allowing opening the door with missiles or a power bomb."
```
osse101, line 714: And so it doesn't use a kill strat into a gmode setup strat.  Usually you can still work around it if that's what logic calls for.

osse101 (approved):

PR #1562, opened 2024-06-22
Green Brinstar remote runways and temp blue

blkerby:


region/brinstar/green/Green Brinstar Main Shaft.json
```diff
@@ -2645,6 +3057,62 @@
...
+      "link": [5, 7],
+      "name": "Come in Shinecharging, Leave With Temporary Blue (Partial Runway)",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 5,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canLongChainTemporaryBlue",
```
osse101, line 3089: For placement sure, but this is 2 temp blues.
Note about falling through the shot block - wait to morph until you are through and then you can bounce on the single tile platform to reach the door.

region/brinstar/green/Green Brinstar Main Shaft.json
```diff
@@ -2645,6 +3057,62 @@
...
+      "link": [5, 7],
+      "name": "Come in Shinecharging, Leave With Temporary Blue (Full Runway)",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 12,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canLongChainTemporaryBlue",
```
osse101, line 3070: Full runway would do a 4highMidairMorph

osse101 (commented):
- I don't know if there is extra handling for mockballing on just the doorframe, but implementation side problem.

region/brinstar/green/Green Hill Zone.json
```diff
@@ -681,6 +827,22 @@
...
+      "link": [2, 2],
+      "name": "Leave With Temporary Blue",
+      "requires": [
+        {"canShineCharge": {
+          "usedTiles": 17,
+          "openEnd": 1
+        }},
+        "canChainTemporaryBlue",
+        "canXRayTurnaround",
+        "canSpringBallJumpMidAir"
```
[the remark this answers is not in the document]
blkerby: I tested that it works with a spring ball jump with no HiJump. Added HiJump + Ice option now.

region/brinstar/green/Green Brinstar Main Shaft.json
blkerby: It was for placement, yeah, because I had a hard time with blindly unmorphing onto that single tile platform. But with your note, it becomes a lot easier. Added that and changed the requirement to just "canChainTemporaryBlue".

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -287,6 +287,67 @@
...
+        }
+      },
+      "collectsItems": [5],
+      "devNote": "The item could be avoided by using a shorter landing runway and enough speed"
+    },
+    {
+      "link": [1, 1],
+      "name": "Leave Space Jumping",
+      "requires": [
+        "canPreciseSpaceJump"
```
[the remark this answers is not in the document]
blkerby: Yeah, added "collectsItems" in a bunch more places now, to be safe. I know there would still be more ways to avoid the item, but I'm not sure it's worth the hassle at this point to fully model it, given the amount of duplication of the strats that this would need.

region/brinstar/green/Green Brinstar Main Shaft.json
blkerby: Added.

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -323,16 +333,17 @@
       "link": [1, 1],
       "name": "Leave Space Jumping",
       "requires": [
-        "canPreciseSpaceJump"
+        "canTrickyJump"
```
osse101, line 336: Should this and `Leave Spinning` requirements be swapped?

blkerby: The "SpaceJump" item requirement is implicit in the exit condition "leaveSpaceJumping" so it's not needed here. Whereas for the "leaveSpinning" strat, we have to include it explicitly. 

The "Leave Spinning" runway length didn't look right though, so I just updated it: when going for max speed (rather than shortcharging) you have to use a little less than full runway in order to be able to Space Jump early enough to control your position through the door as this exit condition requires.

PR #1563, opened 2024-06-23
Add comeInWithBlueSpringBallBounce in Final Missile Bombway

blkerby:
Normally a temporary blue into spring ball bounce could be used here instead. By bouncing all the way instead of using temporary blue, it allows keeping a flash suit, so it seemed worth adding.


osse101:
You have to really think to find rooms with left doors that can be reached in speedball with springball bounces but not mockball.  But landing site has one at least.

PR #1564, opened 2024-06-23
More ways to break Climb bomb blocks

blkerby:


region/crateria/central/Climb.json
```diff
@@ -1141,6 +1199,19 @@
         "can4HighMidAirMorph"        
       ]
     },
+    {
+      "link": [4, 6],
+      "name": "Blue Spring Ball Bounce",
+      "entranceCondition": {
+        "comeInWithBlueSpringBallBounce": {
+          "maxExtraRunSpeed": "$3.F",
```
kjbranch, line 1207: This all looks good to me, I just don't know how to review the `maxExtraRunSpeed` on [3, 6] and [4, 6]

kjbranch (approved):

region/crateria/central/Climb.json
osse101: Yeah I see that there is a max allowable speed when doing a bottom frame jump.  Jumping a little before the door is going to be possible in almost all cases; a door frame only mockball into springball bounce is roughly the case where you can't.

I believe that I want a new insane tech for speedballing into raised 1 tile gaps since these two strats are so hard to do.  I don't know that there are other similar situations though, if we write big pink spospo skip as a tempBlue bounce instead of a speedball.

osse101 (approved):

region/crateria/central/Climb.json
blkerby: Another option would be to group the blue springball bounce strats here in with the reusable notable "Climb Morph Tunnel SpeedBall", as a way to put it into Insane?

osse101: With a setup where you do get about the same speed for the bounce this is reasonable.  Without, its very tedious for extreme but you could still do a chainTempBlue.  Jump+speedballing through the lower hole is actually a beyond strat, top door is just an annoying insane strat.

blkerby: I guess the option to chain temp blue would only be unavailable if you need to preserve a flash suit. But in general looks like the blue spring ball bounce strats may be ok for Extreme, if I understand what you're saying?

I added run speed constraints for the tunnel speedball strats now, and put a `canBeExtremelyPatient` requirement on the lower one to place it in Beyond.

PR #1565, opened 2024-06-26
Refine some Crateria Supers cross-room strats

blkerby:
I'm envisioning "canTrickySpringBallBounce" here as maybe an Expert tech, for cases requiring a high degree of precision like bouncing into morph tunnels? We could put canInsaneJump on ones that are even more difficult, e.g. where you have to do it while moving with significant speed and possibly from another room (vs. doing it from a stop in temporary blue)? For cases that aren't too precise but are tricky I assume we can use "canTrickyJump" (combined with "canSpringBallBounce") to put them in Very Hard, an example being bouncing across a whole room like the bottom of West Ocean?

### DIFF
--- a/tech.json
+++ b/tech.json
@@ -456,7 +456,7 @@
                           "name": "canTrickySpringBallBounce",
                           "techRequires": ["canSpringBallBounce"],
                           "otherRequires": [],
-                          "note": "Performing spring ball bounces in tricky situations, such as squeezing into a Morph tunnel while bouncing to keep blue speed."
+                          "note": "Performing spring ball bounces in tricky situations, such as precisely platforming while bouncing to keep blue speed."
                         }
                       ]
                     },
--- a/tech.json
+++ b/tech.json
@@ -450,7 +450,15 @@
                       "name": "canSpringBallBounce",
                       "techRequires": ["canLateralMidAirMorph"],
                       "otherRequires": ["SpringBall"],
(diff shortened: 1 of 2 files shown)


tech.json
```diff
@@ -450,7 +450,15 @@
...
                       "techRequires": ["canLateralMidAirMorph"],
                       "otherRequires": ["SpringBall"],
-                      "note": "Using a lateral mid-air morph to bounce off a surface with Spring Ball while retaining previous momentum."
+                      "note": "Using a lateral mid-air morph to bounce off a surface with Spring Ball while retaining previous momentum.",
+                      "extensionTechs": [
+                        {
+                          "name": "canTrickySpringBallBounce",
+                          "techRequires": ["canSpringBallBounce"],
+                          "otherRequires": [],
+                          "note": "Performing spring ball bounces in tricky situations, such as squeezing into a Morph tunnel while bouncing to keep blue speed."
```
osse101, line 459:
```suggestion
                          "note": "Performing spring ball bounces in tricky situations, such as precisely platforming while bouncing to keep blue speed."
```

osse101: I see `canTrickySpringBallBounce` as springballing through a room like pre-ridley, gauntlet, or LN-pillars.  The climb supers situation is more about landing the single sinking morph into the hole, but its worded here too closely to the `canTemporaryBlue` bounce into a morph tunnel.  But it still makes sense for it to apply here.

PR #1566, opened 2024-06-28
Update m3-room.schema.json

ryannewington:
Fixes the "note" property of "reusableRoomwideNotable" to use the common definition in "m3-note.schema.json#/definitions/note"

### DIFF
--- a/schema/m3-room.schema.json
+++ b/schema/m3-room.schema.json
@@ -2339,13 +2339,9 @@
             ],
             "pattern": "^(.*)$"
           },
-          "note": {
-            "$id": "#/properties/reusableRoomwideNotable/items/properties/note",
-            "type": ["string", "array"],
-            "title": "Strat Description",
-            "description": "A common description for all strats that use this reusable strat",
-            "default": "",
-            "pattern": "^(.*)$"
+         "note": {
+            "$ref" : "m3-note.schema.json#/definitions/note",
+            "$id": "#/properties/enemies/items/properties/note"
           },
           "devNote": {
             "$ref" : "m3-note.schema.json#/definitions/devNote",


schema/m3-room.schema.json
```diff
@@ -2339,13 +2339,9 @@
...
-          "note": {
-            "$id": "#/properties/reusableRoomwideNotable/items/properties/note",
-            "type": ["string", "array"],
-            "title": "Strat Description",
-            "description": "A common description for all strats that use this reusable strat",
-            "default": "",
-            "pattern": "^(.*)$"
+         "note": {
+            "$ref" : "m3-note.schema.json#/definitions/note",
+            "$id": "#/properties/enemies/items/properties/note"
```
blkerby, line 2344:
```suggestion
            "$id": "#/properties/reusableRoomwideNotable/items/properties/note"
```

blkerby:
I'm not sure about this change, since it removes the more specific "description" of the "note" field in this context. And it seems there wouldn't be a way to use "$ref" while properly keeping that "description": https://github.com/json-schema-org/json-schema-spec/issues/239.

What do you think @osse101 ?

osse101:
I view the `reusableRoomwideNotable` as a (name, note) pair and the general "devNote" is attached.  Where a ref note can be attached to many things.  So I don't see a ref note as something that necessarily should be used in this instance.  

A ref note attached to strats can mean tips, other strats, data, partial instruction, or formal instruction.  A reusableRoomwideNotable note can be summarized instruction or formal common instruction.  In some ways they are the same but there are also distinctions.

From reading the linked issue they suggest not using a ref so that we can keep the unique `description`.

ryannewington:
Hi guys,

Apologies, seems like I mistook this as a mistake, but it's a deliberate difference. Was looking at the note/devnote pairs while validating some models generated from this source and came across the discrepancy. Happy to cancel this PR.

blkerby:
@ryannewington No need to apologize, thanks for taking an interest in the project!

PR #1568, opened 2024-06-28
Add canTrickyJump reqs to Moat speedy jump and airball

blkerby:
These both have a 2-frame window for the jump.


kjbranch (approved):

osse101:
With even more run speed you have more possible jump frames before the door that will cross over, but then you need to know from how you hit the transition if you will make it or not where being wrong can be a softlock.

PR #1570, opened 2024-06-29
Kraid Brinstar remote runways

blkerby:


region/brinstar/kraid/Kraid Eye Door Room.json
```diff
@@ -290,6 +309,109 @@
...
+        }
+      },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}]
+    },
+    {
+      "link": [1, 3],
+      "name": "Leave With Mockball (Space Jump, Door Open)",
+      "requires": [
+        {"doorUnlockedAtNode": 1},
+        "SpaceJump"
```
osse101, line 334: These are similar to `canPreciseSpaceJump`, but you are going slow with a health refill right here for when you hit the spikes.
So canTrickyJump?

region/brinstar/kraid/Warehouse Entrance.json
```diff
@@ -430,6 +466,45 @@
...
+          }
+        }
+      }
+    },
+    {
+      "link": [3, 3],
+      "name": "Leave With Mockball",
+      "requires": [],
+      "exitCondition": {
+        "leaveSpinning": {
```
osse101, line 487: Leave With Mockball

region/brinstar/kraid/Warehouse Kihunter Room.json
```diff
@@ -406,6 +434,39 @@
...
+          {"and": [
+            {"obstaclesNotCleared": ["D"]},
+            {"canShineCharge": {
+              "usedTiles": 31,
+              "openEnd": 1
+            }}    
+          ]}  
+        ]},
+        "canChainTemporaryBlue",
+        "canTrickySpringBallBounce"
```
osse101, line 459: Are all bounce-into-morph-tunnel-and-use-springball's a trickyspringballbounce?  Because there's nothing special about this one.

region/brinstar/kraid/Warehouse Kihunter Room.json
```diff
@@ -795,6 +907,7 @@
     }
   ],
   "devNote": [
-    "FIXME: An ice moonfall might be possible to clip past the bottom left door shell?"
+    "FIXME: An ice moonfall might be possible to clip past the bottom left door shell?",
+    "FIXME: Leaving with temporary blue through the bottom door (node 2) is possible but seems really bad."
```
osse101, line 911: yeah its really bad

region/brinstar/kraid/Warehouse Kihunter Room.json
blkerby: To me getting a bounce into a tunnel feels pretty tough to get consistent. And here if you miss then you have to redo the setup from the other room. But I assume it gets better with practice. And the difficulty of the setup is already covered by the "canChainTemporaryBlue". If it could be done without a lengthy setup, then it would seem fine in Very Hard? So I guess it makes sense to not put "canTrickySpringBallBounce" on all of these, yeah

osse101: Bouncing into a tunnel is tough to get consistent.  Big pink->charge is the only one with a setup I like and that is very recent.  It's annoying to retry so I could see the general idea of doing it being expert.

This strat is [5, 3] that is labelled to only use [3].  It is an in-room shortcharge, so the friendlier kind to retry.  You have one tile of space to bounce into the hole, but that is more than(probably) some strats that run up closer to the tunnel like in Zeela room.

blkerby: It starts at 3 under the assumption that you would begin the strat by opening the door before going back left through the tunnel; this prevents the need to add an explicit `unlocksDoors`. In general I try to have strats with an exit condition start at the door node when possible, since I think it may be easier for players to find the strats there; but we haven't necessarily been consistent about it and it may be debatable which way is better. In this case you could also start at node 5 and unlock/open the door as you get to it while in temp blue, so either way seems logically valid. 

I guess the main drawback of doing it this way is it's going to require extra steps (e.g. in the spoiler log) that aren't necessary, because it will say to break the blocks first (e.g. with another temp blue) and go to the door before going back. If we cared about timings (e.g. for the escape, or for turning the room heated) then we would want it to be able to do it directly, which would mean starting at node 5. Thoughts?

PR #1573, opened 2024-07-01
Climb - Add logic for rising lava

kjbranch:
There are some strats that arent perfect; they basically all deal with going to door 4 moderately slowly, or 4 is locked.


blkerby:
The lava only rises when Zebes is awake. I guess the uses of h_RemoveClimbLava could be replaced with a new helper that has an "or" with Zebes asleep + canRiskPermanentLossOfAccess?

PR #1576, opened 2024-07-05
Boyon Gate and Crab Hole remote runways

blkerby:
Crab Hole has new applications with cross-room space jumps and spring ball bounces. The spring ball bounces split into different runway requirements based on uncontrolled vs. controlled bounces. And most of the strats also split based on whether Speedbooster is equipped. This also includes refinements of the existing cross-room jump strats.

### DIFF
--- a/region/maridia/outer/Crab Hole.json
+++ b/region/maridia/outer/Crab Hole.json
@@ -1122,7 +1122,7 @@
         }
       },
       "requires": [
-        "canLongChainTemporaryBlue"
+        "canChainTemporaryBlue"
       ],
       "exitCondition": {
         "leaveWithTemporaryBlue": {}
@@ -1344,7 +1344,7 @@
         }
       },
       "requires": [
-        "canLongChainTemporaryBlue"
+        "canChainTemporaryBlue"
       ],
       "exitCondition": {
         "leaveWithTemporaryBlue": {}
--- a/region/maridia/outer/Crab Hole.json
+++ b/region/maridia/outer/Crab Hole.json
@@ -786,8 +786,7 @@
       "requires": [
         "SpeedBooster",
         "canCrossRoomJumpIntoWater",
-        "canMomentumConservingMorph",
-        "canDoubleSpringBallJumpMidAir"
+        "canMomentumConservingMorph"
       ],
       "note": [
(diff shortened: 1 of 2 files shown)


region/maridia/outer/Crab Hole.json
```diff
@@ -954,6 +1333,23 @@
...
+      "link": [3, 2],
+      "name": "Come in Shinecharging, Leave With Temporary Blue",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 1,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canLongChainTemporaryBlue"
```
osse101, line 1346: Should this not be `Long`?

region/maridia/outer/Crab Hole.json
```diff
@@ -749,6 +1111,23 @@
...
+      "link": [2, 3],
+      "name": "Come in Shinecharging, Leave With Temporary Blue",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 1,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canLongChainTemporaryBlue"
```
osse101, line 1124: and here

region/maridia/outer/Crab Hole.json
```diff
@@ -657,6 +772,183 @@
...
+      "entranceCondition": {
+        "comeInWithSpringBallBounce": {
+          "remoteAndLandingMinTiles": [[3, 1]],
+          "movementType": "uncontrolled"
+        }
+      },
+      "requires": [
+        "canCrossRoomJumpIntoWater",
+        "canMomentumConservingMorph",
+        "canDoubleSpringBallJumpMidAir"
```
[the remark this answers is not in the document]
blkerby: It's only with SpeedBooster equipped that the springball jump can be needed. Speeds $1.2 and $1.3 are the ones that require a spring ball jump if you don't get the last-frame bounce. Speed $1.2 is the max speed you can get with a jump from a 3-tile runway with 1 closed end. Requiring a last-frame bounce seemed like it could increase the difficulty too much; though maybe this could also be handled through randomizer-side leniency on the runway speed. And in any case, it would still be an option to do the spring ball jump even if it's not included as a requirement, so I guess it doesn't matter too much :shrug:

PR #1577, opened 2024-07-05
Fix Sciser spelling

blkerby:


osse101:
Looks like `enemyDamage` needs a test for valid enemies

PR #1581, opened 2024-07-10
Most of rest of West Maridia remote runways

blkerby:
I still have Mt. Everest and Red Fish Room to do after this, but Everest is turning out to be fairly complicated so I thought maybe it's better to open this PR now and do Everest in a separate one afterward.

PR #1583, opened 2024-07-12
Everest remote runways

blkerby:
This is based on #1581 (since the new comeInSpinning condition is needed), so that should be merged first.


region/maridia/outer/Mt. Everest.json
```diff
@@ -2572,6 +3012,66 @@
...
+        }}
+      ],
+      "exitCondition": {
+        "leaveShinecharged": {
+          "framesRemaining": 115
+        }
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": [
```
osse101, line 3034:
```suggestion
        {"types": ["super", "powerbomb"], "requires": []},
        {"types": ["missiles"], "requires": [
```
Open it from 2, then speedy jump up

region/maridia/outer/Mt. Everest.json
```diff
@@ -2572,6 +3012,66 @@
...
+        }}
+      ],
+      "exitCondition": {
+        "leaveWithTemporaryBlue": {
+          "direction": "right"
+        }
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": [
```
osse101, line 3064:
```suggestion
        {"types": ["super", "powerbomb"], "requires": []},
        {"types": ["missiles"], "requires": [
```

region/maridia/outer/Mt. Everest.json
```diff
@@ -3097,10 +3621,48 @@
...
+        }}
+      ],
+      "exitCondition": {
+        "leaveWithTemporaryBlue": {
+          "direction": "left"
+        }
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": [
```
osse101, line 3647: You can land on the door and open it while chainingTempBlue.  It could help compared to falling through at full speed too.

region/maridia/outer/Mt. Everest.json
```diff
@@ -1875,6 +2049,114 @@
...
+            "steepUpTiles": 5,
+            "steepDownTiles": 5
+          },
+          "maxExtraRunSpeed": "$3.8"
+        }
+      }
+    },
+    {
+      "link": [4, 4],
+      "name": "Leave With Temporary Blue",
```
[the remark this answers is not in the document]
blkerby: Refined this to describe multiple possible runways that could be used..

Added a couple more suitless cross-room strats now, coming in shinecharging at 1 or 4 and leaving with temporary blue at 2 or 3. I think for the Gravity strats, the in-room runways (along with the 1->5 cross-room one) already provide good coverage? I guess it depends on how much we care about including alternatives to gravity jumping, in case there's some scenario where that's not an option.

Short-charging using the runway at 11 from left-to-right I think is shorter than what we've been putting in elsewhere; haven't been putting in runways that have effective length of 12 tiles or less; we could add them but it seems like Beyond territory? In this case if you wanted to blue space jump to the right door I think you would use the left-most island instead, since that is longer, a little more than 14 tiles. Is it worth adding to avoid a gravity jump?

osse101 (approved):

PR #1585, opened 2024-07-16
Aqueduct remote runways

blkerby:


region/maridia/inner-pink/Aqueduct.json
```diff
@@ -887,6 +988,79 @@
         {"shinespark": {"frames": 40, "excessFrames": 7}}
       ]
     },
+    {
+      "link": [2, 7],
+      "name": "Temporary Blue Gravity Jump",
+      "requires": [
+        "h_canShineChargeMaxRunway",
+        "canGravityJump",
+        "canChainTemporaryBlue"
```
osse101, line 997: HiJump should be enough to only use a regular tempblue grav jump

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -1529,6 +2032,35 @@
...
+      "name": "Temporary Blue Chain",
+      "entranceCondition": {
+        "comeInWithTemporaryBlue": {
+          "direction": "left"
+        },
+        "comesThroughToilet": "any"
+      },
+      "requires": [
+        "canChainTemporaryBlue",
+        "can4HighMidAirMorph"
```
osse101, line 2046: With gravity off, the jump morphs aren't tight

region/maridia/inner-pink/Aqueduct.json
blkerby: Yeah, good find, it does work.

osse101 (commented):
I tried out the [2,1] strat.  The [2,5] double springball jump is on a bit of the harder side for extreme but fits.  The [5,9] upper path double SBJ is harder than that, so insane.  The springflings aren't bad but can make you drop the temp blue.

[5,7] starting from 2 is mostly reasonable.  Snails can get in the way, but you can avoid them.  Breaking the right speed blocks to clip up is a bit tricky, but fair for extreme.  An extreme trick into another extreme trick with jumpIntoCFClip is mean, but there is PB leniency for that ☘️

PR #1588, opened 2024-07-18
Colosseum remote runways

blkerby:
Some things worth noting:
- Updated the "Colosseum Bomb Jump Water Escape" notable to take into account that it can also be used in the middle and right side of the room.
- There's an application of some tech of using a pause buffer to continue chaining temporary blue after unmorphing in the air (i.e. an unmorph other than the soft landing type). I remembered seeing this in a bobbob video a while back but forgot about it until recently. It could probably be made into its own "tech", will start keeping an eye out for other applications.

### DIFF
--- a/region/maridia/inner-pink/Colosseum.json
+++ b/region/maridia/inner-pink/Colosseum.json
@@ -655,19 +655,19 @@
         "comeInGettingBlueSpeed": {
           "length": 1,
           "openEnd": 1,
-          "maxExtraRunSpeed": "$2.3"
+          "maxExtraRunSpeed": "$1.9"
         }
       },
       "requires": [
         "HiJump",
-        "canInsaneJump",
         "canPreciseSpaceJump",
         "canChainTemporaryBlue"
       ],
       "exitCondition": {
         "leaveWithTemporaryBlue": {}
       },
-      "unlocksDoors": [{"types": ["ammo"], "requires": []}]
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
+      "devNote": "This can be done at greater speed (at least up to $2.3) but would be more difficult."
     },
     {
       "link": [2, 1],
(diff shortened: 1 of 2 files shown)


region/maridia/inner-pink/Colosseum.json
```diff
@@ -998,6 +1125,56 @@
...
+        "canChainTemporaryBlue"
+      ],
+      "exitCondition": {
+        "leaveWithTemporaryBlue": {}
+      },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}]
+    },
+    {
+      "link": [2, 3],
+      "name": "Come in Shinecharging, Leave With Temporary Blue (Gravity Jump, Spring Ball Jump, Pause Buffer)",
```
osse101, line 1150: The sequence looks possible.  The morph->springballjump->unmorph is really quick and the hardest part.  The pause buffer morph is mostly difficult because it follows the springballjump sequence.  Pause buffer morph makes sense as a new tech because its not too well known.  But it could also be included in another tech, depending on if it has uses other than chainTempBlue.  Like how TAS uses a pause buffer morph to grapple fling down through bowling.

region/maridia/inner-pink/Colosseum.json
```diff
@@ -603,6 +648,27 @@
...
+      "name": "Blue Space Jump, Leave With Temporary Blue (Womple Jump)",
+      "entranceCondition": {
+        "comeInGettingBlueSpeed": {
+          "length": 1,
+          "openEnd": 1,
+          "maxExtraRunSpeed": "$1.9"
+        }
+      },
+      "requires": [
+        "HiJump",
```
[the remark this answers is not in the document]
blkerby: Good catch

region/maridia/inner-pink/Colosseum.json
blkerby: Sounds good. In the next PR there turned out to be a lot more applications of the technique so I was adding the new tech there (`canPauseRemorphTemporaryBlue`). Went ahead and copied it over into this PR as well.

PR #1591, opened 2024-07-25
East Ocean: Use Flash Suit

kjbranch:


region/crateria/east/East Ocean.json
```diff
@@ -898,7 +919,12 @@
           "canWalljump",
           "HiJump",
           "SpaceJump",
-          "canIBJ"
+          "canIBJ",
+          "canSpringBallJumpMidAir",
+          {"and": [
+            "canTrickyJump",
```
osse101, line 925: pretty easy to do from the ledge, maybe not obvious?

region/crateria/east/East Ocean.json
```diff
@@ -1050,6 +1127,24 @@
...
+        "canSuitlessMaridia",
+        "h_canUnderwaterCrouchJumpDownGrab",
+        {"useFlashSuit": {}},
+        {"shinespark": {"frames": 9, "excessFrames": 4}}
+      ],
+      "flashSuitChecked": true,
+      "note": [
+        "Crouch jump down grab to get over the first two pillars.",
+        "This can be done by very quickly pressing down after jumping, to prevent the use of the flash suit.",
+        "For the third pillar, Crouch jump to time a damage boost on the Skultera.",
```
osse101, line 1144: Should this say use Flash suit?

region/crateria/east/Crab Maze.json
```diff
@@ -390,7 +390,16 @@
       "link": [3, 1],
       "name": "Base",
       "requires": [
-        "Morph"
+        "Morph",
```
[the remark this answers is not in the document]
kjbranch: I never knew that second crumble was there lol

region/crateria/east/East Ocean.json
kjbranch: it's not too hard, but I felt like it was harder than `Medium`. Maybe it's ok with a careful jump, since it will only be enabled with collectible wall jump?

PR #1595, opened 2024-07-26
Green Maridia remote runways

blkerby:
This isn't quite all of Green Maridia, because Shaktool's Room needs a major overhaul so it can go in a separate PR.

In West Sand Hall, the suitless shinespark strats both directions are tightened to reflect that they can be done tankless. Also the requirements about entering with low speed were removed. These were unnecessary since at high speeds you can just crouch before the door transition and slide through; for the left-to-right direction, it's also not a problem if you slide off the ledge while getting the shinecharge since you can still get into position for the spark (with even more time to spare).


region/maridia/inner-green/West Sand Hall.json
```diff
@@ -623,6 +666,79 @@
...
+      },
+      "requires": [
+        "canSuitlessMaridia",
+        "canPlayInSand",
+        "canTrickyDashJump"
+      ],
+      "note": [
+        "Run into the room using between 11 and 12 tiles of runway in the other room (extra run speed between $2.D and $3.1).",
+        "Jump somewhat soon after entering the room, and land on the pillar just past the last Evir.",
+        "To get a controlled landing, it can help to turn around while still in spin, then break spin immediately after."
```
osse101, line 687: Turning around early and drifting to the platform while slowing down is funny.  Then break spin to stop instantly.

region/maridia/inner-green/West Sand Hall.json
```diff
@@ -180,27 +181,77 @@
...
+          "speedBooster": true
+        }
+      },
+      "requires": [
+        "canSuitlessMaridia",
+        "canTrickyDashJump",
+        "canInsaneJump"
+      ],
+      "note": [
+        "Run into the room using exactly 20 tiles of runway (with closed ends) in the other room, or up to 6 pixels more (getting extra run speed exactly $4.4).",
```
osse101, line 216: that's wild

PR #1596, opened 2024-07-27
Shaktool Room restructure

blkerby:
- This removes node 3 (the event node) in favor of the newer `setsFlags` strat property.
- Two new obstacles are also added, representing the Power Bombs blocks and the sand blocks.
- The sand blocks are modeled separately since they can be cleared using the snails when entering the room from the right, even without Power Bombs, and this can matter for using the runway to get a shinecharge (or do other "remote runway" things, which will be added in a separate PR after this).
- Adjustments are made to account for the fact that it is unsound to treat the blocks as being cleared just because the flag `f_ShaktoolDoneDigging` is set; rather, the room must be entered with the flag already set in order for the blocks to be initially cleared. This is handling by adding self-link strats at the door nodes with a `comeInNormally` entrance condition, to clear the obstacles if the flag is set.
- Grapple teleporting from right-to-left across the room is an interesting case. With the vanilla room behavior, the flag does not get set in this case (since Samus is never located on the right side of the room when the processing happens that would set the flag). But in Map Rando, the flag would get set. Even though the flag is set, the blocks would not be cleared until resetting the room. If the left door is locked, resetting the room might not be possible, so this is an example of why it matters to model the block state (using obstacles) separately from the flag.
- The new tech `canTrickyDodgeEnemies` is copied from PR #1595. I'm thinking of it as probably an Expert-level tech; it's for movements dealing with enemies where you may have to do something complicated/unintuitive, not necessarily requiring an extreme amount of precision, just very tricky.


osse101 (commented):
- still need to try these ways of clearing all the sand with snails.
- What were you thinking `canTrickyDodgeEnemies` means by example?

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -189,38 +199,126 @@
...
+      "name": "Grapple Teleport (Symmetric)",
+      "entranceCondition": {
+        "comeInWithGrappleTeleport": {
+          "blockPositions": [[5, 3], [7, 2]]
+        }
+      },
+      "requires": [
+        "h_ShaktoolSymmetricFlag"
+      ],
+      "setsFlags": ["f_ShaktoolDoneDigging"]
```
kjbranch, line 282: could use a note here and on all the similar strats. the name of the helper is not intuitive.

kjbranch:
clearing the sand sounds more like tricky use enemies, which I think could have a lot of applications and I have considered making it a few times. It could be added to weird g-mode setups, enemy runways, etc

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -230,20 +329,110 @@
...
+        "comeInNormally": {}
+      },
+      "requires": [
+        "f_ShaktoolDoneDigging"
+      ],
+      "clearsObstacles": ["A", "B"]
+    },
+    {
+      "link": [2, 2],
+      "name": "Shaktool Room Bounce the Snails Around the Room",
```
osse101, line 399: I think 
Expert: Gravity or yard hit
Very Hard: Morph and Gravity
Extreme: patience? 

Its pretty easy to miss a shot then just walk into the snail for damage.
Or get slowed by the water when you don't want to be.

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -230,20 +329,110 @@
...
+      "notable": true,
+      "requires": [
+        "canDodgeWhileShooting",
```
kjbranch, line 402:
```suggestion
        "h_ShaktoolCameraFix",
        "canDodgeWhileShooting",
```

region/maridia/inner-green/Shaktool Room.json
blkerby: After playing around with this some more, I'm thinking Morph may not be important to model here. Gravity is the item that makes the big difference in difficulty. With Gravity, you have a lot more freedom in how you can kick the snails around, and with that there doesn't seem to be much risk of getting them both stuck off-camera; having a second snail available to use as a spare helps a lot.

Without Gravity, it's fairly different, as it needs more of a plan. I'm thinking it maybe makes sense to split off the suitless version as a separate notable strat which can be in Expert. And then we don't have to introduce a new tech ("canTrickyDodgeEnemies"). Snail hit seems reasonable to assume for the Expert suitless version, with damageless suitless being assumed in Extreme, whether by patience or by carefully timed movement.

PR #1597, opened 2024-07-30
Acid Snakes Tunnel carry temp blue

blkerby:
For these strats, the amount of heat damage would depend on the amount of run speed, which depends in a complicated way on the amount of runway available in the other room and on shortcharge skill. The heat frames that I put in are based on a conservative assumption of very low run speed. At some point we could add variant strats with lower heat frames by assuming a higher run speed, but I don't think that's too important right now.

Someday I still think it would be neat to add schema support for interpolating heat frames in a continuous way based on runway length, which could be useful in many situations in heated rooms. But that's a problem for another day.

### DIFF
--- a/region/norfair/east/Acid Snakes Tunnel.json
+++ b/region/norfair/east/Acid Snakes Tunnel.json
@@ -119,9 +119,9 @@
   "strats": [
     {
       "link": [1, 2],
-      "name": "Come in Shinecharging, Leave With Temporary Blue",
+      "name": "Come in Getting Blue Speed, Leave With Temporary Blue",
       "entranceCondition": {
-        "comeInShinecharging": {
+        "comeInGettingBlueSpeed": {
           "length": 13,
           "openEnd": 1
         }
@@ -202,9 +202,9 @@
     },
     {
       "link": [2, 1],
-      "name": "Come in Shinecharging, Leave With Temporary Blue",
+      "name": "Come in Getting Blue Speed, Leave With Temporary Blue",
       "entranceCondition": {
-        "comeInShinecharging": {
+        "comeInGettingBlueSpeed": {
           "length": 13,
           "openEnd": 1
         }
--- a/region/norfair/east/Acid Snakes Tunnel.json
+++ b/region/norfair/east/Acid Snakes Tunnel.json
@@ -84,6 +84,8 @@
       "from": 1,
       "to": [
         {"id": 1},
+        {"id": 2},
+        {"id": 3},
         {"id": 4}
       ]
     },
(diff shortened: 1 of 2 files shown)


region/norfair/east/Acid Snakes Tunnel.json
```diff
@@ -115,13 +117,126 @@
     }
   ],
   "strats": [
+    {
+      "link": [1, 2],
+      "name": "Come in Shinecharging, Leave With Temporary Blue",
+      "entranceCondition": {
+        "comeInShinecharging": {
```
osse101, line 124: You can spacejump through [1, 2] to leave speedballing and that would be a comeInGettingBlueSpeed.
These strats make you do a Temporary Blue chain so there is no real difference (with comeInShinecharing)

osse101 (dismissed):
- You can spacejump through [1,2] and [2,1] to speedball out.  Doing the same with [3, 1] could be possible, but would be harder.
- [3, 2] is something I think people would want to do even if it involves going through a extra room than normal.
- `canXRayCancelShinecharge` would be placed a little higher than expected, I think.  It complicates your inputs, and usually only in situations where you want to rush.

blkerby:
Good catch. It's more accurate to have `comeInGettingBlueSpeed` here since the heat frames are based on an assumption of running and jumping into a speedball, not getting a shinecharge (which would be much slower). The logical requirements would be equivalent since the temp blue chain would prevent carrying a flash suit either way, but always best to make it more accurate anyway.

There's a lot of places in the game where you can (blue) space jump in and then mockball/speedball out the opposite door. Moat is another example. The schema isn't yet set up to handle it, because we would need a way to propagate the run speed constraints across a room in the middle. It's been in the back of my mind for a while; and I've been thinking we could add these in a second pass after finishing the first pass of remote-runway strats that we're doing now. Schema for carrying blue speed through vertical transitions is also something that still needs to be designed and added.

Makes sense that in terms of difficulty, carrying temp blue through one more very small room isn't necessarily bad. Main drawback is it would maybe make the logic a little more murky in terms of the player being able to know what is logical or not, since up to this point we've been strictly following a rule of "no comeInWithTemporaryBlue + leaveWithTemporaryBlue" in order to limit blue chains to 2 rooms. Of course logic is already murky for heat runs, so maybe that's not a big deal. A clean solution could maybe be to limit the total length of the chain, based on some measure that is more refined than just number of rooms. We could count the number of hops for example. I didn't want to do that in the beginning because I wasn't confident in my ability to measure what the minimal number of hops would be. But at this point I think I'm getting the hang of it enough to be able to put in reasonable numbers.

PR #1598, opened 2024-08-01
Add schema for strat IDs

blkerby:
After this I can add a script to auto-assign strat IDs.

The purpose of tracking a "nextStratId" is to ensure that a strat ID isn't ever reassigned to a different strat, in case the strat with the largest ID in the room gets deleted.

### DIFF
--- a/schema/m3-room.schema.json
+++ b/schema/m3-room.schema.json
@@ -84,6 +84,12 @@
         }
       ],
       "properties": {
+        "id": {
+          "$id": "#/definitions/strat/properties/id",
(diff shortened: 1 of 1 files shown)


osse101 (approved):
Seems fine.  Saving the nextID value to use is alright.  Maybe, not now, it should be grouped in a metadata/script variables object.  I think it encourages setting stratIDs using a script that updates nextID.  2 PRs adding strats to the same room would need some communication, probably the ability to manually set the stratID, and tests to verify uniqueness.

PR #1599, opened 2024-08-06
Add script to populate strat IDs

blkerby:
After this, I'm planning to create a GitHub Actions workflow that we could use to just click a button to generate/open a PR to apply this script whenever we want.


osse101 (dismissed):
Looks like it does what it is supposed to.  Kinda weird seeing nodes, obstacles, and now strats are start with `id` now.

I would like if `Processing` was only shown on changed files.

There should be tests that `id`s are unique per room and all that later.

PR #1601, opened 2024-08-07
Run scripts: auto-format and generate strat IDs

blkerby:


kjbranch (approved):
`unlocks doors` and `block positions` get expanded more than I'd prefer, but if that is changed, it can be done later.

There are some strats up in the `nodes` section. i.e. unlocking doors. Should those have an id or be ignored for now? 

When adding a new strat, should we ever manually add an id and update the `nextStratId`?

blkerby:
Agree about the `unlocksDoors`, etc. The auto-formatter could use another look later to try to improve these.

I would ignore the remaining strats in the `nodes` section. I think we eventually want to clean those up so there won't be any more strats anywhere other than in the room `strats` array.

I'm not sure about manually adding strat IDs. I imagine we want to choose one of two ways to do this: either 1) make strat IDs required so the tests enforce that they are always populated in each PR (either manually or by running the script), or 2) leave them out and run a script to populate them periodically as separate PRs. There are pros & cons either way. Enforcing that the strat IDs are always populated would keep the data cleaner, and ensure that a strat could always be referenced in the upcoming video site, but it's more hassle, adds a bit more clutter to the review process, and it would mean that two pull requests that add strats on the same room would always conflict with each other. A hybrid approach could be possible where strat IDs are sometimes manually added and sometimes not, but I'm not sure it makes as much sense.

PR #1610, opened 2024-08-20
Bat Cave temporary blue

blkerby:
The `canXRayCancelShinecharge` doesn't help quite as much going from bottom to top, since you have the wait a bit anyway for the path to become clear of Skree projectiles. Videos for these are up already on https://videos.maprando.com. I can associate them to the new strats after strat IDs are added; going to work on a GitHub Action workflow for that soon.

### DIFF
--- a/region/norfair/east/Bat Cave.json
+++ b/region/norfair/east/Bat Cave.json
@@ -184,6 +184,31 @@
     },
     {
       "link": [1, 2],
+      "name": "Come in Shinecharging, Leave With Temporary Blue",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 2,
+          "openEnd": 1
+        }
+      },
+      "requires": [
+        "canLongChainTemporaryBlue",
+        "HiJump",
+        {"or": [
+          {"heatFrames": 575},
+          {"and": [
+            "canXRayCancelShinecharge",
+            {"heatFrames": 415}
+          ]}
+        ]}
+      ],
+      "exitCondition": {
+        "leaveWithTemporaryBlue": {}
+      },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}]
+    },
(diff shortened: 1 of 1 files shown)


region/norfair/east/Bat Cave.json
```diff
@@ -182,6 +182,31 @@
...
+        "comeInShinecharging": {
+          "length": 2,
+          "openEnd": 1
+        }
+      },
+      "requires": [
+        "canLongChainTemporaryBlue",
+        "HiJump",
+        {"or": [
+          {"heatFrames": 570},
```
osse101, line 198: Bouncing over the platforms can save a remorph, and half a second.
Stopping the shinecharge on the farthest right will get hit by a particle.  It's probably easy to avoid.

region/norfair/east/Bat Cave.json
```diff
@@ -303,6 +328,30 @@
...
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 3,
+          "openEnd": 1
+        }
+      },
+      "requires": [
+        "canLongChainTemporaryBlue",
+        {"or": [
+          {"heatFrames": 550},
```
osse101, line 343: Waiting for the bats to delete themselves isn't obvious since you can't see them.
Reaching the door is really awkward.  Bouncing helps, but really isn't safe.

osse101 (commented):
I'm cool with lowering the heatFrame count, or not.

osse101:
Your videos actually look faster than how I do the rooms.

blkerby:
> Your videos actually look faster than how I do the rooms.

Yeah I was grinding it for quite a while. Before this I haven't practiced temp blue chain much under heat damage, so I was taking time to get used to it. If it goes too far then the heat frames could always be adjusted to be more lenient

PR #1613, opened 2024-08-25
Bubble mountain temporary blue

blkerby:
I made videos for all of these strats on https://videos.maprando.com. Takes more time that way, so haven't decided yet whether or not to keep doing this going forward.

### DIFF
--- a/region/norfair/east/Bubble Mountain.json
+++ b/region/norfair/east/Bubble Mountain.json
@@ -183,7 +183,9 @@
         {"id": 1},
         {"id": 2},
         {"id": 3},
+        {"id": 4},
         {"id": 5},
+        {"id": 6},
         {"id": 7},
         {"id": 9},
         {"id": 10}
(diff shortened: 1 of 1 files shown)


region/norfair/east/Bubble Mountain.json
```diff
@@ -1235,6 +1547,45 @@
...
       "flashSuitChecked": true
     },
+    {
+      "link": [2, 7],
+      "name": "Come in Getting Blue Speed, Leave With Temporary Blue",
+      "entranceCondition": {
+        "comeInGettingBlueSpeed": {
+          "length": 0,
+          "openEnd": 1,
+          "minExtraRunSpeed": "$2.6"
```
osse101, line 1557: This gets really hard to save if your run speed if not even that much higher.
With a little more speed than a 3 tap, you don't get the height. And a little more than that reaches the wall before being high enough to clear it.  And a pause remorph is very difficult to do there.

osse101 (commented):
Maybe all of these need `canTrickyJump`.  You are avoiding or landing on single solid tiles or enemies constantly.

`canChainTemporaryBlue` as a movement type tech is an indicator of a general movement level greater than `canTrickyJump`.  I think trickyjump could be made a requirement of chainTemporaryBlue.  The loosely frame perfect jump part of trickyJump is 🤷 

`can4HighMidairMorph` could be considered a core part of chainTemporaryBlue too, and is actually more like can4High doesn't describe the increased difficulty when chainTempBluing that well.

region/norfair/east/Bubble Mountain.json
blkerby: Yeah this needed to have a `maxExtraRunSpeed`. Added that now. With a range of $2.6 to $2.F, it's a 10-frame window in terms of how many frames to hold dash. Above $2.F there's a dropoff in jump speed which would make it not work. We could add a `canTrickyDashJump` if you think that would help? This one is not as precise as the 2-frame window which that tech can imply in the worst case. I have been thinking it would be good to split off an easier version of that tech, to go into Very Hard, for jumps that are tricky but more lenient than 2-frame precise. In any case, this strat has `canLongChainTemporaryBlue` which is a high-difficulty tech which we could maybe assume to include a fairly high level of movement in general, even more so than `canChainTemporaryBlue`?

PR #1614, opened 2024-08-26
Snail Clip - Morph, no X-Ray

kjbranch:
This is an attempt to lower the difficulty of the snail clip strat with just Morph (no x-ray). I don't think the general clip needs to be notable, since it is so well known. Instead, I repurposed the notable to only be used for placing this middle tier version (precise positioning with morph, but no x-ray) in a lower difficulty than `canPreciseCeilingClip`. 

To me, this one feels easier than normal `canPreciseCeilingClip` strats, as there is no precise refreeze to move it a pixel or two at a time with risk of killing the enemy.

### DIFF
--- a/region/maridia/inner-pink/Aqueduct.json
+++ b/region/maridia/inner-pink/Aqueduct.json
@@ -137,10 +137,10 @@
       "note": "Snail Climb without movement items by better coordinating the movements of both the snail and Samus."
     },
     {
-      "name": "Aqueduct Snail Clip",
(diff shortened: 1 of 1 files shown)


osse101:
I like this change.
Snail clip was made notable for the gravity+morph version as it was unclear where to place it, (modern) hard or very hard.
Having a non-notable `canXrayCeilingClip` branch is reasonable and actually more similar to the difficulty level of the clip has been described.  And disabling snail clip without morph but with preciseClipping enabled doesn't seem like something that is needed.

PR #1615, opened 2024-08-27
Misc Fixes: Sponge Bath, WS E-Tank, Green Pirate, Mama Turtle

kjbranch:
I lowered the number of Beetom hits on the ice+x-ray climb to be doable tankless. Alternatively it could be lowered to 19 to require one tank - what do you think?

### DIFF
--- a/region/crateria/west/Green Pirates Shaft.json
+++ b/region/crateria/west/Green Pirates Shaft.json
@@ -1305,7 +1305,7 @@
         "canBeVeryPatient",
         {"enemyDamage": {
           "enemy": "Beetom",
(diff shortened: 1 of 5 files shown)


osse101:
Tankless seemed fine when I had to do it.  I hadn't practiced it anytime recently and was able to do it twice with no tanks.  The leniency is specifically for the second beetom vertical freeze.  
Morph itself is extra leniency that you don't strictly need.  If you don't have Morph or PowerBombs, it may change how we think about hit leniency.  But then `canBeVeryPatient` is there already for the length of the xray climb, so we don't need to worry about it too much.

PR #1616, opened 2024-08-27
Bowling Alley temp blue break bomb blocks

blkerby:
Video for this strat: https://videos.maprando.com/video/308. The ability to retain temporary blue while falling through crumbles could maybe be a separate tech, but it's same kind of soft unmorph as in `canChainTemporaryBlue` so it may be ok to just be assumed as part of that?

This also adds a `canHeroShot` requirement to the existing notable strat "Bowling Alley Missile Chozo Camera Unlock". This won't affect the difficulty level of the strat in Map Rando, since it's already a Very Hard notable.

### DIFF
--- a/region/wreckedship/main/Bowling Alley.json
+++ b/region/wreckedship/main/Bowling Alley.json
@@ -956,13 +956,30 @@
       "requires": [
         "ScrewAttack",
         "Morph",
-        "canCameraManip"
+        "canCameraManip",
+        "canHeroShot"
       ],
       "note": [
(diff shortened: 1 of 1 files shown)


osse101:
Temp bluing under the crumble block can be done by bouncing on the crumble and then unmorphing.  Or jumping off the crumble.  Or speedballing and unmorphing into the room.  So its not hard to do.  And `canChainTemporaryBlue` tends to be required as part of a temp blue entrance condition so its covered under that.

Its a weak hero shot, but you are on a timer.  The camera unlock strat isn't that well known still.  Probably because getting out without a bomb is only expected at higher difficulty levels.

PR #1618, opened 2024-08-28
Add note to LandingSite - BigJumpMidairShinespark

dragonsinth:
https://videos.maprando.com/video/382

I could not make this trick work with a max height jump, even without high jump boots.

### DIFF
--- a/region/crateria/central/Landing Site.json
+++ b/region/crateria/central/Landing Site.json
@@ -1075,7 +1075,7 @@
       "clearsObstacles": ["A"],
       "note": [
         "Store the spark on the right side of the ledge.  Run left and do a big jump towards the gauntlet door and then midair spark at the right time.",
-        "Do not jump to max height or you will likely run out of shinecharge frames, instead short the jump a little.",
+        "Do not jump to max height or Samus will run out of shinecharge frames, instead short the jump a little.",
         "The lines in the background make for a good way to measure height."
       ]
     },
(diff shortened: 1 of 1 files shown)


region/crateria/central/Landing Site.json
```diff
@@ -1075,6 +1075,7 @@
       "clearsObstacles": ["A"],
       "note": [
         "Store the spark on the right side of the ledge.  Run left and do a big jump towards the gauntlet door and then midair spark at the right time.",
+        "Do not jump to max height or you will likely run out of shinecharge frames, instead short the jump a little.",
```
kjbranch, line 1078:
```suggestion
        "Do not jump to max height or Samus will run out of shinecharge frames, instead short the jump a little.",
```

region/crateria/central/Landing Site.json
```diff
@@ -1113,6 +1114,7 @@
       ],
       "note": [
         "Store the spark on the right side of the ledge.  Run left and do a big jump towards the gauntlet door and then midair spark at the right time.",
+        "Do not jump to max height or you will likely run out of shinecharge frames, instead short the jump a little.",
```
kjbranch, line 1117:
```suggestion
        "Do not jump to max height or Samus will run out of shinecharge frames, instead short the jump a little.",
```

kjbranch (approved):

dragonsinth:
@kjbranch thank you, good suggestions

blkerby:
I did some testing and found it's possible to do with a full height jump but it requires jumping somewhat early and the timing feels tighter. And with a shorter jump you don't have as much fall speed which makes it easier to position vertically for the spark. So shorting the jump seems like good advice

PR #1619, opened 2024-08-30
Check horizontal doors without leaveWithRunway

blkerby:
This is just a standalone script to list horizontal door nodes that are lacking a leaveWithRunway strat at the door (i.e. from the door node to itself). If we wanted to enforce this as a check, it could be added to the tests after fixing the existing cases.

### DIFF
--- a/scripts/check_runways.py
+++ b/scripts/check_runways.py
@@ -1,0 +1,21 @@
+import json
+from pathlib import Path
+
+for path in sorted(Path("../region/").glob("**/*.json")):
+    room_json = json.load(path.open("r"))
+    if room_json.get("$schema") != "../../../schema/m3-room.schema.json":
+        continue
+
+    runway_nodes = set()
+    for strat_json in room_json["strats"]:
+        from_node = strat_json["link"][0]
+        to_node = strat_json["link"][1]
+        if from_node != to_node:
+            continue
+        if "exitCondition" in strat_json and "leaveWithRunway" in strat_json["exitCondition"]:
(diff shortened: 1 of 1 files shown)


osse101:
I see farm rooms and blue hopper room missing leaveWithRunway.  The farm rooms seem like it is always better to refill and then run for the door.  Except technically for when you only need like 1 tile of runway, like for a doorstuck.  Doesn't hurt anything to add the remaining self leaveWithRunways.

I don't see value in adding this to the regular tests.  It can be part of the "When nodes move or blocks of strats are shuffled" checklist process.

There may be a potential use for making the script more flexible or duplicated to check for other conditions.  But the other conditions are more situational so it wouldn't work as easy without a detailed ignore list, or alternatively as a way to return a query.

osse101 (approved):

blkerby:
Yeah, in the randomizer, certain entrance conditions like `comeInWithDoorStuckSetup` will only match with a `leaveWithRunway` that has same "from node" as "to node". The reason was that otherwise it would be effectively duplicating the cross-room strats and have to do a calculation to compute heat frames which would be sort of wasteful. I wrote it that way with the expectation that in every case that we have a `leaveWithRunway` with fromNode != toNode we would also have a `leaveWithRunway` from the door node to itself. For a door stuck setup you're not using momentum from running from the other node so to me it didn't make sense to try to match with a strat like that. But it is different from how other exit conditions are handled and I'm kind of ambivalent about it now. Adding the missing `leaveWithRunway` strats does seem like the simplest solution for now.

PR #1621, opened 2024-08-30
Add GitHub Actions workflow to run scripts

blkerby:
This adds a GitHub Action workflow to help us apply autoformatting and to populate strat IDs.

The workflow won't run automatically; it is run by going into the "Actions" tab on GitHub, selecting the "Run Scripts" workflow, and clicking "Run workflow". Within a few seconds, the workflow should open a PR to apply the automated changes. That PR can then be manually reviewed and merged.

### DIFF
--- a/.github/workflows/run_scripts.yml
+++ b/.github/workflows/run_scripts.yml
@@ -1,0 +1,24 @@
+name: "Run scripts"
+on: workflow_dispatch
+jobs:
+  run_standard_scripts:
+    name: "Run standard scripts and open PR"
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


osse101:
Seems ok as far as I can tell.
Should work as a way to avoid conflicting strat IDs.
A generated PR with no changes shouldn't be much issue, if it were to happen.
Important to run this before new releases.

osse101:
`peter-evans/create-pull-request` has a section on authentication, and this repo has a requirement on signing commits. 🤔

blkerby:
Good catch, the commits won't be signed so the generated PRs wouldn't be mergeable. Should be solvable, will dig into it

PR #1625, opened 2024-09-02
Lava Dive thread the needle entry

blkerby:
This is a more efficient way to enter the lava from left-to-right.

Videos:
- With Varia: https://videos.maprando.com/video/4
- With Gravity: https://videos.maprando.com/video/592
- Suitless: https://videos.maprando.com/video/591

Probably the most significant thing is that with Gravity this puts left-to-right in logic with 1 tank. It could already be done without the new entry (with a few energy to spare), but it wasn't quite in logic because the 5->2 strat had just enough extra lenience in the lava frames that it was making it require 201 energy. So those lava frames also get slightly tightened in this PR.

### DIFF
--- a/region/norfair/east/Lava Dive Room.json
+++ b/region/norfair/east/Lava Dive Room.json
@@ -186,6 +186,30 @@
       ]
     },
     {
+      "link": [1, 5],
+      "name": "Lava Dive Thread the Needle Entry",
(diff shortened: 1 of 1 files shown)


osse101:
I found it to be a little harder with gravity.

PR #1626, opened 2024-09-02
Cathedral temporary blue

blkerby:
Videos:
- left-to-right: https://videos.maprando.com/video/236
- right-to-left: https://videos.maprando.com/video/305
- right-to-left with X-ray: https://videos.maprando.com/video/306

The right-to-left strat is deceptively tricky, took a lot of practice to work out a consistent method. Not completely sure if the canBeVeryPatient requirement is needed, as the strat might not be as hard to learn by watching the video or if the player has a better understanding of the Geruta AI. 

At some point I would like if we had a different tech that we could use for strats like this, since it's technically an abuse of canBeVeryPatient, which is likely to create problems later.

Right-to-left might be possible with X-ray and mid-air Spring Ball jumps instead of HiJump, but I didn't want to go there right now.

### DIFF
--- a/region/norfair/east/Cathedral.json
+++ b/region/norfair/east/Cathedral.json
@@ -290,7 +290,6 @@
       "requires": [
         "HiJump",
         "canLongChainTemporaryBlue",
-        "canBeVeryPatient",
         {"heatFrames": 955},
         {"or": [
           "canXRayCancelShinecharge",
@@ -308,9 +307,6 @@
       ],
       "note": [
         "Carefully planned movement is needed to avoid damage from the Gerutas, and to avoid bonking on the overhangs."
-      ],
-      "devNote": [
-        "The canBeVeryPatient requirement is for difficulty placement."
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


osse101:
We could move `canLongChainTemporaryBlue` to insane imo.  Dropping the chain is the hardest part.
Moving through the geometry doesn't seem that complicated but its  a bit hard to judge when slopes cause me to drop the chain.
There could be options for killing the geruta and sova with supers, or plasma or wave.
I don't think the heat damage will scale well at all and I think makes the strat much harder.

blkerby:
Agree on moving `canLongChainTemporaryBlue` to Insane; seems like a better place for it. And then the `canBeVeryPatient` requirement can be removed.

I looked a bit at killing the Gerutas but for me it didn't seem to make it easier; they are often off-camera when you're on the ground so trying to hit them seemed difficult and likely to mess up the chain. Could be I just didn't figure it out correctly. But if the tech is going to be in Insane anyway then maybe there's nothing more to do here?

PR #1628, opened 2024-09-03
Bombway temp blue

blkerby:
This adds some `leaveWithTemporaryBlue` strats that were missed on the first pass.

Videos:
- left-to-right speedball: https://videos.maprando.com/video/622
- right-to-left speedball: https://videos.maprando.com/video/621
- left-to-right bounce: https://videos.maprando.com/video/343
- right-to-left bounce: https://videos.maprando.com/video/344

### DIFF
--- a/region/crateria/central/Final Missile Bombway.json
+++ b/region/crateria/central/Final Missile Bombway.json
@@ -122,7 +122,8 @@
           "runway": {
             "length": 3,
             "openEnd": 0  
-          }
+          },
+          "maxExtraRunSpeed": "$3.8"
         }
       },
       "requires": [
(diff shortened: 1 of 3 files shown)


osse101 (approved):
The speedball unmorph has a maximum speed. Most of the time it won't be a problem.
 
 I don't understand why the springballing unmorph works, but it is much easier than a pause remorph.

blkerby:
> The speedball unmorph has a maximum speed. Most of the time it won't be a problem.
> 
> I don't understand why the springballing unmorph works, but it is much easier than a pause remorph.

Good catch on the max speed, added that now.

I think part of the reason the spring ball unmorph works is because when you're bouncing in a tunnel you're moving down almost the whole time. I just measured it and it looks like you move up for 1 frame and then down for 7 frames. It makes sense because when you bounce off the ground you get the full Y velocity of a jump, whereas when you bonk the ceiling it zeros out your Y velocity so it puts you in free fall, which takes time to gather speed. So when you exit the tunnel, it's very likely that you're moving downward, giving you a window to unmorph before landing.

PR #1630, opened 2024-09-04
A couple of new Alcatraz strats

blkerby:
Videos:
- Leave With Spark: https://videos.maprando.com/video/379
- Leave With Temporary Blue: https://videos.maprando.com/video/369

The Leave With Spark one comes from insomniaspeed showing us here: https://discord.com/channels/1053421401354285129/1053436236628496454/1272215980483022849.

### DIFF
--- a/region/crateria/central/Parlor and Alcatraz.json
+++ b/region/crateria/central/Parlor and Alcatraz.json
@@ -2496,7 +2496,7 @@
     },
     {
       "link": [8, 5],
-      "name": "Leave With Spark",
+      "name": "Leave With Spark (Top Position)",
       "requires": [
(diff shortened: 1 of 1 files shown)


osse101:
https://videos.maprando.com/video/669 very tight though.
A soft unmorph is going to be faster than the `resetFallSpeed`

blkerby:
> https://videos.maprando.com/video/669 very tight though. A soft unmorph is going to be faster than the `resetFallSpeed`

Interesting, I didn't think there would be enough time to make it close enough to the door for bottom position to work. But yeah it works and seems ok for Extreme. Added a separate strat for it now.

PR #1631, opened 2024-09-04
Construction Zone more temp blue

blkerby:
This adds a few strats that were missed on the first pass, because we didn't know about unmorphing coming out of spring ball bounce through a morph tunnel. Do you think we should add a tech for this?

Videos:
- Bottom to top left: https://videos.maprando.com/video/614
- Bottom to top right: https://videos.maprando.com/video/615
- Top left to bottom: https://videos.maprando.com/video/616
- Top right to bottom: https://videos.maprando.com/video/624

The top-right to bottom is the only one that works without X-Ray.

### DIFF
--- a/region/brinstar/blue/Construction Zone.json
+++ b/region/brinstar/blue/Construction Zone.json
@@ -318,6 +318,30 @@
       "flashSuitChecked": true
     },
     {
+      "link": [1, 3],
+      "name": "Come in Shinecharging, Leave With Temporary Blue",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 1,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canLongChainTemporaryBlue",
+        "canSpringBallBounce",
+        "canXRayTurnaround"
+      ],
+      "exitCondition": {
+        "leaveWithTemporaryBlue": {}
+      },
+      "unlocksDoors": [
+        {
+          "types": ["ammo"],
+          "requires": []
+        }
+      ]
+    },
+    {
       "id": 13,
       "link": [1, 4],
       "name": "Base",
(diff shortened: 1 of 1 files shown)


osse101:
I think bouncing into a morph tunnel is similarly difficult to the unmorph.  So I'd attach it to canTrickySpringballBounce or an advanced temp blue tech.
A speedrun strat is to shoot the 4 rows of shot blocks while falling and then soft morph, so that's what I was expecting to see.

blkerby:
I did a little TAS testing now, and it looks like in general there's a 5.5-pixel window for where to land in order to bounce into a morph tunnel. The main thing is knowing to enter using a neutral bounce, because without that it's much more difficult. The leniency for bouncing into a tunnel seems like it may be enough to be fair game for Very Hard, whether or not using Spring Ball to bounce all the way through (vs. just breaking a bomb block at the entrance). Maybe it could be Expert, though that would leave hardly any temporary blue strats at all in Very Hard so I would worry about that making the difficult curve steeper. Either way, it could be good to have a tech for it (bouncing into a tunnel), in order to better describe how to do it and help make it more visible to players. 

If it's ok, maybe we leave it as "canSpringBallBounce" for now for consistency with other rooms? The strats in this room are already going to be Extreme (or Insane probably soon) because of the canLongChainTemporaryBlue requirement so it's not the limiting factor. We could do a separate PR to introduce the new tech and add it everywhere if that makes sense?

osse101:
Bouncing into morph tunnels feels like a problem right now as part of canTemporaryBlue.  Making it into a tech makes the situation better, imo.  And perhaps it can be refactored again at a later time.

The idea of neutral bouncing as part of a springball bounce sequence isn't well known at Very hard.  Only very specific and high level strats (in vanilla or romhack speedruns) use that, so Expert ish.  Bouncing into holes is on a smaller scale and is seen more commonly, in seeds where you'd want to do this.  In conclusion, the setup needs to be more consistent.

PR #1638, opened 2024-09-06
X-Ray Scope Room shinespark to break bomb block

blkerby:
Video: https://videos.maprando.com/video/646

Based on insomniaspeed showing us here: https://discord.com/channels/1053421401354285129/1053436236628496454/1280911335063949475.

### DIFF
--- a/region/brinstar/red/X-Ray Scope Room.json
+++ b/region/brinstar/red/X-Ray Scope Room.json
@@ -98,7 +98,7 @@
     },
     {
       "link": [1, 2],
-      "name": "Shinespark to Break Bomb Block",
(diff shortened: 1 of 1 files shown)


osse101:
https://videos.maprando.com/video/684
You don't need Hijump, but it makes the jump into the morph tunnel easier.  Perhaps `OR` it with an insane jump.

blkerby:
Oh nice, I had tried for a while but given up. Was able to get it now. Looks like the trick is to start close to the wall and to jump without any dash speed. Added a separate strat for this version, since we can also increase the runway in the entranceCondition for this.

PR #1642, opened 2024-09-10
Relax shortcharge requirements for robot clip

blkerby:
Currently this is an Insane-level notable strat. But the shortcharge is not actually very tight and I believe it makes more sense in Expert. To do it while staying completely inside the robot, it can be done with a 4-tap with relatively long taps, 7 frames each (or more).

### DIFF
--- a/region/wreckedship/main/Wrecked Ship East Super Room.json
+++ b/region/wreckedship/main/Wrecked Ship East Super Room.json
@@ -524,7 +524,7 @@
         "canMidAirMorph",
         "canSlowShortCharge",
         "canKago",
-        {"getBlueSpeed": {"usedTiles": 14, "openEnd": 0}}
+        {"getBlueSpeed": {"usedTiles": 15, "openEnd": 1}}
       ],
       "note": [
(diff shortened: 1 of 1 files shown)


osse101 (approved):
My issue with expert is that the trick just doesn't work sometimes, and clearing the ghost and rekagoing between attempts can be annoying.  But expert players are probably fine just dealing with that.

You can jump or store the spark to be pushed back towards the wall to make it a little more forgiving, if you have a great sense of knowing when you turn blue.

blkerby:
Something that could be worth keeping in mind: At Expert level, the only logical way out from the item is with a CF clip, which requires having Power Bombs. So in this case, even with the robot clip being in logic, the player always has an alternative of farming the Covern for a PB drop if they prefer (in the rare case where it is needed); with the buffed drop rates it's a 16% chance, so not so bad. This won't necessarily be true though if there's ever an option that allows keeping progress while dying/reloading. And of course in Extreme there's the possibility of getting out with R-mode, where you may not have Power Bombs (or Bombs) at all.

Issue #711, opened 2023-04-18
Add underwater wall-jumping tech and strats


blkerby:
Add tech for underwater wall-jumping off a single wall (UWU). This should only include the "easier" variant of this insanely difficult tech: namely, assume it is done with HiJump, and with SpaceJump to break the water if applicable, and only on relatively short climbs (up to ~2 tiles).

osse101:
hjbless and hijump+uwu+breakfree is probably not bad on keyword, or with L+R.

osse101:
Added in #887 

For the next level of difficulty we're looking at 
-HiJump UwU  + Break Free
-Slightly longer HiJump Uwu.  I'm not sure where exactly, other than the really long ones.
-Short HiJumpless.  I think Bob does this to get up fish tank top left door.

kjbranch:
Long HiJump Uwu was added in #1485

kjbranch:
This is how bobbob categorized the difficulty of underwater wall jumps (https://discord.com/channels/445948207638511616/534430696635367444/1088857393892442123).
We currently have level 0-3. Do we want to add any more of these? If not, should we close the issue? Alternatively, we could add the tech for them and put them in `Beyond` on Map Rando, with or without strats added to them, so it is clear they are possible but not required.
```
in a logic that i wrote that had underwater walljumping in it, it was split into different "levels":
Level 0:  In a 1-tile wide area (e.g., escaping Watering Hole). This is probably not needed, i just added it since i could. I guess foreseeing complaints of "but i had underwater walljumping turned off".
Level 1:  In a 2-tile wide area with Hi-Jump (e.g., going up Crab Shaft).
Level 2:  In a 2-tile wide area with Hi-Jump, breaking the water surface (e.g., West Sand Hole).
Level 3:  One wall, with Hi-Jump (this is the above).
Level 4:  One wall, with Hi-Jump, breaking the water surface (e.g., doing the above without Space Jump).
Level 5:  In a 2-tile wide area, without Hi-Jump.
Level 6:  One wall, without Hi-Jump. I think I was split on separating this into right/left walls:  without L+R, I don't really think it's that different. But with L+R, right walls are much easier.
Level 7:  One wall, without Hi-Jump, breaking the water surface (only on right walls). This is L+R.
[...]

PR #1643, opened 2024-09-10
Underwater Wall Jump Break Free

kjbranch:
I never actually tested any of these, other than seeing how close I could jump to the water line (they were all reasonable, within ~1 tile).

There are ~~three~~ *two* tidal ones, which I would think would be easier, but I'm not sure: ~~West Cac Alley,~~ Pants, Grapple Tutorial 3

~~I am especially unsure about West Cac, as the wall doesnt go above the water line.~~ Also unsure on the lenience needed in East Cac above the spikes.

~~I could also ask Eddie to test the tidal difficulty and West Cac specifically if you think that would be worthwhile.~~

Edit: @blkerby tested it with tidal water in Pants Room, also let me know that West Cac can be done without Break Free, and is already in logic.

### DIFF
--- a/region/maridia/inner-green/Pants Room.json
+++ b/region/maridia/inner-green/Pants Room.json
@@ -752,6 +752,13 @@
       ]
     },
     {
+      "link": [5, 2],
+      "name": "Underwater Walljump Break Free",
+      "requires": [
+        "canUnderwaterWalljumpBreakFree"
+      ]
+    },
+    {
       "id": 35,
       "link": [5, 4],
       "name": "Base",
--- a/region/maridia/inner-pink/Colosseum.json
+++ b/region/maridia/inner-pink/Colosseum.json
@@ -1207,6 +1207,14 @@
       ]
     },
     {
+      "link": [2, 3],
+      "name": "Underwater Walljump Break Free",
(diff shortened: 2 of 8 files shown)


kjbranch:
Adds `Level 4` underwater wall jumps, #711

blkerby:
Side note: I found the right wall of East Cac Alley is a convenient place to practice the break-free movement on its own. In this place the water line is just the right height where you don't have to do any regular underwater walljumping first. So far I can only manage it at 50% slowdown or lower. The tidal water in Pants Room does seem to help some; you time it where you break free while the tide is going down, so it helps you out a bit, though it seems like probably not enough difference to affect difficulty placement.

osse101 (approved):
I tried east cac alley by the door for a while and couldn't get it.  That and halfie climb were considered for logic years ago.  I'd call them easier than the rest, but I'm not sure by how much.  East cac from the door in a productive way (over the spikes) may be the hardest.

There may be a handful of people who can do a `canUnderwaterWalljumpBreakFree` given enough time.  Playing on Keyboard is going to help alot, for instant turnaround keypresses.  To me its beyond tech until I see more people who are capable of doing it.

PR #1644, opened 2024-09-10
Bug Sand Hole temp blue

blkerby:
Videos:
- left-to-right: https://videos.maprando.com/video/699
- right-to-left: https://videos.maprando.com/video/700


osse101:
https://videos.maprando.com/video/708 suitless way

osse101:
https://videos.maprando.com/video/746 you can do the reverse as well
and there is a max run speed as well.  On all these strats.

PR #1659, opened 2024-09-17
Green Bubbles temp blue

blkerby:
Videos:
- left-to-right, with X-Ray: https://videos.maprando.com/video/817
- left-to-right, without X-Ray: https://videos.maprando.com/video/818
- right-to-left, with X-Ray: https://videos.maprando.com/video/816
- right-to-left, without X-Ray: https://videos.maprando.com/video/815

### DIFF
--- a/region/norfair/east/Green Bubbles Missile Room.json
+++ b/region/norfair/east/Green Bubbles Missile Room.json
@@ -144,6 +144,37 @@
       ]
     },
     {
+      "link": [1, 2],
+      "name": "Come in Shinecharging, Leave With Temporary Blue",
+      "entranceCondition": {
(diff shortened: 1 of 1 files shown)


osse101 (approved):
Did you look into raising the shutter tower to bounce into the morph tunnel?
Most of the time having the extra tile would help, but here I found it harder to get a neutral bounce since you are close to the ceiling.

blkerby:
> Did you look into raising the shutter tower to bounce into the morph tunnel? Most of the time having the extra tile would help, but here I found it harder to get a neutral bounce since you are close to the ceiling.

I didn't think of that. But yeah, makes sense that there might not be enough vertical space for it to help here.

PR #1674, opened 2024-09-26
Pink Maridia (Final): Use Flash Suit

kjbranch:


region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -2117,6 +2162,18 @@
...
+      "link": [3, 2],
+      "name": "Use Flash Suit",
+      "requires": [
+        "Morph",
+        "canUseEnemies",
+        {"useFlashSuit": {}},
+        {"shinespark": {"frames": 18, "excessFrames": 11}}
+      ],
+      "flashSuitChecked": true,
+      "note": "Go under the Oums and spark horizontally through the speed blocks, then wait for the Oums and use them as a platform to jump up the last ledge."
```
osse101, line 2175: Baiting an oum down to climb up with, and then hitbox clipping the rest is fairly easy to do.  Or you could bring all 3 down. Then its a 4 energy spark.

region/maridia/inner-pink/The Precious Room.json
```diff
@@ -494,6 +494,57 @@
...
+      "link": [2, 1],
+      "name": "Long Underwater Wall Jump, Use Flash Suit",
+      "requires": [
+        "canTrickyCarryFlashSuit",
+        "canUnderwaterWalljump",
+        {"useFlashSuit": {}},
+        {"shinespark": {"frames": 12, "excessFrames": 8}}
+      ],
+      "flashSuitChecked": true,
+      "note": "Underwater wall jump to the top middle platform before sparking to save Energy."
```
osse101, line 546: About as hard as notable extreme uwjs.  The risk of using the flash suit is in the case where you break spin and fail the uwj.  Interestingly map rando can assume usage of spin lock, but here we assume using controllers and no spin lock.  Fine for base insane.

region/maridia/inner-pink/West Sand Hole.json
```diff
@@ -360,6 +360,58 @@
...
+        {"or": [
+          {"shinespark": {"frames": 10, "excessFrames": 3}},
+          {"and": [
+            "HiJump",
+            {"shinespark": {"frames": 7, "excessFrames": 3}}
+          ]}
+        ]}
+      ],
+      "flashSuitChecked": true,
+      "devNote": "FIXME: Is it possible to get onto the ledge with a sandfall bounce?"
```
osse101, line 382: Looks like you get close but cannot make it up.

region/maridia/inner-pink/West Sand Hole.json
```diff
@@ -360,6 +360,58 @@
...
+      "name": "Hero Shot Break Block, Use Flash Suit",
+      "requires": [
+        "canPlayInSand",
+        "canHeroShot",
+        {"useFlashSuit": {}},
+        {"shinespark": {"frames": 27, "excessFrames": 5}}
+      ],
+      "flashSuitChecked": true,
+      "note": [
+        "From the bottom of the sand, shoot upwards to break the block then Shinespark up. To use a flash suit in deep sand:",
```
osse101, line 410: You can jump from the sand after shooting to save some health.  Does make it easier to fail, or just sink into the sand.

region/maridia/inner-pink/The Precious Room.json
kjbranch: I was thinking that this being in insane would be ok, but do you think it should also be notable? I could see players wanting all long uwu to be toggleable

osse101: I generally can't do the notable uwjs that are in extreme without spinlock, but would be fine doing tricky flash suit movements.  So the ability to toggle off this strat makes sense.  
Since there is a set of uwjs that are similarly hard, we could look at making them a tech.  I would look at moving current uwjs down to canShortUnderwaterWalljump, since you only need to do it a few times to gain enough height.

PR #1677, opened 2024-09-30
Add canStaggeredWalljump reqs to Bubble Mountain runway walljumps

blkerby:
The strats with 4+ tile runway are in Medium, which doesn't really seem fair. I think you want to do 3 walljumps here, like in this video: https://videos.maprando.com/video/983. The version with 2 walljumps is more difficult, requiring more precise timing on the initial jump and wall jumps, and risks getting knocked down by a Cac spike. With 3 walljumps there's more leniency and it leaves room to adjust the walljumps to avoid Cac spikes. By adding a canConsecutiveWalljump it will put these into Hard, except for the shorter-runway version which is already Very Hard because of canTrickyJump.

### DIFF
--- a/region/norfair/east/Bubble Mountain.json
+++ b/region/norfair/east/Bubble Mountain.json
@@ -1492,7 +1492,7 @@
       },
       "requires": [
         "canPreciseWalljump",
-        "canConsecutiveWalljump",
+        "canStaggeredWalljump",
         "canTrickyJump"
       ]
     },
@@ -1507,7 +1507,7 @@
         }
       },
       "requires": [
-        "canConsecutiveWalljump",
+        "canStaggeredWalljump",
         "canCarefulJump"
       ]
     },
@@ -1522,7 +1522,7 @@
         }
       },
       "requires": [
-        "canConsecutiveWalljump",
+        "canStaggeredWalljump",
         "canCarefulJump"
       ]
     },
--- a/region/norfair/east/Bubble Mountain.json
+++ b/region/norfair/east/Bubble Mountain.json
@@ -1502,7 +1502,7 @@
             "canWallJumpInstantMorph",
             {"enemyDamage": {
               "enemy": "Cacatac",
-              "type": "contact",
+              "type": "spike",
               "hits": 1
             }}
           ]}
@@ -1530,7 +1530,7 @@
             "canWallJumpInstantMorph",
             {"enemyDamage": {
               "enemy": "Cacatac",
-              "type": "contact",
+              "type": "spike",
               "hits": 1
             }}
           ]}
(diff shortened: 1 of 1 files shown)


kjbranch (dismissed):
This looks fine, though `canConsecutiveWalljump` is in Medium

blkerby:
Ah, I forgot about that. `canStaggeredWalljump` might be a better fit

kjbranch:
I think `canConsecutiveWalljump` should probably still be included. A bit more on the fence about `canStaggeredWalljump`, I think this will be one of the easiest uses for it, but I think this is probably good in `Hard`. We will see what Osse thinks.

blkerby:
> I think `canConsecutiveWalljump` should probably still be included.

It's implicitly included since `canStaggeredWalljump` is an extension tech of `canConsecutiveWalljump`

osse101:
I think staggered walljumps describe avoiding king cac.  It doesn't always puff up to shoot spikes, but you can expect it to.

If you want to put it in hard, then there should be more itemset options added too.  Screw attack can jump right up.  Pseudo screw has more complicated inputs.  canwalljumpInstantMorph is the goto easy way to avoid falling back down.  Candodgewhileshooting is close to helping but probably isn't good enough.

The in-room strats can kill the enemies and more leisurely climb the mountain.

region/norfair/east/Bubble Mountain.json
```diff
@@ -1491,8 +1491,22 @@
...
       },
       "requires": [
+        "canTrickyJump",
+        "canConsecutiveWalljump",
         "canPreciseWalljump",
-        "canTrickyJump"
+        {"or": [
+          "canStaggeredWalljump",
+          "ScrewAttack",
+          "canPseudoScrew",
```
kjbranch, line 1500:
```suggestion
          "canWalljumpWithCharge",
```

region/norfair/east/Bubble Mountain.json
```diff
@@ -1506,8 +1520,21 @@
...
       },
       "requires": [
-        "canWalljump",
-        "canCarefulJump"
+        "canCarefulJump",
+        "canConsecutiveWalljump",
+        {"or": [
+          "canStaggeredWalljump",
+          "ScrewAttack",
+          "canPseudoScrew",
```
kjbranch, line 1528:
```suggestion
          "canWalljumpWithCharge",
```

region/norfair/east/Bubble Mountain.json
```diff
@@ -1521,8 +1548,21 @@
...
```
kjbranch, line 1556:
```suggestion
          "canWalljumpWithCharge",
```

kjbranch (approved):

PR #1679, opened 2024-10-01
Add Morph reqs to reserve double damage boosts

blkerby:
This trick seems much more difficult without Morph. I'm not sure if "canOffScreenMovement" captures the difficulty but wanted to at least put something there; it might belong more in Insane though? At least, my reaction time is not quick enough to do the forward input after hearing the hurt sound, so it instead has to be a timing based on knowing how long the reserve refill will last. It could also be my audio latency is not great with my emulator setup, so maybe for others it might not be as bad.

I tried the Cathedral Entrance one morphless too and couldn't get it at all; I think I don't understand how it's supposed to work, a video could probably help. It already had a note about it so I assume it's ok though

### DIFF
--- a/region/maridia/outer/Main Street.json
+++ b/region/maridia/outer/Main Street.json
@@ -3386,10 +3386,13 @@
         "canCrouchJump",
         "canInsaneJump",
         "canReserveDoubleDamageBoost",
-        "Morph",
         {"or": [
           "h_canMaxHeightSpringBallJump",
           "HiJump"
+        ]},
+        {"or": [
+          "Morph",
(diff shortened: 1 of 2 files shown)


osse101:
I was using morph when I did these with hijump.  I'm not sure whether morphless looked doable.  
The precathedral double boost is much harder without morph.  I didn't put much time into trying to get better at it.
A video is going to be mostly obscured by the pause fade.  You should still need to time the pause so that you get good height gain and then be paused at max height and inside the enemy.  Not being underwater would affect how fast you start falling again, so everything may be tighter.  But it should be in insane.

PR #1680, opened 2024-10-01
Glass Tunnel, Event Node Removal, Overhaul

kjbranch:
Primarily, this removes node 7, Moves node 6 to the top platform, removes "A".
It also cleans up strats with missing door unlocks, the room diagram, and various small things.


region/maridia/outer/Glass Tunnel.json
```diff
@@ -434,10 +422,23 @@
         "canCrossRoomJumpIntoWater",
         "canTrickyJump"
       ],
+      "unlocksDoors": [
+        {
+          "types": ["super"],
+          "requires": []
+        },
+        {
+          "types": ["missiles", "powerbomb"],
```
blkerby, line 431: Something to keep in mind is that there are implicit 4 -> 4 strats for unlocking the door in a standard way and going through. I think the logic is potentially unsound since in fact there's no place to stand at node 4 where this could be done. Resetting the room is probably a workaround in most cases, but that could require additional heat frames which wouldn't be accounted for. Possible solutions:
- Add a requirement `{"doorUnlockedAtNode": 4}` to force the door to be unlocked as part of this start.
- Add a `leaveNormally` exitCondition which forces the strat to exit the room through node 4 as part of this strat, which in turns also requires the door to be unlocked.
- Add `"useImplicitDoorUnlocks": false` to the node properties of node 4, to get rid of the implicit door unlock strats at node 4, since they would be likely to be unsound in general. This would have to be combined with adding `unlocksDoors` to every strat that ends at node 4. It's some work but probably the safest way to make sure the logic is completely correct here.

PR #1682, opened 2024-10-02
Fix strat ID uniqueness check

blkerby:


blkerby:
This is a simple fix to a messed-up ID, also fixing the test which allowed this to slip through. Going to go ahead and merge it now (without review) because in its current state the automation that syncs the logic data to the video site is broken, and I need it working again in order to update the Map Rando dev site, which I'm working on now.

PR #1683, opened 2024-10-04
Misc Fixes

kjbranch:
Found some various things that I thought should be changed, while making some strat videos


region/maridia/inner-yellow/Northwest Maridia Bug Room.json
```diff
@@ -517,26 +517,71 @@
...
+          {"and": [
+            "Ice",
+            "Wave",
+            "Spazer"
+          ]},
+          {"enemyDamage": {
+            "enemy": "Menu",
+            "type": "contact",
+            "hits": 2
+          }}
```
osse101, line 539: Straight up, spacejump should be considered the same as walljump and added to the tricky avoid and menu kill strats.  Its not clear if the 2 hits are part of avoiding or killing, but it can easily turn into more either way.

You can build your height with spacejump before activating the Menus to safely make it up, and I'd put it at hard notable since I haven't seen anyone do it yet and it does require an understanding of spacejump jumps.

kjbranch: On lower difficulties, with space jump alone, tanking the Menus doesnt seem reasonable at all, so I updated the number of hits and description to specify it is for killing them.

I moved the direct space/hijump to hard. Do you think it needs to be notable?

kjbranch:
I added a last minute update to Pants Room diagram as well to mention that there is an East Pants Room. I had been meaning to do that for a while

region/maridia/inner-yellow/Northwest Maridia Bug Room.json
osse101: At hard they can get by  and look up the intended way after, at worst.
The jump with hijump is completely fine.

PR #1684, opened 2024-10-04
Refine Escape Room 4 walljumpless Space Jump

blkerby:
This splits off "Reverse Tourian Escape Room 4 Dodge While Climbing" into a second notable, "Reverse Tourian Escape Room 4 Bootless Walljumpless Space Jump", with several strat variants.

Videos:
- Longer Acid Dip: https://videos.maprando.com/video/1080
- Short Acid Dip: https://videos.maprando.com/video/1078
- Plasma: https://videos.maprando.com/video/1081
- Super: https://videos.maprando.com/video/1084
- Left-Side Weave: https://videos.maprando.com/video/1076

Details:
- The "Longer Acid Dip" is only a moderately fast climb, using full-height Space Jumps; this would go in Very Hard (via `canTrickyJump` and the notable).
- The "Short Acid Dip" climbs faster by doing shorter, tighter jumps; this would go in Expert (via `canPreciseSpaceJump`).
- The "Plasma" version is also a fast climb, but uses Plasma to shoot open the door while landing directly in the doorway, avoiding any acid damage. This would also go in Expert, except if using Plasma to also hitbox through the Pirate, which would go in Extreme because of the precision of the timing (since Samus won't have a lot of fall speed here and Pirates have a tall hitbox).
- The "Super" version is similar but uses a Super to kill the last Pirate. This is tighter than "Plasma" (without hitboxing) but seems probably still ok in Expert.
- The "Left-Side Weave" is the most difficult version, falling down the left side and relying on a carefully-timed shot to open the door (just after passing low enough that the shot won't be blocked by the bottom Pirate) and then down-back into it. Failure to get into the door will result in falling deep into the acid. This would go into Extreme (via `canInsaneJump`).
- A few other adjustments are made to "Reverse Tourian Escape Room 4 Dodge While Climbing": some of the cases were redundant, being already covered by the "Protected Shaft Climb" strats, so I removed these. Also it's now requiring `canCarefulJump` in the HiJump + SpaceJump case since I don't think that's fair for Basic (without Screw Attack).


region/tourian/main/Tourian Escape Room 4.json
```diff
@@ -567,6 +551,110 @@
...
       ]
     },
+    {
+      "link": [3, 1],
+      "name": "Reverse Tourian Escape Room 4 Bootless Walljumpless Space Jump (Longer Acid Dip)",
+      "requires": [
+        {"notable": "Reverse Tourian Escape Room 4 Bootless Walljumpless Space Jump"},
+        "SpaceJump",
+        "canTrickyJump",
+        {"acidFrames": 90}
```
kjbranch, line 561: should these have a leave normally exit condition, so the air runway cant be used?

region/tourian/main/Tourian Escape Room 4.json
```diff
@@ -567,6 +551,110 @@
...
+        "Hold left to land directly in the doorway.",
+        "If done quickly enough, it is possible to avoid any acid damage."
+      ]
+    },
+    {
+      "link": [3, 1],
+      "name": "Reverse Tourian Escape Room 4 Bootless Walljumpless Space Jump (Super)",
+      "requires": [
+        {"notable": "Reverse Tourian Escape Room 4 Bootless Walljumpless Space Jump"},
+        "canPreciseSpaceJump",
```
kjbranch, line 618:
```suggestion
        "canPreciseSpaceJump",
        "canTrickyJump",
        "canAutoCancelWeapon",
```
Should precise space jump just require tricky jump?

kjbranch (dismissed):

region/tourian/main/Tourian Escape Room 4.json
blkerby: Good catch. Door unlocks may also be a problem for some of these, will take another look

region/tourian/main/Tourian Escape Room 4.json
blkerby: I think of `canPreciseSpaceJump` as a higher movement tech that kind of requires tricky jumps by definition, so yeah it would probably make sense for it to have canTrickyJump as a dependency yeah; then redundant `canTrickyJump` requirements could be removed from the other strats.

The auto cancel is not required here; it helps simplify the sequence of inputs, but I think it would still be ok in this difficulty without it, so doesn't need to be a logical requirement

osse101 (commented):
- I prefer baiting the right pirates to shoot right since that is safest and works much better for walljump climb.

- Manipulating the middle pirates off the middle wall helps to avoid hits on the decent.  At least on hard we should assume you didn't manipulate the pirates so that you can safely fall down left side.  And you don't have time to wait for them.  Now maybe I'll stop getting hit everytime I do the room.

- Moonfalling right side left shaft is a little faster but may still not line up well with the acid level, and is trickier to control.

- The left side weave can be done by falling straight down on a safe vertical line.  That's more like expert, but the spacejump climb itself is still kinda bad.  I'm pretty sure you can morph-unmorph to land on the door too, instead of taking pirate damage.

- Powerbombing the door open shouldn't cost so many acidframes, unless the acid level was already high I suppose.  You can lay it pretty high.  You need to know it is a PB door in advance though.  Kinda hard to judge how much extra time to put on it.

blkerby:
> I prefer baiting the right pirates to shoot right since that is safest and works much better for walljump climb.

It may be useful for the walljump climb but I didn't find it to help for the Space Jump strats. If you are going fast enough, I found the Pirates never will be able to shoot you with the Space Jump strats, so it didn't seem to be a problem. And if you are going too slow, then they will still shoot you anyway even you baited them to face right; going that slow, the acid will rise much further than the `acidFrames` account for anyway. The other problem is that making the bottom Pirate face right means you're landing further to the right, giving the acid a head start to rise further, which is not good here. Could be I'm misunderstanding something though. 

> Manipulating the middle pirates off the middle wall helps to avoid hits on the decent.  At least on hard we should assume you didn't manipulate the pirates so that you can safely fall down left side.  And you don't have time to wait for them.  Now maybe I'll stop getting hit everytime I do the room.

I'm not sure I understand. None of the variants of this strat seem reasonable for Hard as they all need at least `canTrickyJump` to avoid the Pirate shots? I guess if the Pirates could be killed beforehand it could maybe fit in Hard; but knowing to kill them is also kind of tricky. Maybe it just fits better in Very Hard and higher?

> Moonfalling right side left shaft is a little faster but may still not line up well with the acid level, and is trickier to control.

Yeah I tried that out but didn't find much success.

> The left side weave can be done by falling straight down on a safe vertical line.  That's more like expert, but the spacejump climb itself is still kinda bad.  I'm pretty sure you can morph-unmorph to land on the door too, instead of taking pirate damage.

When I tried before, I found the weave easier than falling straight down, but trying again today falling straight down does seem easier :shrug:  Adjusted the note to be more generic instead of saying to do it one way or the other. Doing a reset fall speed after hitting the bottom Pirate helps a ton, enough to bring it down from Extreme to Expert I think.  Avoiding the bottom Pirate (taking no damage) with reset fall speed also does work and feels like Extreme difficulty. Both seem worth modeling; added cases for them now.

> Powerbombing the door open shouldn't cost so many acidframes, unless the acid level was already high I suppose.  You can lay it pretty high.  You need to know it is a PB door in advance though.  Kinda hard to judge how much extra time to put on it.

Yeah, I thought the same thing but without being able to test it's hard to know exactly what to put, so I just used the standard 110 frames with a FIXME to refine it later.

blkerby:
Additional videos:
- Reset fall speed (with Pirate damage): https://videos.maprando.com/video/1099
- Damageless reset fall speed:  https://videos.maprando.com/video/1100

PR #1685, opened 2024-10-07
Post Crocomire Jump Room - fix missing obstaclesNotCleared id 28

dragonsinth:
https://dev.maprando.com/seed/kzDPKrrYC/data/visualizer/index.html

Step 7 / Grav in PCJR:

Strat: Break Speed Blocks, Broken Power Bomb Blocks
https://dev.maprando.com/logic/room/127/2/2/23

Strat: Leave with Shinecharge - Power Bomb Blocks Intact
https://dev.maprando.com/logic/room/127/2/2/28

I think the second strat is missing an `"obstaclesNotCleared": ["B"]`?

### DIFF
--- a/region/norfair/crocomire/Post Crocomire Jump Room.json
+++ b/region/norfair/crocomire/Post Crocomire Jump Room.json
@@ -602,6 +602,7 @@
       "name": "Leave with Shinecharge - Power Bomb Blocks Intact",
       "requires": [
         {"obstaclesCleared": ["B"]},
+        {"obstaclesNotCleared": ["A"]},
         "canShinechargeMovement",
         {"canShineCharge": {
           "usedTiles": 32,


blkerby:
The change looks good to me. I think we used to be less consistent about including requirements like this; at the time this logic was originally written, `obstaclesNotCleared` might not have even existed as a possible requirement type. If the PB blocks are destroyed, then Samus can leave shinecharged with even more frames remaining, so the way it's written now shouldn't cause any unsoundness in the logic. Still it's confusing for the spoiler to show that the blocks are destroyed and then later that they are intact. Given that we can now express an `obstaclesNotCleared` requirement, it's more accurate for it to be there. The alternative would be for the strat to be renamed to say something like "Power Bomb Blocks Possibly Intact"; but I think I like `obstaclesNotCleared` more

PR #1689, opened 2024-10-08
Adjust Ice Zeb Skip 

blkerby:
This makes a few adjustments to the Ice Zebetite Skip strat:

- Adds an easier (Hard-level) variant of the strat that assumes Ice+Wave, using a Rinka to get i-frames.
- Adjusts the strat to indicate that the `canDownBack` requirement is an alternative to disabling Screw Attack. 
- Assumes that the random turret hit could be avoidable at an Expert level (with `canTrickyDodgeEnemies`). 

For disabling Screw Attack, I added a new logical requirement type `disableEquipment` which allows us to specify exactly which item must be disabled. The idea is hopefully we could eventually migrate all uses of `canDisableEquipment` to using this requirement. Specifying the item can make the logic significantly clearer to the reader, and in some contexts the distinction could be logically significant, not currently in Map Rando, but potentially in other settings, e.g. if the equipment screen is left unpatched (where disabling certain items may not always be possible, depending on what other items are collected) or if the equipment screen is disabled.

The earlier version of the strat combined a Rinka hit with the `canDownBack` variant. I assume this may have referring to the speedrun version of the down-back, where you get i-frames from the top Rinka and then freeze the bottom Rinka from above while down-backing, landing/clipping onto it immediately after it is frozen. For logic purposes, I think that version is trumped by the slower, damageless version that freezes the Rinka from below, gets i-frames from the Zebetite, and and then does a short jump-morph-unmorph to down-back, which to me seems much easier (I couldn't get the other version to work at all).

With the RNG, I had trouble dodging the turret shots consistently, but from Arcade I remember Eddie doing damageless Mother Brain Rooms regularly so I assume with practice it must be possible at some level?

Videos:
- Ice+Wave: https://videos.maprando.com/video/1094
- Ice: https://videos.maprando.com/video/1092
- Ice+Screw Attack: https://videos.maprando.com/video/1093

### DIFF
--- a/region/tourian/main/Mother Brain Room.json
+++ b/region/tourian/main/Mother Brain Room.json
@@ -397,32 +397,44 @@
         "canTrickyUseFrozenEnemies",
         "Morph",
         {"or": [
-          "canTrickyJump",
           {"and": [
-            "canDownBack",
+            "Wave",
             {"enemyDamage": {
               "enemy": "Rinka",
               "type": "contact",
               "hits": 1
+            }},
+            {"enemyDamage": {
+              "enemy": "Mother Brain 1",
+              "type": "turret",
+              "hits": 1
             }}
+          ]},
+          {"and": [
+            "canTrickyJump",
+            {"or": [
+              {"disableEquipment": "ScrewAttack"},
+              "canDownBack"
+            ]}
           ]}
-        ]},
-        {"enemyDamage": {
-          "enemy": "Mother Brain 1",
-          "type": "turret",
-          "hits": 1
-        }}
+        ]}
       ],
       "note": [
         "Glitch through the Mother Brain Zebetites by using a frozen Rinka and i-frames.",
-        "Freeze the Rinka at its spawn location, then spinjump or Down-Back onto it after acquiring i-frames to clip inside of the Zebetite, then jump through.",
+        "Freeze the Rinka at its spawn location, then spin jump or down-back onto it after acquiring i-frames to clip inside of the Zebetite, then jump through.",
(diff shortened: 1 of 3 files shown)


osse101:
The turret shot cycle is going to be easier if the zebs are already cleared like in arcade.  Speedruns are going to do zeb skip with the same speed every time so know where to expect the turret to shoot, and they preshoot where it will go.  If you aren't trying to save seconds, you can watch the turrets tho and wait.  

I prefer the crouch shot without wave because it works with or without plasma.

region/tourian/main/Mother Brain Room.json
```diff
@@ -397,32 +397,47 @@
...
               "type": "contact",
               "hits": 1
             }}
+          ]},
+          {"and": [
+            "canTrickyJump",
+            {"or": [
+              {"disableEquipment": "ScrewAttack"},
+              "canDownBack"
+            ]}
```
osse101, line 413: This part applies to the wave strat and should be pulled up a level.

It seems like you are going for `OR:[wave, trickyjump]` +`OR:[Rinka hit, trickyJump]`

osse101 (commented):
I like `disableEquipment` to better describe what item must be off.  My first reaction  was that a shorter name could be used, `itemOff` | `disabled`, but disabledEquipement is clearest.
It just may imply that every strat has been tested with every item set to make sure they work now.

region/tourian/main/Mother Brain Room.json
blkerby: Good catch, yeah that's what I was going for. Should be fixed now

PR #1690, opened 2024-10-08
Norfair Reserve Tank Room -- define Sovas as Obstacle B, strat 32 "Return Carefully (Sovas Cleared)"

dragonsinth:
https://videos.maprando.com/video/1097

### DIFF
--- a/region/norfair/east/Norfair Reserve Tank Room.json
+++ b/region/norfair/east/Norfair Reserve Tank Room.json
@@ -583,7 +583,7 @@
       "requires": [
         {"obstaclesCleared": ["B"]},
         {"obstaclesNotCleared": ["A"]},
-        {"heatFrames": 570}
+        {"heatFrames": 480}
       ],
       "note": "Return carefully avoiding Dragons."
     },
--- a/region/norfair/east/Norfair Reserve Tank Room.json
+++ b/region/norfair/east/Norfair Reserve Tank Room.json
@@ -577,6 +577,17 @@
       "devNote": "Ice can move through the room more quickly but must spend time refreezing the dragons."
     },
     {
+      "id": 32,
+      "link": [4, 1],
+      "name": "Return Carefully (Sovas Cleared)",
+      "requires": [
+        {"obstaclesCleared": ["B"]},
+        {"obstaclesNotCleared": ["A"]},
+        {"heatFrames": 570}
+      ],
+      "note": "Return carefully avoiding Dragons."
+    },
+    {
       "id": 30,
       "link": [4, 2],
       "name": "Base",
@@ -589,17 +600,6 @@
       "link": [4, 3],
       "name": "Base",
       "requires": []
-    },
-    {
-      "id": 32,
-      "link": [4, 1],
-      "name": "Return Carefully (Sovas Cleared)",
-      "requires": [
-        {"obstaclesCleared": ["B"]},
-        {"obstaclesNotCleared": ["A"]},
-        {"heatFrames": 570}
-      ],
-      "note": "Return carefully avoiding Dragons."
     }
   ],
   "nextStratId": 33,
--- a/region/norfair/east/Norfair Reserve Tank Room.json
+++ b/region/norfair/east/Norfair Reserve Tank Room.json
@@ -78,11 +78,6 @@
       "id": "A",
       "name": "Lava Dragons",
       "obstacleType": "enemies"
-    },
-    {
-      "id": "B",
-      "name": "Sovas",
-      "obstacleType": "enemies"
     }
   ],
   "links": [
@@ -333,13 +328,12 @@
     {
       "id": 15,
       "link": [1, 4],
-      "name": "Avoid Dragon Fireballs, Kill the Sovas",
+      "name": "Avoid Dragon Fireballs",
       "requires": [
         "canCameraManip",
         "canDodgeWhileShooting",
         {"heatFrames": 840}
       ],
-      "clearsObstacles": ["B"],
       "note": [
         "Clear the Sovas while standing in a safe spot from the Dragon Fireballs.",
         "Bring each Dragon on screen one at a time so that there are fewer projectiles to avoid."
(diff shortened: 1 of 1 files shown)


dragonsinth:
Note, will need to conflict resolve w/ #1691

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -583,9 +589,20 @@
...
+    {
+      "id": 32,
+      "link": [4, 1],
+      "name": "Return Carefully (Sovas Cleared)",
+      "requires": [
+        {"obstaclesCleared": ["B"]},
+        {"obstaclesNotCleared": ["A"]},
+        {"heatFrames": 570}
+      ],
+      "note": "Return carefully avoiding Dragons.",
```
dragonsinth, line 602:
```suggestion
      "note": "Return carefully avoiding Dragons."
```

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -570,6 +576,17 @@
...
       "devNote": "Ice can move through the room more quickly but must spend time refreezing the dragons."
     },
+    {
+      "id": 32,
+      "link": [4, 1],
+      "name": "Return Carefully (Sovas Cleared)",
+      "requires": [
+        {"obstaclesCleared": ["B"]},
+        {"obstaclesNotCleared": ["A"]},
+        {"heatFrames": 570}
```
dragonsinth, line 586: copy-pasted frame count from the "kill" strat.. not sure what the right number is

dragonsinth:
```suggestion
        {"heatFrames": 480}
```

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -78,6 +78,11 @@
       "id": "A",
       "name": "Lava Dragons",
       "obstacleType": "enemies"
+    },
+    {
+      "id": "B",
+      "name": "Sovas",
+      "obstacleType": "enemies"
```
dragonsinth, line 85: There are a few more things we probably need to fix up if this moves forwards... a couple of strats that will kill these guys, and we need to mark strats like `Frozen Sova Platform` (27) with `obstaclesNotCleared: B`.  Unsure whether the `canTrickyUseFrozenEnemies` in strat 24 is with the Sova or with the Dragon.

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -328,11 +333,13 @@
     {
       "id": 15,
       "link": [1, 4],
-      "name": "Avoid Dragon Fireballs",
+      "name": "Avoid Dragon Fireballs, Kill the Sovas",
       "requires": [
+        "canCameraManip",
```
dragonsinth, line 338: per @osse101

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -570,6 +577,16 @@
       "clearsObstacles": ["A"],
       "devNote": "Ice can move through the room more quickly but must spend time refreezing the dragons."
     },
+    {
+      "link": [4, 1],
+      "name": "Return Carefully (Sovas Cleared)",
+      "requires": [
+        {"obstaclesCleared": ["B"]},
+        {"obstaclesNotCleared": ["A"]},
```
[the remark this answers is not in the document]
dragonsinth: Ahh true, I had that thought too.

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -570,6 +577,16 @@
...
+        {"heatFrames": 480}
```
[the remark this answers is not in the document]
dragonsinth: 👍 cool.  The main purpose here is to provide _some_ return strat that doesn't require specific equipment.  The base strat needs one of

- VH canTrickyJump
- Basic Space
- Med Screw + canCarefulJump
- M dragons dead + canCarefulJump

Maybe this new strat should also include canCarefulJump to put it in Medium?

Actually.. now that I talk this through, this new strat could actually absorb a lot of what's currently in Base.  Base is kind of a mess right now if you really look through it.  Imagine that removed everything from Base that requires "canCarefulJump", which basically just leaves Space Jump as the base strat.  And then merged everything else here:

```json
{
  "canCarefulJump"
  "or": [
    "canTrickyJump",
    "ScrewAttack",
    {
        "obstaclesCleared": [
          "A", "B"
        ]
    },
    {
        "obstaclesCleared": [
          "B"
        ],
        "heatFrames": 270
    }
}
{
  "heatFrames": 210
}
```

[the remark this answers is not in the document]
dragonsinth: maybe lavaProof should also be a base strat option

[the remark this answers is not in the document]
osse101: Lavaproof options are in [3, 1]

osse101:
- I'm not seeing a situation where its important to track sovas as an obstacle.  
-- Most of the time they are cleared when first passing through the room or ignored.
-- Not killing them would be with ice, space, shinespark(kills 2 of them), bounceball(tricky), or trickyjump.
-- Doing the room quickly also puts the dragons on a good cycle for the return
-- A waiting strat has time to kill the sovas

- There are kill all enemies strats that only require 9 missiles, 3 supers, or 2 powerbombs, and would be safer to use at low difficulties.

- The current problem with this room is people falling in the lava.  A low requirement strat would make this worse.

- A heatproof or high heatframe strat to get out would be helpful, I think.

dragonsinth:
Okay, fair enough.  It sounds like that last bullet point is what we should go for then.  The situation that triggered this (for me) was a Hard seed with first-item Varia.  I immediately went to this room for the two items, which was actually quite easy with Varia.  But it was out of logic, because with Varia only the only return path is VH.

So maybe this whole PR gets replaced with a simple 4-> 1 "Return Carefully"?:
```
canCarefulJump,
{"heatFrames": 480}
```

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -570,6 +571,15 @@
       "clearsObstacles": ["A"],
       "devNote": "Ice can move through the room more quickly but must spend time refreezing the dragons."
     },
+    {
+      "link": [4, 1],
+      "name": "Return Carefully",
+      "requires": [
+        canCarefulJump,
```
dragonsinth, line 578:
```suggestion
        "canCarefulJump",
```

PR #1691, opened 2024-10-08
Norfair Reserve Tank Room -- Damage Boost strat

dragonsinth:
https://videos.maprando.com/video/1096


region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -583,9 +583,40 @@
...
+          "hits": 1
+        }},
+        {"enemyKill": {
+          "enemies": [
+            ["Dragon"],
+            ["Dragon", "Dragon"]
+          ],
+          "explicitWeapons": ["PowerBomb"]
+        }},
+        {"heatFrames": 210}
```
dragonsinth, line 607: This number is a wild ass guess.

dragonsinth:
(Note, will need to conflict resolve w/ #1690

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -360,6 +360,37 @@
...
```
dragonsinth, line 383: this number is a wild ass guess, need help here

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -360,6 +360,37 @@
...
+      ],
+      "clearsObstacles": ["A"],
```
dragonsinth, line 385: (also clears B if we merge #1690)

region/norfair/east/Norfair Reserve Tank Room.json
dragonsinth: Looking at the video, I measured 270 frames from door -> item, so subtracting out the 60 frames from 2 -> 4, I think this number is accurate.

osse101:
My setup is placing the PB on the bubble instead of jumping into the ceiling, and I get ~180 frames to 4.  The strat makes you waste time somewhere placing the second PB; if it is at 2, then its just 20-30f.  But at 3 its more like 60 with another awkward jump morph to stay out of the lava.  
So you may need to change it to be [1, 2] to model the full strat.  With a copy that collects the missile item on the way with a flatley jump.  And I guess another version that flatleys the missile and returns to the door - I get it with 280 frames door to door.

dragonsinth:
@osse101 just wanted to check in with you on this strat.  Do you think JSON adjustments need to be made, or are you wanting me to reshoot the video with the altered PB placement?

osse101:
I would like for you to make a notable for Power Bomb with Damage Boost and attach it to this strat.
I would like this strat to go from [1, 2] instead of to 4.
I'm thinking a duplicate strat that collects 3 along the way could be added to expert with `canTrickyDodgeEnemies`.
And there is a way to finagle a [1, 1] strat which collects both items and Sova energy drops on no Etanks.

Outside the scope of the PR I want to change the Base [1, 4] to be in expert, and the flatley jump on 3 to be an insane jump.
And make a trickyJump version of crossing the room that uses spazer, wave, or plasma to kill the Sovas while quickly crossing the room.

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -360,6 +360,32 @@
         "Then use a second Power Bomb after waiting for the middle Dragon to finishing firing."
       ]
     },
+    {
+      "link": [1, 4],
+      "name": "Power Bomb with Damage Boost to Lower Item",
```
dragonsinth, line 365: named to represent the `[1,4]` strat specifically

region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -586,6 +612,18 @@
...
+      "note": [
+        "Dash jump from the door to the second ledge.",
+        "Jump forward, morph, and place a Power Bomb on the ceiling between the second and third ledges.",
+        "Press right as Samus begins falling. After falling for 1-2 blocks, unmorph.",
+        "This will place Samus precisely in the path of a Dragon fireball to damage boost left.",
+        "Samus can land on the lower item block, just before the block is destroyed in the Power Bomb explosion."
+      ]
+    }
+  ],
   "nextNotableId": 1
```
dragonsinth, line 628:
```suggestion
  "nextNotableId": 2
```

dragonsinth:
@osse101 updated with the notable, and two different variants, a `[1,4]` that does not clear obstacles and a `[1,2]` that does.  Thoughts?

PR #1710, opened 2024-11-05
Mama Turtle Room - new strat: SpaceJump Escape with Walljump

dragonsinth:
https://videos.maprando.com/video/978

### DIFF
--- a/region/maridia/outer/Mama Turtle Room.json
+++ b/region/maridia/outer/Mama Turtle Room.json
@@ -666,7 +666,7 @@
     },
     {
       "link": [3, 1],
-      "name": "Scapejump Escape with Walljump",
+      "name": "SpaceJump Escape with Walljump",
       "requires": [
         "canSuitlessMaridia",
         "SpaceJump",
@@ -678,7 +678,7 @@
       ],
       "note": [
         "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
-        "Spin jump down and collect the item, then walljump into Spacejump to escape without falling into the water.",
+        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water.",
         "For a slightly easier escape, collect the item with a walljump check, then buffer the jump during the fanfare."
       ]
     },
--- a/region/maridia/outer/Mama Turtle Room.json
+++ b/region/maridia/outer/Mama Turtle Room.json
@@ -668,18 +668,18 @@
       "link": [3, 1],
       "name": "SpaceJump Escape with Walljump",
       "requires": [
-        "canSuitlessMaridia",
         "SpaceJump",
         "canPreciseWalljump",
         {"or": [
           "canHeroShot",
-          "h_canUsePowerBombs"
+          "h_canUsePowerBombs",
+          "Spazer",
+          "Wave"
         ]}
       ],
       "note": [
         "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
         "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water.",
-        "One method: press into the wall while spinning to collect the item, and quickly release directional inputs. Use a neutral button such as Dash to cancel the item fanfare, then immediately perform the walljump inputs."
       ]
     },
     {
--- a/region/maridia/outer/Mama Turtle Room.json
+++ b/region/maridia/outer/Mama Turtle Room.json
@@ -679,7 +679,7 @@
       "note": [
         "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
         "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water.",
-        "For a slightly easier escape, collect the item with a walljump check, then buffer the jump during the fanfare."
+        "Press into the wall while spinning to collect the item and immediate release directional inputs.  Use a neutral button such as Dash to cancel the item fanfare, then immediately perform the walljump inputs."
       ]
     },
     {
--- a/region/maridia/outer/Mama Turtle Room.json
+++ b/region/maridia/outer/Mama Turtle Room.json
@@ -679,7 +679,7 @@
       ],
       "note": [
         "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
-        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water.",
+        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water."
       ]
     },
     {
--- a/region/maridia/outer/Mama Turtle Room.json
+++ b/region/maridia/outer/Mama Turtle Room.json
@@ -665,6 +665,24 @@
       ]
     },
     {
+      "link": [3, 1],
+      "name": "Scapejump Escape with Walljump",
+      "requires": [
+        "canSuitlessMaridia",
+        "SpaceJump",
+        "canPreciseWalljump",
+        {"or": [
+          "canHeroShot",
+          "h_canUsePowerBombs"
+        ]}
+      ],
+      "note": [
+        "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
+        "Spin jump down and collect the item, then walljump into Spacejump to escape without falling into the water.",
+        "For a slightly easier escape, collect the item with a walljump check, then buffer the jump during the fanfare."
+      ]
+    },
+    {
       "id": 28,
       "link": [4, 2],
       "name": "Grapple",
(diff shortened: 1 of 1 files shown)


region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
         "The easiest method is to jump over her by jumping on her babies, or to Morph under her."
       ]
     },
+    {
+      "link": [3, 1],
+      "name": "Scapejump Escape with Walljump",
```
dragonsinth, line 669:
```suggestion
      "name": "SpaceJump Escape with Walljump",
```

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
...
+        "SpaceJump",
+        "canPreciseWalljump",
+        {"or": [
+          "canHeroShot",
+          "h_canUsePowerBombs"
+        ]}
+      ],
+      "note": [
+        "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
+        "Spin jump down and collect the item, then walljump into Spacejump to escape without falling into the water.",
```
dragonsinth, line 681:
```suggestion
        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water.",
```

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
...
+        "canPreciseWalljump",
+        {"or": [
+          "canHeroShot",
+          "h_canUsePowerBombs"
+        ]}
+      ],
+      "note": [
+        "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
+        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water.",
+        "For a slightly easier escape, collect the item with a walljump check, then buffer the jump during the fanfare."
```
dragonsinth, line 682:
```suggestion
        "Press into the wall while spinning to collect the item and immediate release directional inputs.  Use a neutral button such as Dash to cancel the item fanfare, then immediately perform the walljump inputs."
```

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
...
+        "canPreciseWalljump",
+        {"or": [
+          "canHeroShot",
+          "h_canUsePowerBombs"
+        ]}
+      ],
+      "note": [
+        "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
+        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water.",
+        "Press into the wall while spinning to collect the item and immediate release directional inputs.  Use a neutral button such as Dash to cancel the item fanfare, then immediately perform the walljump inputs."
```
dragonsinth, line 682:
```suggestion
        "One method: press into the wall while spinning to collect the item, and quickly release directional inputs, Use a neutral button such as Dash to cancel the item fanfare, then immediately perform the walljump inputs."
```

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
...
+        "canPreciseWalljump",
+        {"or": [
+          "canHeroShot",
+          "h_canUsePowerBombs"
+        ]}
+      ],
+      "note": [
+        "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
+        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water.",
+        "One method: press into the wall while spinning to collect the item, and quickly release directional inputs. Use a neutral button such as Dash to cancel the item fanfare, then immediately perform the walljump inputs."
```
osse101, line 682:
```suggestion
```

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
...
+    {
+      "link": [3, 1],
+      "name": "SpaceJump Escape with Walljump",
+      "requires": [
+        "canSuitlessMaridia",
+        "SpaceJump",
+        "canPreciseWalljump",
+        {"or": [
+          "canHeroShot",
+          "h_canUsePowerBombs"
```
osse101, line 676:
```suggestion
          "h_canUsePowerBombs",
          "Spazer",
          "Wave"
```

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
         "The easiest method is to jump over her by jumping on her babies, or to Morph under her."
       ]
     },
+    {
+      "link": [3, 1],
+      "name": "SpaceJump Escape with Walljump",
+      "requires": [
+        "canSuitlessMaridia",
```
osse101, line 671:
```suggestion
```

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
...
+        "SpaceJump",
+        "canPreciseWalljump",
```
osse101, line 673:
```suggestion
        "canWalljump",
        "canCarefulJump",
```

osse101 (commented):
There is a lot of time to walljump, so its not precise or quick.  But I still don't want this in basic (I don't know if medium is reasonable) but that's what I added carefulJump for.

I hold forward into the wall until the text box goes away and then do a walljump.

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
...
+        {"or": [
+          "canHeroShot",
+          "h_canUsePowerBombs",
+          "Spazer",
+          "Wave"
+        ]}
+      ],
+      "note": [
+        "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
+        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water.",
```
dragonsinth, line 682:
```suggestion
        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water."
```

region/maridia/outer/Mama Turtle Room.json
dragonsinth: Soo... either of these formulations puts the strat in medium, which is what I think we want?

IDK, I formulated it the way I did because making sure you spin jump at the top is free, you can spin jump towards the wall.  But if you miss the walljump (which I have) Mama Turtle whacks you.

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -664,6 +664,24 @@
...
+          "canHeroShot",
+          "h_canUsePowerBombs",
+          "Spazer",
+          "Wave"
+        ]}
+      ],
+      "note": [
+        "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
+        "Spin jump down and collect the item, then walljump into SpaceJump to escape without falling into the water."
+      ]
```
kjbranch, line 683: The way that the link is set up, this will be satisfiable if the item is one that is needed for the strat. So this could be possible even if the item is Space, Wall Jump boots, PBs, Morph, Wave, or Spazer. This can be solved cleanly with another node, but that may not be necessary. Looking through all of these to see if it is necessary:
- This hero shot is quite easy relative to a normal hero shot, so I think it's probably fine if its any of the last 4 and you are just forced to do the hero shot even if it's toggled off. (Actually, the other strats here don't require the canHeroShot and just assume it is free.)
- If you have space jump, you can always get to the top right ledge, grab the item while in a spin and wall jump back up, so it should be fine if its wall jump boots.
- It looks like `Reveal Item while Jumping` is the only strat that can get to the item and not be in spin pose, but if you can wall jump, you could go to the top first.

So, I guess the TL;DR is that this method of representing it should be fine, but could be more error prone if incompatible strats are added that get to 3. 
```suggestion
      ],
      "devNote": [
        "If one of the items that is needed to perform this strat is in this location, the strat could be slightly harder.",
        "Samus should always be able to get here in a spin pose or to the ledge above regardless of which item is in this location."
      ]
```

dragonsinth: I'm confused... I thought the item being acquired was _never_ considered in the routing/reachability?  IE, you have to have Space and one of Spazer, Wave, PB, or Hero Shot in advance to acquire whatever is here, and the item content here is irrelevant to reachability.

dragonsinth: Or wait, my mental model is Maprando specific... maybe you're thinking at a higher level about how this data might be used in other contexts?

If that's the case, then you're right.  Again it's beyond my level of comfort, but I think remodeling this room with a "safe" node above 3 is probably long term better.  But we also need a way to model "spinning" state... reaching node 3 and collecting the item while spinning vs. not spinning is what actually requires the hero shot / wave / spazer / pb.  And then escaping from 3 with walljump+spacejump is a separate node transition.

Hypothetically speaking

region/maridia/outer/Mama Turtle Room.json
osse101: My thought is that you have multiple chances to walljump before you reach the water.
I can see the tech needed here being focused on the walljump, so indicating it as a harder than base form walljump would be appropriate.  canCarefulJump isn't that related to what is happening in the strat but is used for general movement checks.
So you can leave it as it is.

PR #1716, opened 2024-11-06
Mama Turtle Room Update pt.2

kjbranch:


region/maridia/outer/Mama Turtle Room.json
```diff
@@ -450,14 +510,34 @@
       ],
       "note": "Letting Mama Turtle sleep, climb the right wall using HiJump (with a down-grab) and escape the water with Space Jump ."
     },
+    {
+      "link": [1, 6],
```
kjbranch, line 514: This is kind of the simplest/most straightforward [1, 6], so I moved it to be earlier in the order

kjbranch:
For Map Rando, I'm thinking the new notable `Shoot Down and Grab Right Item` will be in `Medium`. The `canHeroShot` didn't quite describe the strat, but also bumped this up to `Hard`, which I think is too high.

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -955,44 +1018,87 @@
...
-          "h_canUsePowerBombs",
-          "Spazer",
-          "Wave"
         ]}
       ],
       "note": [
-        "From the safe spot above the item, use a Power Bomb or shoot down the wall to break the item block.",
-        "Spin jump down and collect the item, then walljump back up; or carefully use SpaceJump."
-      ]
+        "From the safe spot above the item, shoot down the wall to break the item block.",
```
dragonsinth, line 1098: seems worth mentioning the PB option in the public note for people following the spoiler map since the weapon choices aren't explicitly in the json anymore

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -777,9 +837,11 @@
       "requires": [
         "canDodgeWhileShooting",
         "canCarefulJump",
+        {"notable": "Shoot Down and Grab Right Item"},
         {"or": [
           "Wave",
           "Spazer",
+          "h_canUsePowerBombs",
```
dragonsinth, line 844: when and where do you lay the pb for this to work? while riding mama turtle up?

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -397,11 +397,43 @@
...
+      "note": [
+        "Requires getting back on Mama Turtle while she is at the right.",
+        "The safest way is to have a normalized fall speed with either a full height jump or falling from the ledge above.",
+        "If suitless, walking off the right ledge when she hits the left wall can be a good indicator to help land on her safely.",
+        "Shrinking Samus' hitbox after touching the turtles back will also usually help.",
+        "While near the right wall, shoot the item and grab it. If the shot misses, carefully fall back down and safely land on Mama Turtle again.",
+        "After grabbing the item, fall on her again to avoid taking a hit; jump to the left at the top of the ride to escape the water safely."
+      ],
+      "devNote": "FIXME: Is it worth adding a suitless bootless strat that follows this and gets back to the left without another hit?"
+    },
```
dragonsinth, line 430: what's the main difference between this and strat 12?

region/maridia/outer/Mama Turtle Room.json
kjbranch: I was removing the PB option here and just having the normal shot always work. I didn't find using a PB to be easier/more obvious than shooting - it felt far enough away from the item that I actually wasn't initially even expecting it to work. I could add it back if that's the way most people figure out how to get it in vanilla, it just didn't seem obvious to me.

region/maridia/outer/Mama Turtle Room.json
kjbranch: This: Ride mama turtle and at the top shoot the item and try to grab it. If you fail, carefully land on her and keep trying. You can also land on her after successfully grabbing the item. 
[1, 6]: Ride her and jump to the top right, then grab the item either the notable way or continue going down. Note she won't be in a safe location

region/maridia/outer/Mama Turtle Room.json
dragonsinth: That is how I always got the item in vanilla. :)

If you don't want to put it in the logic that's fine, I'm just suggesting having it in the notes.

region/maridia/outer/Mama Turtle Room.json
dragonsinth: 👍 gotcha

region/maridia/outer/Mama Turtle Room.json
dragonsinth: Add the text from the other strat? "Alternatively, a Power Bomb can be used from the right ledge."

dragonsinth (commented):
LGTM

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -416,6 +448,34 @@
...
+          }}
+        ]}
+      ],
+      "clearsObstacles": ["A"],
+      "note": [
+        "Requires getting back on Mama Turtle while she is at the right.",
+        "The safest way is to have a normalized fall speed with either a full height jump or falling from the ledge above.",
+        "If suitless, walking off the right ledge when she hits the left wall can be a good indicator to help land on her safely.",
+        "Shrinking Samus' hitbox after touching the turtles back will also usually help.",
+        "At the top of the ride, jump to the top right ledge."
```
osse101, line 476:
```suggestion
        "Shrinking Samus' hitbox after touching the turtles back will also usually help."
```
This strat makes the most sense to me if you collect 3 and are now coming back to get 2 from below.

osse101: The note indicates going to 6, but if you are getting back on top of an angry turtle and are not going to 3, then being positioned for 5 or 2 makes more sense.

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -397,11 +397,43 @@
...
+      "clearsObstacles": ["A"],
+      "note": [
+        "Requires getting back on Mama Turtle while she is at the right.",
+        "The safest way is to have a normalized fall speed with either a full height jump or falling from the ledge above.",
+        "If suitless, walking off the right ledge when she hits the left wall can be a good indicator to help land on her safely.",
+        "Shrinking Samus' hitbox after touching the turtles back will also usually help.",
+        "While near the right wall, shoot the item and grab it. If the shot misses, carefully fall back down and safely land on Mama Turtle again.",
+        "After grabbing the item, fall on her again to avoid taking a hit; jump to the left at the top of the ride to escape the water safely."
+      ],
+      "devNote": "FIXME: Is it worth adding a suitless bootless strat that follows this and gets back to the left without another hit?"
```
osse101, line 429: This strat does mean that you can ride mama turtle again for free after collecting the item.  Either to reach the door safely or to jump to 6.

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -827,7 +891,8 @@
       "link": [5, 2],
       "name": "Base",
       "requires": [
-        "SpaceJump"
+        "SpaceJump",
+        "canPreciseGrapple"
```
osse101, line 895:
```suggestion
        {"or":[
          "SpaceJump",
          "canPreciseGrapple"
        ]}
```

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -711,15 +780,20 @@
       "name": "Grapple",
       "requires": [
         "Grapple",
-        {"or": [
-          "HiJump",
-          "canPreciseGrapple"
-        ]}
+        "HiJump"
```
osse101, line 783: Is there a Grapple+Springball jump option somewhere?

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -955,44 +1020,88 @@
...
+    {
+      "link": [6, 2],
+      "name": "Grapple, Mama Turtle is Sleeping",
+      "requires": [
+        "canUseGrapple",
+        {"obstaclesNotCleared": ["A"]}
       ],
-      "devNote": "Need to consider other transit options: springball, HBJ."
+      "clearsObstacles": ["B"],
+      "note": "Grappling the block does not require canPreciseGrapple if Mama Turtle is asleep, as there is no risk of death on a miss."
```
osse101, line 1033: - If you use a in-room strat to reach 6 and can redo that, this is fine.
- If you cross room shinespark, you can usually go right to 2 instead on a retry.

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -955,44 +1020,88 @@
...
+      "clearsObstacles": ["B"],
+      "note": [
+        "Briefly run on the top ledge on either side (left is slightly easier) then jump and wall jump on the grapple block."
+      ]
+    },
+    {
+      "link": [6, 2],
+      "name": "Tricky Speed HiJump, Walljumpless",
+      "requires": [
+        "h_canBackIntoCorner",
```
osse101, line 1053: You can spinjump into the wall and turn around to position properly.  I've been calling that a trickyjump.

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -955,44 +1020,88 @@
...
+        "HiJump",
+        "canTrickyJump",
+        "SpeedBooster"
+      ],
+      "note": [
+        "Briefly run on the top right ledge then perform a last-frame jump to barely touch the item and then fall.",
+        "It is required to moon-walk or morph turnaround back against the wall (on the left side of the room, it is instead best to press against the wall and then turn around to be 1 pixel away).",
```
osse101, line 1060: I don't see a good way to normalize y subpixels and that is affecting how many frames you can run.

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -955,44 +1020,88 @@
...
+        "Briefly run on the top ledge on either side (left is slightly easier) then jump and wall jump on the grapple block."
+      ]
+    },
+    {
+      "link": [6, 2],
+      "name": "Tricky Speed HiJump, Walljumpless",
+      "requires": [
+        "h_canBackIntoCorner",
+        "HiJump",
+        "canTrickyJump",
```
osse101, line 1055:
```suggestion
        "canInsaneJump",
```
100% doesn't use this as a frame saver in the room because the jump isn't reliable.

region/maridia/outer/Mama Turtle Room.json
```diff
@@ -955,44 +1020,88 @@
...
+        "canWalljump"
+      ],
+      "clearsObstacles": ["B"],
+      "note": [
+        "Briefly run on the top ledge on either side (left is slightly easier) then jump and wall jump on the grapple block."
+      ]
+    },
+    {
+      "link": [6, 2],
+      "name": "Tricky Speed HiJump, Walljumpless",
```
osse101, line 1051: Should an angry turtle be considered?  You could come up with a cue using wall shakes to safely fall after collecting the item.  Or unmorphing on the way down would be safe.

region/maridia/outer/Mama Turtle Room.json
kjbranch: Oh, yeah. This looks like a copy paste error. Good catch

region/maridia/outer/Mama Turtle Room.json
kjbranch: If it's this difficult, I'm thinking of leaving the strat off and they can just get back on Mama Turtle and get to the top left

region/maridia/outer/Mama Turtle Room.json
kjbranch: It looks like the requirements to spark to 2 vs 6 are the same, except 2 requires an extra ~17 energy. Either can be done tankless if you enter the room.

What would you recommend on this one? I could remove the strat (allowing only the precise grapple variant), add more requirements to it (like gravity+walljump), or leave it as is.
Note that a similar problem could occur if we added a spark to 5 to save energy (which would also require something else, like wall jump, hijump, spring, or space)

osse101: I think you can leave it as it is

PR #1717, opened 2024-11-06
Pause Abuse Crateria Supers Spikes, Add resourceConsumed

kjbranch:
This is focused on the possibility of adding pause abuse strats like the given example in Crateria Supers. I'm not very confident on schema changes, so this may or may not be done correctly.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -967,10 +967,19 @@
           "name": "h_pauseAbuseMinimalReserveRefill",
           "requires": [
             "canPauseAbuse",
-            {"resourceAtMost": [{"type": "RegularEnergy", "count": 1}]},
-            {"resourceConsumed": [{"type": "ReserveEnergy", "count": 1}]}
+            {"or": [
+              {"and": [
+                "canPreciseReserveRefill",
+                {"resourceAtMost": [{"type": "RegularEnergy", "count": 1}]},
+                {"resourceConsumed": [{"type": "ReserveEnergy", "count": 1}]}
+              ]},
+              {"and": [
+                {"resourceAtMost": [{"type": "RegularEnergy", "count": 4}]},
+                {"resourceConsumed": [{"type": "ReserveEnergy", "count": 4}]}
+              ]}
+            ]}
           ],
-          "devNote": "Using a pause abuse to tank a hit and having that take Samus to 0 Energy and then be refilled with 1 Reserve Energy."
+          "devNote": "Using a pause abuse to tank a hit and having that take Samus to 0 Energy and then be refilled with minimal Reserve Energy."
         }
       ]
     },
--- a/helpers.json
+++ b/helpers.json
@@ -962,6 +962,15 @@
             "canCarefulJump"
           ],
           "devNote": "canCarefulJump is used to represent leaving the enemy drops, which isn't a propper fit, but it is a tech with a similar difficulty."
+        },
+        {
+          "name": "h_pauseAbuseMinimalReserveRefill",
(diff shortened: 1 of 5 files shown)


schema/m3-requirements.schema.json
```diff
@@ -392,7 +392,7 @@
                   "description": "The type of resource that will be reduced."
                 },
                 "count": {
-                  "$id": "#/definitions/logicalRequirement/items/properties/resourceAvailable/properties/items/count",
+                  "$id": "#/definitions/logicalRequirement/items/properties/resourceAtMost/properties/items/count",
```
kjbranch, line 395: Pretty sure this was a copy/paste error, but maybe there is just something I'm misunderstanding.

schema/m3-requirements.schema.json
```diff
@@ -549,6 +549,42 @@
               }
             }
           },
+          "resourceConsumed": {
+            "$id": "#/definitions/logicalRequirement/items/properties/resourceConsumed",
+            "type": "array",
+            "title": "Resource Available",
```
blkerby, line 555:
```suggestion
            "title": "Resource Consumed",
```

blkerby (approved):

helpers.json
```diff
@@ -962,6 +962,15 @@
             "canCarefulJump"
           ],
           "devNote": "canCarefulJump is used to represent leaving the enemy drops, which isn't a propper fit, but it is a tech with a similar difficulty."
+        },
+        {
+          "name": "h_pauseAbuseMinimalReserveRefill",
+          "requires": [
+            "canPauseAbuse",
+            {"resourceAtMost": [{"type": "RegularEnergy", "count": 1}]},
+            {"resourceConsumed": [{"type": "ReserveEnergy", "count": 1}]}
```
kjbranch, line 971: Looks like Maddo's comment got lost to the ether. I think the main comment was relating to if this single energy drain is too precise for Expert and if it should be increased to 2. (in Extreme it should be fine at 1). 

I don't think it will really matter much, but these are the options that I could do ...
- fix it properly and add a leniency helper to this that adds 1 energy (probably overkill)
- keep it as is (maybe too tight on Expert)
- increase it to 2 (maybe too lenient on higher difficulties)

osse101 (commented):
This is `ammo` but for energy, so its an area of confusion that resourceConsumed doesn't include ammo resources.  It is more generally inclusive to rename ammo to `resource` and add these types there.  The difference is that resourceConsumed types are going to be much less commonly used, so a separation is nice for that reason.

schema/m3-requirements.schema.json
blkerby: Yeah I think `$id` only needs to be set at the top level (where I notice we're using `id`, which apparently is the old Draft 4 way of doing it, instead of `$id`). The other ones could probably all be removed. It could be done with a script. May be worth discussing on #data-dev?

helpers.json
blkerby: Yeah another solution would be to add a tech for exact manual reserve energy control, and then add a leniency (say of 4) without that tech; that could be expressed as an `or` inside the helper. Because of the heat damage leniency, we're probably unlikely to see a scenario where this will matter in a significant way. So it could also be fine how it is. Adding a tech might be a good thing anyway though, just to have a place to document the technique to help players learn it.

kjbranch: Added this. I dont think this helper should be used in heated rooms, but a heated variant could be made.

PR #1718, opened 2024-11-06
Parlor shinecharge updates & new strats

blkerby:
- Tighten shinecharge frames on several existing strats
- Add many new shinecharge/shinespark strats
- Add missing Morph requirements
- Remove a couple unnecessary `canWalljump` requirements.
- In one case (strat 18), convert "Come In Shinecharged" require to "Come in Shinecharging". This makes the strat a bit more useful since the extra runway can now logically be used. Because the strat required 170 frames remaining, it was always going to require a runway connected to the door in the other room, so there's no reason for it not to be "Come In Shinecharging".

### DIFF
--- a/region/crateria/central/Parlor and Alcatraz.json
+++ b/region/crateria/central/Parlor and Alcatraz.json
@@ -610,14 +610,16 @@
     {
       "id": 18,
       "link": [2, 3],
-      "name": "Leave With Spark (Come In Shinecharged)",
+      "name": "Come In Shinecharging, Leave With Spark",
       "entranceCondition": {
-        "comeInShinecharged": {
-          "framesRequired": 170
+        "comeInShinecharging": {
+          "length": 2,
+          "openEnd": 0
         }
       },
       "requires": [
-        "canShinechargeMovementComplex",
+        "Morph",
+        "canShinechargeMovementTricky",
         {"shinespark": {"frames": 6}}
       ],
       "exitCondition": {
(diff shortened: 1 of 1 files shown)


region/crateria/central/Parlor and Alcatraz.json
```diff
@@ -924,6 +1091,72 @@
...
+    {
+      "link": [3, 6],
+      "name": "Come in Shinecharged, Leave With Spark (Top Position)",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 90
+        }
+      },
+      "requires": [
+        "canShinechargeMovementComplex",
```
osse101, line 1138:
```suggestion
        "canShinechargeMovementComplex",
        "canCarefulJump",
```
It's kinda like trickyJump movement, but not hard to do.

osse101 (dismissed):
Do we need to account for `canMidairShinespark` everywhere somehow?  LeaveWithSpark through a bottom uses canHorizontalShinespark always, and leaving through top uses canMidairShinespark usually.  
And most of the time the two cases are part of one strat.

blkerby:
> Do we need to account for `canMidairShinespark` everywhere somehow? LeaveWithSpark through a bottom uses canHorizontalShinespark always, and leaving through top uses canMidairShinespark usually. And most of the time the two cases are part of one strat.

I've been thinking of `canMidairShinespark` as implicitly assumed in `canShinechargeMovementComplex`. If we're ok with that, then it may make sense to add it to the `techRequires` so that it becomes an actual dependency. Even though a mid-air spark isn't required for every use of `canShinechargeMovementComplex`, it can make sense as a tech prerequisite in that it should definitely be learned first.

osse101:
Yes, I think canMidairShinespark expects you to be able to not accidently shinespark while jumping around, which means learning how to midairShinespark so that you know what to avoid.

PR #1719, opened 2024-11-07
Samus Eater temp blue strats

blkerby:
- This adds new strats in Alpha Power Bomb Room and Hellway for using a Samus Eater to gain temporary blue. Insomniaspeed recently showed us how to do this.
- A little more detail is added to the note for the `canSamusEater` tech and its applications, to clarify what the trick is for making it work.

### DIFF
--- a/region/brinstar/red/Alpha Power Bomb Room.json
+++ b/region/brinstar/red/Alpha Power Bomb Room.json
@@ -105,7 +105,33 @@
           "framesRemaining": 100
         }
       },
-      "flashSuitChecked": true
+      "flashSuitChecked": true,
+      "note": [
+        "Land in the Samus Eater while moving forward.",
+        "Continue holding dash and forward to gain a shinecharge while running in place inside the Samus Eater."
+      ]
+    },
+    {
+      "link": [1, 1],
+      "name": "Leave With Temporary Blue (Samus Eater)",
+      "requires": [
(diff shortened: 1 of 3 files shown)


osse101 (commented):
I didn't know the standup effect came from moving when entering the Vileplume, that is very helpful.

The strat feels like insane to me.  I crouched and xrayed a few times but couldn't keep temp blue.  I imagine you get a better feel for when to stop after spending more time on it.

- Is entering on the furthest pixel required for the trick?
- Does tapping for less run speed increase the frame window?

blkerby:
I did some TAS testing on stream yesterday, around here: https://www.twitch.tv/videos/2295334249?t=3h20m20s.

Entering as far left as possible is helpful but not required. Delaying holding dash helps in theory, but at least for the scenario I tested I'm not sure it was helpful in practice. If you could tap on the magic frames like with a short charge, I expect it would help more, but I didn't try it.

blkerby:
I think I messed up in some of my testing by the way. Like I think sometimes I was still holding right too long after the down/shine charge input, which can mess things up. And I'm not sure if I ever actually got in at the far-left pixel

osse101:
I tried it some more and got it a couple of times.  So now I think its a lot like the other extreme strats, but you need to get a feel for the timing before expecting to get it.  So it can cost extra health in attempts.  Just like other extreme tech.

PR #1720, opened 2024-11-07
Mama Turtle Room - use modern collectsItems instead of obstacles B

dragonsinth:
See first commit; second commit is just to sort.


kjbranch:
I'm not really sure what problem this is trying to fix, but I don't really see it as an improvement. There is also a PR already open with a major change to this room, so I don't see it as a good time to have other big changes until that one is merged.

dragonsinth:
yeah, just exploring the direction. def didn't want to create a merge conflict for you.

dragonsinth:
Oh, I didn't answer your question.  My motivation was to:

- reduce the state space (node vs obstacle)
- make the model more intuitive (representing climbing on top of the grapple block via an obstacle is weird; using node to represent different physical locations is more in line with everything else)
- make the model more accurate (just because you got on top of the grapple block once, doesn't mean you're permanently on top of the grapple block in the same way that killing enemies or waking mama turtle would)

kjbranch:
Response to your 4 reasons:
1. I dont think this matters in a room this simple with only 2 obstacles
2. I'm not seeing this as making this more intuitive. The abstract obstacles are used in many rooms, and the logic pages that Map Rando uses makes it easy to see how to get to an item, which wouldnt work with this method.
3. Added this feedback to my PR
4. Not sure what youre referring to in that other branch?

dragonsinth:
Oh wow... I had no idea `resetsObstacles`  was a thing!  Nevermind then.

PR #1721, opened 2024-11-07
East Ocean shinecharge updates

blkerby:
Videos for all these new/updated strats are on https://videos.maprando.com.

### DIFF
--- a/region/crateria/east/East Ocean.json
+++ b/region/crateria/east/East Ocean.json
@@ -152,7 +152,7 @@
     },
     {
       "link": [1, 1],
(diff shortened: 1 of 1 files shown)


osse101 (commented):
Here's another left door shortcharge
https://videos.maprando.com/video/1250

I got 47frames remaining with hijump, and 38 without but with walljump.

blkerby:
> Here's another left door shortcharge https://videos.maprando.com/video/1250
> 
> I got 47frames remaining with hijump, and 38 without but with walljump.

Added.

PR #1726, opened 2024-11-07
Add Blue Brin ETank notable spark

blkerby:
Video: https://videos.maprando.com/video/1247


osse101:
Extreme imo.  But I still only pull it off with 1 or 2 frames remaining.

PR #1730, opened 2024-11-08
Remove useless airball strats in West Sand Hall

blkerby:
Without HiJump, doing an airball makes it harder to make it across to the pillar, compared to just doing a normal jump. This is already reflected in the "minTiles" for the airball strats which were a tile larger for the airball strats than the corresponding regular jump strats. So there doesn't seem to be a purpose for the airball strats to exist.

This also updates the description of the "Jump With Momentum" and "Tricky Jump With Momentum" strats to reflect the runway requirements currently in those strats.

### DIFF
--- a/region/maridia/inner-green/West Sand Hall.json
+++ b/region/maridia/inner-green/West Sand Hall.json
@@ -376,9 +376,7 @@
         "canCarefulJump"
       ],
       "note": [
-        "Gain momentum in the room to the left, then jump in this room in order to land on the first pillar. Ensure that Speed Booster is disabled.",
-        "With a short runway of 5 tiles in the adjacent room, it is required to jump as late as possible at the edge of the runway in this room.",
-        "With any longer of a runway, the jump is much less precise."
+        "Gain momentum in the room to the left, then jump in this room in order to land on the first pillar. Ensure that Speed Booster is disabled."
       ],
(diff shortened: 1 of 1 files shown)


region/maridia/inner-green/West Sand Hall.json
```diff
@@ -376,9 +376,7 @@
         "canCarefulJump"
       ],
       "note": [
-        "Gain momentum in the room to the left, then jump in this room in order to land on the first pillar. Ensure that Speed Booster is disabled.",
-        "With a short runway of 5 tiles in the adjacent room, it is required to jump as late as possible at the edge of the runway in this room.",
-        "With any longer of a runway, the jump is much less precise."
+        "Gain momentum in the room to the left, then jump in this room in order to land on the first pillar. Ensure that Speed Booster is disabled."
```
kjbranch, line 379:
```suggestion
        "Gain momentum in the room to the left, then jump after entering this room in order to land on the first pillar. Ensure that Speed Booster is disabled."
```
I feel like this point should be extra explicit, since it's not obvious and could be confusing since the strat uses `canCrossRoomJumpIntoWater`

kjbranch (dismissed):
This does feel much easier than the air ball. With 5 tiles, you can jump at almost any time in this room to land on the pillar. Only the first two frames on entry don't work, which is not the time in which you would attempt to jump anyway. So I can't imagine the air ball ever being useful here.

PR #1731, opened 2024-11-10
Croc Speedway shinecharges

blkerby:
- This adds many new shinecharge strats and updates the existing ones.
- A new tech `canControlShinesparkEnd` is added for manipulating Samus' regular energy in order to cause a shinespark to end in a desired location.
- Using this tech, the reverse spark with Varia becomes in logic tankless.
- Heat frames are tightened, so the reverse spark from the Save door becomes in logic with 2 ETanks.
- Reserve abuse strats are added, allowing getting through with 1 Reserve Tank from any of the three doors on the right, using the recently added tech `canPreciseReserveRefill` (or 2 Reserve Tanks without that tech but still with `canPauseAbuse`).
- The shinecharge strats from node 6 (with obstacle "A" cleared) are not yet updated. These currently require `h_heatProof`. It will take a little more thought to figure out how to best refine these; it can be done later.

### DIFF
--- a/region/norfair/west/Crocomire Speedway.json
+++ b/region/norfair/west/Crocomire Speedway.json
@@ -128,6 +128,7 @@
       "to": [
         {"id": 2},
         {"id": 3},
+        {"id": 4},
         {"id": 6}
       ]
     },
@@ -135,7 +136,9 @@
       "from": 4,
       "to": [
         {"id": 2},
+        {"id": 3},
         {"id": 4},
+        {"id": 5},
         {"id": 6}
       ]
     },
@@ -144,6 +147,7 @@
       "to": [
         {"id": 1},
         {"id": 2},
+        {"id": 4},
         {"id": 5},
         {"id": 6}
       ]
@@ -192,7 +196,7 @@
     {
       "id": 4,
       "link": [1, 5],
-      "name": "Leave Shinecharged",
+      "name": "Come In Shinecharging, Leave Shinecharged",
       "entranceCondition": {
         "comeInShinecharging": {
           "length": 0,
@@ -200,11 +204,11 @@
         }
       },
       "requires": [
-        {"heatFrames": 140}
+        {"heatFrames": 120}
       ],
       "exitCondition": {
         "leaveShinecharged": {
-          "framesRemaining": 70
+          "framesRemaining": 85
         }
       },
       "unlocksDoors": [
@@ -221,6 +225,99 @@
       ],
       "flashSuitChecked": true,
       "devNote": "FIXME: Entering with too little run speed causes Samus to 'trip', so the runway cannot reliably be used for short shortcharges."
+    },
+    {
+      "link": [1, 5],
+      "name": "Carry Shinecharge",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 125
+        }
+      },
+      "requires": [
+        "canShinechargeMovementComplex",
+        {"heatFrames": 125}
+      ],
+      "exitCondition": {
+        "leaveShinecharged": {
+          "framesRemaining": "auto"
+        }
+      },
+      "unlocksDoors": [
+        {
+          "types": ["super"],
+          "requires": []
+        },
+        {
+          "types": ["missiles", "powerbomb"],
+          "requires": [
+            "never"
+          ]
+        }
+      ],
+      "flashSuitChecked": true
+    },
+    {
+      "link": [1, 5],
+      "name": "Come In Shinecharged, Leave With Spark (Top Position)",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 105
+        }
+      },
+      "requires": [
+        {"heatFrames": 130},
+        {"shinespark": {"frames": 2, "excessFrames": 0}}
+      ],
+      "exitCondition": {
+        "leaveWithSpark": {
+          "position": "top"
+        }
+      },
+      "unlocksDoors": [
+        {
+          "types": ["super"],
+          "requires": []
+        },
+        {
+          "types": ["missiles", "powerbomb"],
+          "requires": [
+            "never"
+          ]
+        }
+      ],
+      "flashSuitChecked": true
+    },
+    {
+      "link": [1, 5],
+      "name": "Come In Shinecharged, Leave With Spark (Bottom Position)",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 110
+        }
+      },
+      "requires": [
+        {"heatFrames": 130},
+        {"shinespark": {"frames": 2, "excessFrames": 0}}
+      ],
+      "exitCondition": {
+        "leaveWithSpark": {
+          "position": "bottom"
+        }
+      },
+      "unlocksDoors": [
+        {
+          "types": ["super"],
+          "requires": []
+        },
+        {
+          "types": ["missiles", "powerbomb"],
+          "requires": [
+            "never"
+          ]
+        }
+      ],
+      "flashSuitChecked": true
     },
     {
       "id": 5,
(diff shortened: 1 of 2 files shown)


region/norfair/west/Crocomire Speedway.json
```diff
@@ -592,6 +886,47 @@
...
+        "comeInShinecharging": {
+          "length": 1,
+          "openEnd": 1
+        }
+      },
+      "requires": [
+        "canShinechargeMovementTricky",
+        "canInsaneJump",
+        "canBePatient",
+        {"enemyDamage": {
```
osse101, line 902: Its a bit easier if you kill the cac - I was using plasma.

region/norfair/west/Crocomire Speedway.json
```diff
@@ -592,6 +886,47 @@
       ],
       "flashSuitChecked": true
     },
+    {
+      "link": [4, 5],
+      "name": "Come In Shinecharging, Leave With Spark",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 1,
```
osse101, line 894: - Using the tile for runway makes storing the shinespark extremely tight.  
- Sliding into the door and off the ledge is a slower fall so not really easier overall
- stopping before the door means you work with fewer frames but the initial platforming is easier.

They were all about equivalent, but the trick to stopping in room took some tries to learn.

region/norfair/west/Crocomire Speedway.json
```diff
@@ -736,20 +1265,83 @@
...
+        {"resourceConsumed": [{"type": "ReserveEnergy", "count": 100}]},
+        {"or": [
+          "canPreciseReserveRefill",
+          {"resourceConsumed": [{"type": "ReserveEnergy", "count": 15}]}          
+        ]},
+        {"resourceAtMost": [{"type": "RegularEnergy", "count": 1}]}
+      ],
+      "clearsObstacles": ["A"],
+      "note": [
+        "Either jump over a Cacatac or tank a spike hit.",
```
osse101, line 1332: I couldn't survive the trip taking a spike hit

region/norfair/west/Crocomire Speedway.json
```diff
@@ -864,6 +1456,32 @@
...
+        "canShinechargeMovementTricky",
+        "canTrickyJump",
+        {"or": [
+          {"enemyDamage": {
+            "enemy": "Cacatac",
+            "type": "spike",
+            "hits": 1
+          }},
+          "canBePatient"  
+        ]},
```
osse101, line 1478:
```suggestion
        ]},
        {"heatFrames": 205},
```

osse101 (commented):
`canControlShinesparkEnd` is like expert tech but sometimes tedious, and even more of a knowledge check for niche strats = extreme.

1. The pause abuse strats are kinda notable and I think heat scaling would keep it balanced, but scaling isn't applied to `resourceConsumed`?
2. The [5, 2] reverse spark is kinda notable.  But its not a surprising or obscure thing to do at high difficulties.

region/norfair/west/Crocomire Speedway.json
blkerby: Oof, yeah another good catch, and another thing to add to the tests. The video shows 209 so I'm thinking of putting 210.

region/norfair/west/Crocomire Speedway.json
```diff
@@ -329,14 +426,64 @@
...
+        ]}
       ],
       "clearsObstacles": ["A"],
       "flashSuitChecked": true,
       "note": [
         "Spark left through the speed blocks, then run to the right and back to get speed to go through the rest.",
-        "The shinespark expects to kill the crumble bridge pirate, to be safe."
+        "If performing the spark with low energy, there is a risk of ending the spark above the spikes or in front of the Pirate;",
+        "to be safe, assuming Samus has heat protection and begins sparking about 3 tiles from the Speed blocks, start with between 72 and 87 energy or at least 105 energy;",
+        "with a buffered crumble jump to the right, up to 93 energy can work."
```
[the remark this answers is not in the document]
osse101: This is missing canShinesparkMidair 🤔 And the difficulty of the midair spark scales with your leniency frames until you are doing a horizontal spark.  And I'd always try to standing spark this if I could.

region/norfair/west/Crocomire Speedway.json
osse101: I think its a little more difficult than other extreme level shinecharge movement strats.

region/norfair/west/Crocomire Speedway.json
osse101: My problem was with the 11 energy pauses.  I could probably eventually do it.

If you farm the cac, this becomes fairly easy, if you hit the shinespark.
full supers, missile/grapple/plasma/wave+spazer

region/norfair/west/Crocomire Speedway.json
osse101: yes, it only felt like it saved 2 or 3 frames.  There are still 2 other very precise jumps.

region/norfair/west/Crocomire Speedway.json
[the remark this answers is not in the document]
blkerby: Yeah, at the amount of shinecharge frames (35) written in the strat, it would need `canMidAirShinespark`. But as long as you have enough leniency frames then it's not needed. Even at Very Hard, the leniency frames would be enough to ensure you could do a standing spark (as long as those frames weren't used in the other room). So I think it works out ok?

With the mid-air spark requirement being unwritten, the strat could be in Medium, except that the notable requirement puts it in Hard. It's probably for the best, because on Medium players would probably have trouble getting tripped up on the stairs.

region/norfair/west/Crocomire Speedway.json
blkerby: With this change, the strat moves down to Expert in the case where the Cac can be farmed for a big energy. I think it's probably fair since it's just regular pause abuse at that point, nothing precise. And even though it's a long sequence of them, the pause abuse here is straightforward since you're not doing any other movement, only holding dash and left.

blkerby: I'm not sure I understand, is there something you want changed here? This is currently a Hard notable but the strats are in Expert by the tech requirements (or Extreme if coming from top-right with no ability to farm the Cac).

Positioning for the spark does seem like the most difficult part of the trick, and that mostly applies whether or not doing pause abuse (though the pause before wind-up does add another distraction). On the other hand, positioning vertically within a 3-tile space for a spark isn't something that we've been bumping higher than Expert in other places? Most cross-room sparks that require specifically "top" or "bottom" position could require that level of precision, but they would usually be in Very Hard based on `canShinechargeMovementComplex`, or sometimes even Hard if it doesn't need `canShinechargeMovementComplex`.

tech.json
```diff
@@ -1901,7 +1901,17 @@
               ],
               "otherRequires": [],
               "note": "Using the Samus echoes that are emitted after a shinespark bonk as a weapon to kill enemies."
-            }
+            },
+            {
+              "name": "canControlShinesparkEnd",
+              "techRequires": [
+                "canShinespark"
```
kjbranch, line 1908: I think this should say that you end the strat with at most 29 energy

blkerby: I think the way that we express the energy requirement may be a bit situational. So I'm not sure it makes sense to include as a requirement inside the tech. It's good to keep in mind though everywhere it is used

kjbranch: Oh, yeah, I was confused. It's needed in one of the strats I'm writing, but you're right, its not always or even usually needed.

region/norfair/west/Crocomire Speedway.json
osse101: I want pause abusing up the speedway to be a new notable.

`[5, 2] Reverse Spark With Reserve Abuse` does seem to get better with more practice, as does timing the repauses.  But getting close to 11 isn't so simple.  I think it fits in Expert, but is directly harder than `[5, 2] Reverse Spark`.
And it needs to be clearer that 0+1 tanks is enough to make this run.

PR #1733, opened 2024-11-11
Crateria Supers spark strats

blkerby:
This adds a couple new "tricky" strats which make it possible to spark across the top of the room left-to-right while conserving enough energy to be able to survive a spike hit tankless on the way back. Later when we have schema support for coming in shinecharged with momentum, lower-difficulty strats for doing this could also be added.

Videos:
- comeInShinecharged: https://videos.maprando.com/video/1314
- comeInShinecharging: https://videos.maprando.com/video/1315

The strats for sparking up the room from the bottom are also refined/tightened.

### DIFF
--- a/region/crateria/central/Crateria Super Room.json
+++ b/region/crateria/central/Crateria Super Room.json
@@ -908,37 +908,60 @@
       "requires": [
         "canTrivialUseFrozenEnemies",
         "h_canShineChargeMaxRunway",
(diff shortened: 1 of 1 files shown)


osse101:
The [1, 4] strats feel like they should be in veryhard, but they are frustratingly inconsistent for me.  Probably like most shinespark into downgrab strats.  So expert seems appropriate.

PR #1744, opened 2024-11-13
Add Missing Ceiling Bomb Jumps and G-Mode Open Green Gates

### LINKED ISSUE #1225: Add ceiling bomb jump strats
The tech video for canCeilingBombJump shows an application in Crocomire's Room, but there's no strat for this in logic. Probably there are other places too where it could be added.

kjbranch:
This should close #1225


region/norfair/east/Spiky Acid Snakes Tunnel.json
```diff
@@ -339,6 +339,14 @@
         "Delay the damage boost from the spikes slightly in order to rise above the lava before moving."
       ]
     },
+    {
+      "link": [1, 2],
+      "name": "Ceiling Bomb Jump Over Spikes",
+      "requires": [
+        "h_heatProof",
+        "canLongCeilingBombJump"
```
osse101, line 347: This has a sloped ceiling on both sides, so it would take diagonal bomb jumping into returning to neutral and staggered IBJ before the longCIBJ

region/norfair/east/Double Chamber.json
```diff
@@ -1298,6 +1298,14 @@
         "Climb the left wall while avoiding falling onto the spikes or through the crumble blocks. The ledge is the same size of that in writg."
       ]
     },
+    {
+      "link": [3, 4],
+      "name": "Ceiling Bomb Jump Over Spikes",
+      "requires": [
+        "h_heatProof",
+        "canCeilingBombJump"
```
osse101, line 1306: sloped ceiling, so a sinking diagonal bomb jump would be the way to do this

region/norfair/east/Spiky Acid Snakes Tunnel.json
kjbranch: Maddo was testing ceiling bomb jump on gentle up slopes and horizontal ceilings and said they felt the same. He didn't try for long, and never finished crossing the room, so they could be a bit different: https://videos.maprando.com/video/1246

I then tested it on gentle down, and got it to work for about 2 tiles before I would fail, so I know it's doable there, but I have no idea how it compares to horizontal. Instead, it looks like an unmorph reset fall speed works here. 

I didn't think DIBJ into IBJ was something we wanted to include, or there would be a couple more uses (i.e. R to L Crateria Supers)

region/brinstar/blue/Blue Brinstar Boulder Room.json
```diff
@@ -775,6 +775,21 @@
         "This strat is easily bypassed by jumping through the door, or if the door can be opened."
       ]
     },
+    {
+      "link": [3, 1],
+      "name": "Bomb Boost Water Escape Into IBJ",
+      "requires": [
+        "canBombJumpWaterEscape",
+        "canJumpIntoIBJ",
+        "h_canUnderwaterCrouchJump",
```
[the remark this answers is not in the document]
kjbranch: I never got it without the crouch, but if it's not required I will remove it. I was planning on using that helper everywhere you crouch in water. Would you like it renamed somehow to not feel like it's different from a normal crouch jump most of the time?

region/norfair/east/Spiky Acid Snakes Tunnel.json
osse101: I tried this out, and you can CIBJ up gentle slopes, so I learned something.  I had too much trouble getting a CIBJ going for the downslope but you can reach the ledge with an unmorph like you said.

DIBJ into IBJ looks like you could get good enough to react to getting the right pixel, but I couldn't read that well enough.  So it'd be like longCIBJ tier and not a priority.

osse101: And I don't see `1246` as linking to a video

region/brinstar/blue/Blue Brinstar Boulder Room.json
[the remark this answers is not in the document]
osse101: tldr; `h_canUnderwaterCrouchJump` -> `h_canUnderwaterCrouchJumpWithFlashSuit`

Practically speaking the `canBombJumpWaterEscape` always does an underwater crouch jump since it uses a soft-unmorph on the rejump, but you can go out of your way to avoid it here.  The problem right here is that bombjumpwaterescapes have a varying quality to their crouchJumps and it can seem like the helper is referencing that.  Crouch jumps might be missing from other places where you bombJumpWaterEscape but not doing them makes the trick harder.

What the helper is trying to say is that a crouch or straight jump can be done in water with a flash suit without activating it.  Knowing that it is possible to do that is a tech and would be described in a `moveWithFlashSuit` tech making this note a reminder and not the main way to document the skill.  The role of this helper is to show that you need this technique more than applying new requirements to the strat and it is still useful in that way but not much different from being in a note.  Pointless point: you can do this without a straight jump by bombing into the wall to start your IBJ.

`h_canUnderwaterCrouchJumpWithFlashSuit` is ugly long but
- it only works in water physics
- only applies to straight jumps and a crouch jump is more generally useful
- has the specific purpose of avoiding `{noFlashSuit}`

region/norfair/east/Double Chamber.json
osse101: It actually looks too far for the DIBJ I was thinking of

region/norfair/east/Spiky Acid Snakes Tunnel.json
kjbranch: Ah, looks like it's listed as a disabled video, since the CIBJ failed part was through and fell. With that link you can then filter to disabled videos and it will show.

kjbranch:
I will make a new PR tomorrow making the `h_canUnderwaterCrouchJumpWithFlashSuit`

PR #1750, opened 2024-11-14
PCJR shinesparks

blkerby:
This is mostly adding strats for sparking through the left door of PCJR, with a few other refinements/fixes. This also adds strats for sparking diagonally to the top-left, using less energy than the horizontal spark currently in logic.

In Landing Site, I noticed that strats that went through the open left door were implicitly requiring unlocking the door a second time, so that is also fixed here.

As usual, videos for all the new stuff are on https://videos.maprando.com

### DIFF
--- a/region/norfair/crocomire/Post Crocomire Jump Room.json
+++ b/region/norfair/crocomire/Post Crocomire Jump Room.json
@@ -72,6 +72,11 @@
       "name": "G-Mode Morph Direct",
       "obstacleType": "abstract",
       "note": "Samus enters the room in direct G-Mode with artificial morph, in order to be able to remote acquire the item."
+    },
+    {
+      "id": "E",
+      "name": "Left door open",
+      "obstacleType": "inanimate"
     }
   ],
   "enemies": [
@@ -159,6 +164,14 @@
         "h_canCrystalFlash"
       ],
       "flashSuitChecked": true
+    },
+    {
+      "link": [1, 1],
+      "name": "Open Door",
+      "requires": [
+        {"doorUnlockedAtNode": 1}
+      ],
+      "clearsObstacles": ["E"]
     },
     {
       "id": 3,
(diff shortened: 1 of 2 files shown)


region/norfair/crocomire/Post Crocomire Jump Room.json
```diff
@@ -503,13 +532,104 @@
...
+      "name": "Leave With Spark (Short Speedy HiJump)",
+      "requires": [
+        "HiJump",
+        {"obstaclesCleared": ["B", "E"]},
+        {"canShineCharge": {
+          "usedTiles": 32,
+          "gentleDownTiles": 6,
+          "openEnd": 1
+        }},
+        "canShinechargeMovementComplex",
```
osse101, line 600: blind shinespark
fairly ok for very hard 🤷

region/norfair/crocomire/Post Crocomire Jump Room.json
```diff
@@ -503,13 +532,104 @@
...
+          "position": "bottom"
+        }
+      },
+      "unlocksDoors": [
+        {
+          "types": ["ammo"],
+          "requires": [],
+          "useImplicitRequires": false
+        }
+      ]
```
osse101, line 587:
```suggestion
      ],
      "note": [
        "After storing a shinespark, start running from the left side of the vertical door and jump so that a shinespark at the apex of the jump will approximately align with the top of the far door."
      ]
```

blkerby: In this strat the PB blocks may be intact, so you wouldn't be able to start that far right. Was this note maybe intended for the "Long Speedy Jump" version?

osse101: Yes sorry, this note would go with the long runway strat.

This short runway strat lines up well on its own but has some 🤔 aspects: jumping right near the end of the jump to have enough speed for jump height, and waiting until just before the shinespark frames expire.  But there are visual cues for both that are good enough.

PR #1753, opened 2024-11-14
Reserve Pause Abuse: Remaining Maridia

kjbranch:
Relies on #1731 for `canControlShinesparkEnd`

### DIFF
--- a/region/maridia/inner-pink/Draygon's Room.json
+++ b/region/maridia/inner-pink/Draygon's Room.json
@@ -246,19 +246,36 @@
           "usedTiles": 22,
           "openEnd": 0
         }},
-        {"enemyDamage": {
-          "enemy": "Draygon",
-          "type": "contact",
-          "hits": 2
-        }},
         {"or": [
-          "HiJump",
+          "Morph",
           {"enemyDamage": {
             "enemy": "Draygon",
             "type": "contact",
             "hits": 1
           }}
         ]},
+        {"or": [
+          "canTrickyJump",
+          {"enemyDamage": {
+            "enemy": "Draygon",
+            "type": "contact",
+            "hits": 1
+          }}
+        ]},
+        {"or": [
+          "HiJump",
+          {"enemyDamage": {
+            "enemy": "Draygon",
+            "type": "contact",
+            "hits": 1
+          }},
+          "h_pauseAbuseMinimalReserveRefill"
+        ]},
+        {"enemyDamage": {
+          "enemy": "Draygon",
+          "type": "contact",
+          "hits": 1
+        }},
         {"shinespark": {"frames": 150}}
       ],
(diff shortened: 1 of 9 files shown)


region/maridia/inner-yellow/Plasma Spark Room.json
```diff
@@ -323,7 +323,8 @@
               "Charge+Wave",
               "Charge+Spazer"
             ]
-          }}
+          }},
+          "h_pauseAbuseMinimalReserveRefill"
```
osse101, line 327: It may be that `{shinespark}` has to be ordered after this.  
It would have been somewhat wrong listing the shinespark before enemyDamage as it was.

region/maridia/inner-pink/Draygon's Room.json
```diff
@@ -246,19 +246,36 @@
...
           }}
         ]},
+        {"or": [
+          "HiJump",
+          {"enemyDamage": {
+            "enemy": "Draygon",
+            "type": "contact",
+            "hits": 1
+          }},
+          "h_pauseAbuseMinimalReserveRefill"
```
osse101, line 272: I don't think these is practical since you need high RegularEnergy throughout the fight to shinespark.  Pauseabusing can be used in other strats to walk through Draygon after a goop instead of dodging, or to escape a grab, and a morphless suitless doesn't sound impossible with it.

osse101 (commented):
My vague idea of the draygon fight is
- 1 hit on the killing shinespark
- 1 hit when she is close to a wall since you are aiming for goop patterns and it is hard to time OR you are blind sparking swoops and need more leniency
- 1 leniency for Very Hard because it is very easy to fumble the spark, or get hit while trying to short charge
- And HiJump improves damage effeciency so it is easier to finish in 3 sparks.  A low health Draygon is when things get out of control.

kjbranch:
After sparking into Draygon, he cornered me, and my options to escape were Gravity + HiJump, Morph, or pause abuse. Then I take another hit at the end of the fight. I was thinking the pause abuse would only be useful if you also have enough in reserves to then take another contact hit at the end of the kill, but I wasn't thinking about the Shinespark frames as well. I think I'll remove the pause abuse to reduce a hit.

region/maridia/inner-yellow/Plasma Spark Room.json
kjbranch: Good catch. I bet there are a lot of these problems in the project. I wonder if it would be possible to search for all strats that take damage and spark - then i could look through and verify them.

kjbranch:
I tried to follow your feedback on Dragon Shinespark kill, but I'm not sure if I got it exactly right. Let me know what you think

osse101 (approved):

PR #1754, opened 2024-11-14
Beta Power Bomb Room: in-room Speed Kill

dragonsinth:


region/brinstar/red/Beta Power Bomb Room.json
```diff
@@ -305,6 +305,28 @@
...
+              "openEnd": 0
+            }
+          }
+        ]}
+      ],
+      "clearsObstacles": ["A"],
+      "note": [
+        "Manipulate the Sidehoppers to a wall and synchronize their jumps.",
+        "Roll away just as they jump towards Samus to the opposite wall.",
+        "Shinecharge to the right, and jump into the Sidehoopers when Samus acquires blue speed."
```
dragonsinth, line 327:
```suggestion
        "Shinecharge towards the Sidehoopers, jumping into them when Samus acquires blue speed."
```

kjbranch (commented):
I dont think this is useful. If you have morph and the skill to get a 16 tile shortcharge, you can just kill the hoppers.

dragonsinth:
Fair. It's probably more than 16 tho, my shortcharges are ok but not wonderful (I don't stutter).

region/brinstar/red/Beta Power Bomb Room.json
```diff
@@ -305,6 +305,28 @@
...
+          "Morph",
+          "canCameraManip",
+          {
+            "canShineCharge": {
+              "usedTiles": 16,
+              "openEnd": 0
+            }
+          }
+        ]}
+      ],
```
osse101, line 322:
```suggestion
        "Morph",
        "canCameraManip",
        {
          "canShineCharge": {
            "usedTiles": 16,
            "openEnd": 0
          }
        }
      ],
```

osse101 (commented):
I don't see how you synchronize the hoppers.  I can do the strat but it doesn't feel safe.  I'd rather keep 2 against the wall and kill whichever one follows me with power beam and then speed kill the remaining.  But that is most of the way towards killing all of them with beam shots.

region/brinstar/red/Beta Power Bomb Room.json
dragonsinth: oops yeah

dragonsinth:
Maybe we should scrap this in favor of a comeInShineCharging based strat?  What would that look like?

dragonsinth:
Gonna close this for now.

PR #1757, opened 2024-11-15
Reserve Pause Abuse WS, Tourian

kjbranch:
Only going for high hitting enemies. Most enemies here are pretty low damage.
Skipped all of the blue hopper hits that are taken to open the left door with a PB or missiles

### DIFF
--- a/region/tourian/main/Blue Hopper Room.json
+++ b/region/tourian/main/Blue Hopper Room.json
@@ -347,11 +347,14 @@
       "link": [1, 2],
       "name": "Take the Damage",
       "requires": [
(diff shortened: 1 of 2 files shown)


region/tourian/main/Blue Hopper Room.json
```diff
@@ -472,11 +475,14 @@
       },
       "requires": [
         "canShinechargeMovementTricky",
```
kjbranch, line 477: Thinking this should be fine? It might require a few more shinecharge frames though. There was a [2, 1] requiring 160 frames, so I left it off there.

PR #1762, opened 2024-11-17
Screw Attack Room shinecharges

blkerby:
This overhauls the shinecharge/shinespark strats in this room, removing the obstacle "C" in favor of the newer `startsWithShineCharge`/`endsWithShineCharge` strat properties used in The Moat. This allows us to express many more strats that collect the item while holding a shinecharge, as well as more ways of using the shinecharge afterward.

As part of this, I got rid of the notable "Descent and Shinespark Escape". It seemed maybe unnecessary now that we have `canShinechargeMovementTricky` and shinecharge frame leniency to express the strat difficulties. There are many different possible difficulties depending on the combination of strats used here (ranging from Very Hard to Extreme), and not all of them involve a descent since you can also enter shinecharged/shinecharging from the left door.

### DIFF
--- a/region/lowernorfair/west/Screw Attack Room.json
+++ b/region/lowernorfair/west/Screw Attack Room.json
@@ -68,11 +68,6 @@
       "id": "B",
       "name": "Bottom Bomb Blocks",
       "obstacleType": "inanimate"
-    },
-    {
-      "id": "C",
-      "name": "Shinecharge from Above",
-      "obstacleType": "abstract"
     }
   ],
   "enemies": [],
@@ -160,7 +155,6 @@
       "link": [1, 1],
       "name": "Crystal Flash",
       "requires": [
-        {"obstaclesNotCleared": ["C"]},
         {"heatFrames": 10},
         "h_canHeatedCrystalFlash",
         {"heatFrames": 10}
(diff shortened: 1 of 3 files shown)


region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -968,45 +1456,207 @@
...
+      "link": [3, 2],
+      "name": "Come In Shinecharging, Leave With Spark",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 3,
+          "openEnd": 1
+        }
+      },
+      "requires": [
+        "canShinechargeMovementComplex",
```
osse101, line 1552: The difficulty depends on your run speed.  Its mostly a problem because you are coming through a door.  Continuing to fall to B is ShinechargeMovementTricky, but I find this 1 tile break to be only annoying.

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -968,45 +1456,207 @@
...
+          "requires": [
+            "never"
+          ]
+        }
+      ]
+    },
+    {
+      "id": 39,
+      "link": [3, 2],
+      "name": "Reset Fall Speed (Quick Shot)",
```
osse101, line 1484: This works well with comeInShinecharging strats.

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -1094,43 +1744,122 @@
...
+    {
+      "link": [3, 4],
+      "name": "Come In Shinecharged, End Shinecharged (Screw Attack)",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 140
+        }
+      },
+      "requires": [
+        "canShinechargeMovementTricky",
```
osse101, line 1833:
```suggestion
        "canShinechargeMovementTricky",
        "ScrewAttack",
```

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -1270,33 +2072,124 @@
           ]
         }
       ],
-      "note": "Diagonally shinespark towards the middle door.",
-      "devNote": [
-        "Obstacle C is indicating that Samus has a shinecharge ready to be used.",
-        "Useful only if you cannot reach the door at 2."
+      "note": "Diagonally shinespark towards the middle door."
```
osse101, line 2075: You climb up and horizontal spark here.

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -1246,18 +1975,91 @@
         {"heatFrames": 60}
       ]
     },
+    {
+      "link": [4, 1],
+      "name": "Start Shinecharged, Leave Shinecharged",
+      "startsWithShineCharge": true,
+      "requires": [
+        "canShinechargeMovementTricky",
```
osse101, line 1983: Why tricky?

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -1270,33 +2072,124 @@
...
+    {
+      "link": [4, 2],
+      "name": "Start Shinecharged, Leave Shinecharged (HiJump)",
+      "startsWithShineCharge": true,
+      "requires": [
+        "HiJump",
+        "canShinechargeMovementComplex",
+        {"or": [
+          "ScrewAttack",
+          {"obstaclesCleared": ["B"]}
```
osse101, line 2152: Is it possible for B to be broken?

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -1094,43 +1744,122 @@
...
+        {"or": [
+          "canMoonfall",
+          {"and": [
+            {"heatFrames": 5},
+            {"shineChargeFrames": 5}
+          ]}
+        ]}
+      ],
+      "endsWithShineCharge": true,
+      "clearsObstacles": ["B"],
```
osse101, line 1804: Makes sense to not count A as broken

region/lowernorfair/west/Screw Attack Room.json
blkerby: I tried it the other day but wasn't able to make it onto the door ledge that way. I assume I'm missing something

region/lowernorfair/west/Screw Attack Room.json
blkerby: Yeah I saw it's possible to break the one block without sliding off the ledge: https://videos.maprando.com/video/1484. Makes sense that it should be easier at lower run speeds, since you would have more frames where you would be in contact with the bomb block before falling off. I think collision oscillation might play a role too, not sure. Do you think something needs to change in the requirements here?

region/lowernorfair/west/Screw Attack Room.json
blkerby: Yeah, I was thinking about how the doorframe CWJ wouldn't work with only the left blocks broken. We could make a separate obstacle for left vs. right block, but I didn't see a scenario where it would matter

region/lowernorfair/west/Screw Attack Room.json
blkerby: Not sure, I think at one point I had in mind that carrying the shinecharge through the item collection should stay Expert like it was with the notable. I changed my mind but then probably forgot I put these in. Downgraded them to `canShinechargeMovementComplex` now.

region/lowernorfair/west/Screw Attack Room.json
blkerby: Good catch, looks like the note got copy-pasted by mistake, removed it now

region/lowernorfair/west/Screw Attack Room.json
blkerby: The idea was in case you fell through with temporary blue from above. I guess there's not quite enough shinecharge frames remaining, but it's only 5 short so maybe it's conceivable if something changes here later. Also someone might want to add an option someday for longer shinecharge timer (e.g. 240 frames); there would be lots of stuff missing in that case, but I figure no harm in having stuff be more complete where we can

region/lowernorfair/west/Screw Attack Room.json
osse101: It would only be situations where you have forced high run speed entering the room like leaving SpeedBoosterHall. Otherwise the leniency should allow you to control your speed well enough.  So I don't think its a concern.

Leaving SpeedBoosterHall with full room run speed is in Hard with HeroShot, when it should be Very hard right2left and either VH or EXP left2right.

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -240,19 +234,216 @@
...
+          ]
+        }
+      ]
+    },
+    {
+      "link": [1, 2],
+      "name": "Come In Shinecharged, Leave With Spark (HiJump, Screw Attack)",
+      "entranceCondition": {
+        "comeInShinecharged": {
+          "framesRequired": 110
```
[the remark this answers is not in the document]
blkerby: Yeah, now that we have a `shineChargeFrames` logical requirement, I've been thinking that to simplify things it may make sense to remove `framesRequired` and `framesRemaining` from the schema of `comeInShinecharged` and `leaveShinecharged`, as the way those interact can be a little confusing. In that case, we would want to add a test to ensure that `shineChargeFrames` is not forgotten from the `requires` when using those entrance/exit conditions.

region/lowernorfair/west/Screw Attack Room.json
blkerby: Oh of course, that makes sense. Added a strat now for this. I found it can work well even if you don't slide off the ledge: https://videos.maprando.com/video/1767

When I was trying it before, I was doing it wrong, trying to spin jump and then mid-air morph-unmorph.

PR #1763, opened 2024-11-18
Halfie Climb grapple teleport into grapple clip

blkerby:
Some grapple teleport strats in Halfie Climb that were missed in the first pass. These all involve getting into the air space in the top left and then grapple clipping to bypass the door shell.

Videos:
- original find by SamLittlehorns: https://videos.maprando.com/video/1470
- from Moat to top-right door (on-camera): https://videos.maprando.com/video/1648
- from Moat to bottom-right door (off-camera): https://videos.maprando.com/video/1649
- from Red Brin Firefleas to top-right door (on-camera): https://videos.maprando.com/video/1651
- from Red Brin Firefleas to bottom-right door (off-camera): https://videos.maprando.com/video/1652

### DIFF
--- a/region/maridia/inner-pink/Halfie Climb Room.json
+++ b/region/maridia/inner-pink/Halfie Climb Room.json
@@ -3385,22 +3385,6 @@
       ]
     },
     {
-      "link": [4, 1],
-      "name": "Grapple Teleport",
-      "entranceCondition": {
-        "comeInWithGrappleTeleport": {
-          "blockPositions": [
-            [5, 3],
-            [7, 2]
-          ]
-        }
-      },
-      "requires": [],
-      "note": [
-        "After teleporting, allow Samus to swing right before releasing Grapple."
-      ]
-    },
-    {
       "id": 116,
       "link": [4, 1],
       "name": "Grapple Teleport Door Lock Skip",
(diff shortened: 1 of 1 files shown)


blkerby:
Added more strats that swing out right of the grapple blocks instead of getting stuck in the air space behind them to the left. Videos for these:

- from Red Brin Firefleas to top-right door (on-camera): https://videos.maprando.com/video/1655
- from Red Brin Firefleas to bottom-right door (off-camera): https://videos.maprando.com/video/1654
- from Moat to top-right door (on-camera): https://videos.maprando.com/video/1657
- from Moat to top-right door (on-camera): https://videos.maprando.com/video/1658

region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -3333,9 +3385,25 @@
       ]
     },
     {
-      "id": 116,
       "link": [4, 1],
       "name": "Grapple Teleport",
+      "entranceCondition": {
```
osse101, line 3390: Seems pointless if there is no doorLockBypass to this strat.
Do `comeInWithGrappleTeleport`s have an implicit "just don't hold the button and you stay at the door"?

osse101 (commented):
The fall out to the right strats appear closer to [3,4] instead of [3,1].  Since you have grapple, you can freely move between the two.  But its more about if you were looking up whether it is possible to grapple teleport from 3 where you would look.  I don't have a problem with listing it as [3, 1].

Some grapple teleports make you do weird movements.  At this point it is moreso expected that you look up how to do a GT that puts you in a wall so we could be better off keeping the number of GT notables small with that in mind.

PR #1765, opened 2024-11-18
Croc Escape G-Mode Setup

kjbranch:


region/norfair/west/Crocomire Escape.json
```diff
@@ -254,6 +254,192 @@
...
+          {"ammo": {"type": "Super", "count": 2}},
+          {"ammo": {"type": "PowerBomb","count": 2}},
+          {"enemyDamage": {
+            "enemy": "Dragon",
+            "type": "fireball",
+            "hits": 5
+          }},
+          {"and": [
+            "h_lavaProof",
+            {"and": [
```
osse101, line 281:
```suggestion
            {"or": [
```

region/norfair/west/Crocomire Escape.json
```diff
@@ -254,6 +254,192 @@
...
+      "flashSuitChecked": true,
+      "note": [
+        "Lure the Geruta to the left door. This may require many jumps to get it to swoop up onto the ledge and not get stuck.",
+        "To safely pass the Geruta without a hit, jump left below it, to get it to swoop left before going up."
+      ],
+      "devNote": "It's possible to avoid the item, but the Geruta will get stuck on it and usually require many swoops to get past it."
+    },
+    {
+      "link": [2, 1],
+      "name": "G-Mode Setup - Get Hit By Geruta, Heatproof, Frozen Geruta Platform",
```
osse101, line 411: The name and description can be confused with riding the geruta across while freezing it repeatedly.

region/norfair/west/Crocomire Escape.json
```diff
@@ -254,6 +254,192 @@
...
+          "canInsaneJump",
+          "Ice",
+          "h_lavaProof",
+          {"enemyDamage": {
+            "enemy": "Geruta",
+            "type": "contact",
+            "hits": 1
+          }},
+          {"and": [
+            "canNeutralDamageBoost",
```
osse101, line 347: I'm not completely sure how this works.  I think the damage boost is for difficulty placement.  And avoiding a second(the first) fireball takes a room entry that is a higher difficulty - or its assumed you kill the other dragon while in the lava.

region/norfair/west/Crocomire Escape.json
kjbranch: I forgot about that one, but I'm not including it, because it requires many more swoops and hits than just crossing the room, as it can take a while to get it in the top left region

PR #1768, opened 2024-11-19
Rename canIframeSpikeJump to canUseIFrames, Cleanup

kjbranch:
Standardized the spelling to `i-frame`, since there were many variations. Let me know if you think the tech name or standardization should be done differently. 

I was thinking it might make sense to rename `canHitbox`, since it is obscurely named and I'm not really sure you are manipulating the enemy hitboxes - thoughts?

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -368,9 +368,8 @@
         {
           "name": "h_IBJFromThorns",
           "requires": [
-            "canIframeSpikeJump",
+            "canUseIFrames",
             "canJumpIntoIBJ",
-            "canCarefulJump",
             {"thornHits": 1},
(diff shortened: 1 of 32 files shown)


osse101 (approved):
TourianEscape3 has the same pass through pirates in the other direction but is lacking a branch for armpumping/dboosting through the middle pack.  

Hitboxing comes from a general gaming term for passing through hitboxes without taking damage.  But in SM I've always heard it applied to using plasma to run through enemies.  Now that hitbox-controllers exist, the term sounds weird.  So its a community name but you can try to come up with something catchy.

GJ finding all of these i-frame uses.

Issue #1540, opened 2024-06-09
Tech: canCrumbleQuickDrop vs canQuickCrumbleEscape


kjbranch:
There was some discussion of adding `canCrumbleQuickDrop` or renaming `canQuickCrumbleEscape` to it, while potentially slightly changing its use to only focus on the drop portion of the strat (in #463). This tech could then be expanded to uses like quickly falling through a crumble block while moving with a shinecharge timer.

There was some discussion about `canQuickCrumbleEscape` representing being stuck on a failed attempt, and it being more useful in representing the Mickey Mouse Multiviola ice clip's crumble escape, where `canCrumbleQuickDrop` wouldn't apply.

Would it be best to have `canCrumbleQuickDrop`, `canQuickCrumbleEscape`, or both?

osse101:
I see `canQuickDrop` being the community name for the technique and `canCrumbleQuickDrop` is a bit of a nicer name.  `canMomentumConservingTurnaround` was something I tried to attach to falling down opening doors, but a quick drop is the same movements.  And quick dropping on top of actively breaking blocks is another case that saves frames that maybe matters somewhere; it's done on the upper left door runway of red tower in any% while waiting for a bomb to explode.

The ability to reliably quickDrop is about very hard.  The value of the tech is that it saves a decent number of frames.  And in those situations there tends to be additional factors which make the strat difficult.  early supers room has the softlock risk, mission impossible has the softlock risk, early supers shinespark is a room reset, etecoon etank is a one way.  `Early Supers Quick Crumble Escape` could be moved to `Very Hard` really and I don't think there needs to be a notable made because it is still closely tied to the general tech.

Rejumping through the crumble block in mickeymouse and etecoon-main-shaft seems unrelated to quick dropping.  Its kinda more of a canQuickRejump tech like canBombJumpWaterEscape.

PR #1769, opened 2024-11-19
Rename canQuickCrumbleEscape to canQuickDrop

kjbranch:
This is a rename to help make this tech useful in more locations - specifically shinecharges through a vertical door.

This was discussed most recently in #1540.

Do we want to name it `canQuickDrop`? 
Or should it even just be combined with `canMomentumConservingTurnaround`?

After this is settled, I can look into adding the tech to a few more locations that were brought up in the issue.

### DIFF
--- a/tech.json
+++ b/tech.json
@@ -637,10 +637,13 @@
         },
         {
           "id": 47,
-          "name": "canQuickCrumbleEscape",
+          "name": "canQuickDrop",
           "techRequires": [],
           "otherRequires": [],
-          "note": "The combination of a crumble quick drop, and landing on a lower surface and jumping back over the crumble block before it re-forms."
+          "note": [
+            "Using a momentum conserving turnaround on a crumble block, an opening doorshell, or any other breaking block, to preserve Samus' momentum.",
+            "This is often used when traversing with a shinecharge, or to quickly fall through a crumble block and jump back through it before it re-forms."
+          ]
         },
         {
           "id": 48,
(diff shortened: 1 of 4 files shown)


tech.json
```diff
@@ -637,10 +637,13 @@
         },
         {
           "id": 47,
-          "name": "canQuickCrumbleEscape",
+          "name": "canQuickDrop",
           "techRequires": [],
           "otherRequires": [],
-          "note": "The combination of a crumble quick drop, and landing on a lower surface and jumping back over the crumble block before it re-forms."
+          "note": [
+            "Using a momentum conserving turnaround on a crumble block, an opening doorshell, or any other breaking block, to preserve Samus' momentum.",
```
osse101, line 644: Using a tech name implies its a requirement.  It'd be better to redeclare the action.
While these are separate techs that is.

osse101 (approved):
The think quickDrop can be combined with momentumConservingTurnaround.
The only problem is that quickDrop is a really good recognizable name.

There are subtleties to working with crumble blocks and opening doors but I think they are higher level than this tech or not important.

kjbranch:
I was thinking this might make sense as an extension of `canMomentumConservingTurnaround`, and then I kept the reference in the description. 

While these will likely stay in the same difficulty, a quick drop is often a bit more risk - since the crumble will reform, or a bit more difficulty - since you might need to turnaround multiple times on a doorshell.

Once this is merged, I think #1540 can be closed. All of the quick drop examples mentioned there already have `canQuickDrop`. It mentioned a quick rejump on the Mickey Mouse Clip and the Green Brin Bottom Ice Clip, but those are already notable. I could add an insane jump to them though, if you think that is warranted.

PR #1773, opened 2024-11-20
Enemies Cleanup - Remove dropRequires, farmCycles

kjbranch:
These farming properties are not used, and the new farming methods won't be in this format. Is it worth removing them or should we keep them in case it has some use as a reference?

### DIFF
--- a/region/brinstar/green/Early Supers Room.json
+++ b/region/brinstar/green/Early Supers Room.json
@@ -68,14 +68,7 @@
       "groupName": "Early Supers Zeb",
       "enemyName": "Zeb",
       "quantity": 1,
-      "homeNodes": [1],
-      "farmCycles": [
-        {
-          "name": "Crouch over spawn point",
-          "cycleFrames": 120,
-          "requires": []
-        }
-      ]
+      "homeNodes": [1]
     },
     {
       "id": "e2",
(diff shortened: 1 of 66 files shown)


osse101:
I don't think the farmCycle values are accurate enough to preserve for reference sake.

The original model provided descriptions of the room features so that things like farm strats could have been automatically figured out by visiting the node.  But all of the details are on connections now and not shared with other links.

Enemies aren't used either.  I find it to be a handy reference when adding enemyDamage or enemyKill events.  And I think its more likely to model single random drops by using `enemies` than by adding enemy farm strats to every strat, if finer item drop details become desired.

osse101 (approved):

blkerby:
Looks good, I agree it makes sense to remove these.

As a side note, the concept of a `farmCycle` with a measured `cycleFrames` is appealing, and I'm interested in bringing it back in the future as a strat property, and generalizing it to apply to the situation where a room reset is involved and allowing multiple enemy kills to be involved in each cycle. It could help us more accurately model farming strats, allowing them to properly interact with patience tech and account for whether the adjacent room is heated. Because otherwise it's a somewhat overwhelming amount of possible combinations to try to model manually. Automating the computation would come with a cost of making it harder for players to interpret the logic, but that can be eased by providing a calculator tool like was done with the boss calculator.

PR #1781, opened 2024-11-21
More ways to open Caterpillar Room green gate with G-mode

blkerby:
This also fixes a typo in a few places.

### DIFF
--- a/region/brinstar/red/Caterpillar Room.json
+++ b/region/brinstar/red/Caterpillar Room.json
@@ -2170,7 +2170,8 @@
         {"or": [
           "h_canArtificialMorphMovement",
           "h_canArtificialMorphBombThings"
-        ]}
+        ]},
+        "canOffScreenMovement"
       ],
       "exitCondition": {
         "leaveWithGMode": {
@@ -2253,22 +2254,9 @@
         "Exit G-Mode while under the gate to open it without a Super.",
         "The camera will not scroll through the morph tunnel, but Samus will still be visible;",
         "position where Samus appears to be standing on air one tile to the right of the ledge.",
+        "It is possible to test the location, as there is a block on the ceiling one tile left of the gate.",
         "With Morph, Samus can fix the camera by going through the morph tunnel."
       ]
-    },
-    {
-      "link": [7, 6],
-      "name": "Carry G-Mode Up Elevator",
-      "requires": [
-        "canEnterGMode"
-      ],
-      "exitCondition": {
-        "leaveWithGMode": {
-          "morphed": false
-        }
-      },
-      "flashSuitChecked": true,
-      "note": "Exit G-Mode near the door to pass the gate."
     }
   ],
   "nextStratId": 96,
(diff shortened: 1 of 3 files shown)


kjbranch:
The camera wont follow through the morph tunnel, so I think they would need `canOffScreenMovement` to open the gate.

blkerby:
> The camera wont follow through the morph tunnel, so I think they would need `canOffScreenMovement` to open the gate.

Good point. Split these off into separate strats, to be able to put `canOffScreenMovement` on the version that opens the gate. 

I tested the strat now, and Samus is still visible, with a nice visual cue for where to stand. Here's a video: https://videos.maprando.com/video/1818. But makes sense that it still needs `canOffScreenMovement`.

This could be a good use case for node-local state if we ever add support for that, to be able to reduce the amount of duplication in strat requirements. An obstacle could work too though it's a bit overkill; or a G-mode junction at the right can work. Let me know if you'd rather have it done one of those ways.

region/brinstar/red/Caterpillar Room.json
```diff
@@ -2214,6 +2323,27 @@
...
+      "requires": [
+        "Morph",
+        "canOffScreenMovement"
+      ],
+      "clearsObstacles": ["A"],
+      "flashSuitChecked": true,
+      "note": [
+        "Exit G-Mode while under the gate to open it without a Super.",
+        "The camera will not scroll through the morph tunnel, but Samus will still be visible;",
+        "position where Samus appears to be standing on air one tile to the right of the ledge."
```
kjbranch, line 2345: Something like this could be added to the others as well, though Samus will only sometimes have morph with them.
```suggestion
        "position where Samus appears to be standing on air one tile to the right of the ledge.",
        "Samus can then use the morph tunnel to fix the camera."
```

kjbranch (dismissed):

kjbranch:
A g-mode morph junction at the tunnel would actually work nicely. Then the strats could go up to the elevator, right, or right and open the gate

osse101 (dismissed):
I wanted to change 5,6 to 5,7.  But actually 5,6 is missing a offscreenMovement requirement.

region/brinstar/red/Caterpillar Room.json
```diff
@@ -2214,6 +2230,45 @@
...
+      "note": [
+        "Exit G-Mode while under the gate to open it without a Super.",
+        "The camera will not scroll through the morph tunnel, but Samus will still be visible;",
+        "position where Samus appears to be standing on air one tile to the right of the ledge.",
+        "With Morph, Samus can fix the camera by going through the morph tunnel."
+      ]
+    },
+    {
+      "link": [7, 6],
+      "name": "Carry G-Mode Up Elevator",
```
blkerby, line 2261: I'm probably missing something, but I don't understand what this strat is for. What would be the purpose of going to this G-mode morph tunnel junction only to then go up the elevator? It looks like there are already strats that go from each of the doors to the elevator (including one from node 5).

In this case, it could make more sense to define node 7 as being at the right door rather than in the morph tunnel, since its only purpose is to either go past the gate or to open the gate? At least with how it's written now, I'm not seeing a purpose to go left from node 7?

kjbranch: i was mostly thinking because this is where the off screen movement starts, but I don't think it should really matter. I think [7, 6] should be removed

PR #1789, opened 2024-11-22
Remove $id in schema

kjbranch:
Removed all `$id` references in the schema files.
Converted the top level `id` in each schema file to a `$id`.

Is this what was wanted?


blkerby (approved):
This looks good. I confirmed that schema validation still works (and catches errors), and schema-related functionality in VSCode (e.g. tooltips and auto-complete) also appears unaffected.

PR #1794, opened 2024-11-23
Migrate json schema validation to use 'referencing' instead of RefResolver

blkerby:
The validation script has been giving warnings that RefResolver is deprecated in favor of the `referencing` package. This PR migrates over to the new, recommended package.

What prompted this is that since #1789 fixed the top-level `$id` property, the RefResolver started recognizing it and fetching the schema files using HTTP GET requests, treating the `$id` as a URL, and using the responses as the schemas to apply  in validation. This initially appeared to work because those URLs actually are valid references to the schema files on GitHub. But once we start introducing schema changes locally, then it doesn't work because it's using the schemas from the master branch instead of the local copies. This was considered a bad behavior of the RefResolver, and it no longer happens after switching to the `referencing` implementation: https://python-jsonschema.readthedocs.io/en/stable/referencing/#other-key-functional-differences.

### DIFF
--- a/tests/asserts/validate.py
+++ b/tests/asserts/validate.py
@@ -7,8 +7,10 @@ import json
 import re
 from pathlib import Path
 import jsonschema.validators
+import referencing
+from referencing import Registry, Resource
 from jsonschema.validators import Draft7Validator
-from jsonschema import validate, RefResolver
+from jsonschema import validate
 from jsonschema import exceptions as JSONSchemaExceptions
 
 schemas = {}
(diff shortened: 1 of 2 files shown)


tests/asserts/validate.py
```diff
@@ -7,8 +7,10 @@
 import re
 from pathlib import Path
 import jsonschema.validators
+import referencing
```
osse101, line 10: These two aren't used since they are being `from`ed too.

osse101 (commented):
Seems to work 🤷

tests/asserts/validate.py
blkerby: True. Removed now

PR #1795, opened 2024-11-24
More tests

blkerby:
This depends on #1794, so it should be merged first.

This adds tests to check that a `shinespark` logical requirement occurs in every strat with a `leaveWithSpark` exit condition, and that every `or` branch is covered. This turned up three strats that were missing a `shinespark`; these are fixed here.

Tests are also added to check for a `heatFrames` requirement for every strat in heated rooms, again ensuring that every `or` branch is covered. Here, "h_heatProof" and a number of helpers that indirectly include heat frames are also counted. Exceptions are carved out for common situations where it is normal for no heat frames to be included (such as "leaveWithRunway" at a door node, "comeInWithGMode", and grapple teleports that bypass a door lock). For all other "false positive" cases, I'm including a `{"heatFrames": 0}` to satisfy the test and also as a way of documenting that the lack of heat frames is intentional. 

The heat frames test turned up a few cases where heat frames were missing. The only really egregious case was the Cathedral "Tricky Platforming (Come In Running)" strat from 1 to 4, which I apparently messed up long ago. The other cases only involve relative small amounts of heat damage. One of the Multiviola ice clip strats (from 9 to 4) was missing a "h_heatProof" requirement, though it shouldn't be impactful since there's already a "h_heatProof" requirement in the 6 to 9 strat.

I also deleted a few useless "Leave Normally" strats here rather than adding `{"heatFrames": 0}` to them. The remaining "Leave Normally" strats (in unheated rooms) can be cleaned up in a separate pull request, when I'm planning to document the implicit leaveNormally and comeInNormally strats and add boolean properties to toggle them.

### DIFF
--- a/region/lowernorfair/east/Lower Norfair Spring Ball Maze Room.json
+++ b/region/lowernorfair/east/Lower Norfair Spring Ball Maze Room.json
@@ -429,7 +429,7 @@
       "link": [3, 3],
       "name": "Open Door",
       "requires": [
-        {"heatFrames": 0}
+        {"heatFrames": 20}
       ],
       "clearsObstacles": ["C"]
     },
--- a/region/lowernorfair/east/Lower Norfair Spring Ball Maze Room.json
+++ b/region/lowernorfair/east/Lower Norfair Spring Ball Maze Room.json
@@ -292,15 +292,6 @@
       ]
     },
     {
-      "id": 7,
-      "link": [2, 2],
-      "name": "Leave Normally",
-      "exitCondition": {
-        "leaveNormally": {}
-      },
-      "requires": []
-    },
-    {
       "id": 8,
       "link": [2, 2],
       "name": "Shinespark",
@@ -437,7 +428,9 @@
       "id": 17,
       "link": [3, 3],
       "name": "Open Door",
-      "requires": [],
+      "requires": [
+        {"heatFrames": 0}
+      ],
       "clearsObstacles": ["C"]
     },
     {
--- a/tests/asserts/keywords.py
+++ b/tests/asserts/keywords.py
@@ -253,6 +253,28 @@ def check_node_covered_in_unlocks_doors(strat, node_id):
     missing_types = {"missiles", "super", "powerbomb"}.difference(types)
     return missing_types
 
+def check_shinespark_req(req):
+    if isinstance(req, dict):
+        if "shinespark" in req:
+            return True
+        if "and" in req:
+            return any(check_shinespark_req(v) for v in req["and"])
+        if "or" in req:
+            return all(check_shinespark_req(v) for v in req["or"])
+
+def check_heat_req(req):
+    if isinstance(req, str):
+        if req in ["h_heatProof", "h_canHeatedCrystalFlash", "h_canHeatedLavaCrystalFlash", "h_LowerNorfairElevatorDownwardFrames",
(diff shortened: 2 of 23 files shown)


blkerby:
Cathedral Entrance 4->5 with no heat frames also seemed a bit weird, though it may be ok since it seems the heat frames are accounted for in getting to 4. The asymmetry doesn't seem right (i.e. in how the 4->3 requires heat frames while 4->5 does not); probably node 4 should just be removed and an obstacle used instead but I didn't want to try to tackle a room overhaul like that here.

tests/asserts/keywords.py
```diff
@@ -718,6 +743,23 @@ def process_strats(src, paramData):
                                 msg = f"🔴ERROR: Strat ID {strat_id} is not less than nextStratId ({next_strat_id}):{stratRef}"
                                 messages["reds"].append(msg)
                                 messages["counts"]["reds"] += 1
+                        if heated and not check_heat_req({"and": strat["requires"]}):
+                            if fromNode == toNode and "leaveWithRunway" in strat.get("exitCondition", []):
+                                # Ok since there is implicit heat frames in leavesWithRunway, and it is normal
+                                # if no explicit ones to be present for a strat going from the door to itself.
+                                pass
+                            elif "comeInWithGMode" in strat.get("entranceCondition", []) or "gModeRegainMobility" in strat:
```
osse101, line 751: A gmode strat off of a junction probably will have heat frames when you end gmode.  Junction to junction isn't likely.

There could be a case for Gmode->Ridley->Leave through open door to save the boss flag.  GT too.  ~~But I've also heard the flag is set when drops appear and that would take releasing GMode too 🤔~~ 
I don't know how you do the fight damageless, but you keep the flag on boss kill.  There's also time to end gmode and exit the room before the (open)door closes if you had to do it that way.

region/lowernorfair/east/Lower Norfair Spring Ball Maze Room.json
```diff
@@ -437,7 +428,9 @@
       "id": 17,
       "link": [3, 3],
       "name": "Open Door",
-      "requires": [],
+      "requires": [
+        {"heatFrames": 0}
```
osse101, line 432:
```suggestion
        {"heatFrames": 20}
```
[3, 7] that opens the door first would add two turnarounds.  It may mess with your alcoon dodge strat too so its something to look at for the room later.

osse101 (dismissed):

tests/asserts/keywords.py
blkerby: Yeah, it's a good point, the "comeInWithGMode" exception is a little too broad since some of these strats should be having heat frames. I changed it now to carve out the G-mode exceptions more narrowly, only applying now in these cases:
1. Strat has both "comeInWithGMode" and "leaveWithGMode", so it's spending the entire time in G-mode.
2. Strat has "gModeRegainMobility", which also will be spending the entire time in G-mode.
This didn't cause any new strats to fail the test. I think this is expected since the G-mode applications in heated rooms have not yet been done.

kjbranch: I dont think there are any g-mode usage strats in heated rooms yet.

It could also be fine to have strats that come in with g-mode and end at a item node or a g-mode junction, but it's fine just manually adding 0 frames to those as well - upper Norfair rooms are usually much simpler, so I dont think there will be a lot of g-mode junctions, and there arent many items in heated rooms.

kjbranch (approved):
The logic looks good to me. I didn't look through the python stuff.

PR #1796, opened 2024-11-24
Obstacle tests for resetRoom

blkerby:
This depends on #1794, so it should be merged first. I also expect this to have conflicts with #1784, so I can resolve this once that is merged.

A while ago we had a soundness bug in the past in PCJR, where logic thought you could enter shinecharged to break the speed blocks, use the Mella farm to refill, then use that energy to spark up. The problem was that a `resetRoom` requirement on its own does not affect the obstacle state, when in reality basically it always should. We fixed that specific case already but there has still been potential for the same type of error to happen in other places. To solve this problem, this PR adds a test to check that every strat with a `resetRoom` mentions every obstacle in the room in either the `resetsObstacles` or `clearsObstacles` of that strat. Usually the obstacles will be reset since that's what resetting the room does, but in some cases they may get cleared again as part of executing the strat, so that's why it allows either one, as long each obstacle is covered in one or the other. There were many cases where this test did not pass, and this PR fixes all of them.

The other thing is that there were some strats with an exit condition and also a `resetsObstacles` or `clearsObstacles`. A strat with an exit condition will have left the room at the end of the strat, so it does not make sense for it to be affecting the obstacle state. As far as the randomizer is concerned, these were probably harmless since it would have been ignoring these anyway, but it can create confusion so it seemed good to clean this up: the PR adds a test to check that an exit condition does not occur in the same strat with a `resetsObstacles` or `clearsObstacles`. The case where a strat has an exit condition is also an exception where the previous test does not apply: i.e., even if a strat has a `resetRoom` requirement, if it has an exit condition then it should not include `resetsObstacles` or `clearsObstacles`.

### DIFF
--- a/region/brinstar/red/Caterpillar Room.json
+++ b/region/brinstar/red/Caterpillar Room.json
@@ -1382,11 +1382,12 @@
       "name": "Zero Farm",
       "requires": [
         {"resetRoom": {
-          "nodes": [3]
+          "nodes": [2, 3]
         }},
         {"refill": ["PowerBomb"]}
       ],
-      "resetsObstacles": ["A", "B"],
+      "resetsObstacles": ["A"],
+      "clearsObstacles": ["B"],
       "devNote": "FIXME: Other nodes could be used to reset the room, with additional requirements."
     },
     {
@@ -2048,7 +2049,8 @@
         {"resetRoom": {
           "nodes": [5]
         }},
-        {"refill": ["Energy", "Super", "PowerBomb"]}
+        {"refill": ["Energy", "PowerBomb"]},
+        {"partialRefill": { "type": "Super", "limit": 10 }}
       ],
       "resetsObstacles": ["A", "B"],
       "devNote": "It is possible to roll off of the platform above to avoid a mid-air morph."
--- a/region/brinstar/red/Caterpillar Room.json
+++ b/region/brinstar/red/Caterpillar Room.json
@@ -1386,6 +1386,7 @@
         }},
         {"refill": ["PowerBomb"]}
       ],
+      "resetsObstacles": ["A", "B"],
       "devNote": "FIXME: Other nodes could be used to reset the room, with additional requirements."
     },
     {
@@ -1812,6 +1813,7 @@
         }},
         {"refill": ["Energy", "Super"]}
       ],
+      "resetsObstacles": ["A", "B"],
       "devNote": "FIXME: Other nodes could be used to reset the room, with additional requirements."
     },
     {
@@ -2048,6 +2050,7 @@
         }},
         {"refill": ["Energy", "Super", "PowerBomb"]}
       ],
+      "resetsObstacles": ["A", "B"],
       "devNote": "It is possible to roll off of the platform above to avoid a mid-air morph."
     },
     {
--- a/region/crateria/west/Green Pirates Shaft.json
+++ b/region/crateria/west/Green Pirates Shaft.json
@@ -311,7 +311,8 @@
           "type": "Missile",
           "limit": 20
         }}
-      ]
+      ],
+      "resetsObstacles": ["A"]
     },
     {
       "id": 77,
@@ -331,6 +332,7 @@
           "limit": 6
         }}
       ],
+      "clearsObstacles": ["A"],
       "devNote": "Killing the beetoms will also clear the top pirate for small health gain."
     },
     {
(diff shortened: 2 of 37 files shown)


blkerby:
> It's worth noting that `resetRoom` is not clearly defined and is being used, in several instances, in a way which Map Rando isn't supporting. I plan to look into fixing those problems soon.

Yeah, I think the earlier interpretation was that resolving the `resetRoom` included requirements for getting to the doors where the room could be reset; but that potentially requires a full traversal of the room graph, and implementing that would be orders of magnitude more computation than any any other logical requirement type. So in the randomizer we never supported that, and we've shifted to more of an interpretation that `resetRoom` should only include the requirements for unlocking the door, and for reentering the door on the other side; requirements for getting to the door would need to be included explicitly as part of the strat requirements. But the requirements for reentering the door on the other side is something that has been neglected, and the randomizer is currently just treating it as free. This can create soundness problems in case the other side is Tourian Blue Hoppers or Mickey Mouse left door, for instance, so we'll want to sort that out eventually. I guess usually it just works out that you have another door that you can use instead, so it hasn't resulted in an unbeatable seed as far as we know.

region/crateria/west/Green Pirates Shaft.json
```diff
@@ -1341,6 +1344,7 @@
           "nodes": [2, 3, 4]
         }}
       ],
+      "resetsObstacles": ["A"],
```
kjbranch, line 1347: This strat needs some work, but it's out of scope. I'll see what I can do when I work on resetRoom

region/lowernorfair/east/Three Musketeers' Room.json
```diff
@@ -146,7 +146,8 @@
           "limit": 10
         }},
         {"partialRefill": {"type": "Super", "limit": 4}}
-      ]
+      ],
+      "resetsObstacles": ["A"]
```
kjbranch, line 150: The kihunters are A. Not sure if this should be clears or resets. Since youre heatproof maybe clears would be better?

region/maridia/inner-yellow/Plasma Room.json
```diff
@@ -219,7 +219,8 @@
           "nodes": [1]
```
kjbranch: space+plasma should be an acceptable option as well.

region/norfair/east/Cathedral.json
```diff
@@ -727,7 +727,8 @@
         }},
         "h_heatProof",
         {"refill": ["Energy", "Missile", "Super"]}
-      ]
+      ],
+      "resetsObstacles": ["A"]
```
kjbranch, line 734: A is the geruta, with heatproof it could probably be cleared. not sure which is better

region/tourian/main/Metroid Room 4.json
```diff
@@ -127,6 +127,7 @@
         ]},
         "h_canCrystalFlash"
       ],
+      "resetsObstacles": ["A"],
```
kjbranch, line 130: Not liking this one. I'm thinking the resetRoom is instead supposed to be a come in normally option

kjbranch (commented):
This review is ignoring the python stuff and the recent commit

region/brinstar/red/Caterpillar Room.json
```diff
@@ -2048,6 +2050,7 @@
         }},
         {"refill": ["Energy", "Super", "PowerBomb"]}
       ],
+      "resetsObstacles": ["A", "B"],
```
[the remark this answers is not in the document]
blkerby: Ok, lowered to 10 for now. Could be lowered further as needed.

region/lowernorfair/east/Three Musketeers' Room.json
blkerby: Done.

region/norfair/east/Cathedral.json
blkerby: Yeah for this farm I think I had in mind just farming the one Sova near the door, since the Geruta can be a pain to deal with if you don't have a good beam. So probably leave the Geruta reset I guess? I notice there's not even any strat in logic to kill it with a weak beam, so that should probably be looked into at some point.

region/tourian/main/Metroid Room 4.json
blkerby: Ok, split that off into a separate strat.

region/brinstar/red/Caterpillar Room.json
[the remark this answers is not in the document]
kjbranch: Looks like you missed that this clears B, but maybe it doesnt matter, since you can reset the room one last time and not go to the bottom

kjbranch (approved):

region/maridia/inner-yellow/Plasma Room.json
blkerby: Done.

PR #1801, opened 2024-11-27
Tweak auto-formatting script for more compactness

blkerby:
This makes some adjustments to the autoformatting script, allowing more values to be formatted on a single line:
- Character limit for a single-line value increased from 50 to 80
- Relaxed requirements on data types allowed to nest inside a value formatted on one line.
- Properties of "unlocksDoors" are now excluded from the rules about certain field names (e.g. "requires") being disallowed from being a one-liner.

The most common things impacted by these changes are "unlocksDoors", "canShineCharge", and "enemyDamage". I wasn't sure about "enemyDamage" being on one line; I think it maybe looks ok, but if it isn't liked, it could be added to the `non_one_line_keys`.

### DIFF
--- a/scripts/format_json.py
+++ b/scripts/format_json.py
@@ -3,65 +3,84 @@
 import json
 
 # Keys whose values should always be expanded to separate lines
-non_one_line_keys = {"requires", "techRequires", "otherRequires", "and", "or", "to"}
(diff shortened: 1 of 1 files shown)


osse101 (approved):
Looks nicer to me
- unlock doors does become multiline on the complex cases
- explicitWeapons can have multiple options, but they are usually fully described by the strat name and strat alternatives.

PR #1806, opened 2024-12-01
Refine Pillar Room shinespark strats

blkerby:


region/lowernorfair/east/Pillar Room.json
```diff
@@ -608,17 +695,131 @@
...
+      "link": [2, 1],
+      "name": "Come In Shinecharging, Shinespark",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 3,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canShinechargeMovement",
```
osse101, line 750:
```suggestion
        "canShinechargeMovement",
        "canHorizontalShinespark",
```

region/lowernorfair/east/Pillar Room.json
```diff
@@ -608,17 +695,131 @@
...
+      "name": "Come In Shinecharged, Shinespark",
       "entranceCondition": {
         "comeInShinecharged": {}
       },
       "requires": [
-        {"shineChargeFrames": 15},
-        "canMidairShinespark",
-        {"shinespark": {"frames": 78, "excessFrames": 4}},
-        {"heatFrames": 210}
+        "canShinechargeMovement",
```
osse101, line 799:
```suggestion
        "canShinechargeMovement",
        "canHorizontalShinespark",
```

region/lowernorfair/east/Pillar Room.json
```diff
@@ -608,17 +695,131 @@
...
+        {"or": [
+          {"and": [
+            {"shinespark": {"frames": 79, "excessFrames": 5}},
+            {"heatFrames": 105}
+          ]},
+          {"and": [
+            {"shinespark": {"frames": 79, "excessFrames": 16}},
+            "canCarefulJump",
+            {"or": [
+              "canTrickyJump",
```
osse101, line 716: I can't tell if  this frame count lands on the pillar or ends slightly after it.  Landing on the pillar and avoiding any acid damage is an insane jump or trickyJump+walljump.

osse101 (approved):

region/lowernorfair/east/Pillar Room.json
blkerby: It can cover both cases. If your spark ends close enough to the pillar you can land on it. Otherwise you can land on either the high part of the floor or on the door platform. Avoiding all acid damage seems possible in all cases. 

Adjusted the requirements for the jump from the pillar as you suggested. Also added a HiJump alternative to walljump for escaping the acid in the comeInShinecharged/comeInShinecharging cases, where the acid will be higher; in the  comeInWithSpark case, the acid is low enough that it doesn't affect your jump, so getting out is free.

PR #1807, opened 2024-12-01
Shinecharge tests

blkerby:
This adds a test to ensure that if a strat has a `leaveShinecharged` exit condition (or `"endsShineCharged": true`) then it accounts somewhere for the requirement to gain the shinecharge.

The cases that were failing are fixed here. None of them were a significant problem, as they all involved a relatively long runway. In checking these, I also took the opportunity to refine the ones in West Ocean: adding videos, a note, tightening some frames, and adding one more strat (which leaves with frames remaining whereas the existing strat only leaves with a spark).

### DIFF
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -243,12 +243,16 @@ __Example:__
 
 #### shineChargeFrames object
 
-A `shineChargeFrames` object represents the need for Samus to have the given amount of shinecharge frames remaining; after this requirement, the new amount of shinecharge frames remaining is updated by subtracting away the given amount of frames.
+A `shineChargeFrames` object represents the need for Samus to have at least the given amount of shinecharge frames remaining. After this requirement, the new amount of shinecharge frames remaining is updated by subtracting away the given amount of frames.
 
-For this requirement to be satisfied, one of the following must be true:
-- It must be preceded by a `canShineCharge` requirement in the same strat.
-- The strat must have `startsWithShineCharge` set to true, and connect with an immediately preceding strat with `endsWithShineCharge` set to true.
-- The strat must have a `comeInShinecharged` or `comeInShinechargedJumping` entrance condition.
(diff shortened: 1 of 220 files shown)


osse101 (approved):
This checks that a shinesparkFrame consumer has a source for frames, but how important would it be for a strat with a source `canShineCharge{}` also uses the shinespark with a shinespark or exit condition?  Without them, it does indicate the requirements were written incorrectly.

For some reason I thought this was checking the `shinespark` requirement for a way to generate the shinespark too, moreso in cases without an exit condition.

blkerby:
> This checks that a shinesparkFrame consumer has a source for frames, but how important would it be for a strat with a source `canShineCharge{}` also uses the shinespark with a shinespark or exit condition? Without them, it does indicate the requirements were written incorrectly.
> 
> For some reason I thought this was checking the `shinespark` requirement for a way to generate the shinespark too, moreso in cases without an exit condition.

Yeah I was planning on addressing that other side of things next in a separate PR. Thinking more about it, it might make more sense to integrate all these checks into a single pass through the strat requirements (and entrance/exit conditions) from top to bottom, tracking the state as it goes. That could also allow ensuring a proper sequencing between production and consumption of a shinecharge. I was focused on cross-room strats, but the checks can apply to in-room strats as well. Temporary blue is another valid way to consume a shinecharge, aside from shinesparking. 

Things can get a little tricky, because for in-room strats it's possible that a shinecharge and its consumption happens inside an `or` branch, where coverage then wouldn't necessarily be required in the other branchs; I'm not sure if strats ever actually do that, and typically it would be better split off into a separate strat in a case like that, but it still could be a logically valid possibility. I think we also sometimes have two `canShineCharge` requirements or two `shinespark` requirements in sequence, e.g. with one of them being in an `or`, and that should probably also be considered valid.

PR #1809, opened 2024-12-01
Rename and Broaden h_EtecoonDoorSpawnFix

kjbranch:
Currently Map Rando removes the misplaced door closing animation in Etecoon Shaft, and fixes the closing animation. This will enable a fix the closing animation of all 3 bad doors.

I think it would make sense for Map Rando to not use this helper and to revert back to the vanilla door close in Etecoon shaft (but fix the misplaced doorshell in Main Shaft).

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -248,11 +248,13 @@
           ]
         },
         {
-          "name": "h_EtecoonDoorSpawnFix",
+          "name": "h_DoorImmediatelyClosedFix",
           "requires": [ "never" ],
           "devNote": [
-            "In Vanilla, there is a bug placing the door at the top of the Etecoon Shaft in the wrong location.",
-            "This primarily affects the closing animation and prevents return through the door while in direct g-mode."
+            "In Vanilla, there is a bug placing the closing door at the top of the Etecoon Shaft in the wrong location.",
+            "Additionally, the top right door in both Pink Brinstar Power Bomb Room and Halfie Climb Room are immediately closed.",
+            "This affects the closing animation and prevents return through the door while in direct g-mode.",
+            "This also prevents getting doorstuck, although that doesn't seem to be important on any of these locations."
           ]
         },
         {
--- a/helpers.json
+++ b/helpers.json
@@ -253,8 +253,8 @@
           "devNote": [
             "In Vanilla, there is a bug placing the closing door at the top of the Etecoon Shaft in the wrong location.",
             "Additionally, the top right door in both Pink Brinstar Power Bomb Room and Halfie Climb Room are immediately closed.",
-            "This affects the closing animation and prevents return through the door while in direct g-mode.",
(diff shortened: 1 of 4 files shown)


helpers.json
```diff
@@ -248,11 +248,13 @@
...
         {
-          "name": "h_EtecoonDoorSpawnFix",
+          "name": "h_DoorImmediatelyClosedFix",
           "requires": [ "never" ],
           "devNote": [
-            "In Vanilla, there is a bug placing the door at the top of the Etecoon Shaft in the wrong location.",
-            "This primarily affects the closing animation and prevents return through the door while in direct g-mode."
+            "In Vanilla, there is a bug placing the closing door at the top of the Etecoon Shaft in the wrong location.",
+            "Additionally, the top right door in both Pink Brinstar Power Bomb Room and Halfie Climb Room are immediately closed.",
+            "This affects the closing animation and prevents return through the door while in direct g-mode.",
```
blkerby, line 256:
```suggestion
            "This affects the closing animation and prevents return through the door while in direct G-mode.",
```

helpers.json
```diff
@@ -248,11 +248,13 @@
...
+            "This also prevents getting doorstuck, although that doesn't seem to be important on any of these locations."
```
blkerby, line 257:
```suggestion
            "This also prevents getting doorstuck, although that doesn't seem to be important in any of these locations."
```

blkerby (approved):

region/brinstar/green/Green Brinstar Main Shaft.json
```diff
@@ -3960,7 +3960,7 @@
         }
       },
       "requires": [
-        "h_EtecoonDoorSpawnFix"
+        "h_DoorImmediatelyClosedFix"
```
osse101, line 3963: Does the door spawn need to be fixed for it to block going back through?  Like it won't be there without a spawnFix patch?

kjbranch: In vanilla, the door "closes" one screen to the left, as well as already being instantly closed. 

Apparently all doors are already closed and then immediately afterwards there is a PLM drawn over them, and the one you just came through is opened and then starts closing. 

That's why in indirect g-mode, all doors are closed. In direct, PLMs start processing a bit before the jank happens, so the door you came through is replaced with the open door PLM, and items and gates are added.

kjbranch: In Etecoon shaft, the closing PLM was placed a screen to the left, and it was never caught because the door is permanently grey, so it didnt really get testing.

Apparently, each door has a bit that says whether or not a door should close, so for transitions without a door cap it is turned off. Pink PB room was probably messed up because the door on the right is Big Pink, and it has the bit turned off because there is no door there. 

Halfie Climb is pretty surprising though

PR #1811, opened 2024-12-02
Add more comprehensive tests on shinecharge, blue, and shinespark states

blkerby:
This introduces more comprehensive checks relating to shinecharges, temporary blue, and shinesparks, and fixes issues in strats that were failing the new checks. The new checks involve tracking Samus' current possible states (shinecharged, blue, shinesparking) in the strat from top to bottom, starting with the entrance condition if applicable (or "startsWithShineCharge"), then the `requires`, and finally the exit condition if applicable (or "endsWithShineCharge"), ensuring that each point Samus is in an appropriate state for the given requirement or condition.

The way this works, the tests can now be picky about the order of the requirements; for example, a `canShinecharge` requirement would need to come before a corresponding `shinespark` requirement, rather than the other way around. Many of the fixes simply involve reordering the requirements in order to satisfy the stricter checks. Others involve just adding missing "h_canShineChargeMaxRunway" or "h_getBlueSpeedMaxRunway" requirements.

Cases where the fixes have a logically significant effect to some degree:

- Crateria Super Room: Strat 22, "X-Mode BlueSuit", was missing a shinespark requirement for the superjump. I added one with 7 shinespark frames, based on the video.
- Aqueduct: Strat 25, "Temporary Blue Chain (Come in Running)", was missing a "canWaterShineCharge" requirement.
- Halfie Climb: Strat 83, "Suitless Stutter Water Shinecharge, Leave With Temporary Blue", had a `shinespark` requirement which didn't belong.
- Kassiuz Room: Strat 10, "Shinespark Deep Stuck X-Ray Climb", wrongly had entrance condition "comeInWithDoorStuckSetup" instead of "comeInShinecharged".
- Screw Attack Room: Strat 104, "Come In Shinecharged, End Shinecharged (Screw Attack)" had "canTemporaryBlue" requirement which didn't belong
- Double Chamber: Strat 84, "Leave With Temporary Blue", was missing "leaveWithTemporaryBlue" exit condition.
- Crocomire's Room: Strat 51, "Use Flash Suit", had a "or" where an "and" should have been.
- Ice Beam Gate Room: Strat 38, "Come in Shinecharging", should have been "Come in Getting Blue Speed" (with "comeInGettingBlueSpeed").
- Grapple Beam Room: the "Shinespark Escape" strats are reworked to use "startsWithShineCharge"/"endsWithShineCharge" similarly to Screw Attack Room, removing the obstacle "A". This will affect the difficulty placement of some combinations of strats. It also fixes a soundness issue where it could have expected coming in shinecharging from the top door, grabbing the item, then sparking through the bottom door without accounting for the time required to open its lock (if it requires Missiles or Power Bombs to open).

Helpers are added to help separate water shinecharge & water blue speed strats, to more accurately account for when a flash suit can be maintained. 

A helper "h_shinechargeSlideTemporaryBlue" is also added. For now this is to help the tests handle Screw Attack Room, where a shinecharge is maintained simultaneously with temporary blue (whereas "canTemporaryBlue" would normally be interpreted by the tests as waiting out any shinecharge frames). Having this shinecharge slide helper might also help in identifying these strats in case later we might want to add a separate tech for this or modify its requirements.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -547,6 +547,38 @@
           ]
         },
         {
+          "name": "h_shinechargeSlideTemporaryBlue",
+          "requires": [
+            "canTemporaryBlue",
+            "canShinechargeMovementTricky"
+          ],
+          "devNote": [
+            "This represents sliding off the ledge while gaining a shinecharge, which leaves Samus falling with temporary blue."
+          ]
+        },
+        {
+          "name": "h_waterGetBlueSpeed",
+          "requires": [
+            {"tech": "canWaterShineCharge"},
+            "SpeedBooster"
+          ],
+          "devNote": [
+            "This is a way to require the `canWaterShineCharge` tech but gaining only blue speed, not a shinecharge.",
+            "A flash suit can be maintained through this."
+          ]
+        },
+        {
+          "name": "h_stutterWaterGetBlueSpeed",
+          "requires": [
+            {"tech": "canStutterWaterShineCharge"},
+            "SpeedBooster"
+          ],
+          "devNote": [
+            "This is a way to require the `canStutterWaterShineCharge` tech but gaining only blue speed, not a shinecharge.",
+            "A flash suit can be maintained through this."
+          ]
+        },
+        {
           "name": "h_canGateGlitch",
           "requires": [
             "canGateGlitch",
--- a/schema/m3-requirements.schema.json
+++ b/schema/m3-requirements.schema.json
@@ -725,7 +725,7 @@
             "properties": {
               "frames": {
                 "type": "integer",
-                "minimum": 1,
+                "minimum": 0,
                 "title": "Shinespark Frames",
                 "description": "The duration (in frames) of the shinespark that must be executed. This defines the energy cost of the shinespark assuming it is not interrupted by running down to 29 energy."
               },
--- a/region/brinstar/kraid/Warehouse Kihunter Room.json
+++ b/region/brinstar/kraid/Warehouse Kihunter Room.json
@@ -718,13 +718,13 @@
         {"or": [
           {"and": [
             "canSlowShortCharge",
-            "canSpeedball",
-            {"getBlueSpeed": {"usedTiles": 17, "openEnd": 0}}
+            {"getBlueSpeed": {"usedTiles": 17, "openEnd": 0}},
+            "canSpeedball"
           ]},
           {"and": [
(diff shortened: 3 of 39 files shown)


kjbranch (commented):
Looked through most of it. Looks like a good change.

schema/m3-requirements.schema.json
```diff
@@ -725,7 +725,7 @@
             "properties": {
               "frames": {
                 "type": "integer",
-                "minimum": 1,
+                "minimum": 0,
```
osse101, line 728: The Rmode case tries to be a 0 cost shinespark so I could see not trying to write it as 1 frame with 1 excess.

helpers.json
```diff
@@ -546,6 +546,38 @@
...
+          "requires": [
+            "canTemporaryBlue",
+            "canShinechargeMovementTricky"
+          ],
+          "devNote": [
+            "This represents sliding off the ledge while gaining a shinecharge, which leaves Samus falling with temporary blue."
+          ]
+        },
+        {
+          "name": "h_waterGetBlueSpeed",
```
osse101, line 560: Only useful in Botwoon Etank I believe.  The aqueduct cases still lose their flash suit. Its fine to have around.  `h_stutterWaterGetBlueSpeed` has some extra uses in aqueduct that don't lose the flash suit.

osse101 (commented):
It all looks good but I want to know the difference between `h_shinechargeSlideTemporaryBlue` and `canTempBlue by sliding` to the tests.

blkerby:
> I want to know the difference between h_shinechargeSlideTemporaryBlue and canTempBlue by sliding to the tests.

The tests treat `h_shinechargeSlideTemporaryBlue` as leaving you in a "shinecharging" state, where the shinecharge must be consumed. If we want to use this in a case where the shinecharge is unused, we could either 
1) make a variant helper (e.g. `h_slideTemporaryBlue`) to represent that you're getting the temporary blue but don't care about the shinecharge,
2) add a helper to represent that shinecharge frames expire without being used (e.g. `h_shinechargeExpire`), or
3) use a `canTemporaryBlue` requirement, which the tests already treat as converting a "shinecharging" state into a "blue" state.

osse101:
Then we have to remember that canTemporaryBlue means blue state and h_shinechargeSlideTemporaryBlue is for doing something while blue but then doing something with shinecharge after.  And that sounds like how it should be.  We'll see if anything gets mixed up in their use down the line.

PR #1812, opened 2024-12-02
Use New Implicit Strats, Electric Death, Blue Boulder

kjbranch:
This is a test to make sure I'm doing it properly before doing several rooms at once.
- This allows the removal of a spawnAt in Blue Boulder, and makes Electric Death require the down grab to leave the room.
- I am letting the G-mode return strats use the implicit strats, even though they need a down grab or something else. Do we think that's fine?
- For room diagrams: if we get rid of the spawnAts, should they still have a red arrow on entry? should this arrow be used if there is any come in/leave normally that is disabled?


region/brinstar/blue/Blue Brinstar Boulder Room.json
```diff
@@ -564,25 +589,6 @@
         }
       }
     },
-    {
-      "id": 31,
-      "link": [2, 2],
-      "name": "Prepared Entry",
```
blkerby, line 570: I'm curious about the reasoning for removing this strat. Is it just because the prepared entry is harder than getting out of the water, so it doesn't have much logical purpose?

If it were kept I guess it should have a `{"doorUnlockedAtNode": 2}` requirement. I see some value in maybe keeping it, as it seems like it may still be a potentially useful strat for players?

region/brinstar/blue/Blue Brinstar Boulder Room.json
```diff
@@ -479,15 +493,26 @@
         }
       },
       "requires": [
+        {"or": [
+          "canWalljump",
+          "HiJump",
```
blkerby, line 498: HiJump and the other options aside from `canWalljump` are redundant, since `canPreciseWalljump` is assumed in every case anyway?

blkerby:
>     * I am letting the G-mode return strats use the implicit strats, even though they need a down grab or something else. Do we think that's fine?

For high-level tech like G-mode, I think it can make sense to implicitly assume low-level movement tech like down grab. It can help reduce clutter. The likelihood of ever wanting a game option that patches out down-grab seems very slim.

>     * For room diagrams: if we get rid of the spawnAts, should they still have a red arrow on entry? should this arrow be used if there is any come in/leave normally that is disabled?

I'm not sure that the red arrow necessarily makes sense anymore, because there's not generally going to be a uniquely defined place for it to point to. I think there should always be some `comeInNormally` strat at every door node (implicit or explicit), which is something that should probably be added to the tests, but there could be multiple of them and their destinations might not all be the same. Also the destination could be the door node itself (as with the "Prepared Entry" strat being deleted in this room) and drawing an arrow from the door to itself probably would probably be confusing.

region/brinstar/blue/Blue Brinstar Boulder Room.json
blkerby: You're probably right that it doesn't have much benefit at the moment. If we cared about timings, it might matter because it can be faster than falling in and jumping back out. The Electric Death one would be more niche since the only use for a prepared entry there would be to reset the room more quickly, compared to Blue Boulder where it can help you get across the room faster. I'm fine with it either way though 🤷‍♂️

blkerby (dismissed):

PR #1813, opened 2024-12-02
Farm cycle schema and examples

blkerby:
This is a proposed schema for a new way of representing farming strats, along with a couple examples.

I tried to be relatively accurate in these examples, but in general I think that for `cycleFrames` it is not so important to have the frame counts be tight compared to `heatFrames`. In the end, all these numbers are going to have a "patience" factor applied, which is somewhat arbitrary anyway.

Unrelated to the main change in the PR, I also added a little clarification to the `shineChargeFrames` doc.

### DIFF
--- a/region/brinstar/blue/Construction Zone.json
+++ b/region/brinstar/blue/Construction Zone.json
@@ -90,11 +90,26 @@
       "name": "Geemer Farm",
       "requires": [
         "h_ZebesIsAwake",
-        {"resetRoom": {
-          "nodes": [1, 2]
-        }},
-        {"partialRefill": {"type": "Missile", "limit": 10}},
-        {"partialRefill": {"type": "Energy", "limit": 100}}
+        {"or": [
+          {"and": [
+            {"resetRoom": {"nodes": [1, 2]}},
+            {"frames": 150}
+          ]},
+          {"and": [
+            "Morph",
+            {"resetRoom": {"nodes": [3]}},
+            {"or": [
+              {"frames": 510},
+              {"and": [
+                "Wave",
+                {"frames": 340}
+              ]}
+            ]}
+          ]}
+        ]}
+      ],
+      "farmCycleDrops": [
+        {"enemy": "Geemer (blue)", "count": 2}
       ]
     },
     {
--- a/region/brinstar/blue/Construction Zone.json
+++ b/region/brinstar/blue/Construction Zone.json
@@ -93,16 +93,16 @@
         {"or": [
           {"and": [
             {"resetRoom": {"nodes": [1, 2]}},
-            {"frames": 150}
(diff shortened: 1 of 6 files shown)


region/brinstar/blue/Construction Zone.json
```diff
@@ -90,11 +90,26 @@
...
       "requires": [
         "h_ZebesIsAwake",
-        {"resetRoom": {
-          "nodes": [1, 2]
-        }},
-        {"partialRefill": {"type": "Missile", "limit": 10}},
-        {"partialRefill": {"type": "Energy", "limit": 100}}
+        {"or": [
+          {"and": [
+            {"resetRoom": {"nodes": [1, 2]}},
```
kjbranch, line 95: looks like `nodes` is the only thing that can exist inside of resetRoom. should this be collapsed to something like `{"resetRoom": [1, 2]}`? Or will there ever be anything else that could be needed in there (neighboring environment, runway length, etc)?

kjbranch (approved):

osse101 (commented):
I mostly wonder how cycleFrames will scale.  Construction zone is a very technical room where you shoot between tiles to build a path and often times walljump morph up or clear too much and wait for the blocks to respawn.  And also lower difficulties are expected to burn more resources so would spend more time farming them back.

`ammo cost per cycle` should be able to be supported with this?

Otherwise its an informational strat and leaves things like a stopping point to the randomizer.

blkerby:
> I mostly wonder how cycleFrames will scale. Construction zone is a very technical room where you shoot between tiles to build a path and often times walljump morph up or clear too much and wait for the blocks to respawn. And also lower difficulties are expected to burn more resources so would spend more time farming them back.
> 
> `ammo cost per cycle` should be able to be supported with this?
> 
> Otherwise its an informational strat and leaves things like a stopping point to the randomizer.

I was thinking that we would apply a leniency multiplier to the cycleFrames, probably the same as the heat damage multiplier but it could also be adjusted separately if desired. The multiplier wouldn't apply to the door transition times (which the randomizer would add on top of the cycleFrames) so because of that it won't scale as heavily when the farm is quick/simple.

If ammo is used as part of a cycle, I think it could go in as a normal `ammo` logical requirement. I'm not sure that the randomizer would be able to support using strats like that in the near future, but it's at least something that could be expressed in the logic data.

blkerby:
To determine the logical number of cycles, the randomizer would also take into account any enabled patience tech, which should change the threshold for the total amount of time budgeted for the farm. I'm not sure what the exact numbers should be, but I figured I would try to roughly calibrate it to match the existing farms when using no patience tech, then take roughly 2x that with "canBePatient", 4x with "canBeVeryPatient", 8x with "canBeExtremelyPatient".

PR #1815, opened 2024-12-02
Add Node: Top of Crab Shaft

kjbranch:
There was never a spawnAt on this door, even though there probably should have been. This prevents the need for one.


blkerby (approved):

region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -1212,68 +1105,132 @@
...
+        ]}
+      ],
+      "flashSuitChecked": true
+    },
+    {
+      "id": 28,
+      "link": [4, 3],
+      "name": "Crab Ice Clip Door Lock Skip",
+      "requires": [
+        {"notable": "Crab Ice Clip Door Lock Skip"},
```
osse101, line 1189: Odd that these requirements didn't reference the tedious crab climb.
Starting from 4, the logic steps would tell you how to get up here before doing the crab clip.  And it would only be wrong if it wanted you to use SpeedBooster to climb.

osse101 (dismissed):

PR #1816, opened 2024-12-03
Remove spawnAt: Toilet, Big Pink

kjbranch:


region/brinstar/pink/Big Pink.json
```diff
@@ -1147,48 +1165,13 @@
...
-      "flashSuitChecked": true
-    },
     {
       "id": 45,
-      "link": [5, 14],
+      "link": [5, 13],
       "name": "Base",
+      "entranceCondition": {
+        "comeInNormally": {}
+      },
```
kjbranch, line 1174: just realized, is this going to have an implicit [5, 5] come in with mockball, that will override this?

blkerby: There would be a [5, 5] implicit comeInWithMockball, yes. I think you'll want to override it by setting `"useImplicitComeInWithMockball": false` and adding a [5, 13] comeInWithMockball strat.

Probably we should add tests that ensure if one of these "useImplicit" properties is overridden then all of them should be, to make sure cases aren't overlooked.

blkerby (approved):

osse101 (approved):
Toilet will still be special cased to work around these strats and carry state from the previous rooms 🤷

kjbranch:
Good point. I think Map Rando does special things while traversing this room. That needs to be documented and maybe these come in/leave normally arent useful - though they might be worth keeping for projects trying to use a minimal usable amount of the sm-json-data

blkerby:
Looks like the strats doc does mention how the Toilet can be skipped over when matching entrance/exit conditions: https://github.com/vg-json-data/sm-json-data/blob/master/strats.md#comes-through-toilet. It could maybe be worth highlighting more clearly or in more places. I think having the basic comeInNormally+leaveNormally strats in Toilet is probably good, to be safe, even though technically they are redundant. Like Kyle said, some projects might not fully implement the special behavior for skipping over Toilet, so providing coverage of the base strats is helpful in that case.

PR #1817, opened 2024-12-03
Wrecked Ship Main Shaft break bomb block with Ice and Screw

blkerby:
Bobbob showed us these a while ago.

This also fixes the "Temporary Blue Chain (In-Room)" strat to clear the bomb block obstacle.

Videos:
- Atomic Ice Moonfall Screw Attack Clip: https://videos.maprando.com/video/2359
- Covern Wall Ice Screw Clip: https://videos.maprando.com/video/2365

### DIFF
--- a/region/wreckedship/main/Wrecked Ship Main Shaft.json
+++ b/region/wreckedship/main/Wrecked Ship Main Shaft.json
@@ -861,11 +861,68 @@
         {"obstaclesNotCleared": ["B"]},
         {"canShineCharge": {"usedTiles": 13, "openEnd": 1}}
       ],
+      "clearsObstacles": ["C"],
       "unlocksDoors": [
         {"nodeId": 3, "types": ["missiles", "super"], "requires": []},
         {"nodeId": 3, "types": ["powerbomb"], "requires": ["never"]}
       ],
       "flashSuitChecked": true
+    },
+    {
+      "link": [3, 8],
+      "name": "Atomic Ice Moonfall Screw Clip",
+      "requires": [
+        {"notable": "Atomic Ice Moonfall Screw Clip"},
+        "f_DefeatedPhantoon",
+        {"enemyDamage": {"enemy": "Atomic", "type": "contact", "hits": 1}},
+        "canUseIFrames",
+        "canTrickyUseFrozenEnemies",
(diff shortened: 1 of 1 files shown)


region/wreckedship/main/Wrecked Ship Main Shaft.json
```diff
@@ -861,12 +861,69 @@
...
+    {
+      "link": [3, 8],
+      "name": "Covern Wall Ice Screw Clip",
+      "requires": [
+        {"notable": "Covern Wall Ice Screw Clip"},
+        {"not": "f_DefeatedPhantoon"},
+        {"or": [
+          "canRiskPermanentLossOfAccess",
+          {"and": [
+            {"notable": "Atomic Ice Moonfall Screw Clip"},
```
kjbranch, line 907: technically this is missing some of the requirements from above (`canUseIFrames`, `canEnemyStuckMoonfall`), but I guess if it's notable then it should only be turned on if you can do those?

kjbranch (dismissed):

region/wreckedship/main/Wrecked Ship Main Shaft.json
blkerby: Yeah that's what I was thinking although it's a little unclear exactly how this should work. I remember we considered defining a tech dependencies property for notables but abandoned the idea because it seemed like it would be too much hassle and be redundant with the strat `requires`. We might want to think about that again though.

Map Rando is also not currently handling tech dependencies in a very good way: it will let you enable tech without its dependencies being enabled, and in that case the tech will effectively not really be enabled. It's also checking all the tech dependencies at runtime during each graph traversal which is not very efficient compared to resolving them ahead-of-time.

region/wreckedship/main/Wrecked Ship Main Shaft.json
```diff
@@ -861,12 +861,69 @@
...
+            {"enemyDamage": {"enemy": "Atomic", "type": "contact", "hits": 1}}
+          ]}
+        ]},
```
osse101, line 910:
```suggestion
        ]},
        "canMidairWiggle",
```
Quickly turning back around in order to activate screw attack is a wiggle.

osse101 (dismissed):

PR #1819, opened 2024-12-03
Remove spawnAt Bug Sand Hole

kjbranch:


region/maridia/inner-yellow/Bug Sand Hole.json
```diff
@@ -285,18 +250,25 @@
...
+        "comeInNormally": {}
       },
       "requires": [
+        "canPrepareForNextRoom",
         "SpaceJump",
-        "canPrepareForNextRoom"
+        {"or": [
+          {"doorUnlockedAtNode": 1},
+          {"doorUnlockedAtNode": 3}
+        ]}
```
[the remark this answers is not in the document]
kjbranch: Come in space jumping, then the only way to land outside of the water is if either door can be opened with a shot or a Super. Regardless of which door you can land in, you can get to 3 (as you can then open any door, such as a PB door from 1).

It will still require unlocking 3 to go through it, so landing at 1 and unlocking 3 will still be accurate this way.

region/maridia/inner-yellow/Bug Sand Hole.json
```diff
@@ -469,8 +456,42 @@
...
+        }
+      },
+      "requires": [
+        "canPrepareForNextRoom",
+        "canTrickyJump",
+        "canPlayInSand"
+      ],
+      "note": [
+        "Jump through the door to cross most of the room, then jump off of the sand to cross the rest.",
+        "Crossing the room with enough speed to cross with one jump will set Samus up to be grabbed by the Yapping Maw."
```
blkerby, line 492:
```suggestion
        "With enough speed, crossing the whole room without landing in the sand is also possible, but Samus must still jump quickly after landing to avoid being grabbed by the Yapping Maw."
```

blkerby:
The new "Jump Through Room" strats seem reasonable for Very Hard. I wonder if the naked sand jump notable ("Bug Sand Hole: Sand Jumps") should maybe be bumped up to Expert, as it feels quite a bit harder?

region/maridia/inner-yellow/Bug Sand Hole.json
```diff
@@ -819,10 +874,44 @@
...
```
blkerby, line 893:
```suggestion
        "With enough speed, crossing the whole room without landing in the sand is also possible, but Samus must still jump quickly after landing to avoid being grabbed by the Yapping Maw."
```

osse101:
> The new "Jump Through Room" strats seem reasonable for Very Hard. I wonder if the naked sand jump notable ("Bug Sand Hole: Sand Jumps") should maybe be bumped up to Expert, as it feels quite a bit harder?

The whole room is in a situation where the strats should be marked more difficult than they at first seem, so I could see bumping it for that reason.  Jumping from sand to door, from door to distant sand, or shooting a yapping maw with ice or supers all take learning their attack timing and that's basically very hard.  `Sand Jumps` is that plus being accurate with hitting the middle bug, and its also probably harder if it doesn't have a drop to delay its next spawn.

PR #1823, opened 2024-12-04
Split off bubble mountain Frozen Ripper Squeeze

blkerby:
This strat didn't seem fair in Basic. Split it off from the base strat since it seems worth having its own note.

Video by Sam: https://videos.maprando.com/video/2424

### DIFF
--- a/region/norfair/east/Bubble Mountain.json
+++ b/region/norfair/east/Bubble Mountain.json
@@ -3463,13 +3463,13 @@
     },
     {
       "link": [8, 9],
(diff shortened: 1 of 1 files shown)


osse101:
I would say it is more intuitive to jump over the ripper and freeze it below you.
The tripper squeeze version is close to needing `canTwoTileSqueeze`.

blkerby:
> I would say it is more intuitive to jump over the ripper and freeze it below you.

Oh yeah, that works. Updated the strat name and note to include that version first.

PR #1824, opened 2024-12-04
Remove spawnAt: West Ocean, Fast Pillar, Ice Gate

kjbranch:


region/lowernorfair/east/Fast Pillars Setup Room.json
```diff
@@ -1605,45 +1634,74 @@
...
-        "comeInWithGMode": {
-          "mode": "direct",
-          "morphed": true
-        }
+        "comeInWithDoorStuckSetup": {}
       },
       "requires": [
-        "canArtificialMorph"
+        "canPrepareForNextRoom",
+        {"heatFrames": 80},
```
kjbranch, line 1683: Before, the doorstuck and come in jumping strats were combined and shared the 80 frames before. I'm not sure if this is actually accurate though.

osse101: I had 54 frame room times for both doorstuck and comeInJumping.

region/norfair/west/Ice Beam Gate Room.json
```diff
@@ -180,57 +167,146 @@
...
+      ]
+    },
+    {
+      "link": [1, 1],
+      "name": "Prepared Entry",
+      "entranceCondition": {
+        "comeInNormally": {}
+      },
+      "requires": [
+        "canPrepareForNextRoom",
```
osse101, line 207: At this height you can react in time and don't need to be prepared.  But the springball option would need a movement tech too for placement.

kjbranch: Added a careful jump to the spring ball option. You're right that it's really not that bad with this much space. I think its probably on the upper end of medium (assuming you dont know what room this will be). Do you think that is the right difficulty?

I kept the spring ball option free when coming in with mockball, as youre already morphed so it is trivial (you would also have the mockball tech turned on, so it's probably never a worry anyway).

osse101: Feels too hard for Medium to me.  Hard doesn't have a great movement tech for placement but it could be like `canLateralMidAirMorph` since there is a canMidairMorph component.

blkerby: If you walk into this room without knowing what's coming, I don't think there's really enough time to react by morphing and bouncing, at the Hard difficulty level. That would seem more like Expert to think/react that quickly in a niche scenario like this. Morphing and bouncing is probably not very intuitive as a reaction anyway; I think most players would just try to crumble jump. To me a `canPrepareForNextRoom` requirement seems appropriate for the Spring Ball option. The `canCrumbleJump` option could be ok without `canPrepareForNextRoom`, since crumble jumping doesn't require as quick of a reaction.

osse101: I agree with that.  The panic reaction wouldn't be to morph.

PR #1827, opened 2024-12-05
Some tricky Bat Cave bottom-to-top strats

blkerby:
This splits off the "Frozen Skree" strat from the base strat, and tightens the heat frames on other variants of the Base strat by adding "tricky" versions. 

Videos:
- Frozen Skree - With Varia (by Sam): https://videos.maprando.com/video/2395
- Frozen Skree - Suitless: https://videos.maprando.com/video/2458
- Tricky HiJump: https://videos.maprando.com/video/2462
- Tricky Spring Ball Jump: https://videos.maprando.com/video/2461

### DIFF
--- a/region/norfair/east/Bat Cave.json
+++ b/region/norfair/east/Bat Cave.json
@@ -186,14 +186,19 @@
           {"and": [
             "canWalljump",
             {"heatFrames": 180}
-          ]}          
+          ]},
+          {"and": [
+            "canTrickySpringBallJump",
+            {"heatFrames": 195}
+          ]}
         ]}
       ],
       "note": [
         "Enter the room with a Missile selected on auto-cancel, holding angle-up.",
         "Fire a Missile shot to kill the first Skree, then run right and immediately fire a beam shot to destroy the shot block.",
         "Run off the edge and down-grab onto the next platform.",
-        "Run and jump directly up where the shot block used to be, moving quickly enough to dodge the Skree projectiles by going under them."
+        "Run and jump directly up where the shot block used to be, moving quickly enough to dodge the Skree projectiles by going under them.",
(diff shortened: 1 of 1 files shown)


region/norfair/east/Bat Cave.json
```diff
@@ -133,6 +143,33 @@
         "While avoiding Skree projectiles, jump to the next bubble platform followed by jumping directly up where the shot block used to be."
       ]
     },
+    {
+      "link": [1, 2],
+      "name": "Tricky HiJump",
+      "requires": [
+        "HiJump",
+        "canTrickyJump",
+        {"heatFrames": 200}
```
osse101, line 152: The speedrun strat runs below the skree particles.  Using a missile reduces the shot cooldown making it easier to time the next shot.  The hard part is shooting the shot block, and your strat does that too in the same way.  I don't actually recall if having a beam makes it more likely to hit the shot block or the adjacent skree.
https://videos.maprando.com/video/2463

blkerby: Oh yeah that seems really difficult but it works. This can be done with a walljump or spring ball jump too:
- https://videos.maprando.com/video/2472
- https://videos.maprando.com/video/2474

For all these auto-cancel strats, it's a complex, rapid sequence of inputs with not much room for error, feels like Extreme? For me at least it took some grinding of the room to choreograph exactly what to do, because it all needs to happen so quickly that it can't just be treated like normal movement. Even then I still fail most of the time 🤷‍♂️ 

Added a strat with `canTrickyDodgeEnemies` and `canInsaneJump`. If you think it's ok for Expert then the `canInsaneJump` could be removed. I think we should try to not let it be influenced much by its role in speedruns, since even on high settings not everyone playing the randomizer will be familiar with the speedrun strats

PR #1829, opened 2024-12-05
Elevator Implicit Strats, Remove Heated spawnAts

kjbranch:


osse101 (dismissed):

region/lowernorfair/east/Main Hall.json
```diff
@@ -410,39 +382,73 @@
       ]
     },
     {
-      "id": 23,
-      "link": [3, 3],
-      "name": "Crystal Flash",
+      "id": 29,
+      "link": [4, 4],
+      "name": "Come Down",
```
blkerby, line 387:
```suggestion
      "name": "Base (Come Down)",
```

By starting the strat name with "Base", the randomizer will omit these strat names from the spoiler routes. I think it also just makes sense since this is the basic way of entering the room.

region/lowernorfair/east/Main Hall.json
```diff
@@ -410,39 +382,73 @@
...
+        "h_MainHallElevatorFrames"
       ],
       "flashSuitChecked": true
     },
     {
       "id": 24,
-      "link": [3, 4],
-      "name": "Base",
+      "link": [4, 4],
+      "name": "Go Up",
```
blkerby, line 399:
```suggestion
      "name": "Base (Go Up)",
```

region/norfair/east/Lower Norfair Elevator.json
```diff
@@ -148,6 +115,14 @@
       ],
       "flashSuitChecked": true
     },
+    {
+      "link": [1, 2],
+      "name": "Base",
+      "requires": [
+        {"heatFrames": 60}
```
blkerby, line 122:
```suggestion
        {"heatFrames": 55}
```

55 will make it consistent with the Carry Shinecharge strat.

region/norfair/east/Lower Norfair Elevator.json
```diff
@@ -393,6 +368,14 @@
       },
       "flashSuitChecked": true
     },
+    {
+      "link": [2, 1],
+      "name": "Base",
+      "requires": [
+        {"heatFrames": 60}
```
blkerby, line 375:
```suggestion
        {"heatFrames": 55}
```

region/norfair/east/Lower Norfair Elevator.json
```diff
@@ -688,58 +671,180 @@
     },
     {
       "id": 31,
-      "link": [3, 1],
+      "link": [4, 1],
       "name": "Base",
       "requires": [
-        {"heatFrames": 30}
+        {"heatFrames": 40}
```
blkerby, line 677:
```suggestion
        {"heatFrames": 50}
```
I don't think 40 is possible. 44 is the best I could get. 50 will match the Carry Shinecharge strat. There's an asymmetry here (based on where Samus' shot spawns?), where technically you could put 45 for the left door and 50 for the right door. I've just been doing them symmetrically using the conservative value that works for both doors, with the idea that this way these won't need modification in case we ever want to support horizontal mirroring.

I believe that for the in-game times shown in the Practice Hack here, the timer starts when the elevator ride ends, not when first entering the room; I notice I get the same results whether fast elevators option is enabled or not.

region/norfair/east/Lower Norfair Elevator.json
```diff
@@ -688,58 +671,180 @@
...
       "id": 32,
-      "link": [3, 2],
+      "link": [4, 2],
       "name": "Base",
+      "entranceCondition": {
+        "comeInNormally": {}
+      },
       "requires": [
-        {"heatFrames": 30}
+        {"heatFrames": 40}
```
blkerby, line 738:
```suggestion
        {"heatFrames": 50}
```

region/norfair/east/Lower Norfair Elevator.json
```diff
@@ -688,58 +671,180 @@
...
       ],
-      "devNote": "60 frames to cross the room, but 40 to and from the elevator, so the other 10 frames are listed in 4->3."
+      "flashSuitChecked": true
     },
     {
       "id": 33,
-      "link": [3, 4],
-      "name": "Base",
+      "link": [4, 4],
+      "name": "Go Down",
```
blkerby, line 794:
```suggestion
      "name": "Base (Go Down)",
```

region/norfair/east/Lower Norfair Elevator.json
```diff
@@ -688,58 +671,180 @@
...
+        "h_LowerNorfairElevatorDownwardFrames"
+      ],
+      "exitCondition": {
+        "leaveNormally": {}
+      }
+    },
+    {
+      "id": 42,
+      "link": [4, 4],
+      "name": "Come Up",
```
blkerby, line 805:
```suggestion
      "name": "Base (Come Up)",
```

blkerby:
I think ideally the elevator node in Main Hall and Lower Norfair Elevator would be renumbered from node 4 to node 3, to avoid a gap in the numbers, but that doesn't need to be changed in this PR. Sometime I'll write a script for renumbering nodes.

region/lowernorfair/east/Main Hall.json
kjbranch: I thought it had to have the exact name as `Base`. I was fighting to see which strats to have with that name. I think there are a lot of strats that should have their name changed to clean up the spoiler logs.

This does feel weird and hacky though, just checking the start of the strat name, I wonder if it should be a separate field?

kjbranch:
Renumbering nodes requires editing a lot of stuff, including the map tiles accessed for each node. I think we should move the map tile data into this project before doing that.

PR #1830, opened 2024-12-06
Remove spawnAt: Ridley, Amphitheatre, Screw

kjbranch:


region/lowernorfair/east/Amphitheatre.json
```diff
@@ -254,6 +296,26 @@
...
+        {"or": [
+          "ScrewAttack",
+          {"heatFrames": 60}
+        ]},
+        {"heatFrames": 300}
+      ],
+      "clearsObstacles": ["A"],
+      "note": [
+        "Falling into the room while holding right is safe at low horizontal speed.",
+        "Pirates can be shot with any weapon to prevent them from firing."
```
osse101, line 316: I've been calling "shoot at pirates to prevent them from firing" `canDodgeWhileShooting`.  And its pretty hard to avoid to avoid damage without it - from the pirate, or the acid - or using a tech for trust that you can jump over the pirate.  Something to look at later.

osse101 (dismissed):
"Land in door frame" is also a "Spinjump Entry" from what I can tell.  So it could confuse someone.

region/lowernorfair/east/Amphitheatre.json
kjbranch: Agreed, it would be good in a separate PR. I wouldn't want to add it and inadvertently remove Basic strats: There should be variants that fall then use space jump, or plasma, or take a hit, etc.

PR #1833, opened 2024-12-08
Some Crateria farms

blkerby:
I went ahead and put videos for these farm strats on https://videos.maprando.com, since I needed to do testing anyway to get `cycleFrames` values. They're not optimized as heavily as heat frames but I wanted them to be at least somewhat reasonable.

In cases where it seemed important, I included a "Both Doors" variant that resets the room twice, at opposite ends of the room. In this case the enemy drop counts are doubled, cycle frames are the total for traversing both directions, and it's understood that the randomizer will add extra `cycleFrames` for each `resetRoom` requirement to account for the time taken by the door transitions.

### DIFF
--- a/region/crateria/central/Climb.json
+++ b/region/crateria/central/Climb.json
@@ -2130,21 +2130,44 @@
       "name": "Space Pirate Farm",
       "requires": [
         "h_ZebesIsAwake",
-        {"resetRoom": {
-          "nodes": [1, 5]
-        }},
-        {"refill": ["Energy", "Missile"]},
         {"or": [
           "h_ClimbWithoutLava",
           "h_lavaProof",
           {"obstaclesNotCleared": ["B"]}
+        ]},
+        {"or": [
+          {"and": [
+            {"resetRoom": {"nodes": [1]}},
+            {"cycleFrames": 2350}
+          ]},
+          {"and": [
+            {"resetRoom": {"nodes": [2]}},
+            {"or": [
+              "h_canUseMorphBombs",
+              "ScrewAttack"
+            ]},
+            {"cycleFrames": 2270}
+          ]},
+          {"and": [
+            {"resetRoom": {"nodes": [3]}},
+            "h_canUseMorphBombs",
+            {"cycleFrames": 2695}
+          ]},
+          {"and": [
+            {"resetRoom": {"nodes": [4]}},
+            "h_canUseMorphBombs",
+            {"cycleFrames": 2480}
+          ]},
+          {"and": [
+            {"resetRoom": {"nodes": [5]}},
+            {"cycleFrames": 1905}
+          ]}
         ]}
       ],
-      "resetsObstacles": ["A", "B"],
-      "devNote": [
-        "The Space Pirates (wall) drop Supers and Power Bombs each at a 0.4% rate, too low to put into logic.",
-        "FIXME: Other nodes could be used to reset the room, with additional requirements."
-      ]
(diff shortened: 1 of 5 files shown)


region/crateria/central/Climb.json
```diff
@@ -2130,21 +2130,44 @@
...
+            {"cycleFrames": 2695}
+          ]},
+          {"and": [
+            {"resetRoom": {"nodes": [4]}},
+            "h_canUseMorphBombs",
+            {"cycleFrames": 2480}
+          ]},
+          {"and": [
+            {"resetRoom": {"nodes": [5]}},
+            {"cycleFrames": 1905}
```
osse101, line 2163: Arcade shows that energy free shinesparks + moonfall lets you farm this room pretty fast.

region/crateria/east/Bowling Alley Path.json
```diff
@@ -228,12 +228,81 @@
...
+            "Spazer",
+            {"cycleFrames": 505}
+          ]},
+          {"and": [
+            {"or": [
+              "canDodgeWhileShooting",
+              "Ice"
+            ]},
+            {"cycleFrames": 870}
+          ]}
```
osse101, line 295:
```suggestion
          ]},
          {"and": [
            "canUseGrapple",
            {"cycleFrames": 370}
          ]},
```

region/crateria/east/Bowling Alley Path.json
```diff
@@ -228,12 +228,81 @@
...
+            "Spazer",
+            {"cycleFrames": 1010}
+          ]},
+          {"and": [
+            {"or": [
+              "canDodgeWhileShooting",
+              "Ice"
+            ]},
+            {"cycleFrames": 1740}
+          ]}
```
osse101, line 258:
```suggestion
          ]},
          {"and": [
            "canUseGrapple",
            {"cycleFrames": 740}
          ]},
```

region/crateria/central/Crateria Power Bomb Room.json
```diff
@@ -86,24 +86,37 @@
...
+          {"and": [
+            {"or": [
+              "canDodgeWhileShooting",
+              "Ice"
+            ]},
+            {"cycleFrames": 1210}
+          ]}
+        ]}
       ],
+      "farmCycleDrops": [{"enemy": "Alcoon", "count": 3}],
```
osse101, line 117: This is a farm you would take on no items if you needed PBs.
And is very fast with missiles.

osse101 (commented):
Flyway stands out as a room where you almost need canDodgeWhileShooting to farm and cycleFrame scaling may be a standin for that.  Otherwise there would be a spazer option.

Climb and pitroom are going to scale more harshly with your movement fundamentals than most rooms.  So will BAP without gravity as you stumble in the water, particularly in the sense of stumbling into enemy damage.

region/crateria/central/Crateria Power Bomb Room.json
blkerby: Added a Missile option now. Currently the randomizer wouldn't use it, because it's not set up to trade off ammo types like that, but that's something that can be refined, and in any case it belongs as an option in the logic data.

region/crateria/east/Bowling Alley Path.json
blkerby: Good catch, I've been forgetting about Grapple in general when working out these strats, need to remember it going forward.

PR #1838, opened 2024-12-11
Remove Incorrect MB G-Mode Return

kjbranch:


blkerby:
Worth noting that SamLittlehorns reported an unbeatable seed due to this: https://dev.maprando.com/seed/q2LcpZrYB/

The behavior that removes the wall on entry is specific to Map Rando, but even in vanilla you wouldn't be able to return after entering in indirect G-mode. So removing these strats is good.

I assume we'll probably need a helper to handle the randomizer-specific behavior of the door, e.g. being able to go back at all after entering. But that can be addressed later.

PR #1848, opened 2024-12-15
Script to import map tile data from room_geometry.json

blkerby:
This is the starting point for a project to include map-tile data in `sm-json-data`. I'm using a script here to import the data from the Map Rando [room_geometry.json](https://github.com/blkerby/MapRandomizer/blob/6177311d46d36f1da6b2fc4702d240b2c8c7c145/room_geometry.json). I'll open a second PR committing the results of executing the script (#1849).

After this a few rooms will need to be manually added as they were not included in `room_geometry.json`: all Ceres rooms, Homing Geemer Room, and East Pants Room.

Next steps in the project then include:
- Importing node-level map-tile data. This data is used in the Map Rando spoiler map to determine which map tiles to highlight on each step. Currently this data often gets broken, whenever nodes are added, deleted, or redefined. It will be easier to maintain as part of the `sm-json-data` project, since then the node map-tile data can be updated at the same time as corresponding node logic changes, and tests can be used to ensure it is kept complete. This data can also have applications to other projects.
- Adding more data to describe features of each map tile, including walls, passages, doors, elevators, slopes, and items. The idea would be to include enough detail to make it possible to render the map tiles in either their vanilla form or their Map Rando form. Having this data in a structured format can help in removing Map Rando's dependence on the ROM during seed generation.

### DIFF
--- a/schema/m3-room.schema.json
+++ b/schema/m3-room.schema.json
@@ -1438,6 +1438,18 @@
         }
       }
     },
+    "mapTileMask": {
+      "type": "array",
+      "title": "Room map tile mask",
+      "description": "Map tiles which are accessible in normal ways (excluding X-ray climbing, etc.)",
+      "items": {
+        "type": "array",
+        "items": {
+          "type": "integer",
+          "enum": [0, 1]
+        }
+      }
+    },
     "nodes": {
       "type": "array",
       "title": "Nodes; Doors, Elevators, Item Locations, etc",
--- a/scripts/format_json.py
+++ b/scripts/format_json.py
@@ -3,7 +3,7 @@
 import json
 
 # Keys whose values should always be expanded to separate lines
-non_one_line_keys = {"requires", "techRequires", "otherRequires", "and", "or", "to", "id"}
+non_one_line_keys = {"requires", "techRequires", "otherRequires", "and", "or", "to", "id", "mapTileMask"}
 
 # Keys whose children are exceptions to the rule about `non_one_line_keys`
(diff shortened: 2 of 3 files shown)


osse101:
- I was wondering if there would be technically license considerations.  This is more of an upstream PR.  And the meaning of significant code usage is vague.

- Keeping a soft link to Map Rando in a file is kinda odd.  But also not a concern or anything.  You said you would be using the sm-json version of the map data, so will maprando keep room_geometry.json?  My point here is about saving the script in this repo, and its probably better to have it here.

- I think the room data is unedited from vanilla map data.

- I believe it is desired to have room data included as part of this project.

osse101 (approved):

blkerby:
> * I was wondering if there would be technically license considerations.  This is more of an upstream PR.  And the meaning of significant code usage is vague.

I created the Map Rando room geometry data so I think it's no problem; I'm offering it up to be included in the `sm-json-data` project. I hadn't noticed until now that the two projects have different licenses. I can't remember if there was a specific reason for choosing MIT license in Map Rando; maybe it was what was requested for Archipelago.

>     * Keeping a soft link to Map Rando in a file is kinda odd.  But also not a concern or anything.  You said you would be using the sm-json version of the map data, so will maprando keep room_geometry.json?  My point here is about saving the script in this repo, and its probably better to have it here.

Makes sense. There's no real need for it to stay in the repo, so it could always be deleted. I just included it to give clarity about the import process.

PR #1850, opened 2024-12-15
Script to import node tiles from room_geometry.json

blkerby:
This is similar to the script in #1848 but applied to map tiles at the node level, i.e. data that associates a set of map-tiles to each node. The map tiles can overlap across nodes, but all the room map tiles must be covered by at least one node.


osse101 (approved):
- To avoid being incorrect, multiple nodes should be able to share a map tile.  And it looks like that is how it works.

- To describe the location of a node in heated rooms  I would want more precision.  Room Diagrams can't work with this level of detail either.  It does put general node locations on the map and that has potential value 🤷

osse101:
It adds a lot of clutter to some rooms.  Could you encode it a different way?

blkerby:
> It adds a lot of clutter to some rooms. Could you encode it a different way?

In the `room_geometry.json` it is instead encoded as a list of map coordinates. It was a suggestion from Kyle to put it in this format for easier visualization. But it's a trade-off with compactness. And in very large rooms I'm not sure how helpful the visualization is anyway :man_shrugging: I would be fine with either format, or maybe there is another alternative that we haven't thought of yet.

osse101:
I think I would prefer it be compact.  Like how room_geometry has it is fine.
It is only the nodes where this is added, so its not that big a deal for skimming strats.
If you remove the 0's for spaces the 1's stand out a lot more.  But adds a bunch more processing and handling.

blkerby:
Pros/cons of the method currently proposed in this PR:
- PRO: Matches the format of the room-level `mapTileMask`. This consistency can make things a bit simpler overall.
- PRO: Probably more intuitive to read and understand at first glance. It's a sort of diagram that visually shows the reader where the node is, which can be a supplement to the node name in helping the reader understand roughly where it is (e.g. in case they don't have the room diagram PNG open).
- PRO: Probably more compact in a few cases like Croc Speedway where a single node may have a lot of tiles.
- PRO: Probably less error-prone. With listing map coordinates, I would expect it to be easier to make unnoticed mistakes like having a coordinate off by one or swapping the X and Y values.
- CON: Significantly less compact in general

I still think either could be fine. Map coordinates would be more compact. Overall the space taken up by the node data is still small compared to the strats, and not hard to scroll past. 

A third alternative which would take up the same amount of space but with a little more clarity: the node tiles could use 0 for tiles not belonging to the room, 1 for tiles belonging to the room but not the node, and 2 for tiles belonging to the node. That would also make it a little easier to define node tiles by copy-pasting the room-level `mapTileMask` and then just writing "2"s in the tiles belonging to the node. Being able to see the surrounding 0/1 values gives a visual reference point, making it easier to see where the node fits in the room, which should further reduce the potential for errors.

osse101:
I think it looks a lot nicer with the 2s

PR #1857, opened 2024-12-20
Blue Brin ETank low-energy fast walljump sparks

blkerby:
This lowers the requirements on the existing strat "Walljump Up Shaft and Shinespark" from Expert to Very Hard, then adds two more variants (Expert and Extreme) for a low-energy spark, requiring HiJump or not, respectively.

Videos:
- HiJump: https://videos.maprando.com/video/2830
- Bootless: https://videos.maprando.com/video/2832

### DIFF
--- a/region/brinstar/blue/Blue Brinstar Energy Tank Room.json
+++ b/region/brinstar/blue/Blue Brinstar Energy Tank Room.json
@@ -1011,15 +1011,15 @@
     {
       "id": 51,
       "link": [4, 2],
-      "name": "Walljump Up Shaft and Shinespark",
+      "name": "Wall Jump Spark",
       "requires": [
         {"or": [
           "h_canUsePowerBombs",
           {"obstaclesCleared": ["C"]}
         ]},
-        "canShinechargeMovementComplex",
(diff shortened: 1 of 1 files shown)


osse101:
I like putting a notable on this because it is not clear that you can break the block with <29 energy.  I think the movement to break the block with hijump is fine for very hard, and without hijump for expert.  At expert you may not make it right up to the block, but that will be only a few points of excess frames damage.  With a few tries you can make it since its not really tight.
I wouldn't make 2 notables.

blkerby:
> I like putting a notable on this because it is not clear that you can break the block with <29 energy. I think the movement to break the block with hijump is fine for very hard, and without hijump for expert. At expert you may not make it right up to the block, but that will be only a few points of excess frames damage. With a few tries you can make it since its not really tight. I wouldn't make 2 notables.

Makes sense. Adjusted the requirements to push them down a difficulty tier, and combined into a single notable. There's still more shinespark variations that could be added here, e.g. walljumpless jump into mid-air spark, possibly with a speedy jump or Space Jump; I just put in a FIXME to add those later.

PR #1866, opened 2024-12-31
Norfair Reserve Power Bomb with Double Damage Boost

blkerby:
This is just based on the video: https://videos.maprando.com/video/1115. I didn't try it out, so let me know if I missed something.

### DIFF
--- a/region/norfair/east/Norfair Reserve Tank Room.json
+++ b/region/norfair/east/Norfair Reserve Tank Room.json
@@ -170,8 +170,11 @@
         "canHorizontalDamageBoost",
         {"enemyDamage": {"enemy": "Dragon", "type": "fireball", "hits": 2}},
         "canFlatleyJump",
-        {"tech": "canUsePowerBombLag"},
         "canTrickyJump",
+        {"or": [
+          {"tech": "canUsePowerBombLag"},
+          "canInsaneJump"
+        ]},
         {"heatFrames": 400}
       ],
       "collectsItems": [3]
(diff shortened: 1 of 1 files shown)


region/norfair/east/Norfair Reserve Tank Room.json
```diff
@@ -158,6 +158,22 @@
...
+      "name": "Power Bomb with Double Damage Boost (Collect Right Item)",
+      "requires": [
+        {"notable": "Power Bomb with Damage Boost"},
+        {"enemyKill": {
+          "enemies": [["Dragon"], ["Dragon", "Dragon"]],
+          "explicitWeapons": ["PowerBomb"]
+        }},
+        "canHorizontalDamageBoost",
+        {"enemyDamage": {"enemy": "Dragon", "type": "fireball", "hits": 2}},
+        "canFlatleyJump",
```
osse101, line 172:
```suggestion
        "canFlatleyJump",
        {"tech": "canUsePowerBombLag"},
        "canTrickyJump",
```

osse101 (commented):
I don't think `canUsePowerBombLag` was around when these strats were added.  But my thoughts were that without a PB explosion, grabbing the item without falling in the lava would be `canInsaneJump`, and with the extra lag just an expertish strat.

blkerby:
> I don't think `canUsePowerBombLag` was around when these strats were added. But my thoughts were that without a PB explosion, grabbing the item without falling in the lava would be `canInsaneJump`, and with the extra lag just an expertish strat.

Makes sense. Added `canInsaneJump` as an alternative to `canUsePowerBombLag` now.

PR #1872, opened 2025-01-03
Crocomire Speedway G-Mode Setups

kjbranch:
One of the right setups (the rest look pretty similar): https://videos.maprando.com/video/3085

If you're interested, here are the two ways I found to get the Multiviola to go up, both times it gets stuck:
- https://videos.maprando.com/video/3088
- https://videos.maprando.com/video/3087

### DIFF
--- a/region/norfair/west/Crocomire Speedway.json
+++ b/region/norfair/west/Crocomire Speedway.json
@@ -150,6 +150,7 @@
             "FIXME: It is possible to stored fall speed clip through the door at 3 by using a fast run speed turnaround to slide down the stairs."
           ]
         },
+        {"id": 4},
         {"id": 6}
       ]
     },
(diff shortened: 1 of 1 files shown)


region/norfair/west/Crocomire Speedway.json
```diff
@@ -397,6 +475,49 @@
...
+        "h_heatProof",
+        "canBePatient",
+        "canCameraManip",
+        "SpeedBooster",
+        {"or": [
+          "canTrickyDodgeEnemies",
+          {"enemyDamage": {"enemy": "Multiviola", "type": "contact", "hits": 3}}
+        ]},
+        {"or": [
+          "canTrickyUseFrozenEnemies",
```
osse101, line 491: How do you continue breaking the speed blocks and not kill the viola without ice?

kjbranch: I can add it to the note. I wasn't sure if it was worth including all 5 setups or not. I felt like it was nice to know that if you shinecharge too far then you are still fine, but maybe its unhelpful?

osse101: I like having the possibilities documented.  Watching a clip is going to be better than reading whatever is written in the note, imo.  If the options get documented in some other way, they can be removed from the note making it more readable.
Moving the viola off camera so you can run under it is the intuitive way you'd do it so I don't think it needs to be said.

PR #1876, opened 2025-01-03
G-Mode Setups - Finish Remaining Upper Norfair

kjbranch:
Sam found that there was a missing G-Mode immobile strat in Bubble Mountain, so I added that as well.

### DIFF
--- a/region/norfair/west/Crumble Shaft.json
+++ b/region/norfair/west/Crumble Shaft.json
@@ -429,29 +429,16 @@
       "link": [2, 1],
       "name": "G-Mode Setup - Get Hit By Sova",
       "requires": [
-        {"heatFrames": 4000},
+        {"heatFrames": 2060},
         {"or": [
-          "SpaceJump",
           "h_canUseSpringBall",
           {"and": [
             "canStaggeredWalljump",
-            {"or": [
-              "canCrumbleJump",
-              "h_heatProof"
-            ]}
+            {"heatFrames": 100}
           ]},
           {"and": [
-            "canStaggeredWalljump",
-            "canTrickyUseFrozenEnemies",
-            {"or": [
-              "Wave",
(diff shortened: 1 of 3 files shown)


region/norfair/west/Crumble Shaft.json
```diff
@@ -406,6 +425,50 @@
...
+    {
+      "link": [2, 1],
+      "name": "G-Mode Setup - Get Hit By Sova",
+      "requires": [
+        {"heatFrames": 4000},
+        {"or": [
+          "SpaceJump",
+          "h_canUseSpringBall",
+          {"and": [
+            "canStaggeredWalljump",
```
osse101, line 437: The easiest way I saw to do this was to follow behind the Sova and pass it when you are near the top by jumping off the side of the top-middle crumble blocks.  I don't really see how the extra requirements are meant to be used.
And my room times were 36 seconds for staggered wall jump, 36 for spacejump and 35 for springball.

region/norfair/west/Ice Beam Snake Room.json
```diff
@@ -440,6 +544,60 @@
...
+    {
+      "link": [2, 2],
+      "name": "G-Mode Setup - Get Hit By Sova",
+      "entranceCondition": {
+        "comeInNormally": {}
+      },
+      "requires": [
+        {"or": [
+          {"and": [
+            {"heatFrames": 600},
```
osse101, line 556:
```suggestion
            {"heatFrames": 2340},
```
And my cue to shoot the super is when the 3rd sova begins going off camera.

region/norfair/west/Crumble Shaft.json
kjbranch: I was climbing the left wall. this does seem much better

PR #1877, opened 2025-01-04
G-Mode Setups in Lower Norfair

kjbranch:


region/lowernorfair/east/Plowerhouse Room.json
```diff
@@ -608,6 +630,25 @@
...
       "flashSuitChecked": true
     },
+    {
+      "link": [2, 2],
+      "name": "G-Mode Setup - Get Hit By Zebbo",
+      "entranceCondition": {
+        "comeInNormally": {}
+      },
+      "requires": [
+        {"heatFrames": 375}
```
osse101, line 640: I found doing a loop around the left holtz as the way I would do this.  That gives a room time of 150frames.  Otherwise I think you would synchronize the two holtzes but I find that just as tricky.

region/lowernorfair/east/Red Kihunter Shaft.json
```diff
@@ -601,6 +601,18 @@
...
+      "link": [3, 3],
+      "name": "G-Mode Setup - Get Hit By Kihunter",
+      "requires": [
+        {"heatFrames": 150}
+      ],
+      "exitCondition": {
+        "leaveWithGModeSetup": {}
+      },
+      "flashSuitChecked": true,
+      "devNote": "FIXME: Could probably reduce the heat frames if coming from above or below."
```
osse101, line 614: That wouldn't be reliably faster I wouldn't think.

PR #1878, opened 2025-01-04
Etecoon ETank Wall Ice Clip X-Ray Climb

blkerby:
If I remember, BobBob showed us this a long time ago.

Video: https://videos.maprando.com/video/3103

### DIFF
--- a/region/brinstar/green/Etecoon Energy Tank Room.json
+++ b/region/brinstar/green/Etecoon Energy Tank Room.json
@@ -1065,13 +1065,17 @@
       ],
       "note": [
         "Freeze a Zebbo inside the wall above the right side of the spikeway, close to the ceiling thorns.",
-        "Face right, equip X-Ray, and jump straight up into the ceiling thorns.",
+        "Face right, equip X-Ray, move at least a few pixels away from the wall, and jump straight up into the ceiling thorns.",
         "If successful, Samus will be knocked back into the wall and end up standing inside it.",
         "Use X-Ray immediately to prevent additional thorn hits.",
         "If Samus is high enough to be touching the thorns, then i-frames will continue to be refreshed once per second.",
-        "While X-Ray climbing, be careful to only release X-Ray soon after i-frames are refreshed."
+        "While X-Ray climbing, be careful to only release X-Ray soon after i-frames are refreshed.",
+        "Crouch before pressing X-Ray again."
       ],
       "devNote": [
+        "Standing at least 3 pixels away from the wall ensures that knockback pushes Samus fully up against the ceiling;",
+        "there is also no problem with standing further right, even hanging off the ledge:",
+        "as long as Samus is still in range to take damage from the thorns above, they will be able to knock Samus into the wall.",
         "It is possible to freeze the Zeebo lower, but if Samus is too low, it will be necessary to take 3 thorn hits.",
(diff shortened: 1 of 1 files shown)


region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -1051,6 +1052,33 @@
         "At the beginning of the climb, avoid pressing left without X-Ray being held, to prevent triggering the transition of the bottom door."
       ]
     },
+    {
+      "link": [4, 2],
+      "name": "Wall Ice Clip X-Ray Climb",
+      "requires": [
+        {"notable": "Wall Ice Clip X-Ray Climb"},
+        "canWallIceClip",
+        {"thornHits": 1},
```
osse101, line 1061: I often lose iframes once while standing solidly inside the thorns, in a spot that doesn't deal damage.
Screenshot 2025-01-04 072005
Then I take an unavoidable 2nd hit when I crouch.

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -1051,6 +1052,33 @@
...
+        "canXRayClimb",
+        "canUseIFrames"
```
osse101, line 1063:
```suggestion
        "canUseIFrames",
        "canNeutralDamageBoost"
```

osse101 (commented):
I remember what Bob showed looked crazy so he must have used the beetom instead of the thorns.
This feels like the least notable version of damage boosting into a wall stuck out there.

region/brinstar/green/Etecoon Energy Tank Room.json
blkerby: Interesting, I don't know exactly what's going but was able to reproduce it now. It never happened to me before, apparently because after starting to climb I always crouched before pressing X-Ray again. Added a line in the notes to say to do this.

blkerby:
> This feels like the least notable version of damage boosting into a wall stuck out there.

Yeah, I don't disagree. I figured it's maybe worth still being notable for now, to help draw attention to it since it's probably not something that very many players would be aware of as a possibility? 

The other trick I was wanting to investigate for a while was possibly luring a Beetom through the spikeway while doing X-Ray turnarounds through it, to be able to ice clip up through the crumbles using it. The wall ice clip strat has the same equipment requirements and I think should be strictly better as it's faster and doesn't have the Beetom damage that you would expect to take while setting up the ice clip. A Beetom could be used for a moondance setup at the bottom-left door though; or, if you can bring two of them, maybe one can be used to clip while bringing the other one to the top for a moondance setup through one of the top doors.

blkerby:
I also checked that you can bring a Beetom along in the X-Ray climb. The thorns protect you from damage but once you get above them you will take hits. The damage seems predictable, since the enemy frame counter pauses while X-Ray is used, so you're just trying to minimize the frames with X-Ray not held, while not missing any stand-ups. If Wave is available, then the Beetom can be frozen most of the time while climbing, which eliminates almost all of the damage. Once at the top, a Beetom could be used for either G-mode setups or moondance at either door. You could also leave a Beetom at the top, fall back down, farm, and climb back up again :laughing: So it might make sense to model it as an obstacle. I think I'll create an issue for it since I don't think I'm up for tackling it all right now. Green Pirates Shaft is probably another room where you could bring a Beetom along while X-Ray climbing.

PR #1882, opened 2025-01-05
Worst Room In The Game shinecharges

blkerby:
The existing 2->4 shinespark strat seemed too difficult for Medium. Added a `canMidairShinespark` requirement to put it in Hard; even though it's technically possible to do this from the ground with no mid-air spark, I don't think that makes it significantly easier. One thing to keep in mind is that when sparking from from too far right (as a player on Medium would likely do), there's significant risk of not just taking the hit from the fireball but also getting knocked back down by it. Assuming a mid-air spark also allows us to tighten the shinespark frame requirements without having to double the amount of variants further. 

### DIFF
--- a/region/lowernorfair/east/The Worst Room In The Game.json
+++ b/region/lowernorfair/east/The Worst Room In The Game.json
@@ -245,6 +245,118 @@
       ]
     },
     {
+      "link": [1, 3],
+      "name": "Come In Shinecharging, Leave Shinecharged",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 3,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canShinechargeMovementTricky",
+        {"shineChargeFrames": 125},
+        {"heatFrames": 125}
+      ],
+      "exitCondition": {
+        "leaveShinecharged": {}
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],
+      "flashSuitChecked": true
+    },
+    {
+      "link": [1, 3],
+      "name": "Come In Shinecharging, Leave With Spark",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 3,
+          "openEnd": 0
+        }
+      },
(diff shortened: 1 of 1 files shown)


region/lowernorfair/east/The Worst Room In The Game.json
```diff
@@ -244,6 +244,118 @@
...
+      "name": "Come In Shinecharging, Leave Shinecharged",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 3,
+          "openEnd": 0
+        }
+      },
+      "requires": [
+        "canShinechargeMovementTricky",
+        {"shineChargeFrames": 125},
```
osse101, line 258: I couldn't get the downback strat to be reliable.  Stopping the shinecharge with a slide off the edge was hard enough, but more often than not I would hit the pirate instead of passing over it.

And the downback could be represented as a requirement

osse101 (commented):
I thought I moved the [2,4] spark to very hard, but I guess I left the extra flame hit as its own option in medium.

region/lowernorfair/east/The Worst Room In The Game.json
blkerby: Makes sense, added `canDownBack` and `canInsaneJump` requirements and added a separate variant for going down the right side, as an `or`.

I found that hitting the pirate could be avoided by starting the down-back early enough to press against the 2-tile floating platform, and continuing holding it almost until landing. For me the main problem that caused a pirate hit was ending the down-back too soon (by pressing forward) since that expands Samus' hitbox. 
Holding the down-back too long, Samus will turn around left after landing, which messes up the strat, but surprisingly doesn't take a pirate hit.

At first I was also testing killing the Pirate with temporary blue while falling. In that case you have to not press against the floating platform since that would remove the blue. In the end going over the Pirate seemed better, and it's faster too. Getting the slide at all is still difficult though, yeah.

blkerby:
> I thought I moved the [2,4] spark to very hard, but I guess I left the extra flame hit as its own option in medium.

Yeah it seemed kind of close to the border of where we'd want to add `canShinechargeMovementComplex`. I was thinking it's probably ok in Hard, with the flame hit for lenience.

PR #1884, opened 2025-01-05
Refine Botwoon Quicksand Hidden Tunnel logic

blkerby:
If entering from a 4-tile-wide sand transition above, it is surprisingly difficult to reach the tunnel. And I think 4-tile-wide entrance is consistently what happens in Map Rando whenever the tunnel would be logically useful. It feels like at least an Expert-level mid-air morph, but with only one attempt, so probably an Extreme notable.

I left a FIXME to handle different types of entrances. Even with the vanilla transition, it's not trivial and would probably be at least a Very Hard trick; a Bull can mess you up, or you can easily move too far off the top ledge and not have enough time to move back to the tunnel, since the acceleration is so slow in the sand. With the vanilla transition, though, there's no reason to ever use the tunnel because you could just enter from the opposite side above. In Map Rando, Aqueduct would not be a possible entrance above as the rooms won't fit together. That leaves Bug Sand Hole as the only possible wide entrance that could come into play, but I don't know if it can actually occur; I don't think I've seen it before.

I tested the entrance through Toilet, and confirmed it can work with a lateral mid-air morph from an air room above, but it requires a certain amount of run speed, so it would be situational whether you would be able to get that run speed or if the room above is an air room at all. Since we don't currently have a way to model that, for now I just removed the Toilet variations and left a FIXME for it.

### DIFF
--- a/region/maridia/inner-pink/Botwoon Quicksand Room.json
+++ b/region/maridia/inner-pink/Botwoon Quicksand Room.json
@@ -216,7 +216,8 @@
         "If Samus falls below the tunnel, even to the ledge just below, it will no longer be possible to reach."
       ],
       "devNote": [
-        "FIXME: Depending on the room above, the 'can4HighMidAirMorph' requirement might be unnecessary."
+        "FIXME: Depending on the room above, the 'can4HighMidAirMorph' requirement might be unnecessary.",
+        "FIXME: Entering through the Toilet can be possible with a lateral mid-air and enough run speed from an air room above."
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


blkerby:
I also found that after crossing through the tunnel, if you need to exit below in a specific way (e.g. spin jumping across to the opposite side) then a second Bull hit may be unavoidable, unless a Power Bomb is available to clear them.

blkerby:
Videos:
- With Sprite Features (Samus on Top): https://videos.maprando.com/video/3137
- Without Sprite Features: https://videos.maprando.com/video/3138

dragonsinth (commented):
LGTM, thank you!

kjbranch (approved):
Could also be done easily with a flash suit. I wonder if you could come in shinecharged and spark diagonally to the upper ledge as well. (i.e. from Botwoon E-Tank or Bug Sand Hole)

PR #1889, opened 2025-01-09
Boyon Gate and Crab Hole shinecharges

blkerby:

### DIFF
--- a/region/maridia/outer/Crab Hole.json
+++ b/region/maridia/outer/Crab Hole.json
@@ -250,6 +250,30 @@
       ]
     },
     {
+      "link": [1, 2],
+      "name": "Come in Shinecharging, Leave With Spark (Gravity, Sciser Hit)",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 2,
+          "openEnd": 1
+        }
+      },
+      "requires": [
+        "Gravity",
+        "Morph",
+        "canShinechargeMovementTricky",
+        {"enemyDamage": {"enemy": "Sciser", "type": "contact", "hits": 1}},
+        {"shinespark": {"frames": 9, "excessFrames": 0}}
+      ],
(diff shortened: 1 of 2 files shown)


region/maridia/outer/Crab Hole.json
```diff
@@ -2114,21 +2706,149 @@
...
+        "canShinechargeMovementComplex",
         {"shineChargeFrames": 100}
       ],
       "exitCondition": {
         "leaveShinecharged": {}
       },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],
```
osse101, line 2727: Where these doors are capless, I think we have been treating them as not-lockable (by ammo).  But someone could always put a door cap on it.  So they should also add the unlock requirements themselves in that case?

osse101 (dismissed):
- Looks like you treated the Boyon Gate hall door->door runway as good enough to use instead of the green gate runway.
- And Crab Hole has door requirements on the exposed transitions.

blkerby:
> Looks like you treated the Boyon Gate hall door->door runway as good enough to use instead of the green gate runway.

Yeah, the door runway is 1.5 tiles shorter, 2.5 if the other door can't be unlocked, so it does seem worth modeling using the open-gate runway. Added strats for those now.

region/maridia/outer/Crab Hole.json
blkerby: Yeah, the way the tests are currently written, it allows unlocksDoors on those but doesn't require them. I think it could make sense to change that and start including them consistently. I think the only case that wouldn't make sense to include would be the sand transitions. 

There's a bunch of cases across many rooms, so maybe will defer that to a separate PR later.

PR #1890, opened 2025-01-10
Add canHeatedCrystalFlash tech

blkerby:


osse101:
It seems fair to not require heatedCF tech if you have heatproof.  And that would apply to lava and lavaproof too.

blkerby:
> It seems fair to not require heatedCF tech if you have heatproof. And that would apply to lava and lavaproof too.

For sure. I was originally thinking that too, just forgot about it while writing it up. Adjusted the helpers now; also reverted adding `canHeatedCrystalFlash` requirements on strats that already required heatproof.

Also, I was forgetting about the fact that `canHeatedCrystalFlash` has the CF ammo requirements built into it, and it's important not to have multiple CF requirements that would double up those requirements. So I went back and made adjustments for that too.

blkerby:
Full-room video demonstrating a slightly better than the new logically expected heat frames for the Crystal Flash clip in Lower Norfair Spring Ball Maze Room: https://videos.maprando.com/video/3453. 

I bound angle-down to the Practice Hack "update timers" function and am using it to capture in-game times for key points in the strat, e.g. landing after the CF (where the heat frames in the helper `h_canHeatedCrystalFlash` would end), and entering the last morph tunnel which is where I think we want the junction node 5 to be defined.

PR #1891, opened 2025-01-12
Glass Tunnel shinecharges (etc.) 

blkerby:

Glass Tunnel still needs cross-room temporary blue strats added too. Will come back to that later.

### DIFF
--- a/region/maridia/outer/Crab Tunnel.json
+++ b/region/maridia/outer/Crab Tunnel.json
@@ -279,10 +279,6 @@
         "canStutterWaterShineCharge",
         "canDodgeWhileShooting",
         {"ammo": {"type": "Super", "count": 1}},
-        {"or": [
-          "HiJump",
-          "canStationaryLateralMidAirMorph"
-        ]},
         "canChainTemporaryBlue"
       ],
       "exitCondition": {
--- a/region/maridia/outer/Crab Tunnel.json
+++ b/region/maridia/outer/Crab Tunnel.json
@@ -230,7 +230,7 @@
     {
       "id": 11,
       "link": [1, 2],
-      "name": "Stutter Shinecharge Through The Gate",
+      "name": "Stutter Shinecharge Through The Gate, Leave Shinecharged",
       "entranceCondition": {
         "comeInRunning": {
           "speedBooster": true,
@@ -243,7 +243,13 @@
         "canStutterWaterShineCharge",
         "canDodgeWhileShooting",
         {"ammo": {"type": "Super", "count": 1}},
-        {"shineChargeFrames": 110}
+        {"or": [
+          {"and": [
+            "HiJump",
+            {"shineChargeFrames": 95}
+          ]},
+          {"shineChargeFrames": 110}
+        ]}
       ],
       "exitCondition": {
         "leaveShinecharged": {}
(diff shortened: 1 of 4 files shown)


region/maridia/outer/Fish Tank.json
```diff
@@ -1382,6 +1382,30 @@
...
+    {
+      "link": [4, 1],
+      "name": "Come In Shinecharged, Leave With Spark",
+      "entranceCondition": {
+        "comeInShinecharged": {},
+        "comesThroughToilet": "any"
+      },
+      "requires": [
+        "canShinechargeMovementTricky",
+        {"shineChargeFrames": 125},
```
osse101, line 1394: This is the enterSpinning number.  Suitless enterspinning doesn't work for water->water, like mainstreet.  Gravity helps in that case and also helps a bit in air->water.
Suitless, you can't land and jump spark into the door.
Gravity alone can exit bottom, and with hijump can exit top.

region/maridia/outer/Glass Tunnel.json
```diff
@@ -264,6 +264,136 @@
...
+      "name": "Come In Shinecharging, Leave Shinecharged (Gravity)",
+      "entranceCondition": {
+        "comeInShinecharging": {
+          "length": 2,
+          "openEnd": 1
+        }
+      },
+      "requires": [
+        "Gravity",
+        "canShinechargeMovementComplex",
```
osse101, line 278: "f_MaridiaTubeBroken" throughout

region/maridia/outer/Crab Tunnel.json
```diff
@@ -258,6 +264,37 @@
...
+      },
+      "requires": [
+        {"notable": "Stutter Shinecharge Through The Gate"},
+        "canShinechargeMovementComplex",
+        "canStutterWaterShineCharge",
+        "canDodgeWhileShooting",
+        {"ammo": {"type": "Super", "count": 1}},
+        {"or": [
+          "HiJump",
+          "canStationaryLateralMidAirMorph"
```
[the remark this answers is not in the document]
blkerby: Makes sense, removed this requirement.

region/maridia/outer/Fish Tank.json
blkerby: Changed it to `comeInShinechargedJumping` now, which will force the room above to have air physics (really it ought to allow Gravity as an option too); I wasn't thinking about vertical transitions when we made that entrance condition, but I think it should work.

The video is using 112 frames (https://videos.maprando.com/video/3312) though it might need another frame or two to get in bottom position. I think the gap between that and 125 may be enough to avoid issues for now, since for an air room above it should be possible to enter spinning by jumping instead of falling, which generally shouldn't cost more than those 10 frames? In a case like Tourian Escape Room 1 where you can slide off directly in the transition, jumping costs more than 10 frames, but there's so much extra time available that it still works. It's not ideal, and should be refined later, but I'm not seeing any example yet where it would make something logically expected that isn't doable?

region/maridia/outer/Glass Tunnel.json
blkerby: Good catch, added those now.

region/maridia/outer/Fish Tank.json
blkerby: Even if it stayed as a `comeInShinecharged` entrance condition, I think Main Street wouldn't have created a problem because you can't leave suitless with 125 frames remaining anyway; at least, the best I could get was 113 frames remaining.

tests/asserts/keywords.py
```diff
@@ -1159,6 +1160,7 @@ def strat_err_fn(msg):
                                     msg = f"🔴ERROR: Strat name contains '{phrase}' but there is no {entrance_condition_name} entrance condition:{stratRef}"
                                     messages["reds"].append(msg)
                                     messages["counts"]["reds"] += 1
+                                break
```
osse101, line 1163: This deals with "Come In Shinecharged Jumping" and "Come In Shinecharged" containing the same string.  The phrase and requirement matching still happens once, so it seems fine.  Maybe some weirdness happens later with more related phrase entries.

PR #1893, opened 2025-01-13
Refine PCJR Mella ice clip

blkerby:
This PR removes the X-Ray variant of the Mella Ice Clip, making "h_canPreciseIceClip" (Expert) required. It also adds more detail to the notes/devNotes about how to do the trick using Eddie's normalized setup. If this is merged, we would also move the corresponding notable up from Very Hard to Expert difficulty.

HisBeardNis brought this up to me the other day in his stream: he mentioned he had been doing an Extreme seed but spent 3 hours to get this clip which is currently in Very Hard if X-Ray is available, as it was in his case.

I took a look and to me even with X-Ray it feels pretty bad to try to freestyle it; I don't think it's really ok in Very Hard. It's much better to use Eddie's normalization trick, and in that case X-Ray is irrelevant. The setup for the normalization is kind of complex and many things can go wrong: e.g., freezing the Mella at the wrong height, moonwalking back the wrong amount of pixels, missing the refreeze on the Mella, getting stuck on top of it with Samus in the wrong horizontal position, not being able to land on the Mella after breaking the crumble if it already thawed, getting a standing jump off of it instead of a spin-jump, or bonking the re-formed crumble by not spin jumping soon enough. With a full understanding of everything, it can be done reliably, but I think it's a bit too much for Very Hard.

To avoid the canMoonwalk requirement, technically a freestyle version (with or without X-Ray) or some alternative kind of normalized setup could be an option, but it would need additional requirements to place it at the right difficulty. I left it as a FIXME for now.

### DIFF
--- a/region/norfair/crocomire/Post Crocomire Jump Room.json
+++ b/region/norfair/crocomire/Post Crocomire Jump Room.json
@@ -1381,10 +1381,7 @@
       "requires": [
         {"notable": "Mella Ice Clip"},
         "Morph",
-        {"or": [
-          "h_canXRayMorphIceClip",
-          "h_canPreciseIceClip"
-        ]},
+        "h_canPreciseIceClip",
         {"or": [
           "canConsecutiveWalljump",
           {"and": [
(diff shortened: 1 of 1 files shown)


osse101:
If you can bring a mella that's at about the right height to the block, then its completely fine for very hard.
If you walk the two high mellas over to the block and take a hit then the horizontal position isn't hard.  This falls apart with the height gained or lost per movement.  Sometimes its easy, most times it doesn't line up.  Adding consistency would be using either the eddie strat or the mella ice climb strat, so we're better off doing it that way.

PR #1894, opened 2025-01-13
Fix Big Pink G-Mode Bug, Add Reqs Fish Tank Spark

kjbranch:
The Fish Tank spark could still use work (testing and excess frames), I just thought these should be added in order to be able to reset the room


region/maridia/outer/Fish Tank.json
```diff
@@ -1267,6 +1267,12 @@
         "comesThroughToilet": "any"
       },
       "requires": [
+        {"or": [
+          "Gravity",
+          "HiJump",
+          "canSpringBallJumpMidAir",
```
osse101, line 1273: Gravity, HiJump, and Springball will use the more direct strats.  Avoiding the pirate technically has some requirements that you may be avoiding with this shinespark.

osse101 (dismissed):

kjbranch:
Refined the both shinespark strat and the strats to cross the room without the spark.

region/maridia/outer/Fish Tank.json
```diff
@@ -1990,9 +2006,97 @@
...
+      "requires": [
+        "canSuitlessMaridia",
+        {"or": [
+          {"and": [
+            "HiJump",
+            "canSpringBallJumpMidAir"
+          ]},
+          {"and": [
+            "HiJump",
+            "canDodgeWhileShooting"
```
osse101, line 2042: I land 'not with the pirate' and `h_canCrouchJumpDownGrab`

region/maridia/outer/Fish Tank.json
```diff
@@ -1990,9 +2006,97 @@
...
+          ]},
+          {"and": [
+            "canTrickyJump",
+            "canSpringBallJumpMidAir",
```
osse101, line 2046: TrickyJump lets you stand next to the pirate.
- `h_canCrouchJumpDownGrab` can reach the ledge from there.
- Or springball itself can reach the next ledge.

region/maridia/outer/Fish Tank.json
```diff
@@ -1990,9 +2006,97 @@
...
+      "name": "Suitless",
+      "requires": [
+        "canSuitlessMaridia",
+        {"or": [
+          {"enemyKill": {
+            "enemies": [["Pink Space Pirate (standing)"]],
+            "explicitWeapons": ["Plasma"]
+          }},
+          {"and": [
+            "canUseIFrames",
```
osse101, line 2070: The placement of this is a bit fuzzy because you need to rush a certain amount with the water physics, or you get interrupted on your jump.  Its probably ok.

region/maridia/outer/Fish Tank.json
```diff
@@ -1990,9 +2006,97 @@
...
+            }}
+          ]}
+        ]},
+        {"or": [
+          "HiJump",
+          "canSpringBallJumpMidAir",
+          "h_canCrouchJumpDownGrab",
+          {"and": [
+            "Plasma",
+            "canTrickyUseFrozenEnemies"
```
osse101, line 2084: canUseFrozenEnemies

kjbranch: I was on the fence about this since it's a bit non-intuitive, and if you kill the pirate you likely cant reset the room. But I think its fine to lower it. I also split the tank/kill strats, it looked like tank+spring was not great.

PR #1900, opened 2025-01-17
Some Pink Maridia shinecharges

blkerby:
Colosseum is the main room here that was missing lots of strats. I also refined some strats and tightened frames in Aqueduct and a few other rooms.

As usual, videos are on https://videos.maprando.com.

### DIFF
--- a/region/maridia/inner-pink/Colosseum.json
+++ b/region/maridia/inner-pink/Colosseum.json
@@ -715,7 +715,7 @@
         "comeInShinecharged": {}
       },
       "requires": [
-        "Gravity",
+        "canGravityJump",
         "HiJump",
         "canShinechargeMovementTricky",
         {"shineChargeFrames": 90},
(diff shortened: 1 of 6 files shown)


region/maridia/inner-pink/Colosseum.json
```diff
@@ -700,6 +708,21 @@
       ],
       "note": "Jumping to the first platform is relatively tight without a wall jump and requires disabling HiJump."
     },
+    {
+      "link": [2, 1],
+      "name": "Come In Shinecharged, Shinespark (HiJump Gravity Jump)",
```
osse101, line 713: There's a clip for gravity+walljump, but there isn't a strat for it, on [2. 1]

blkerby: This clip?: https://videos.maprando.com/video/3624

I think there's no reason to include a strat for it, since with Gravity + canConsecutiveWalljump you can already cross the room more easily without a shinespark? Sparking would require canConsecutiveWalljump too.

PR #1901, opened 2025-01-17
Clean up some notable names

blkerby:
This simplifies a few notable names that seemed more verbose than necessary. I started with notables referenced in the v117 release notes, then did a few more. It's not a complete pass though.

I also added a `canPreciseSpaceJump` requirement on the East Ocean left-to-right strat with Space Jump only.

Also I notice we have a lot of notables for cross-room shinecharge/shinespark strats that probably don't need to be notable anymore, something that could be cleaned up later.

### DIFF
--- a/region/norfair/east/Frog Speedway.json
+++ b/region/norfair/east/Frog Speedway.json
@@ -483,9 +483,9 @@
     {
       "id": 20,
       "link": [2, 1],
-      "name": "Shot Block Overload (Speedless Speedway)",
+      "name": "Shot Block Overload",
       "requires": [
-        {"notable": "Shot Block Overload (Speedless Speedway)"},
+        {"notable": "Shot Block Overload"},
         "Wave",
         {"or": [
           "Spazer",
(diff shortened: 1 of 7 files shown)


region/norfair/east/Frog Speedway.json
```diff
@@ -842,7 +842,7 @@
   "notables": [
     {
       "id": 1,
-      "name": "Shot Block Overload (Speedless Speedway)",
+      "name": "Shot Block Overload",
```
osse101, line 845: I like "Speedless Speedway" more

blkerby: Ok, reverted this room. "Speedless Speedway" doesn't seem as descriptive of what is happening in these strats, compared to "Shot Block Overload", but at least it conveys that it's done without Speedbooster, which is useful. The room name isn't too long here, so we can just leave both descriptors without much issue.

kjbranch (approved):

PR #1907, opened 2025-01-21
Main Street G-Mode Cleanup

kjbranch:
This is mostly just a restructure of the current g-mode overload speed block strats. I added two new junctions and added more strats to get into position. It's also worth noting that there are no strats in this PR to grab the item without Ice, but there are strats added that get into position without it.

This lays the groundwork to (hopefully) be able to easily add the other ways to grab the item:
- wave+ice and wave is the item
- PB + ice
- Kago +ice (immobile, CF, E-Tank item)
- R-Mode standup (CF + no ice)

### DIFF
--- a/region/maridia/outer/Main Street.json
+++ b/region/maridia/outer/Main Street.json
@@ -973,157 +973,6 @@
       "devNote": "FIXME: Add a midair version requiring 0 tanks"
     },
     {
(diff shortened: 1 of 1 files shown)


osse101 (approved):
Seems fine

PR #1909, opened 2025-01-23
G-Mode Main St Pt 2. Add Many Variations

kjbranch:
This was a pain, but I think these strats will be quite impactful. It will be fun to see more seeds have SpeedBooster here.

Some strats looked like they would fit under `canBeVeryPatient`. I didn't want to complicate things yet by thinking about immobile at 3. I'm not great at IBJ, but it looked maybe possible to get PLMs overloaded in time. Starting at the very bottom will maybe make things too tight.

### DIFF
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -525,6 +525,16 @@ __Example:__
 ]}
 ```
 
+#### itemCollectedAtNode object
+An `itemCollectedAtNode` object represents the need to have collected the item at a given node in the same room. For example, the existance of such an item could unintentionally be used to overload PLMs in G-mode assuming the item has spawned.
+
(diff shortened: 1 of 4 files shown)


osse101 (commented):
good stuff
------------
- Crystal flashing with high etank count can be hard to damage done quickly
- Underwater walljumps are on some strats that I'd rather call `notable:underwaterwalljumps`
- The notable `GMode then use frozen crab` is misnamed when using a reserve tank trigger since you don't need ice.

kjbranch:
I added `canBePatient` to all of these strats. It probably should have it either way, but it will also help if there is an extra slow variant that isn't properly requiring `canBeVeryPatient`.

PR #1910, opened 2025-01-23
Artificial Morph Double Spring, G/P/Y Maridia

kjbranch:
This turned out to be quite a bit easier than a normal double spring ball jump, but it's obscurity makes it seem fine to share the same tech.

I'll work on Outer and other water rooms next.

### DIFF
--- a/region/maridia/inner-yellow/Watering Hole.json
+++ b/region/maridia/inner-yellow/Watering Hole.json
@@ -225,6 +225,98 @@
       "flashSuitChecked": true
     },
     {
+      "link": [1, 1],
+      "name": "G-Mode Morph - Remote Acquire Left Item",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "direct",
+          "morphed": true
+        }
+      },
+      "requires": [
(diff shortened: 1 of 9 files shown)


region/maridia/inner-yellow/Watering Hole.json
```diff
@@ -224,6 +224,98 @@
...
+            "canTrickyJump",
+            "h_canArtificialMorphBombs",
+            "h_canArtificialMorphSpringBall"
+          ]},
+          {"and": [
+            "canInsaneJump",
+            "h_canArtificialMorphBombs"
+          ]},
+          {"and": [
+            "h_canArtificialMorphPowerBomb",
```
osse101, line 249: Can't tell if its intended, but PBs + DoubleSpringball jump does cleanly kill the enemies and lets you move through the room.

osse101 (commented):
I've added 1 or 2 of these double springball jumps starting from morph to east ocean and a sandpit.  And they are a little easier than regular double springball jumps, but the tech knowledge around how to do the trick is the same.

PR #1913, opened 2025-01-24
Add Remaining Artificial Morph Double Spring Ball Jumps 

kjbranch:
Skipped Main St, since there is an open PR for it. I'll add them separately. Looks like the room is fully traversable with double spring.


region/maridia/outer/Fish Tank.json
```diff
@@ -2196,8 +2281,20 @@
       "id": 7,
       "name": "Insane Naked Jump Over Pirate (Left to Right)",
       "note": "This is particularly precise, and requires subpixel precision."
+    },
+    {
+      "id": 8,
+      "name": "Artificial Morph, Double Spring Ball Jump onto Frozen Fish",
```
osse101, line 2287: Its not crazy hard; the timing is somewhat lenient but I didn't see a good cue for it.   Resetting the room if you miss is the only really bad part to this, and setting up artificial morph again.

region/norfair/east/Lava Dive Room.json
```diff
@@ -442,6 +442,27 @@
         "Then Walljump again to exit the Lava and reach the left ledge."
       ]
     },
+    {
+      "link": [2, 3],
```
osse101, line 446: The part that is off is ending at [3] instead of [1].  
At 3 you get some heat frames for running to and opening the door and works better for missile doors.
But the artificial morph could open a PB door.

osse101 (dismissed):

region/norfair/east/Lava Dive Room.json
kjbranch: I didnt realize this would be possible without Varia. My earlier attempts looked like it would require it, but thats cool. I'll throw 50 heat frames to escape the room and put a note on it to come back to it. Havent completely decided how to deal with heated g-mode and exiting the room yet.

region/maridia/inner-yellow/Watering Hole.json
```diff
@@ -239,21 +239,17 @@
...
+              "canTrickyJump",
+              "h_canArtificialMorphSpringBall",
+              {"enemyDamage": {"enemy": "Zeb", "type": "contact", "hits": 1}}
+            ]}
           ]},
-          {"and": [
-            "h_canArtificialMorphBombs",
-            {"enemyDamage": {"enemy": "Zeb", "type": "contact", "hits": 1}}
-          ]}
+          "canInsaneJump"
```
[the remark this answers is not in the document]
kjbranch: I had almost deleted the Gravity option when I later found out that it could be done with a tight crouch jump down grab. Now that I see HiJump is also viable, I figure the strat should just be removed. 

It was cool to see a remote acquire 2 items simultaneously, I wasn't sure if that was possible until working on this.

PR #1914, opened 2025-01-25
A Few Misc Fixes: Kzan, Room Diagram, Leniency

kjbranch:
Added some leniency to Croc Escape frozen Geruta strats; hopefully didn't go too far.
Most of these room diagrams were missing an obstacle marker.


region/norfair/west/Crocomire Escape.json
```diff
@@ -576,7 +583,11 @@
       "requires": [
         "canTrickyUseFrozenEnemies",
         "HiJump",
-        {"heatFrames": 1150}
+        {"heatFrames": 1150},
```
osse101, line 586: This 1150 could be lowered to 1050.
Also there is RNG on the bat doing a big jump.  So I think we'd want to account for that by saying extreme can use the RNG number, and you otherwise take 1500 frames, for the bat to catch on the left ledge and not need the big jump.

And adding an or for avoiding the bat when you first pass it on the right was a good idea.

The springball version looks ok.

osse101: It needs to describe passing it on the right side now.  You can jump over the live Geruta, freeze it, or freeze it and turn off ice.  Or get hit I guess.  The requirements are otherwise placing the strat in hard, and the leniency is doing a lot of work with keeping the Geruta on camera while you move it. 
Can cameraManip would be good to add here too.

kjbranch: Oh, I see now. I hadn't realized you could jump over it reliably, and I thought the frames were for just freezing it. The contact hit I had included in the spring ball variant was because I was jumping into it occasionally near the left while waiting for a good swing.

Issue #1006, opened 2023-08-02
Add cross-room grapple jump strats


blkerby:
There are multiple places where one can use grapple to bring momentum and/or grapple-jump state across rooms.

Example by insomniaspeed: https://discord.com/channels/1053421401354285129/1053436236628496454/1136374862416449658

Leaving through top of Everest should also be an option. Also left of Halfie Climb?

I think it should be possible to represent these using the existing jumpway schema except we just need to add a new jumpwayType.

- Top left of Aqueduct: insomniaspeed video: https://discord.com/channels/1053421401354285129/1053436236628496454/1261141046570651648
- Many applications from BaronRed: https://discord.com/channels/1053421401354285129/1053436236628496454/1261206175303729165

osse101:
Halfie climb has grapple blocks.

Exiting top of everest
Entering Red fish from below
Entering Crab shaft mid door
Entering Fish tank bottom left
Entering Broken Maridia Tube?
Entering east cac alley right side
Entering West cac alley left side

The Jumpway system would be pretty similar.  Find out distance from door and momentum needed.  
A horizontal transition grapple jump would be much harder than everest->red fish.  Could be a tech.

blkerby:
Another example from insomniaspeed, entering Pseudo Plasma Spark from below: https://discord.com/channels/1053421401354285129/1073621773184159764/1309532378149949482

blkerby:
@osse101  I'm getting ready to write up the logic for cross-room grapple jumps, and was interested in getting your input first on how to place their difficulty. I created about 200 videos the past few days, covering essentially all the combinations of setups & applications that I know of. The reason it's so complex is because even for the same application (destination room), the location of the grapple block in the setup room can significantly influence what type of movement is required; so every combination has to be individually tested and assigned appropriate tech requirements. So far there are a total of 22 setup locations (4 vertical, 9 leftward, 9 rightward) and 36 application locations (5 vertical, 11 leftward, and 20 rightward). 

Here's what I'm thinking so far:

- Very Hard: Simplest cases where the speed and angle required through the door isn't precise, and all you have to do is grapple jump in the next room using HiJump boots. Examples:
    - https://videos.maprando.com/video/4179
    - https://videos.maprando.com/video/4134
- Expert: The setup may be somewhat precise, you may need to perform a turnaround in the next room (but not too precisely), may need to grapple jump without HiJump, and may need to shoot open a door above while grapple jumping. Examples:
[...]

PR #1922, opened 2025-01-29
Cross-room grapple jumps

blkerby:
This addresses #1006. It adds 
- Tech `canPreciseGrappleJump`, intended for Extreme difficulty, to fit in between `canGrappleJump` (Expert) and `canTrickyGrappleJump` (Insane).
- Exit conditions `leaveWithGrappleSwing` and `leaveWithGrappleJump`, and corresponding entrance conditions `comeInWithGrappleSwing` and `comeInWithGrappleJump`. 
- Strats for setups and applications of cross-room grapple swings & grapple jumps.

Some things left out for now:
- Heat frames for the setups and applications. I just assume `h_heatProof` because it generally didn't seem very reasonable to expect getting it on a small number of attempts. It could be refined later.
- Application to grapple jump up Crumble Shaft. This would only be to save heat frames compared to crumble jumping. But again, since we're just assuming `h_heatProof` for these strats, it would serve no purpose. With grapple jumping here, it is particularly tricky to avoid the crumble blocks and Sovas.
- Grapple jumping while carrying a flash suit. If we considered this, there also would be a couple more applications, in Grapple Beam Room and Caterpillar Room. But maintaining a flash suit while executing a grapple fling and grapple jump seems very difficult, and I didn't want to think about it, so it's just treated as not being possible.
- Grapple jumping after a grapple teleport. There are some good strats to be added here, but this PR is already big so it's just left out for now.

The main difference between the two sets of exit/entrance conditions is that `leaveWithGrappleSwing` encodes information about the location of the block, to be able to account for Samus' momentum through the transition; whereas `leaveWithGrappleJump` assumes you are going up through a vertical door with no horizontal momentum, so only Samus' position in the doorway matters while the location of the Grapple block is irrelevant and potentially unknown (e.g. in case of grapple jumping up through the door of Fish Tank or Glass Tunnel, where the block would be in a previous room).

Compared to what we did with grapple teleports and Samus Eater teleports, the exit/entrance conditions are using a more verbose format to list information about the usable Grapple blocks. This was necessary because there are some cases where blocks in the same position can behave differently due to other features of the room. This format created the opportunity to include a `note` field where we can specify the relevant room names, which I think makes the strats a lot more readable. If we wanted, we could convert grapple teleports and Samus Eater teleports to use a similar format later.

In the strats, I'm not listing every Grapple block that could be used, only a minimal set needed to cover all applications. Similarly, the `obstructions` field on the `blocks` is not used comprehensively, only where needed to disambiguate cases where the setup behaves differently in spite of the block being in the same position as another room; it was only needed in Red Brinstar Firefleas Room. In principle, we could also add a field to describe the fast-crumbling block in Pants Room, but it wasn't needed since this combination of block location and environment (water) was already unique. If the data is ever applied to larger sets of room (e.g. mirrored or with randomized environments) then this would all need to be looked at closely again.


region/maridia/inner-yellow/Maridia Elevator Room.json
```diff
@@ -953,6 +986,36 @@
...
+      "entranceCondition": {
+        "comeInWithGrappleSwing": {
+          "blocks": [
+            {"position": [7, 3], "note": "The Precious Room"},
+            {"position": [7, 2], "note": "The Moat"}
+          ]
+        }
+      },
+      "requires": [
+        "canTrickyGrappleJump"
```
osse101, line 1001: These two feel easier than other tricky's.
Pant's start was still hard.

region/maridia/outer/Mt. Everest.json
```diff
@@ -1637,6 +1649,31 @@
...
+        "comesThroughToilet": "any"
+      },
+      "requires": []
+    },
+    {
+      "link": [2, 8],
+      "name": "Grapple Jump (Left Side)",
+      "entranceCondition": {
+        "comeInWithGrappleJump": {
+          "position": "right"
```
osse101, line 1668:
```suggestion
          "position": "left"
```

region/maridia/outer/Mt. Everest.json
```diff
@@ -4246,6 +4306,23 @@
         "In the model, that'll be represented by going 2 -> 9 -> 5 via two shinespark strats."
       ]
     },
+    {
+      "link": [9, 5],
+      "name": "Leave With Grapple Swing",
```
osse101, line 4311: Is there a leaveWithGrappleJump missing?
I use the blocks below the door to align with the sides, if that matters.

osse101 (commented):
I only tested a few and did not check the block numbers.
- Red Brin Firefleas was very hard to exit with the obstruction.  Wasn't able to do it.
- Pants room might not let you get up to full fling speed.  But you noted that where it came up.
- Screw may be able to help in a few cases.  Or using other kill enemy weapons.
- The note saying which room has the block arrangement is very helpful.

region/maridia/inner-yellow/Maridia Elevator Room.json
blkerby: Agreed, moved those two down to `canPreciseGrappleJump` now.

region/maridia/outer/Mt. Everest.json
blkerby: It would be possible to leave with that condition but currently I don't think it would add anything new. Aside from Everest itself, the two rooms with `comeInWithGrappleJump` entrance condition are Main Street and Red Fish Room, and those also have entrance conditions for `comeInWithGrappleSwing` from the Everest location. Added a devNote now to explain.

strats.md
```diff
@@ -1332,6 +1387,59 @@ __Example:__
...
+A `comeInWithGrappleSwing` implicitly comes with a `canPreciseGrapple` requirement, which includes the Grapple item requirement. If a mid-air morph or grapple jump is required, it would need to be specified explicitly in the `requires` of this strat or the matching strat in the other room.
+
+#### Example
+```json
+{
+  "name": "Grapple Jump",
+  "entranceCondition": {
+    "comeInWithGrappleSwing": {
+      "normalPhysicsBlockPositions": [[8, 3], [7, 3]],
+      "waterPhysicsBlockPositions": [[-1, 5]]
```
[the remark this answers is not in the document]
blkerby: Yup. Good catch, fixed now.

blkerby:
> I only tested a few and did not check the block numbers.
> 
>     * Red Brin Firefleas was very hard to exit with the obstruction.  Wasn't able to do it.

Yeah, it's tricky but once you get the feel for it I think it's not the worst. For me the main thing was to get used to the pattern of swinging back and forth while extending Grapple when moving right, and retracting it when moving left, to avoid the obstruction, like in this video: https://videos.maprando.com/video/4382. 

>     * Screw may be able to help in a few cases.  Or using other kill enemy weapons.

Gave it a try just now in Kassiuz Room, and it looks like when in the grapple jumping state, Screw Attack does not kill the enemy nor protect you. There were other possible ways to kill the Choot though (Missile, Super, Plasma) so I added those.

In Main Street I didn't find a way yet to kill the Skultera while continuing grapple jumping (from either the right or bottom door), though it wouldn't surprise me if there's something that can be done. I saw that it's technically possible to dodge it when entering from the bottom with HiJump but it seems very precise; added a devNote about it. Power Bombs is something I didn't try yet so maybe that could work.

osse101 (approved):
Weird that screw doesn't work in grapple state 🤷

PR #1929, opened 2025-01-31
Video Site Finds Pt. 1

kjbranch:
These are fixes to some of the problems we found from videos on the video site. Most of them are from Sam's videos.

### DIFF
--- a/region/brinstar/red/Beta Power Bomb Room.json
+++ b/region/brinstar/red/Beta Power Bomb Room.json
@@ -387,13 +387,28 @@
         }
       },
       "requires": [
-        "Plasma",
-        "Wave"
+        {"or": [
+          {"and": [
+            "Plasma",
+            "Wave"
+          ]},
+          {"and": [
+            "canInsaneJump",
+            {"or": [
+              "Plasma",
+              {"ammo": {"type": "Super", "count": 6}}
+            ]}
+          ]}
+        ]}
       ],
       "clearsObstacles": ["A"],
+      "note": [
+        "Come in with stored fall speed and get past or kill the Sidehoppers without getting hit.",
+        "It is possible to open the door to back up and quickly kill the hoppers, or to run under them and clip without getting hit."
+      ],
       "devNote": [
         "Coming in with blue speed would be an alternative way to get past the hoppers, but this may be tricky to model.",
-        "Other weapon combinations are possible, but not reliable due to movement being limited."
+        "FIXME: It is possible to get past the sidehoppers and moonfall without getting hit."
       ]
     },
     {
@@ -406,13 +421,28 @@
         }
       },
       "requires": [
-        "Plasma",
-        "Wave"
+        {"or": [
+          {"and": [
+            "Plasma",
+            "Wave"
+          ]},
+          {"and": [
+            "canInsaneJump",
+            {"or": [
+              "Plasma",
+              {"ammo": {"type": "Super", "count": 6}}
+            ]}
+          ]}
+        ]}
       ],
       "clearsObstacles": ["A"],
+      "note": [
+        "Come in with stored fall speed and get past or kill the Sidehoppers without getting hit.",
+        "It is possible to open the door to back up and quickly kill the hoppers, or to run under them and clip without getting hit."
+      ],
       "devNote": [
         "Coming in with blue speed would be an alternative way to get past the hoppers, but this may be tricky to model.",
-        "Other weapon combinations are possible, but not reliable due to movement being limited."
+        "FIXME: It is possible to get past the sidehoppers and moonfall without getting hit."
       ]
     },
     {
--- a/region/brinstar/red/Beta Power Bomb Room.json
+++ b/region/brinstar/red/Beta Power Bomb Room.json
@@ -408,7 +408,8 @@
       ],
       "devNote": [
         "Coming in with blue speed would be an alternative way to get past the hoppers, but this may be tricky to model.",
-        "FIXME: It is possible to get past the sidehoppers and moonfall without getting hit."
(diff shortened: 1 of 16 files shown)


region/crateria/central/Landing Site.json
```diff
@@ -1268,9 +1283,10 @@
     {
       "id": 55,
       "link": [4, 3],
-      "name": "SpeedyJump Into SpringballJump",
+      "name": "Tricky HiJump SpringBall Jump",
       "requires": [
-        "canTrickyDashJump",
+        "canTrickyJump",
```
kjbranch, line 1288: Moved this down to VH. Do you think it should be moved to Hard?

osse101: Seems mechanically fine for hard, very similar to the movement to reach pantry item.
I don't expect people to do it, but it could tighten health counts to where they start thinking about it.

region/maridia/inner-pink/East Sand Hole.json
```diff
@@ -422,6 +422,29 @@
...
+        "canSuitlessMaridia",
+        "HiJump",
+        "canCrouchJump",
+        "canMidAirMorph",
+        "canTrickyJump",
+        "canConsecutiveWalljump",
+        "canPreciseWalljump"
+      ],
+      "flashSuitChecked": true,
+      "note": "Midair Morph through the bottom right Crumble Blocks. Wall Jump and break free of the waterline to get to the item.",
```
osse101, line 445: The middle column is more direct and fewer steps.

region/maridia/inner-pink/East Sand Hole.json
```diff
@@ -422,6 +422,29 @@
...
+      "devNote": "FIXME: This may need Space Jump or a way to break free, like the notable in West Sand Hole."
```
osse101, line 446: You do not need to break free.

region/tourian/main/Metroid Room 4.json
```diff
@@ -270,11 +270,16 @@
       "link": [1, 2],
       "name": "Easy Evade",
       "requires": [
+        "canDodgeWhileShooting",
```
osse101, line 273: This removes a rinka hit lower.

region/brinstar/red/Beta Power Bomb Room.json
```diff
@@ -387,13 +387,28 @@
...
       ],
       "clearsObstacles": ["A"],
+      "note": [
+        "Come in with stored fall speed and get past or kill the Sidehoppers without getting hit.",
+        "It is possible to open the door to back up and quickly kill the hoppers, or to run under them and clip without getting hit."
+      ],
       "devNote": [
         "Coming in with blue speed would be an alternative way to get past the hoppers, but this may be tricky to model.",
-        "Other weapon combinations are possible, but not reliable due to movement being limited."
+        "FIXME: It is possible to get past the sidehoppers and moonfall without getting hit."
```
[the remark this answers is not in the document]
blkerby: Yeah, it is possible with nothing: https://videos.maprando.com/video/4448. Getting down is consistent. Getting back out is the hard part, not sure if it can reliably be done damageless or not.

region/norfair/east/Double Chamber.json
```diff
@@ -1171,6 +1174,40 @@
...
+      "link": [3, 4],
+      "name": "Ripper I-Frames, Spike Jump",
+      "requires": [
+        "canPrepareForNextRoom",
+        "canUseIFrames",
+        {"enemyDamage": {"enemy": "Ripper 2 (green)", "type": "contact", "hits": 1}},
+        {"or": [
+          "canHorizontalDamageBoost",
+          {"and": [
+            "HiJump",
```
osse101, line 1187: I'm not sure how you're meant to do this without falling on the crumbles.  Do you hit the ripper on the right and then again on the left, or activate the ripper by walljumping on the right wall and then blind jump to where it will be on the left?
Wither way doesn't use hijump.

region/norfair/east/Double Chamber.json
```diff
@@ -1171,6 +1174,40 @@
...
+      ],
+      "note": [
+        "Run off or perform a short hop off the ledge into the Ripper and damage boost to the left of the spikes.",
+        "Using the i-frames from the Ripper, quickly jump from the spikes to the item.",
+        "With HiJump, it is also possible to avoid the damage boost and land right of the crumble blocks and get a good jump to the left."
+      ],
+      "devNote": [
+        "FIXME: It is possible to avoid the Ripper and jump off of the crumble blocks.",
+        "With a Spring Ball jump, this requires an very precise jump and very quick pause to get as much horizontal speed as possible before the second jump.",
+        "With HiJump + Wall Jump, it is possible to crumble spin jump from the first crumbles to the second crumbles and then wall jump."
```
osse101, line 1208: This involves running on the crumbles while they break to gain speed.
I would do it by falling through the crumbles and jumping back up off the kago first so I could see where the blocks are.

region/norfair/east/Double Chamber.json
kjbranch: Hit the Ripper, fall to the left of the crumbles, get a running jump off the spikes and wall jump on the top left ledge. 
Only the HiJump variant can land right of the crumbles, I couldnt get the spring ball variant to jump far enough left

osse101: That feels fine to do, I just didn't think it would reach.
With springball you want to morph faster for some springfling, or run on the crumbles a bit.  And that is going to be harder than damage boosting on the bug.  And if you want to overcomplicate the strat, you can bounce on the crumbles before touching the ripper to be on the left side with iframes.

PR #1933, opened 2025-02-02
Fix/refine Draygon Microwave Fight strat

blkerby:
This adds a Charge requirement which appears to have been missing. Also adds `canTrickyDodgeEnemies` as an alternative to some of the other requirements.

This was prompted by Sam's video which does a 1-cycle with Charge+Plasma beam only and without breaking the turrets: https://videos.maprando.com/video/4356

### DIFF
--- a/region/maridia/inner-pink/Draygon's Room.json
+++ b/region/maridia/inner-pink/Draygon's Room.json
@@ -303,7 +303,14 @@
       "requires": [
         "h_canNavigateUnderwater",
         {"or": [
-          "canTrickyDodgeEnemies",
+          {"and": [
+            "canTrickyDodgeEnemies",
+            {"or": [
+              "Gravity",
+              "canInsaneJump",
+              {"enemyDamage": {"enemy": "Draygon", "type": "turretProjectile", "hits": 1}}
+            ]}
+          ]},
           "h_canBreakThreeDraygonTurrets"
         ]},
         "canXRayWaitForIFrames",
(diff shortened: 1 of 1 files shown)


region/maridia/inner-pink/Draygon's Room.json
```diff
@@ -302,10 +302,15 @@
       "name": "Microwave Fight",
       "requires": [
         "h_canNavigateUnderwater",
-        "h_canBreakThreeDraygonTurrets",
+        {"or": [
+          "canTrickyDodgeEnemies",
+          "h_canBreakThreeDraygonTurrets"
```
osse101, line 307:
```suggestion
        {"or": [
          {"and": [
            "canTrickyDodgeEnemies",
            {"or": [
              "Gravity",
              "canInsaneJump",
              {"enemyDamage": {"enemy": "Draygon", "type": "turretProjectile", "hits": 1}}
            ]}
          ]},
          "h_canBreakThreeDraygonTurrets"
```
- You can't react dodge the turrets, so you need to know what you are doing.  Waiting on the left side of the room works well
- Dying in 1 hit is very uncomfortable.  Being able to survive a hit helps a lot.
- Turrets can RNG snipe you right when draygon swoops and there is nothing you can do.

region/maridia/inner-pink/Draygon's Room.json
```diff
@@ -302,10 +302,15 @@
...
+        ]},
         "canXRayWaitForIFrames",
+        "Charge",
         "Plasma",
         {"or": [
+          "canTrickyDodgeEnemies",
```
osse101, line 313: 1 rounding with only plasma is bordering extreme, but you feel that its your own fault for not getting it.

PR #1935, opened 2025-02-02
Fix to Double Chamber Hijumpless Spike Speedjump

blkerby:
The strat was missing a wall jump requirement, making the logic unsound. While fixing it, went ahead and also added the Ripper damage boost variant. Videos:
- Spike hit: https://videos.maprando.com/video/4447
- Ripper hit: https://videos.maprando.com/video/4446

### DIFF
--- a/region/norfair/east/Double Chamber.json
+++ b/region/norfair/east/Double Chamber.json
@@ -1258,10 +1258,16 @@
       "name": "Hijumpless Spike Speedjump",
       "requires": [
         {"notable": "Hijumpless Spike Speedjump"},
+        {"or": [
+          {"spikeHits": 1},
+          {"and": [
+            {"enemyDamage": {"enemy": "Ripper 2 (green)", "type": "contact", "hits": 1}},
+            "canHorizontalDamageBoost"
+          ]}
+        ]},
         "canUseIFrames",
-        {"spikeHits": 1},
-        "SpeedBooster",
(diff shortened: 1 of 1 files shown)


blkerby:
> The ripper option wasn't added because it makes the trick a little harder. I'd want to put a tech for doing an advanced horizontal damage boost on it, or an insane jump for hitting the ripper in an odd place.

Fair enough. Added a `canInsaneJump` and a note explaining to get the boost from a low position.

> The spike hit should have canHorizontalDamageBoost tech too. You need to control how the damage boost moves samus, and in this case you need to be pushed back and down.

To get pushed back and down, it only requires holding forward (i.e. left) before landing. So I was thinking it's not exactly a `canHoizontalDamageBoost` which would mean holding jump and timing a backward press during knockback. At least, we haven't been putting that requirement on other strats where you hold forward, like crossing Crateria Super Room with 1 hit or Bowling Alley with 3 hits. I think knowing to hold forward was understood as part of the `canIFrameSpikeJump` tech (now `canUseIFrames`); though that tech doesn't actually describe this technique, so maybe it should be added to the tech description?

osse101 (approved):
You're right that you probably get used to this behaviour by running across spikes.  I also think people don't think about being deliberate with forward boosting, since it only helps in specific cases.  Makes it hard to place as a tech.

PR #1945, opened 2025-02-11
Add G-Mode to Croc Escape

kjbranch:
Maddo said the tests could be edited to allow these helpers instead of heat frames.

### DIFF
--- a/region/norfair/west/Crocomire Escape.json
+++ b/region/norfair/west/Crocomire Escape.json
@@ -136,7 +136,6 @@
         "canTrickyUseFrozenEnemies",
         {"or": [
           "HiJump",
-          "canSpringBallJumpMidAir",
           "canTrickyDodgeEnemies"
         ]},
         "h_HeatedGModeOpenDifferentDoor"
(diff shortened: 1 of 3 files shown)


region/norfair/west/Crocomire Escape.json
```diff
@@ -549,7 +796,14 @@
...
         "canUnmorphBombBoost",
         "h_canCrouchJumpDownGrab",
-        {"heatFrames": 800}
+        {"heatFrames": 800},
+        {"or": [
+          "canInsaneJump",
+          {"and": [
+            {"heatFrames": 250},
+            "h_additionalBomb"
+          ]}
```
kjbranch, line 805: I added a leniency PB attempt in Expert. I could remove it or increase it; whatever you think is best.

osse101 (approved):
- A CF option would indeed enable more ways of doing the room.  As would lava a lava path.  Shinesparking both ways would be pretty advanced.  

- Most hot rooms probably benefit from a CF so that you can end Gmode and open a door. Probably mainly the rooms where you do a GMode thing like walking through this green gate.

- I think unmorph bomb boost does need more leniency everywhere.  And the other strats that have a `retry OR harderTech` structure like bomb jump water escape, walljump morph bomb boost.  I think not springball bomb jump or can bomb horizontally.

PR #1947, opened 2025-02-13
G-Mode: Remaining Upper Norfair West

kjbranch:
Also Post Crocomire Power Bomb Room

### DIFF
--- a/region/norfair/west/Ice Beam Tutorial Room.json
+++ b/region/norfair/west/Ice Beam Tutorial Room.json
@@ -266,10 +266,7 @@
             "h_canArtificialMorphSpringFling",
             "canTrickyJump"
           ]},
-          {"and": [
-            "h_canArtificialMorphHBJ",
-            "canInsaneJump"
-          ]},
+          "h_canArtificialMorphHBJ",
           {"and": [
             "Gravity",
             "h_canArtificialMorphIBJ",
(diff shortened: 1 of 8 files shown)


region/norfair/west/Ice Beam Tutorial Room.json
```diff
@@ -360,6 +409,79 @@
...
+            "Morph",
+            "canTrickyUseFrozenEnemies"
+          ]},
+          {"and": [
+            "h_canArtificialMorphIBJ",
+            "h_canArtificialMorphBombHorizontally",
+            "canTrickyUseFrozenEnemies"
+          ]},
+          {"and": [
+            "h_canArtificialMorphHBJ",
```
osse101, line 437: I did this with an IBJ+bomb horizontally on the right, and a bomb horizontally into unmorph+down back on the left.  I couldn't make the jump with the bomb alone.  I had rolling speed when getting hit by the bomb.

The left side HBJ looks like you sit on the farthest pixel over and just time the HBJ right...  It's harder than a regular HBJ.  You can also start an IBJ facing forward or backwards with the same setup and horizontal boost from there.  Or DIBJ.

I don't like the consistency of any of these options.

kjbranch: We don't yet have a tech for rolling speed + bomb horizontally. I think there would be quite a few places it could be used, but it feels rough.

I added a dev note and the dibj option. Do you think it should be changed further?

PR #1960, opened 2025-02-27
Add canHeatedGMode

kjbranch:
Its uses should be able to be added implicitly. I'm not expected that entering and exiting a heated room while already in G-mode to require this, but might not matter much either way (i.e. enterring Caterpillar Room in direct, using the elevator into LN Elevator Room and returning to despawn the gate)

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -1067,8 +1067,18 @@
           ]
         },
         {
+          "name": "h_heatedGMode",
+          "requires": [
+            {"or": [
+              "h_heatProof",
+              "canHeatedGMode"
+            ]}
+          ]
+        },
+        {
           "name": "h_DirectHeatedGModeLeaveSameDoor",
           "requires": [
+            "h_heatedGMode",
             {"heatFrames": 1}
           ],
           "note": "Leaving the same door only adds one heat frame. This is likely only useful if remote acquiring an item or opening a yellow door."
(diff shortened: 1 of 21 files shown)


blkerby:
It would probably be good to update `strats.md` to describe the rules for when this tech will be implicit. Probably in the sections for `leaveWithGModeSetup` and `comeInWithGMode`

blkerby:
I pushed a commit to document when `canHeatedGMode` is implicitly required. For this it seemed helpful to define a helper `h_heatedGMode`, and to use that in the various helpers that exit G-mode and go through a door. Things to note:
- The implicit `h_heatedGMode` requirement only exists when a `comeInWithGMode` matches with `leaveWithGModeSetup`, and one of the two rooms is heated. If the setup room is heated, then managing reserve energy can be quite tricky, and it will likely not be possible to retry. If the destination room is heated, then Samus will instantly die if failing to get G-mode. So both those cases are what we're trying to cover with `canHeatedGMode`.
- The implicit `h_heatedGMode` requirement doesn't exist when a `comeInWithGMode` matches with `leaveWithGMode`. This is because heat does not create any difficulty when you already have G-mode and carry it into another room. So this allows us to account for the case of going up and down the elevator from Caterpillar Room to Lower Norfair Elevator.
- Putting `h_heatedGMode` requirement on the door-exit helpers is because this is the other place where a difficulty happens relating to heated G-mode. Usually the `h_heatedGMode` would already be implicit in the `comeInWithGMode` too; but in case you entered through an open door (e.g. Crateria Tube) with `leaveWithGMode`, it wouldn't, so that's why it matters to put the requirement on these helpers.
- By avoiding making a special exception for the elevator `comeInWithGMode`+`leaveWithGMode` case, this approach is flexible and can also work in case, for example, there are ever strats that enter a heated room and pass through G-mode junctions before leaving the room while still in G-mode.

osse101 (dismissed):

PR #1961, opened 2025-02-28
In-room spikesuits

blkerby:
This is a first pass at adding strats for spikesuit setups, only covering cases where an in-room runway can be used to gain the shinecharge. So this isn't including cross-room strats (comeInShinecharged or comeInShinecharging), nor cases where X-mode could be used to gain the shinecharge. I'm assuming that an X-mode + spikesuit combo might deserve a new (Insane) tech.

Someday I'm still hoping to adapt something like the `failures` schema, to create a more general way of expressing lenience.


region/lowernorfair/east/Lower Norfair Fireflea Room.json
```diff
@@ -407,6 +407,31 @@
...
+        {"spikeHits": 1},
+        {"or": [
+          "h_SpikeSuitSpikeHitLeniency",
+          {"and": [
+            {"resourceCapacity": [{"type": "PowerBomb", "count": 1}]},
+            {"resetRoom": {"nodes": [2]}},
+            {"or": [
+              "SpaceJump",
+              "HiJump",
+              "canPreciseWalljump"
```
kjbranch, line 425:
```suggestion
              "canPreciseWalljump",
              "canSpringBallJumpMidAir",
              "canIBJ"
```

region/wreckedship/main/Bowling Alley.json
```diff
@@ -1138,6 +1139,25 @@
         "Use X-Ray or a Crystal Flash to stand up, followed by a Partial Floor clip to jump through the Crumble blocks."
       ]
     },
+    {
+      "link": [4, 4],
+      "name": "Gain Flash Suit (Spikesuit)",
+      "requires": [
+        "f_DefeatedPhantoon",
```
kjbranch, line 1146:
```suggestion
        "f_DefeatedPhantoon",
        {"obstaclesNotCleared": ["B"]},
```

kjbranch (approved):
Looks good. 

I went through the rooms to see which ones could be included but weren't (just to document them, not that they should be added here):
- Noob Bridge - sounds like you couldnt get it; might be doable with Ice+Wave frozen Zeela platform to get another tile, or gmode overload crumbles
- Red Firefleas - very short
- Croc Speedway - (can it be done without g-mode?)
- Speedbooster Hallway
- Draygon
- Lava Dive (lava proof, space?)
- Wasteland?
- Electric Death
- WS Main Shaft
- WS East Super
- Basement

osse101 (approved):
- hotarubi room maybe takes more figuring out compared to the rest of this list, but the not-yet-added spikesuits are going to be even harder.  And the trick will always be baseline very difficult.

- Noob bridge in the distant past was done with a bomb jump into the spikes or springball.  I got it with a quickdrop + HiJump.

- https://videos.maprando.com/video/5401  Shooting the block or falling through the crumble worked.

region/lowernorfair/east/Lower Norfair Fireflea Room.json
blkerby: Spring Ball jump and IBJ were considered too slow to be reasonable as a way to farm between attempts. They're also left out of the "Fireflea and Fune Farm" strat. I don't know though; maybe they could be ok to add? @osse101 What do you think?

osse101: In my vanilla map randomizer experience, farming firefleas takes much longer than you would think.  And gets a lot more complicated if you want all the drops instead of 2health+1PB.
IBJ or springballjump would be fine for the ridley claw jump.  But you have to do another speedyjump into itemTech to reach the Fune.  And doing both is too much.  But if logic is officially only requiring 1 trip or 1 and a bit of fireflea farming, you could do that.
Spikesuits are not that likely to hit in 3 tries but I wouldn't want to require more effective health either.

kjbranch: I was thinking that the spike suit leniency is only 2 extra spike hits, so that would mean that you're not needing to farm a lot of Energy. But you do need to use a PB to get down there and reset and use another PB to turn the lights back on. I think IBJ is probably not worth including, but spring may be?

osse101: With springball its a trickydash jump, and that's not fun in the dark.  I'd be more ok with `canJumpIntoIBJ` getting up.  Pausing a bunch is just less fun than IBJing too.
The first PB gives two Fune drops and that pays for itself.  The farming sequence is pretty slow though.
It's `canSpikesuit` level of difficulty & movement so they can make it work, but I'd rather require the extra health.  Also, it feels bad to only get one try at the spikesuit per shortcharge.
It looks like the way it plays out, you make one attempt and then farm 90% of your health back and then resetRoom.

PR #1965, opened 2025-03-06
G-Mode: Double Chamber

kjbranch:
This was a particularly difficult room, as it's difficult to traverse and the gate, item, and Kago all made it worse.
I think the two nodes ended up being worthwhile. I didn't see a good place to add another, but if you have any ideas for how to clean it up, I could redo the room.

### DIFF
--- a/region/norfair/east/Double Chamber.json
+++ b/region/norfair/east/Double Chamber.json
@@ -86,14 +86,14 @@
     },
     {
       "id": 7,
-      "name": "G-Mode Junction (Bottom Left)",
+      "name": "G-Mode Junction (Bottom Right)",
       "nodeType": "junction",
       "nodeSubType": "g-mode",
       "mapTileMask": [
         [1, 1, 1, 1],
         [1, 1, 2, 2]
       ],
-      "note": "Represents being at the bottom right with G-Mode and Morph or artificial morph."
+      "note": "Represents being at the bottom right with G-Mode and Morph."
     }
   ],
   "obstacles": [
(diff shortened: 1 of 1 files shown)


region/norfair/east/Double Chamber.json
```diff
@@ -723,6 +1016,55 @@
...
+          ]},
+          "canTrickyDashJump",
+          {"and": [
+            {"canShineCharge": {
+              "usedTiles": 28,
+              "gentleUpTiles": 3,
+              "gentleDownTiles": 3,
+              "openEnd": 0
+            }},
+            "canMidairShinespark",
```
osse101, line 1044: I don't know where we are jumping to or sparking.  Contextually you pass through a descending platform.  But it seems like canShinechargeMovementComplex + a description.

osse101 (commented):
Yeah its tricky when you need to do every node in [1, 2, 3, 4] to every other node and also have [direct, indirect].
Something that stands out is the "up through crumbles" strats not going through 7.  But its only a couple of cases where that would help.  Alternatively 7 could have a more specific description.

region/norfair/east/Double Chamber.json
kjbranch: This is just sparking up, similar to strat 13. It's not super likely to be useful until CF options are added, but I pointed out the speed options, since they are a bit more obscure. There's also the energy free shinespark option which could make it applicable.

kjbranch: It's because there is a refill which isn't modeled and then heat damage (+ shinespark damage) taken while reaching the door. This strat is just saying all of those cancel out, partially to make it so that this strat won't be used to gain energy. Without modeling the refill I dont see any other way to model the spark, though we could change it from 1/1 frames/excess frames to 0/0 or 17/17; not sure which of the three would be most clear (also not sure if 0 would validate).

PR #1968, opened 2025-03-10
Refine Oasis jump-through-top-door strats, add detailNotes schema

blkerby:
- Update the notes of the strats that use Screw Attack to jump up through the top door of Oasis.
- Add blue speed variations that don't require Screw Attack.
- Add schema for `detailNotes`.

The existing notes here very much overstate how much precision is needed in the amount of runway to use. There's a wide range that works well, with only a slight benefit to using the minimal amount of 13 tiles. The blue speed version was also not so bad: with a 2-tap, I could consistently get run speed in a good range, and from there the trick is the same as the version with Screw Attack.

Videos for the blue speed version:
- left-to-right: https://videos.maprando.com/video/5482
- right-to-left: https://videos.maprando.com/video/5483

### DIFF
--- a/region/maridia/inner-green/Oasis.json
+++ b/region/maridia/inner-green/Oasis.json
@@ -569,7 +569,7 @@
       ],
       "detailNote": [
         "The optimal amount of extra run speed is $3.4 or $3.5, by using 13 tiles of runway or about a half tile more;",
-        "this gives a 7-window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
+        "this gives a 7-frame window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
         "Extra run speeds between $3.6 and $3.F can also work, but with shorter windows.",
         "A perfect boomerang is always required for the second turnaround, regardless of where in the window it is done.",
         "It is recommended to buffer the spin break, by holding angle (up or down), jump, and forward through the transition,",
@@ -610,7 +610,7 @@
       ],
       "detailNote": [
         "Lower run speeds are generally better, as long as the extra run speed is at least $3.4.",
-        "this gives up to a 7-window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
+        "this gives up to a 7-frame window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
         "A perfect boomerang is always required for the second turnaround, regardless of where in the window it is done.",
         "It is recommended to buffer the spin break, by holding angle (up or down), jump, and forward through the transition,",
         "then switching from forward to backward, with at most 1 frame of neutral in between, while still holding jump and angle."
@@ -1217,7 +1217,7 @@
       ],
       "detailNote": [
         "The optimal amount of extra run speed is $3.4 or $3.5, by using 13 tiles of runway or about a half tile more;",
-        "this gives a 7-window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
+        "this gives a 7-frame window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
         "Extra run speeds between $3.6 and $3.F can also work, but with shorter windows.",
         "A perfect boomerang is always required for the second turnaround, regardless of where in the window it is done.",
         "It is recommended to buffer the spin break, by holding angle (up or down), jump, and forward through the transition,",
@@ -1258,7 +1258,7 @@
       ],
       "detailNote": [
         "Lower run speeds are generally better, as long as the extra run speed is at least $3.4.",
-        "this gives up to a 7-window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
+        "this gives up to a 7-frame window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
         "A perfect boomerang is always required for the second turnaround, regardless of where in the window it is done.",
         "It is recommended to buffer the spin break, by holding angle (up or down), jump, and forward through the transition,",
         "then switching from forward to backward, with at most 1 frame of neutral in between, while still holding jump and angle."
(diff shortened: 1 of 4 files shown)


region/maridia/inner-green/Oasis.json
```diff
@@ -560,12 +560,61 @@
...
+        "Longer runway lengths can also work, up to 17 tiles, but will make the trick somewhat more precise.",
+        "Use angle to break spin, and do a momentum conserving turnaround either on the same frame or one frame later;",
+        "this first turnaround should be done early enough that the turnaround is complete by the time Samus clears the two-tile passage in the middle of the room.",
+        "Do a second momentum conserving turnaround must be done somewhat late, and by switching from holding left to holding right one frame later (a perfect 'boomerang').",
+        "Shoot open the door, and use a third momentum conserving turnaround on it as it opens."
       ],
-      "devNote": "This does not have collision oscillation. It is possible to do this without Screw Attack, but is extremely precise."
+      "detailNote": [
+        "The optimal amount of extra run speed is $3.4 or $3.5, by using 13 tiles of runway or about a half tile more;",
+        "this gives a 7-window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
```
kjbranch, line 572:
```suggestion
        "this gives a 7-frame window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
```

region/maridia/inner-green/Oasis.json
```diff
@@ -560,12 +560,61 @@
...
+      "note": [
+        "Use a 2-tap shortcharge and jump into the room with a last-frame jump, breaking the bomb block with blue speed, and making it all the way through the top door.",
+        "Use angle to break spin, and do a momentum conserving turnaround either on the same frame or one frame later;",
+        "this first turnaround should be done early enough that the turnaround is complete by the time Samus clears the two-tile passage in the middle of the room.",
+        "Do a second momentum conserving turnaround must be done somewhat late, and by switching from holding left to holding right one frame later (a perfect 'boomerang').",
+        "Shoot open the door, and use a third momentum conserving turnaround on it as it opens."
+      ],
+      "detailNote": [
+        "Lower run speeds are generally better, as long as the extra run speed is at least $3.4.",
+        "this gives up to a 7-window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
```
kjbranch, line 613:
```suggestion
        "this gives up to a 7-frame window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
```

region/maridia/inner-green/Oasis.json
```diff
@@ -1159,12 +1208,61 @@
...
+        "Ideally, gain speed using between 13 and 13.5 tiles of runway (with an open end).",
+        "Longer runway lengths can also work, up to 17 tiles, but will make the trick somewhat more precise.",
+        "Use angle to break spin, and do a momentum conserving turnaround either on the same frame or one frame later;",
+        "this first turnaround should be done early enough that the turnaround is complete by the time Samus clears the two-tile passage in the middle of the room.",
+        "The second momentum conserving turnaround must be done somewhat late, and by switching from holding forward to holding backward one frame later (a perfect 'boomerang').",
+        "Shoot open the door, and use a third momentum conserving turnaround on it as it opens."
+      ],
+      "detailNote": [
+        "The optimal amount of extra run speed is $3.4 or $3.5, by using 13 tiles of runway or about a half tile more;",
+        "this gives a 7-window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
```
kjbranch, line 1220:
```suggestion
        "this gives a 7-frame window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
```

region/maridia/inner-green/Oasis.json
```diff
@@ -1159,12 +1208,61 @@
...
+      "note": [
+        "Use a 2-tap shortcharge and jump into the room with a last-frame jump, breaking the bomb block with blue speed, and making it all the way through the top door.",
+        "Use angle to break spin, and do a momentum conserving turnaround either on the same frame or one frame later;",
+        "this first turnaround should be done early enough that the turnaround is complete by the time Samus clears the two-tile passage in the middle of the room.",
+        "The second momentum conserving turnaround must be done somewhat late, and by switching from holding forward to holding backward one frame later (a perfect 'boomerang').",
+        "Shoot open the door, and use a third momentum conserving turnaround on it as it opens."
+      ],
+      "detailNote": [
+        "Lower run speeds are generally better, as long as the extra run speed is at least $3.4.",
+        "this gives up to a 7-window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
```
kjbranch, line 1261:
```suggestion
        "this gives up to a 7-frame window for the first turnaround, and typically a 4-frame window for the second turnaround if the first turnaround is done early enough.",
```

kjbranch (commented):
Just looked through this quickly - I don't know much about the strat, other than watching Eddie work on it. 

The `detailNote` looks good. Is it going to be available everywhere `note` was available? How are you intending to show this information on Map Rando?

blkerby:
> The `detailNote` looks good. Is it going to be available everywhere `note` was available?

For now this is only adding it as a strat property. It could be added in more places though.

> How are you intending to show this information on Map Rando?

I think for now it can just be shown as a second paragraph. We could consider maybe collapsing it later.

kjbranch:
I think `detailNote` would also make sense on tech. Probably not needed anywhere else. It would make sense to eventually go through devNotes and convert them to detailNotes. I can look through that sometime if that would be helpful

osse101 (commented):
- The new description helps a lot
- The blue speed version is completely reasonable

My vision of note/detail note would have note much simpler "crossroom jump+precise+momentumTurnarounds".  If I want to toggle the strat on/off, its better to be super consise.  If you want to know how to do it, this level of detail is great for a note.  Harder strats perhaps should have a bigger note?  You need to know where it can go wrong, and essentially are going to learn and practice the strat before including it in logic.  But detailNote here is still a step up in detail and there is a good separation here.

PR #1970, opened 2025-03-10
Business Center temp blue

blkerby:
Gaining a shinecharge in-room would also be possible, but I didn't want to do it; could be added later.

### DIFF
--- a/region/norfair/west/Business Center.json
+++ b/region/norfair/west/Business Center.json
@@ -183,6 +183,7 @@
         {"id": 1},
         {"id": 2},
         {"id": 3},
+        {"id": 4},
         {"id": 5},
         {"id": 6},
         {"id": 7},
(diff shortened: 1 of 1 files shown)


region/norfair/west/Business Center.json
```diff
@@ -1972,6 +2319,29 @@
...
+        "comeInShinecharging": {
+          "length": 8,
+          "openEnd": 0,
+          "gentleUpTiles": 4
+        }
+      },
+      "requires": [
+        "canLongChainTemporaryBlue",
+        {"or": [
+          "HiJump",
```
osse101, line 2335: Does this have a shorter runway, with the overhang?

blkerby: Oh right, yeah that needs to be split into a separate strat since the usable runway is 1 tile less in that case. Done.

PR #1972, opened 2025-03-11
Ice Beam Gate Room anti-shortcharge

blkerby:
- Adds a notable strat Ice Beam Gate Room: Anti-Short Charge Hero Shot.
- Adds a property `minExtraRunSpeed` to `leaveWithRunway`, to represent the new strat.
- Adds a "Cross-Room Blue Speed Jump" strat in East Sand Hall, which can match with the new strat in Ice Beam Gate Room. 
- Adds "comeInNormally" entrance conditions to Ice Beam Gate Hall strats that rely on the gates still being up (since we don't have an obstacle to represent their state).

There's already a "Cross-Room Blue Speed Jump, Leave With Temporary Blue" strat in East Sand Hall. The only reason the more simple "Cross-Room Blue Speed Jump" didn't already exist is because I was thinking it was covered by the "Cross-Room Speed Jump", which doesn't need as much runway. But that one has a specific run speed requirement which can't be met coming from Ice Beam Gate Room.

The "anti-short charge" feels fairly unintuitive, and the player may only get one attempt, so I was thinking the notable would be placed in either Extreme or Insane?

### DIFF
--- a/region/maridia/inner-green/East Sand Hall.json
+++ b/region/maridia/inner-green/East Sand Hall.json
@@ -220,7 +220,7 @@
     {
       "id": 8,
       "link": [1, 2],
-      "name": "Cross-Room Speed Jump (Left to Right)",
+      "name": "Cross-Room Speed Jump over Evirs",
       "entranceCondition": {
         "comeInRunning": {
           "speedBooster": true,
@@ -229,7 +229,7 @@
         }
       },
       "requires": [
-        {"notable": "Cross-Room Speed Jump"},
+        {"notable": "Cross-Room Speed Jump over Evirs"},
         "canPlayInSand",
         "canCrossRoomJumpIntoWater",
         "canTrickyDashJump"
@@ -243,7 +243,7 @@
     {
       "id": 9,
       "link": [1, 2],
-      "name": "Cross-Room SpringBall Jump (Left to Right)",
+      "name": "Cross-Room Spring Ball Jump",
       "entranceCondition": {
         "comeInRunning": {
           "speedBooster": true,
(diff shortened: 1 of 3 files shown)


region/maridia/inner-green/East Sand Hall.json
```diff
@@ -679,6 +679,28 @@
         "A little higher run speed can also work but may require releasing jump near the peak to avoid getting caught on the overhang."
       ]
     },
+    {
+      "link": [2, 1],
+      "name": "Cross-Room Blue Speed Jump",
```
osse101, line 684: The name sounds a lot like the notable

region/maridia/inner-green/East Sand Hall.json
```diff
@@ -679,6 +679,28 @@
...
+        "comeInGettingBlueSpeed": {
+          "length": 1,
+          "openEnd": 1,
+          "minExtraRunSpeed": "$4.4"
+        }
+      },
+      "requires": [
+        "canPlayInSand",
+        "canCrossRoomJumpIntoWater",
+        "canInsaneJump"
```
osse101, line 695: This strat would be easier than the cross room speed jump, in expert

region/maridia/inner-green/East Sand Hall.json
blkerby: True. Created a separate notable for this now: "Cross-Room Blue Speedy Jump (Right to Left)". And renamed the existing notable "Cross-Room Speed Jump" to "Cross-Room Speed Jump over Evirs", to make the difference more clear. And a few other little clean-ups to the names. For example, I tried to remove most instances of "Left to Right" or "Right to Left" from strat names where the direction doesn't make a significant difference.

region/maridia/inner-green/East Sand Hall.json
blkerby: Makes sense. So we can put the new notable in Expert: "Cross-Room Blue Speed Jump (Right to Left)". I did also test it again left-to-right but found it doesn't really work, or at least not any better than just doing the (Extreme) regular speedy jump over the Evirs. The trouble with left-to-right is its hard to get high enough to clear the pillar at the end.

PR #1973, opened 2025-03-12
Most of rest of West Norfair temp blue

blkerby:
This just leaves Crocomire Speedway for a separate PR. I'm also skipping Ice Beam Tutorial Room and Ice Beam Snake Room, which probably have some dumb things that could be done to carry temp blue through them.

This also includes a new mockball strat for Ice Beam Gate Room left-to-right, using Speedbooster to do it with a much shorter other-room runway, by delaying the mockball until right before the gate.

### DIFF
--- a/region/norfair/west/Ice Beam Acid Room.json
+++ b/region/norfair/west/Ice Beam Acid Room.json
@@ -257,8 +257,20 @@
         }
       },
       "requires": [
-        "canLongChainTemporaryBlue",
-        {"heatFrames": 370}
+        {"or": [
+          {"and": [
+            "canBlueSpaceJump",
+            {"heatFrames": 215}
+          ]},
+          {"and": [
+            "canTrickySpringBallBounce",
+            {"heatFrames": 215}
+          ]},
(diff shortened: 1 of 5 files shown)


region/norfair/west/Ice Beam Acid Room.json
```diff
@@ -246,6 +246,43 @@
...
+        "comeInGettingBlueSpeed": {
+          "length": 2,
+          "openEnd": 1,
+          "minExtraRunSpeed": "$1.D"
+        }
+      },
+      "requires": [
+        {"or": [
+          {"and": [
+            "canBlueSpaceJump",
```
osse101, line 262: This should have a max extra run speed.  Maybe springball needs one too.
These are reminiscent of pit room.

blkerby: Cool, just added a max speed now across both strats. The canLongChainTemporaryBlue variants might not need it, but it didn't seem worth splitting off into a separate strat

PR #1975, opened 2025-03-15
Add missing Grapple Teleport strats, use 'Door Lock Skip' in names

blkerby:
- Add a handful of missing Grapple Teleport strats.
- Use "Door Lock Skip" consistently in strat names involving grapple teleports that bypass a door shell.
- Remove an unsound grapple teleport strat in Mt. Everest. I think I was fooled by the Practice Hack's door portal behavior when testing this before. Approaching the morph tunnel transition from the back will usually put Samus OOB in the next room. There could possibly be some in-bounds applications, which we could add later if discovered.
- Clarifications to a few notes.

### DIFF
--- a/region/brinstar/blue/Blue Brinstar Boulder Room.json
+++ b/region/brinstar/blue/Blue Brinstar Boulder Room.json
@@ -560,7 +560,7 @@
     {
       "id": 27,
       "link": [2, 1],
-      "name": "Grapple Teleport",
+      "name": "Grapple Teleport Door Lock Skip",
       "entranceCondition": {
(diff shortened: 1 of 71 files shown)


kjbranch:
> * Remove an unsound grapple teleport strat in Mt. Everest. I think I was fooled by the Practice Hack's door portal behavior when testing this before. Approaching the morph tunnel transition from the back will usually put Samus OOB in the next room. There could possibly be some in-bounds applications, which we could add later if discovered.

There are a handful of rooms that could work, but none of them would be able to geometrically connect to Everest.

osse101 (approved):

PR #1977, opened 2025-03-16
Refine Precious Room shinecharge/shinespark strats

blkerby:
I noticed that during the second seed of Nito's recent 24hr stream, he got Gravity early by sequence breaking the Precious Room stutter water shinecharge, doing it with 2 runway tiles when the logic required 2.4375.

I dug into it a little to see how the runway length affects the frame window, and concluded that we could probably use some more tech to help place strats into an appropriate difficulty level. We moved "canStutterWaterShinecharge" down to Very Hard a while ago, but in Precious Room with only 2.4375 tiles I think it's too tough for Very Hard, and would fit better in Expert. With only 2 tiles, I think it fits in Extreme, because this case needs a double-frame-perfect stutter: releasing for exactly 3 frames, and repressing on the last possible frame.

- New tech `canPreciseStutterWaterShinecharge` (for Expert) and `canTrickyStutterWaterShinecharge` (for Extreme).
- New node (4) representing being at the bottom-right of the room with a shinecharge. The new node is helpful since there are many ways to get the shinecharge there, and also many things to do with it once you have it.
- Split the existing stutter water shinecharge into 3 variants, requiring a different tech based on runway length.
- Switch to using the `comeInStutterShinecharging` rather than `comeInRunning`. The difference currently doesn't matter for the randomizer, but it's more accurate since it is in fact a different way of entering the room.
- Added a regular water shinecharge strat, which was missing.
- Tightened the shinecharge frames and shinespark frames.
- Added a `leaveShinecharged` variant that doubles back out the bottom-left door with frames remaining.
- Likewise, add a `leaveWithTemporaryBlue` variant that doubles back out the bottom-left door.
- Added `detailNote` to the schema for tech.
- Added more information on the `canStutterWaterShinecharge` tech, to recommend using 5 tiles of runway. Based on my testing, this is the optimal length to use, at least when applied to entering Precious Room. Going lower than 5 the stutter gradually gets more precise (narrower frame windows for the release and repress). Using 4 or 6 (or somewhere in between) is not too much worse, but 7 or higher seems to not work at all (for this room).

The new tech can of course be applied in more rooms, so the plan is to cover other rooms in follow-up PRs.

### DIFF
--- a/region/maridia/inner-pink/The Precious Room.json
+++ b/region/maridia/inner-pink/The Precious Room.json
@@ -53,6 +53,18 @@
         [1, 0],
         [1, 0]
       ]
+    },
+    {
+      "id": 4,
+      "name": "Bottom Right Shinecharged",
+      "nodeType": "junction",
+      "nodeSubType": "junction",
+      "note": "Gaining a shinecharge at the bottom-right of the room.",
+      "mapTileMask": [
+        [1, 1],
+        [1, 0],
+        [2, 0]
+      ]
     }
   ],
   "enemies": [],
@@ -72,13 +84,21 @@
       "from": 2,
       "to": [
         {"id": 1},
-        {"id": 2}
+        {"id": 2},
+        {"id": 4}
       ]
     },
     {
       "from": 3,
       "to": [
         {"id": 1}
+      ]
+    },
+    {
+      "from": 4,
+      "to": [
+        {"id": 1},
+        {"id": 2}
       ]
     }
   ],
@@ -184,48 +204,47 @@
       ]
     },
     {
-      "id": 9,
-      "link": [2, 1],
-      "name": "Shinespark",
-      "entranceCondition": {
-        "comeInShinecharging": {
-          "length": 12,
-          "openEnd": 0
-        }
-      },
-      "requires": [
-        "Gravity",
-        {"shinespark": {"frames": 40, "excessFrames": 8}}
-      ]
-    },
-    {
       "id": 10,
       "link": [2, 1],
-      "name": "Shinespark, Come in Shinecharged",
+      "name": "Come in Shinecharged, Shinespark (Gravity)",
       "entranceCondition": {
         "comeInShinecharged": {}
       },
       "requires": [
-        {"shineChargeFrames": 40},
         "Gravity",
-        {"shinespark": {"frames": 40, "excessFrames": 8}}
+        {"or": [
+          {"and": [
+            {"shineChargeFrames": 20},
+            {"shinespark": {"frames": 40, "excessFrames": 8}}
+          ]},
+          {"and": [
+            {"shineChargeFrames": 30},
+            "canMidairShinespark",
+            {"shinespark": {"frames": 33, "excessFrames": 8}}
+          ]}
+        ]}
       ],
       "flashSuitChecked": true
     },
     {
       "id": 11,
       "link": [2, 1],
-      "name": "Suitless Shinespark",
+      "name": "Come In Shinecharged, Shinespark (Suitless)",
       "entranceCondition": {
         "comeInShinecharged": {}
       },
       "requires": [
-        {"shineChargeFrames": 60},
+        {"shineChargeFrames": 40},
         "canSuitlessMaridia",
-        {"shinespark": {"frames": 40, "excessFrames": 8}}
+        {"or": [
+          {"shinespark": {"frames": 40, "excessFrames": 8}},
+          {"and": [
+            "canMidairShinespark",
+            {"shinespark": {"frames": 37, "excessFrames": 8}}
+          ]}
+        ]}
       ],
-      "flashSuitChecked": true,
-      "note": "It takes a bit more time to set up the spark when suitless."
+      "flashSuitChecked": true
     },
     {
       "id": 12,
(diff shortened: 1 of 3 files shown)


region/maridia/inner-pink/The Precious Room.json
```diff
@@ -785,44 +757,102 @@
...
+        {"shineChargeFrames": 0}
+      ],
+      "endsWithShineCharge": true
+    },
+    {
+      "link": [2, 4],
+      "name": "Stutter Water Shinecharge",
+      "entranceCondition": {
+        "comeInStutterShinecharging": {
+          "minTiles": 4
```
kjbranch, line 803: This is more lenient than some other stutter water shinecharges, but it sounds like you'll look into the other rooms on another PR

region/maridia/inner-pink/The Precious Room.json
```diff
@@ -785,44 +757,102 @@
...
       ],
-      "exitCondition": {
-        "leaveWithSpark": {}
+      "endsWithShineCharge": true,
+      "note": "Use between 4 and 5 tiles of runway, to make the stutter as lenient as possible.",
+      "detailNote": [
+        "With this length of runway, the following timings for the stutter are possible:",
+        "1. Release forward for 3, 4, or 5 frames, repress forward on the last possible frame;",
+        "2. Release forward for 2 or 3 frames, repress forward on the 2nd-to-last possible frame;",
+        "3. Release forward for 1 pr 2 frames, repress forward on the 3rd-to-last possible frame."
```
kjbranch, line 816:
```suggestion
        "3. Release forward for 1 or 2 frames, repress forward on the 3rd-to-last possible frame."
```

region/maridia/inner-pink/The Precious Room.json
```diff
@@ -785,44 +757,102 @@
...
       "requires": [
-        "h_canCrystalFlash"
+        "canTrickyStutterWaterShineCharge",
+        {"shineChargeFrames": 0}
       ],
-      "flashSuitChecked": true
+      "endsWithShineCharge": true,
+      "note": [
+        "With only a runway of 2 tiles (open end) in the other room, this requires a double-frame-perfect setup:",
+        "release forward for exactly 3 frames, and repress forward on the last possible frame before the transition."
```
kjbranch, line 854: I couldnt get this one to work even with frame advance. Not sure how I'm counting it differently

osse101: image_2025-03-19_141308449

This is pretty much how I've been doing the trick and I feel better at it than double frame perfect would suggest.  The 3 tile version is less normalized and doesn't feel different enough for me.  But it could just be a me thing.

osse101 (commented):
Is there enough need for 3 techs?  It should currently be where crab tunnel, main street, precious, and crab shaft are the hard ones & almost every stutterwater shinecharge is set to 2~2.4 tiles runway.  So I think it would be a 3 tile version and a doorframe version for most cases?

blkerby:
> Is there enough need for 3 techs? It should currently be where crab tunnel, main street, precious, and crab shaft are the hard ones & almost every stutterwater shinecharge is set to 2~2.4 tiles runway. So I think it would be a 3 tile version and a doorframe version for most cases?

To me, it feels like a large difference in difficulty between the 2 vs. 2.4375 tile version. Just tested it out with Construction Zone (with 2.4375 tiles) to Precious Room, took me 3 tries to get it the first time; next successes took 5, 3, 9, 9, 3, 1, 3, 4, and then 3 tries. Tried it out from Early Supers (2 tiles) to Precious Room, took me 17 tries to get first success, then 32 tries, then 16 tries. I'm sure with a lot of practice it can be possible to get consistent at it, but it feels like an Extreme trick, especially considering the setup room could be heated (e.g. Crumble Shaft).

Halfie Climb, West Sand Hall Tunnel, and Oasis would be more examples of hard ones (the last two cases especially if the opposite door can't be unlocked).

On the other hand, after testing out the 4-tile version (from Baby Kraid Room into Precious), it doesn't seem as much easier as I was expecting based on the frame windows. My number of tries to get successes were 6, 1, 2, 6, 3, 2, 3, 4, 2, 2, and 8. It maybe shouldn't be treated differently from 2.4375 tiles after all, probably fits better in Expert than Very Hard.

Going into Botwoon ETank Room with at least 2 tiles of runway, I can get it essentially every time (1,2,1,1,1). I think that's probably more the difficulty that we want for Very Hard. And Main Street would of course also fit there in Very Hard, at least with 2 tiles of runway; with only 1 tile it's quite a bit harder, feels more like Expert.

I think using 3 tech may be justified here because of how large the differences in difficulty seem to be. The idea is for Very Hard to be able to give players a more gentle introduction to the tech by only including the easy cases, the bulk of the rest going into Expert, with just the really tough ones (needing double frame-perfect stutter) in Extreme.

osse101:
I agree with going over the stutterwatershinecharge strats and sorting them into very hard / expert / extreme.

If we could use `canInsaneJump` to indicate movement ability, it would be nice to not create a new tech.  But tricky SWSC handles the notable strats that were created better.  The tech situations is less important than categorizing the strats.

Precious room with 2 or 2.4 tiles felt no different to me.  The frame window differences are very minor.  But that is likely only a me thing.

blkerby:
> I agree with going over the stutterwatershinecharge strats and sorting them into very hard / expert / extreme.
> 
> If we could use `canInsaneJump` to indicate movement ability, it would be nice to not create a new tech. But tricky SWSC handles the notable strats that were created better. The tech situations is less important than categorizing the strats.
> 
> Precious room with 2 or 2.4 tiles felt no different to me. The frame window differences are very minor. But that is likely only a me thing.

Ok, removed the `canTrickyStutterWaterShineCharge` tech and put in a `canInsaneJump` for now. I don't really love `canInsaneJump` being used for things that aren't jumps, given that its name and description is only about jumps. But I know that this done in a lot of places currently, and same for `canTrickyJump`, so it's something that can be sorted out later.

Something I've thought about for a long time is allowing a tech to have different "tiers", which essentially could be a way to organize related tech, so that instead of "canGrappleJump", "canPreciseGrappleJump", "canTrickyGrappleJump", you could have just one tech but have different "levels" of it, and the player could select among those levels. Or maybe they could still exist as separate tech in the logic data, and they could just be presented as a group somehow in the randomizer. But I'm not sure this is compatible with the way the tech are currently grouped in the randomizer, by difficulty level, which also has its advantages. Just something to think about. I think there should be some way to improve it eventually, and then it might be less of a concern to have many variants of a tech for different difficulty levels.

PR #1982, opened 2025-03-18
Refine door lock ceiling clips, add G-mode variants

kjbranch:


osse101 (dismissed):
These strats make you do a hard trick (GMode) and another hard trick (doorLockSkip).  And doing two hard tricks in a sequence is something that I think a randomizer would want to manage.  Right now, that is just something to keep in mind for the future.

`canSkipDoorLock` also feels like there are easy and hard versions.  Knocking down crabs is very difficult, and use atomics is not.

osse101:
The cosine room remote acquires should be changed to a [1, 1] structure, but that is for another PR.

PR #1983, opened 2025-03-21
G-Mode: Screw Attack, Acid Statue

kjbranch:


region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -1288,6 +1467,93 @@
...
+              {"and": [
+                "HiJump",
+                "canSpringBallJumpMidAir"
+              ]}
+            ]}
+          ]}
+        ]},
+        {"heatFrames": 0}
+      ],
+      "clearsObstacles": ["A"],
```
osse101, line 1518: You could want to break both A and B.

osse101 (commented):
There's the theoretical path of breaking both sets of bomb blocks in screw room to collect the item and pick your door.  It would take a very specific room layout if you also use artificial morph and don't have varia.

The acid being there in indirect acid room was neat.

PR #1985, opened 2025-03-22
G-Mode: Golden Torizo

kjbranch:


region/lowernorfair/west/Golden Torizo's Room.json
```diff
@@ -479,6 +622,35 @@
...
+          "morphed": true
+        }
+      },
+      "requires": [
+        "h_canArtificialMorphLongIBJ",
+        {"or": [
+          "f_DefeatedGoldenTorizo",
+          {"and": [
+            "canTrickyDodgeEnemies",
+            "h_canArtificialMorphJumpIntoIBJ"
```
osse101, line 640: OR: can staggered IBJ.
Double IBJ is fast enough to outpace GT, or you can get him to jump away and double IBJ up.
Note: When double IBJing you have to switch to single IBJS to overload PLMS before touching the crumbles, or do it twice.

region/lowernorfair/west/Golden Torizo's Room.json
```diff
@@ -550,6 +939,69 @@
...
+            ]}
+          ]}
+        ]},
+        {"heatFrames": 0}
+      ],
+      "setsFlags": ["f_DefeatedGoldenTorizo"],
+      "flashSuitChecked": true,
+      "note": [
+        "Start the fight with 31 Missiles or fewer in order to prevent GT from using its laser attack, which will enable Samus to stand next to it.",
+        "Having GT start with his Chozo Ball attack also will make it so there are drops which can be collected at the end of the fight, otherwise they will have been overloaded.",
```
osse101, line 999: You probably don't get drops with this strat.

region/lowernorfair/west/Golden Torizo's Room.json
```diff
@@ -233,6 +257,110 @@
...
+        {
+          "types": ["powerbomb"],
+          "requires": [{"heatFrames": 110}]
+        }
+      ],
+      "collectsItems": [3],
+      "flashSuitChecked": true,
+      "note": [
+        "Avoid stepping on all three crumble blocks before moving through the item to overload PLMs, then go through the crumble blocks.",
+        "Alternatively, it is possible to touch the item then overload PLMs by using bomb spreads to overload PLMs before going through.",
```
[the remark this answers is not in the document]
kjbranch: I was mostly worried about it choosing this strat when the item isnt there, because of the OR. I'll reword it in the feedback commit

region/lowernorfair/west/Golden Torizo's Room.json
kjbranch: When i was doing it, I did get drops in their spawn location, but only if that was his first attack

kjbranch:
I wasn't seeing how staggered was helpful, as the main problem was not being high enough fast enough. Double bomb jump works if you do it quickly - you only need to do it for a little bit. But when going up the left to overload PLMs then up the right, you have less time, so it felt like canTrickyGMode to me (so i pulled that in)

PR #1987, opened 2025-03-25
Refine Oasis shinecharges

blkerby:
Similar to Precious Room, this adds new nodes to represent having just gained a shinecharge while running into the room. It helps because there are many ways to gain the shinecharge and also many ways to use it. With this way of doing it, many strats are covered that were missed before. I also tightened shinecharge frames and shinespark frames.

Probably the most interesting addition is the way to do the stutter shinecharge using only 1 tile of runway. It requires being able to open the opposite door, and it needs a not-so-intuitive subpixel normalization in the other room, along with a double frame-perfect stutter. This one I think fits in Insane, with `canBeVeryPatient` (could be improved with a more specific tech later). I created videos for these, showing all 3 ways that I know of to do the subpixel normalization.

I was thinking about making the Insane one a notable, but it's a little awkward since the same technique is going to be applicable in other rooms (e.g. West Sand Hall Tunnel), making it seem better to handle it with tech.

### DIFF
--- a/region/maridia/inner-green/Oasis.json
+++ b/region/maridia/inner-green/Oasis.json
@@ -1577,8 +1577,8 @@
         {"shineChargeFrames": 0}
       ],
       "unlocksDoors": [
-        {"nodeId": 2, "types": ["super", "missiles"], "requires": []},
-        {"nodeId": 2, "types": ["powerbomb"], "requires": ["never"]}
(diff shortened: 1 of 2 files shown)


osse101 (commented):
It would be possible to support `unlocksDoors: missiles` better now that shinesparkFrames can be added as a requirement there; I think the order of operations works.  For LeaveWithSpark, the health cost would be minorly off.

PR #1989, opened 2025-03-27
Colosseum side platform applications

blkerby:
- Adds side platform strats for going from the bottom right of Colosseum to the top right.
- Fixes a small bug in a Crab Hole side platform strat.
- Adds more notes to some setups. There are still more details that could added, especially for describing setups that only use part of the available runway (since this can sometimes be easier than using a full runway).
- Adds "environment" property to platforms in the entrance condition `comeInWithSidePlatform`, to be able to specify if it's required for the neighboring room to be water (with Gravity). Where applicable, doing a Gravity jump before the transition makes both momentum-conserving turnarounds and ceiling mockballs easier due to the slower underwater animations, in addition to its main effect of preserving greater vertical speed. In some cases it can make a strat possible that wouldn't be possible at all from an air room with the same geometry. I didn't add a corresponding property to the exit condition `leaveWithSidePlatform` because I think it can be reasonably inferred from the `doorEnvironments` on the door node.

### DIFF
--- a/region/maridia/inner-pink/Colosseum.json
+++ b/region/maridia/inner-pink/Colosseum.json
@@ -1195,6 +1195,283 @@
       "note": "Uses a runway of two tiles in the adjacent room."
     },
     {
+      "link": [2, 3],
+      "name": "Side Platform Cross Room Jump",
+      "entranceCondition": {
+        "comeInWithSidePlatform": {
+          "platforms": [
+            {
+              "minHeight": 2,
+              "maxHeight": 2,
+              "minTiles": 27.4375,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [
+                "canMomentumConservingTurnaround"
+              ],
+              "note": ["This applies to Dust Torizo Room and Noob Bridge."],
+              "devNote": [
+                "The jump could also be done with a shorter runway, with a tricky dash jump using run speed $4.0 or $4.1.",
+                "But it wouldn't apply to any more rooms in the game."
+              ]
+            },
+            {
+              "minHeight": 3,
+              "maxHeight": 3,
+              "minTiles": 28.2472,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [
+                "canMomentumConservingTurnaround"
+              ],
+              "note": ["This applies to Double Chamber."],
+              "devNote": [
+                "The jump could also be done with a shorter runway, with a tricky dash jump using run speed $4.0 or $4.1.",
+                "But it wouldn't apply to any more rooms in the game."
+              ]
+            },
+            {
+              "minHeight": 1,
+              "maxHeight": 1,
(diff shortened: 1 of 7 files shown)


region/maridia/inner-pink/Colosseum.json
```diff
@@ -1194,6 +1194,283 @@
...
+              "obstructions": [[3, 0]],
+              "environment": "water",
+              "requires": [
+                "canTrickyDashJump",
+                "canInsaneJump",
+                "canGravityJump",
+                "canMomentumConservingTurnaround"
+              ],
+              "note": [
+                "This applies to Draygon's Room.",
```
osse101, line 1292: This took a lot of playing around with to see where the extra height comes from.  moving forward and back, pausing later jumping earlier timing the turnaround.  So its trickier to understand than others, but not particularly hard.

region/maridia/inner-pink/Colosseum.json
```diff
@@ -1194,6 +1194,283 @@
...
+            {
+              "minHeight": 3,
+              "maxHeight": 3,
+              "minTiles": 39.4375,
+              "speedBooster": true,
+              "obstructions": [[3, 2]],
+              "requires": [
+                "canMomentumConservingTurnaround"
+              ],
+              "note": ["This applies to Metal Pirates Room."]
```
osse101, line 1339: A lot nicer if you turn around before the door.

region/maridia/inner-pink/Colosseum.json
```diff
@@ -1194,6 +1194,283 @@
...
+            },            
+            {
+              "minHeight": 3,
+              "maxHeight": 3,
+              "minTiles": 39.4375,
+              "speedBooster": true,
+              "obstructions": [[5, 2]],
+              "requires": [
+                "canInsaneJump",
```
osse101, line 1348: They're both expert to me.  The morph is kinda harder but is also more natural to do once I understood it.

region/maridia/inner-pink/Colosseum.json
```diff
@@ -1194,6 +1194,283 @@
...
+              "maxHeight": 2,
+              "minTiles": 31,
+              "speedBooster": true,
+              "obstructions": [[3, 0]],
+              "requires": [
+                "canInsaneJump",
+                {"or": [
+                  "canMomentumConservingTurnaround",
+                  {"and": [
+                    "canMomentumConservingMorph",
```
osse101, line 1432: I found the morph to be easier.

But a LeaveWith springball bounce should be the way you traverse these nodes.  Even better with a unmorph momentumConservingMorph.

osse101 (commented):
The springball+speed strats tend to be easier if you springball bounce on the bottom of the door frame instead of wiggling on the top of the door.

region/maridia/inner-pink/Colosseum.json
osse101: I think it still fits.

PR #1997, opened 2025-04-01
G-Mode, Second Pass: Crateria East and West

kjbranch:


region/crateria/east/West Ocean.json
```diff
@@ -2979,13 +2938,13 @@
...
-      "name": "G-mode Overload PLMs by PBing Missile Item",
+      "link": [17, 1],
+      "name": "G-mode Overload PLMs by Power Bombing Morph Maze Item",
       "requires": [
-        {"notable": "G-mode Overload PLMs by PBing Missile Item"},
+        {"notable": "G-mode Overload PLMs by Power Bombing Morph Maze Item"},
         "canEnterGMode",
-        {"itemNotCollectedAtNode": 10},
         "h_canUsePowerBombs",
+        "canInsaneJump",
```
osse101, line 2947: This wasn't a hard jump, from when I had to do this last.

region/crateria/east/West Ocean.json
```diff
@@ -3024,12 +2983,13 @@
             ]}
           ]}
         ]},
-        "h_canArtificialMorphPowerBomb"
+        "h_canArtificialMorphPowerBomb",
+        "canInsaneJump"
```
osse101, line 2987: same here

region/crateria/west/Gauntlet Energy Tank Room.json
```diff
@@ -946,6 +978,30 @@
...
+        "canInsaneJump",
+        "canOffScreenMovement"
+      ],
+      "collectsItems": [3],
+      "flashSuitChecked": true,
+      "note": [
+        "Roll through the item to overload PLMs then quickly fall into the morph tunnel, unmorph at the correct spot and jump out before the acid touches Samus.",
+        "Exit G-mode and roll back into the tunnel to fix the camera and return avoiding the acid again."
+      ],
+      "devNote": "It's not possible to do this with acid damage, as Samus needs to be in direct G-mode, and a CF would solidify the bomb blocks."
```
osse101, line 1003: Could you overload PLMs and then CF?

region/crateria/west/Gauntlet Energy Tank Room.json
```diff
@@ -946,6 +978,30 @@
...
+        {"itemNotCollectedAtNode": 3},
+        "canRiskPermanentLossOfAccess",
+        "Morph",
+        "canInsaneJump",
+        "canOffScreenMovement"
+      ],
+      "collectsItems": [3],
+      "flashSuitChecked": true,
+      "note": [
+        "Roll through the item to overload PLMs then quickly fall into the morph tunnel, unmorph at the correct spot and jump out before the acid touches Samus.",
```
osse101, line 1000: being "quick" doesn't seem important

region/crateria/east/West Ocean.json
kjbranch: It was mostly used for the obscurity and precision of the PB, and if someone wanted to play with just an easier set of g-mode strats, or if the easier strats were moved down into Expert. But I think you're right, that this isnt the way to do it. I'll remove it and check if there were any others that were added like this in Crateria and remove them and think more on how it should be addressed.

region/crateria/west/Gauntlet Energy Tank Room.json
kjbranch: looks like it is!

Issue #1001, opened 2023-08-01
New Field for Specific Types of Strats (wall jump avoid, one way)


kjbranch:
Some strats will never be useful unless wall jumps are disabled; for example a g-mode artificial morph into a morphless IBJ to avoid a wall jump. 

There are two interesting problems that these strats create: 1) They increase the number of irrelevant strats on a link that may make reading the logic more confusing. 2) They are never notable, even if they should be, because they are for a niche scenario.

It may be nice to add another field on a strat, like `WallJumpAvoid` = `true`, with a default of `false` such that these strats may be filtered out of logic and notable strat viewers unless the player is wishing to play with wall jumps disabled. 

Similarly, some strats are only useful for getting to an item, where the only way out is that this item enables the player to escape; for example a g-mode artificial morph through a morph tunnel to get an item, where the only way out is if that item is Morph. These strats may only be desirable to view if a randomizer allows these types of strats in its item placement logic. These could have a field, like `OneWay` = `true`.

osse101:
Sounds like `Tags`
```JSON
"tags": [ "WallJumpAvoid", "OneWay", "Patient", "SoftlockRisk" ]
```

kjbranch:
I really like this. It's clean and scales well if we need a few more.

kjbranch:
It could be nice to add a tag for artificial morph strats that are only valid without real morph, so that the tech canArtificialMorph can be applied to them. Alternatively, the tech can just be added to them, but there are a handful of strats that require artificial morph and real morph isnt a substitute.

PR #1998, opened 2025-04-02
Add wallJumpAvoid to strat and notable schema

blkerby:
This addresses a long-standing issue (#1001) about marking strats that are only useful in walljumpless contexts. This PR just adds a new boolean property `wallJumpAvoid` to strats and notables, along with an example. Later PRs would go through and populate this throughout the logic.

Based on the discussion in the issue, one alternative that was considered was implementing a more general `tag` property. However, after thinking about this some more, it seems that a simple tag would not work well for the other examples mentioned in the issue. For example, to model a strat that softlocks you at an item location, we would want to model more information such as the specific item and/or tech that could get you out of the softlock. The existing "failures" schema is also relevant to modeling softlock risk, and we may want to refine/expand it someday, but a simple tag is again not going to be enough. And "patience" is already modeled through tech, so it seems unnecessary to model as a tag. A boolean property has the advantage of being a more structured format (compared to a bag of strings), making it more straightforward to consume.

The plan is to use these properties in Map Rando to help reduce the clutter on the Generate page as well as the logic pages. The notable property `wallJumpAvoid` would apply to filtering notables on the Generate page; initially we may simply filter out the walljumpless notables from showing there, though eventually we could add some control to toggle them on or off. The strat property `wallJumpAvoid` can be applied to filter strats on logic pages for rooms and tech, and we already have a place to add a toggle for that.

### DIFF
--- a/region/tourian/main/Tourian Escape Room 4.json
+++ b/region/tourian/main/Tourian Escape Room 4.json
@@ -658,6 +658,7 @@
           "requires": [{"acidFrames": 110}]
         }
       ],
+      "wallJumpAvoid": true,
       "note": [
         "Climbing the shaft with Space Jump is slower than other methods, so it is necessary to move quickly in order to minimize acid damage.",
         "After reaching the top, fall down the right side to land on the platform below the door, taking a dip in acid before jumping into the door."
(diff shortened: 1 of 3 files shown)


osse101 (approved):
- Hopefully it is clear that a walljump being too difficult, whatever the tech level, still allows for easier alternative strats 😺 

- I like what this change lets us do.  And if there is a grouping structure in the future we want to use, I don't think that would be difficult to switch to.

PR #1999, opened 2025-04-03
Aqueduct shinecharges

blkerby:
This refines the water shinecharge strats entering from the bottom-left door of Aqueduct. This time I took a closer look at how the runway length affects the amount of frames remaining with "Leave Shinecharged" strats. It turned out to make a fairly big difference. Making 5 strats for the various runway lengths might seem like a bit of overkill, but there are applications for each case; maybe someday we'll be able to reference the runway length in the `requires` so that they wouldn't have to all be separate strats. Once this PR is merged, I plan to go back and apply the same refinement to `leaveShinecharged` variants in the other rooms (e.g. Precious Room, West Sand Hall Tunnel, Oasis).

Example applications:
1 tile, Screw Attack Room: https://videos.maprando.com/video/5923
2 tiles, Waterway: https://videos.maprando.com/video/5922
3 tiles, Colosseum: https://videos.maprando.com/video/5925
4 tiles, Below Spazer: https://videos.maprando.com/video/5924
5 tiles, Moat: https://videos.maprando.com/video/5927

### DIFF
--- a/region/maridia/inner-pink/Aqueduct.json
+++ b/region/maridia/inner-pink/Aqueduct.json
@@ -1028,6 +1028,7 @@
       "requires": [
         "canPreciseStutterWaterShineCharge",
         "canShinechargeMovementTricky",
+        "canInsaneJump",
         {"shineChargeFrames": 130}
       ],
       "exitCondition": {
@@ -1056,6 +1057,7 @@
       "requires": [
         "canPreciseStutterWaterShineCharge",
         "canShinechargeMovementTricky",
+        "canInsaneJump",
         {"shineChargeFrames": 140}
       ],
       "exitCondition": {
@@ -1085,7 +1087,12 @@
       "requires": [
         "canPreciseStutterWaterShineCharge",
         "canShinechargeMovementTricky",
-        {"shineChargeFrames": 150}
+        "canInsaneJump",
+        {"shineChargeFrames": 145},
+        {"or": [
+          "canBeVeryPatient",
+          {"shineChargeFrames": 5}
+        ]}
       ],
       "exitCondition": {
         "leaveShinecharged": {}
@@ -1099,7 +1106,11 @@
         "the ideal timing for the stutter is to release forward for 3 or 4 frames, and repress forward on the last possible frame before the transition.",
         "Other timings can work, but will gain the shinecharge further from the door, leaving fewer shinecharge frames remaining.",
         "After gaining the shinecharge, continue holding forward in order to make Samus stand.",
-        "Then turnaround spin jump back toward the door, and continue spin jumping to reach the transition with a shinecharge."
+        "Then turnaround spin jump back toward the door, and continue spin jumping to reach the transition with a shinecharge.",
+        "Alternatively, with greater difficulty but saving some shinecharge frames,",
+        "after performing the turnaround spin jump, briefly moonwalk back, then retain momentum by boomeranging forward",
+        "(switching from pressing backward to pressing forward on the next frame),",
+        "and arm pump to reach the door quickly."
       ],
       "devNote": [
         "The same shinecharge frames could be achieved with a closed end runway (effective length of 2.4375),",
@@ -1117,7 +1128,12 @@
       "requires": [
         "canPreciseStutterWaterShineCharge",
         "canShinechargeMovementTricky",
-        {"shineChargeFrames": 160}
+        "canInsaneJump",
+        {"shineChargeFrames": 150},
+        {"or": [
+          "canBeVeryPatient",
+          {"shineChargeFrames": 10}
+        ]}
       ],
       "exitCondition": {
         "leaveShinecharged": {}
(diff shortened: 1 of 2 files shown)


region/maridia/inner-pink/Aqueduct.json
```diff
@@ -927,6 +1012,204 @@
...
+      ],
+      "exitCondition": {
+        "leaveShinecharged": {}
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],
+      "note": [
+        "With only a runway of 4 tiles (open end) in the other room,",
```
osse101, line 1069: I tried 4 and 5 tiles for a while and couldn't get close to leaving with frames with either.
I was able to leave with ~15 frames using the moonwalk trick instead of jumping (moonwalk then switch to running forward on the next frame).

These strats don't scale down since you lose so many frames to a suboptimal input, so they need stronger requirements.

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -927,6 +1012,204 @@
...
+        "comeInStutterShinecharging": {
+          "minTiles": 1
+        }
+      },
+      "requires": [
+        "canPreciseStutterWaterShineCharge",
+        "canInsaneJump",
+        "canBeVeryPatient",
+        "canShinechargeMovementTricky",
+        {"shineChargeFrames": 175}
```
osse101, line 1150: You can get about 20 frames with the moonwalk trick, and at these requirements its probably the better approach

osse101: https://videos.maprando.com/video/6014

blkerby: Interesting, looking into the moonwalk trick, then will update the requirements & notes

blkerby: With a better stutter, I was able to get 30 shinecharge frames remaining: https://videos.maprando.com/video/6037. This is using a new, easier setup for the stutter which I found by accident (doing an arm pump by pressing up). It eliminates the need for the complicated subpixel normalization (to $3FFF), instead just using subpixels of $7FFF which can be obtained with the standard normalization, by backing into the corner below the door frame, then jump and mid-air turnaround (though that doesn't work here in Blue Boulder Room, because of the water). It gives you the shinecharge a little further from the door but could be worth it.

Aside from the large impact on the 1-tile runway setup, it also helps some with the 2-tile and 3-tile setups, so I updated those as well. The shinecharge frames are not really too optimized (e.g. put it in as 20 frames remaining for the 1-tile runway, though >30 is possible); I didn't want to push it too far since these strats are already very difficult. Could always be tightened later though.

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -927,6 +1012,235 @@
...
+        }
+      },
+      "requires": [
+        "canPreciseStutterWaterShineCharge",
+        "canShinechargeMovementTricky",
+        "canInsaneJump",
+        {"shineChargeFrames": 150},
+        {"or": [
+          "canBeVeryPatient",
+          {"shineChargeFrames": 10}
```
osse101, line 1135: At extreme it feels to me like 
- you can leave with 10 frames if you moonwalk boomerang
- otherwise you can't reach the door with frames (<5)

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -927,6 +1012,235 @@
...
+      "entranceCondition": {
+        "comeInStutterShinecharging": {
+          "minTiles": 5
+        }
+      },
+      "requires": [
+        "canPreciseStutterWaterShineCharge",
+        "canShinechargeMovementTricky",
+        "canInsaneJump",
+        {"shineChargeFrames": 130}
```
osse101, line 1032: 150 at extreme

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -927,6 +1012,235 @@
...
+      "entranceCondition": {
+        "comeInStutterShinecharging": {
+          "minTiles": 3
+        }
+      },
+      "requires": [
+        "canPreciseStutterWaterShineCharge",
+        "canShinechargeMovementTricky",
+        "canInsaneJump",
+        {"shineChargeFrames": 145},
```
osse101, line 1091: 160 at extreme

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -927,6 +1012,235 @@
...
+      "entranceCondition": {
+        "comeInStutterShinecharging": {
+          "minTiles": 4
+        }
+      },
+      "requires": [
+        "canPreciseStutterWaterShineCharge",
+        "canShinechargeMovementTricky",
+        "canInsaneJump",
+        {"shineChargeFrames": 140}
```
osse101, line 1061: 155 at extreme

osse101 (commented):
- Moonwalk boomerang does need to be a new tech, at some point.  At extreme I'd imagine.
- Getting the perfect room entry took many many tries, focusing on the 1 tile runway.  But then leaving the room still takes multiple precise inputs (stopping around the right place, the crouch-stand-turnaround spin, the second jump or moonwalk trick)
- I timed the strats without grinding much, and they seem like they are like 20frames slower if you aren't moving perfectly.

region/maridia/inner-pink/Aqueduct.json
blkerby: I'm interpreting this as increasing `{"shineChargeFrames": 10}` to `{"shineChargeFrames": 15}` in the `or`? Then without canBeVeryPatient, the logic will say it uses 165 frames, meaning 170 with the Extreme lenience.

I find the moonwalk boomerang harder to execute compared to getting a perfect stutter, but it makes sense that this is going to vary by player; with more practice probably I could get more comfortable with it.

PR #2001, opened 2025-04-03
Add wallJumpAvoid to notable strats

kjbranch:
Next I am planning on opening another PR to clean up and delete some notables.

### DIFF
--- a/region/brinstar/green/Green Hill Zone.json
+++ b/region/brinstar/green/Green Hill Zone.json
@@ -1113,6 +1113,7 @@
         {"notable": "Grapple Teleport Fling to Right (from Moat)"},
         "canInsaneJump"
       ],
+      "wallJumpAvoid": true,
       "note": [
         "After teleporting, extend the Grapple, and swing back and forth to fix the camera and then to gain momentum.",
         "A precisely timed release of Grapple will allow Samus to fling onto the ledge on the right."
@@ -1132,6 +1133,7 @@
         "canTrickyGrappleJump",
         "canInsaneJump"
       ],
+      "wallJumpAvoid": true,
       "note": [
         "After teleporting, swing back and forth to fix the camera.",
         "Swing to the right by soft-bouncing against the door followed by fully extending the Grapple Beam.",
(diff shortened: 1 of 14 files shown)


region/brinstar/green/Green Hill Zone.json
```diff
@@ -1458,6 +1460,7 @@
     {
       "id": 5,
       "name": "Grapple Teleport Fling to Right (from Moat)",
+      "wallJumpAvoid": true,
```
osse101, line 1463: Wondering how this is used, being a property of the notable and of strats using the notable.
It provides a descriptor on both so we can better see what the strat and notable each expect.
And the notables and individual strats are listed in different places.  So maybe that's it.  As a requirement you generally only want it used in one place, but as a meta property it is consumed it different ways.

region/maridia/inner-pink/East Cactus Alley Room.json
```diff
@@ -1421,11 +1424,13 @@
     {
       "id": 1,
       "name": "G-Mode Morph, IBJ, HBJ",
+      "wallJumpAvoid": true,
```
osse101, line 1427: This is an artificial morph strat, to get through the tunnel.

region/brinstar/green/Green Hill Zone.json
blkerby: My plan is that the notable-level property will affect filtering on the Generate page, while the strat-level property will affect filtering of strats on the room logic pages. Eventually we may also list notables on room logic pages, in which case the notable-level property would affect filtering there too. Both properties would also affect the randomizer item placement logic: wallJumpAvoid strats and notables would be excluded from the graph traversal if collectible wall jump is not enabled.

It's possible that some notables might only sometimes be a wall-jump avoid, in which case the notable itself wouldn't need a wallJumpAvoid property, while some of the strats that use it might. If a notable does have the wallJumpAvoid property, then any strats that require it should also have the wallJumpAvoid property (unless the notable is only required within an `or` in the strat), which is something that could make sense to add as a test. Conversely, if all strats that use a notable have wallJumpAvoid, then the notable should too.

PR #2003, opened 2025-04-04
Removing and refining lower-difficulty notables

kjbranch:
Looking through all notables up to VH.
This is a draft of what I think could be changed with some of the notables, with an attempt to remove some. I'm assuming that some of these will need to be reverted, I was just looking at any that seemed like they could be adjusted. I will add some comments to some of them to give some of my thoughts, and we can start a discussion.


region/brinstar/red/Red Tower.json
```diff
@@ -2463,20 +2463,32 @@
...
         "then using well-timed short space jumps to go up as rapidly as possible before the block respawns."
       ]
     },
+    {
+      "link": [9, 5],
+      "name": "Space Jump Climb Between the Rippers",
+      "requires": [
+        "canPreciseSpaceJump",
+        "canTrickyDodgeEnemies"
+      ],
```
kjbranch, line 2486: just looked at the video, this looks much harder, so this would push it up to from VH to expert, and the expert variant would not be notable

region/crateria/central/Landing Site.json
```diff
@@ -1859,14 +1858,6 @@
         "The jump can be set up by using the full runway with a one-tap shortcharge, with the tap being at the top of the lowest slope."
       ]
     },
-    {
-      "id": 4,
-      "name": "Blue Space Jump (Bottom)",
-      "note": [
-        "Starting near the right runway, run through the bomb block passage, then jump right after exiting.",
-        "Using HiJump and space jump, Samus is able to elevate enough to break through the bomb blocks blocking the Gauntlet entrance."
-      ]
```
kjbranch, line 1868: This one is not really very useful, as you can shinespark from the plateau then refill at the ship. I think it would mostly be for players who can't really short charge, or preserving a flash suit - so it doesn't seem notable

region/crateria/east/The Moat.json
```diff
@@ -1137,11 +1135,6 @@
     }
   ],
   "notables": [
-    {
-      "id": 1,
-      "name": "SpringBall Bounce",
-      "note": "From an open doorway or adjacent room, run, jump, lateral midair morph on the way down, then bounce off the pedestal to get to the other side."
-    },
```
kjbranch, line 1144: This one is very interesting and has vanilla uses, but it's not really much more notable over any other spring ball bounce. Instead of being notable, we switched the tech video be this (instead of bubble mountain)

region/crateria/west/Gauntlet Entrance.json
```diff
@@ -914,6 +914,7 @@
         {"obstaclesCleared": ["A"]},
         "canMidairShinespark",
         "canCameraManip",
+        "canTrickyDodgeEnemies",
```
kjbranch, line 917: this would move this notable up to expert

region/lowernorfair/east/Lower Norfair Fireflea Room.json
```diff
@@ -898,11 +898,10 @@
     {
       "id": 37,
       "link": [4, 5],
-      "name": "Ice Bridge (Right to Left)",
+      "name": "Ice Bridge",
       "requires": [
-        {"notable": "Ice Bridge"},
         "canTrickyUseFrozenEnemies",
-        "canCarefulJump"
+        "canTrickyJump"
```
kjbranch, line 904: this isn't too useful, so this would stay in VH, but not be notable

region/lowernorfair/east/Red Kihunter Shaft.json
```diff
@@ -813,7 +813,7 @@
       "requires": [
         {"notable": "KiHunter Dodge"},
         "h_canNavigateHeatRooms",
-        "canTrickyJump",
+        "canTrickyDodgeEnemies",
```
kjbranch, line 816: This would push this up to Expert (the `Kihunter Manipulation from Middle Door` is also in expert)
looking at the video, it looked like expert

region/norfair/west/Crumble Shaft.json
```diff
@@ -281,6 +281,10 @@
       "note": [
         "Jump and aim down to lower the camera so the middle platform Sova starts moving.",
         "Then freeze it by aiming down, shoot the shot block, and use the Sova as a stable platform."
+      ],
+      "devNote": [
+        "Wall jumping and using the same frozen Sova, or falling onto the Sova just below the item and then jumping to the top Sova could be added,",
+        "but this would require an obstacle to ensure they haven't fallen or died, and is not really easier than a crumble jump."
       ]
```
kjbranch, line 288: this felt too complex for hard, and the crumble jump (VH) was easier than this, but I think it's just getting back without falling that would push it up. 
Most players would probably just do the crumble jump here, and you kind of have multiple attempts, so it feels fine to stay in hard.
I was also concerned of if the Sova's were moved or killed, but maybe its not too likely. Freezing the Sova below while falling with two quick shots was also a little precise/unintuitive.

kjbranch (commented):
Went through all these notables with Maddo and tested a couple with him, but he might want to review this as well.

region/norfair/west/Crumble Shaft.json
osse101: - I haven't seen this used much
- Falling is punishing enough  that there are situations where you would take the time to do the more reliable strat.  Many people die checking crumble shaft in vanilla map races.
- The ice platforms seem more important if you don't have walljump or the runway to jump directly to the item.  Walljumping off the crumbles is safer than crumble jumping.

blkerby:
Looks good. Getting rid of notables that are no longer useful (along with an ability to filter out wall-jump-avoid ones) will help reduce the clutter on the Generate page, making it easier for players to go through them. I think in the past we had some bias toward making things notable because it was the only way to put a video on a strat; but now videos can go on any strat. And we also have more tech now that can be used as an alternative to notables.

region/brinstar/red/Red Tower.json
kjbranch: Practiced this and it's only the last Ripper that's annoying. I found that space jumping against one wall was easier than weaving back and forth, like the video.

region/brinstar/pink/Pink Brinstar Hopper Room.json
```diff
@@ -1308,6 +1308,7 @@
       "requires": [
         {"notable": "Shinespark to Ride the Elevator"},
         "canHorizontalShinespark",
+        "canUseSpeedEchoes",
```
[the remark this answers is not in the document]
kjbranch: This one is just adding the tech. Though, I was considering combining the 3 notables that raise the elevator eventually - it's not too important for now, since it will be split off as a walljumpless notable

region/brinstar/red/Red Tower.json
```diff
@@ -2463,18 +2463,44 @@
...
+        "Alternatively, this can be done by shooting the block while falling right as it goes off screen,",
+        "then using well-timed short space jumps to go up as rapidly as possible before the block respawns."
+      ]
+    },
+    {
+      "link": [9, 5],
+      "name": "HiJump, Space Jump Climb Between the Rippers",
       "requires": [
-        {"notable": "Climb with Space Jump"},
         "SpaceJump",
```
osse101, line 2485:
```suggestion
        "SpaceJump",
        "HiJump",
```

PR #2005, opened 2025-04-04
Removing and refining "Expert" difficulty notables

kjbranch:
Similar to before, this is very much a draft and any of it can easily be reverted. I'll add some comments again as well.

### DIFF
--- a/region/brinstar/red/Red Brinstar Fireflea Room.json
+++ b/region/brinstar/red/Red Brinstar Fireflea Room.json
@@ -522,9 +522,8 @@
       "link": [2, 1],
       "name": "X-Ray Access Ice Bridge",
       "requires": [
-        {"notable": "X-Ray Access Ice Bridge"},
         "canTrickyUseFrozenEnemies",
-        "canCarefulJump",
+        "canTrickyDodgeEnemies",
         {"spikeHits": 3},
         {"or": [
           "canUseIFrames",
@@ -1030,7 +1029,6 @@
       "link": [4, 1],
       "name": "X-Ray Access Tricky SpringBall Jump",
       "requires": [
-        {"notable": "X-Ray Access Tricky SpringBall Jump"},
         "canTrickySpringBallJump",
         "canSpringFling"
       ],
@@ -1091,7 +1089,6 @@
       "link": [4, 1],
       "name": "X-Ray Access SpikeJumps with HiJump",
       "requires": [
-        {"notable": "X-Ray Access SpikeJumps with HiJump"},
         "HiJump",
         "canTrickyJump",
         {"or": [
@@ -1134,40 +1131,11 @@
       ]
     },
     {
-      "id": 3,
-      "name": "X-Ray Access Ice Bridge",
-      "note": [
-        "Cross the gaps by freezing multiple enemies or by refreezing a Waver while standing on it.",
-        "Be careful running on frozen enemies.",
-        "Freeze the final Waver high and walljump up, or wait for the second Waver to come back."
-      ]
-    },
-    {
       "id": 4,
       "name": "X-Ray Access Damage Boost",
       "note": [
         "A precise strat that involves crossing both gaps by dboosting off Wavers.",
         "The Wavers are also used for iframes to avoid taking damage from any spikes."
-      ]
-    },
-    {
-      "id": 5,
-      "name": "X-Ray Access Tricky SpringBall Jump",
-      "note": [
-        "Jump from the safe spot on the spikey stairs and use the momentum change from equipping SpringBall to move closer to the door's platform.",
-        "Then SpringBall jump to reach the door."
-      ]
-    },
-    {
-      "id": 6,
-      "name": "X-Ray Access SpikeJumps with HiJump",
-      "note": [
-        "Jump to the door leading to X-Ray either by jumping from spikes or with a very precise jump.",
-        "Cross the room by Jumping atop the spikes.",
-        "Taking enemy damage can be used to reduce some spike damage.",
-        "The final jump can either be made by jumping from the bottommost spike stair or from the safe stair.",
-        "Setup run speed atop the spike stair by holding forward to be pushed backward.",
-        "The safe stair jump is much harder, but can be made without every pixel of runway."
       ]
     }
   ],
--- a/region/crateria/central/Landing Site.json
+++ b/region/crateria/central/Landing Site.json
@@ -826,25 +826,6 @@
       ]
     },
     {
-      "id": 28,
-      "link": [3, 1],
-      "name": "Blue Space Jump (Top)",
-      "requires": [
-        {"notable": "Blue Space Jump (Top)"},
-        "canBlueSpaceJump",
-        "canTrickyDashJump",
-        {"getBlueSpeed": {"usedTiles": 30, "openEnd": 2}}
-      ],
-      "clearsObstacles": ["A"],
-      "note": [
-        "This strat is fairly precise. All jumps should be as small as possible.",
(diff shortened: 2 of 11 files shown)


region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -522,9 +522,8 @@
       "link": [2, 1],
       "name": "X-Ray Access Ice Bridge",
       "requires": [
-        {"notable": "X-Ray Access Ice Bridge"},
         "canTrickyUseFrozenEnemies",
-        "canCarefulJump",
+        "canTrickyDodgeEnemies",
```
kjbranch, line 526: This would keep this in Expert, but it would not be notable. It does require ice and 3 spikes, so you'll still typically go for the d-boost strat instead. This one is awkward though, because this can still be expected if the d-boost variant is turned off. So it could be kept notable if that's going to be a problem.

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -1030,7 +1029,6 @@
       "link": [4, 1],
       "name": "X-Ray Access Tricky SpringBall Jump",
       "requires": [
-        {"notable": "X-Ray Access Tricky SpringBall Jump"},
         "canTrickySpringBallJump",
         "canSpringFling"
```
kjbranch, line 1033: the spring fling keeps it in expert, but is no longer notable. The spring fling is a bit of a stretch though. I wonder if it should also have some sort of quick morph?

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -1091,7 +1089,6 @@
       "link": [4, 1],
       "name": "X-Ray Access SpikeJumps with HiJump",
       "requires": [
-        {"notable": "X-Ray Access SpikeJumps with HiJump"},
         "HiJump",
         "canTrickyJump",
```
kjbranch, line 1093: removes this from being notable, which moves the spike variant from expert to VH

region/crateria/central/Landing Site.json
```diff
@@ -825,25 +825,6 @@
...
-        {"getBlueSpeed": {"usedTiles": 30, "openEnd": 2}}
-      ],
-      "clearsObstacles": ["A"],
-      "note": [
-        "This strat is fairly precise. All jumps should be as small as possible.",
-        "Using at least one-tap short charge, jump off the ledge and descend as much as possible while keeping space jump active.",
-        "Once approaching the wall, be sure to jump again to continue through all of the bomb blocks.",
-        "If HiJump is available, it is much easier to use the blue space jump strat from the ground."
-      ],
-      "devNote": "Although there is a larger runway available, minimizing the speed is important."
```
kjbranch, line 845: this strat was kinda useless and especially shouldn't be notable. I combined it with the other strat as a note for an alternate variant instead

region/crateria/central/Landing Site.json
```diff
@@ -1840,16 +1829,6 @@
         "Aiming down or morphing before hitting the wall can help make the jump slightly less precise."
       ]
```
kjbranch, line 1830: I was actually wondering if there was a good way to remove this one as well, but didn't look like it.

region/maridia/inner-pink/East Cactus Alley Room.json
```diff
@@ -811,16 +811,22 @@
         "comeInWithBombBoost": {}
       },
       "requires": [
-        {"notable": "Cross Room Jump with Spring Ball, Bomb Boost"},
         "canSpringBallBombJump",
         "canCrossRoomJumpIntoWater",
+        "canInsaneJump",
```
kjbranch, line 816: pushes it up to extreme and removes it from being notable

region/maridia/inner-pink/West Sand Hole.json
```diff
@@ -704,12 +704,13 @@
       "link": [7, 5],
       "name": "Suitless Bootless IBJ",
       "requires": [
-        {"notable": "Suitless Bootless IBJ"},
         "canJumpIntoIBJ",
         "canBombJumpWaterEscape",
-        "canCrumbleJump"
+        "canCrumbleJump",
+        "canInsaneJump"
```
kjbranch, line 710: also pushes it up to extreme and removes it from being notable

region/norfair/crocomire/Post Crocomire Farming Room.json
```diff
@@ -545,7 +545,6 @@
       "link": [3, 1],
       "name": "Speedy Jump",
       "requires": [
-        {"notable": "Speedy Jump"},
```
kjbranch, line 548: moved this to be the tech video (instead of warehouse entrance, since that's just a walljumpless strat)

region/norfair/east/Cathedral Entrance.json
```diff
@@ -855,9 +855,8 @@
     {
       "id": 33,
       "link": [4, 1],
-      "name": "Speedjump (Right to Left)",
+      "name": "Speedjump, Through the Shot Block",
       "requires": [
-        {"notable": "Speedjump (Right to Left)"},
         "canTrickyDashJump",
         {"heatFrames": 150}
```
kjbranch, line 861: this is not just a walljumpless strat, as it should also save heat frames

region/brinstar/red/Red Brinstar Fireflea Room.json
osse101: You don't need much of a quick morph with a second fling.  It does make timing the jump harder without a quick morph.

region/brinstar/red/Red Brinstar Fireflea Room.json
osse101: The problem with this in VH is that no one other than me does this strat, so they would expect logic to want them to damage boost over.  
The practical difference between the two is that you can retry the hijump jump once you make it this deep into the room, but the damage boost strat you cannot.  And dboosting to this point and not to the door doesn't look modelled, so you have extra health if you do that.

region/crateria/central/Landing Site.json
```diff
@@ -1167,7 +1148,15 @@
         }}
       ],
       "clearsObstacles": ["A"],
-      "note": "Use the lines in the background to determine the height of the breakable blocks."
+      "note": "Use the lines in the background to determine the height of the breakable blocks.",
+      "detailNote": [
+        "It is also possible to do this from the top right door's runway.",
+        "Although there is a larger runway available above, minimizing the speed is important,",
+        "so the runway is only slightly longer than the plateau, but the Space Jump is significantly more precie.",
```
osse101, line 1155:
```suggestion
        "so the runway is only slightly longer than the plateau, but the Space Jump is significantly more precise.",
```

osse101 (approved):

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -1029,13 +1029,22 @@
...
+          "canLateralMidAirMorph",
+          "can4HighMidAirMorph"
+        ]}
       ],
       "note": [
-        "Jump from the safe spot on the spikey stairs and use the momentum change from equipping SpringBall to move closer to the door's platform.",
-        "Then SpringBall jump to reach the door."
-      ]
+        "Spring Ball jump from the safe spot on the spikey stairs. Moving over far enough left before the peak of the jump is the trickiest part of the jump.",
+        "There are multiple things which can help get further left:",
```
osse101, line 1042: It will help to have clips of these approaches.

osse101 (dismissed):

PR #2008, opened 2025-04-05
Maridia Elevator Room L->R side platform jumps

blkerby:
This strat is only useful in a situation where wall-jump is unavailable and you want to preserve a flash suit (and so can't crouch-jump down-grab).

Removing the "canInsaneMidAirMorph" requirement from the Baby Kraid Room side platform setup, since depending on the runway length used, it is not necessarily that difficult. The "canInsaneMidAirMorph" can still be applied to individual applications as needed.

### DIFF
--- a/region/maridia/inner-yellow/Maridia Elevator Room.json
+++ b/region/maridia/inner-yellow/Maridia Elevator Room.json
@@ -278,6 +278,121 @@
       "devNote": "This is mainly useful as an alternative to crouch-jump down-grab in order to preserve a flash suit."
     },
     {
+      "link": [1, 2],
+      "name": "Side Platform Cross Room Jump",
+      "entranceCondition": {
+        "comeInWithSidePlatform": {
+          "platforms": [
+            {
+              "minHeight": 1,
+              "maxHeight": 1,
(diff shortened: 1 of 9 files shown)


region/maridia/inner-yellow/Maridia Elevator Room.json
```diff
@@ -277,6 +277,121 @@
...
+              "obstructions": [[3, 0]],
+              "requires": [
+                "canMomentumConservingMorph"
+              ],
+              "note": ["This applies to Noob Bridge, Metroid Room 1, and Statues Hallway."]
+            },
+            {
+              "minHeight": 3,
+              "maxHeight": 3,
+              "minTiles": 39.4375,
```
osse101, line 340: I count 38 tiles 🤔 
And it should be displayed more clearly to not use the full runway.

blkerby: Yeah, I had the length wrong, went through and fixed it everywhere now.

Also updated the "minTiles" in this strat to reflect the recommended length, and added more detail to the note.

PR #2013, opened 2025-04-08
Mother Brain left-to-right clips

blkerby:
In the first few months of Map Rando, we heard that cemo knew a way to clip past Mother Brain from the left, using a Crystal Flash, moondance, and ice clipping. Finally got around to working out the details. This can be interesting because with G-mode artificial morph it's possible without Morph. Compared to the Halfie Climb grapple teleport (the only other known way to do Mother Brain without Morph), this technique is much more broadly applicable.

Videos:
- https://videos.maprando.com/video/6000
- https://videos.maprando.com/video/6001
- https://videos.maprando.com/video/6002

### DIFF
--- a/region/tourian/main/Mother Brain Room.json
+++ b/region/tourian/main/Mother Brain Room.json
@@ -1081,9 +1081,8 @@
         "canTrickyDodgeEnemies",
         "canMoondance",
         "canWallIceClip",
-        {"ammo": {"type": "Missile", "count": 18}}
+        {"spikeHits": 1}
       ],
-      "setsFlags": ["f_MotherBrainGlassBroken"],
       "note": [
         "Perform a moondance while killing the Rinkas, carefully avoiding being hit by them.",
         "After 176 moonfalls, Samus will clip down into the tank.",
--- a/region/tourian/main/Mother Brain Room.json
+++ b/region/tourian/main/Mother Brain Room.json
@@ -92,6 +92,15 @@
       "mapTileMask": [
         [1, 2, 1, 1]
       ]
+    },
+    {
+      "id": 8,
+      "name": "Standing on Top Left of Mother Brain Tank",
+      "nodeType": "junction",
+      "nodeSubType": "junction",
+      "mapTileMask": [
+        [2, 1, 1, 1]
+      ]
     }
   ],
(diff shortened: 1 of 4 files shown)


region/tourian/main/Mother Brain Room.json
```diff
@@ -304,6 +342,89 @@
...
+          "morphed": true
+        }
+      },
+      "requires": [
+        {"notable": "Moondance Clip"},
+        {"not": "f_MotherBrainGlassBroken"},
+        "h_canArtificialMorphMovement",
+        "h_canArtificialMorphCrystalFlash",
+        {"resourceMaxCapacity": [{"type": "RegularEnergy", "count": 299}]},
+        {"autoReserveTrigger": {"minReserveEnergy": 100}},
```
osse101, line 409: - The main use is to gain a flash suit and I don't believe that is tied to Moondance Clip when exiting out the left.  Though, that is a trick that can be added to half the rooms with enemies.
- I guess the Mother Brain encounter logic ensures you have the health to do the fight after autoReserveTrigger

osse101 (approved):

region/tourian/main/Mother Brain Room.json
blkerby: Yeah, I don't know that the flash suit would be useful here, but it is a side effect so I included it. The main use of the strat would be to break the glass from the left with artificial morph without needing suits. With the "disableable etanks" QoL, it would be possible to turn off all the ETanks while the CF fills reserves, then pause and turn them back on right before the reserve triggers. So in that case you could end with 400 energy minus the Rinka hit and spike hit (if wall ice clipping through to the right), so enough to do the fight with Varia. But with breaking the glass on the left, leaving and coming back after refilling is also an option.

If the goal were to leave with a flash suit, I imagine it could be done without artificial morph movement, by doing the CF on the door ledge. That could be a separate strat sometime. Like you said, there will be a lot of strats like that.

One interesting thing that isn't addressed in this PR but is in the back of my mind is that after getting to the right of the tank (node 3), initiating the fight is not the only option, as you could also traverse the room in reverse and leave out the right door. Currently the logic for that would require the Zebetites to already be destroyed, but that can be refined later.

PR #2017, opened 2025-04-15
G-Mode: Metal Pirates

kjbranch:


osse101 (commented):
Why is [4, 2] 190 heat frames while [4, 1] is 45?

kjbranch:
> Why is [4, 2] 190 heat frames while [4, 1] is 45?

In vanilla, you have to shoot the left door before leaving through either door. so 4->2 stops at 1 on the way. It could use a helper  so that it's 45 for each in Map Rando, but I just wasn't sure if it mattered enough. Do you think it would be worth adding?

osse101:
Ok, I see it is only on the strats that clear the pirates.  

I don't think it works as intended then.  You must visit the door node to unlock it and set the lock as cleared, so it would still use [4, 1].  Writing something that visits 1 then runs to 2 without taking (45 + 200) probably isn't worth doing.

Couldn't you shoot the door then run to the other door while still in GMode?  Maybe you miss drops doing that.

kjbranch:
> Couldn't you shoot the door then run to the other door while still in GMode? Maybe you miss drops doing that.

That's what I thought too. Apparently you need to exit G-mode before shooting the door or it doesnt work. Something to do with it still being grey when you shoot it, I would think

PR #2019, opened 2025-04-16
Blue Brinstar Boulder shinespark

blkerby:


kjbranch (commented):
i think this should require can shinecharge movement. With more shinecharge frames remaining, it becomes easy to accidentally hit the boulder

osse101:
I think not shinesparking into the boulder should be canTrickyJump, as that is what we use to indicate you know where the end of the bridge is.

PR #2020, opened 2025-04-16
G-Mode: Red Kihunter Shaft

kjbranch:


region/lowernorfair/east/Red Kihunter Shaft.json
```diff
@@ -548,6 +550,65 @@
...
+        {
+          "types": ["powerbomb"],
+          "requires": [{"heatFrames": 110}]
+        }
+      ],
+      "flashSuitChecked": true,
+      "note": [
+        "Overload PLMs by rolling through the camera scroll block just right of the bomb blocks,",
+        "then roll through the morph tunnel and use a bomb boost or Spring Ball to escape the tunnel and get to the door.",
+        "Without a way to boost out of the tunnel, exit G-mode and use a partial floor clip then ceiling clip to jump out of the tunnel.",
```
osse101, line 603: Its better described as a 2 turn xray climb.  The partial floor clip would need the blocks gone, but using xray to stand up makes them solid again.

osse101 (approved):

PR #2021, opened 2025-04-16
G-Mode: Three Musketeers

kjbranch:


region/lowernorfair/east/Three Musketeers' Room.json
```diff
@@ -472,6 +494,57 @@
...
+        ]},
+        {"heatFrames": 0}
+      ],
+      "flashSuitChecked": true,
+      "note": [
+        "Avoiding the acid completely requires precise movement. Either jump into IBJ or use a precise Spring Fling.",
+        "Getting a well-timed pause to trigger immediately after Samus jumps can give enough horizontal momentum to cross the second Hibashi and land in the alcove,",
+        "however, a quick pause with a second spring fling on the descent can help."
+      ],
+      "devNote": "This may be possible with a very precise double bomb IBJ."
```
osse101, line 546: https://videos.maprando.com/video/6063

To reach 5 while morphed you would need gravity and a few acid frames.

osse101 (dismissed):

region/lowernorfair/east/Three Musketeers' Room.json
kjbranch: That's great, thanks! 
(Fortunately, getting to 5 while morphed is surprisingly not helpful, since you can't overload PLMs and go through the bomb block, and Kihunters are immune to PBs)

PR #2028, opened 2025-04-22
Fast Pillars Setup cross-room tricky dash jump

blkerby:
I created a bunch of videos a while ago for doing side platform jumps at this location. But then today when I came back to it, I realized that with 1 tile of air runway, you can do this tricky dash jump. It's dangerous (risk of hitting of the Space Pirate) but probably not any more than the side platform jumps, so with this I think there's not much reason to add strats for the side platform jumps.

Video: https://videos.maprando.com/video/6177

### DIFF
--- a/region/lowernorfair/east/Fast Pillars Setup Room.json
+++ b/region/lowernorfair/east/Fast Pillars Setup Room.json
@@ -574,6 +574,22 @@
       ]
     },
     {
+      "link": [1, 4],
+      "name": "Come in Running, Tricky Dash Jump",
+      "entranceCondition": {
+        "comeInRunning": {
+          "minTiles": 1,
+          "speedBooster": true
+        }
+      },
+      "requires": [
+        "canTrickyDashJump",
+        "canInsaneJump",
+        "canMidairWiggle",
+        {"heatFrames": 140}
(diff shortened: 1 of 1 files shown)


osse101:
You safely jump over the lower pirate so it's not like you need an instant turnaround, but it takes some practice to land on the platform.  The big risk is not hitting the `TrickyDashJump` speed value, because you get lasered.  I don't know that it needs wiggle tech, but it looks like a wiggle.

This strat has `canInsaneJump`, so there are easier side platform entries.  Most of the side platform entries are safer.  If you can have too much speed, it's safe.  If you think you are close to entering with the right speed but are short it is a pirate hit.

Then the long runway side platforms could be added.  But not the ones that are `canInsaneJump` hard.  Metal Pirates is what I was testing with.

blkerby:
Removed the wiggle tech, and added a side platform strat. Dust Torizo Room and Metal Pirates Room were the only setups where it seemed useful. Let me know if you notice others that would make sense to add. Side platform jumps involving ceiling mockball through the door can be possible, but the fact that you then can't control your jump height makes it hard to avoid Pirate damage; so I didn't think those were worth adding.

PR #2029, opened 2025-04-23
Screw Attack Room left-to-right side platform jumps

blkerby:
This is mainly useful as an alternative to shinesparking, in order to preserve a flash suit. It can also save some energy.

The test failure should be resolved after #2026 is merged (which can be done first).

Updated the Baby Kraid Room right-side setup note. Using the overhang makes the setup faster, and doing 4 arm pumps give you a 2nd possible frame for the morph, something I had missed before.

### DIFF
--- a/region/lowernorfair/west/Screw Attack Room.json
+++ b/region/lowernorfair/west/Screw Attack Room.json
@@ -629,11 +629,11 @@
                   "openEnd": 1
                 }},
                 {"or": [
-                  "canMomentumConservingMorph",
                   {"and": [
-                    "canBeVeryPatient",
-                    "canInsaneJump"
-                  ]}
+                    "canLateralMidAirMorph",
+                    "canTrickyJump"
+                  ]},
+                  "canInsaneJump"
                 ]},
                 {"heatFrames": 80}
               ],
@@ -666,6 +666,7 @@
                   "openEnd": 1
                 }},
                 "canLateralMidAirMorph",
+                "canTrickyDashJump",
                 {"heatFrames": 70}
               ],
               "note": [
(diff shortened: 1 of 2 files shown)


region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -611,6 +611,131 @@
...
+              "minTiles": 17,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [
+                {"getBlueSpeed": {
+                  "usedTiles": 15,
+                  "openEnd": 1
+                }},
+                {"or": [
+                  "canMomentumConservingMorph",
```
osse101, line 632: It feels to me like it is an airball and not really ceilingMorphTech.  I'm morphing after the door.

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -611,6 +611,131 @@
...
+              "maxHeight": 2,
+              "minTiles": 27.4375,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [
+                {"getBlueSpeed": {
+                  "usedTiles": 15,
+                  "openEnd": 1
+                }},
+                "canLateralMidAirMorph",
```
osse101, line 668: This one is a bit odd to rank.  But I think I'll go with `canTrickyDashJump`.  
It's tighter movement than an airball so there is expertish movement on the morph which may come from the `15` tile shortcharge.  A momentumConservingMorph helped once but not usually.

Basically you need to control your shortcharge and do the stutter+early 2 tap.  It feels like you aim for a low height type jump with the amount of speed you get.

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -611,6 +611,131 @@
...
+              ]
+            },
+            {
+              "minHeight": 3,
+              "maxHeight": 3,
+              "minTiles": 37.4375,
+              "speedBooster": true,
+              "obstructions": [[3, 0]],
+              "requires": [
+                "canMomentumConservingMorph",
```
osse101, line 688: Works for placement.  I didn't touch the overhang most of the time.  It's hard to describe how to best do the 2 tap or if its even better than full speed.

region/lowernorfair/west/Screw Attack Room.json
blkerby: Right, there's a dropoff in jump height at extra run speed $3.2, so that's the reason for aiming to get more than that much speed. Updated the note to clarify, and added a `canTrickyDashJump` requirement.

PR #2030, opened 2025-04-23
Grapple Tutorial Room 3 side platform jumps

blkerby:
This is mainly useful if for some reason Gamet damage boost is not possible (e.g. if their damage value is increased). In any case, logically it can at least save a bit of energy.


osse101 (approved):
This kind of room entry would make the springball bounce across much easier.  But I think tech difficulty makes that unusable compared to an airball.

Enemizer support 👀

PR #2032, opened 2025-04-23
Bubble Mountain side platform jumps

blkerby:
This finishes off a first pass at left-to-right side platform applications.

### DIFF
--- a/region/norfair/east/Bubble Mountain.json
+++ b/region/norfair/east/Bubble Mountain.json
@@ -1111,6 +1111,194 @@
       ]
     },
     {
+      "link": [2, 1],
+      "name": "Side Platform Cross Room Jump",
+      "entranceCondition": {
+        "comeInWithSidePlatform": {
+          "platforms": [
+            {
+              "minHeight": 1,
+              "maxHeight": 1,
+              "minTiles": 13,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [
+                {"or": [
+                  "HiJump",
+                  {"and": [
+                    "canTrickyDashJump",
+                    "canTrickySpringBallJump"
+                  ]}
+                ]},
+                {"or": [
+                  "canTrickyDodgeEnemies",
+                  {"enemyDamage": {"enemy": "Waver", "type": "contact", "hits": 1}}
+                ]}
+              ],
+              "note": ["This applies to Warehouse Entrance."]
+            },
+            {
+              "minHeight": 2,
+              "maxHeight": 2,
+              "minTiles": 11.4375,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [
+                {"or": [
+                  {"and": [
+                    "HiJump",
+                    "canHorizontalDamageBoost",
+                    {"enemyDamage": {"enemy": "Waver", "type": "contact", "hits": 1}}    
+                  ]},
+                  {"and": [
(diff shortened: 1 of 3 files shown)


region/norfair/east/Bubble Mountain.json
```diff
@@ -1110,6 +1110,194 @@
...
+              "requires": [
+                {"or": [
+                  "HiJump",
+                  {"and": [
+                    "canTrickyDashJump",
+                    "canTrickySpringBallJump"
+                  ]}
+                ]},
+                {"or": [
+                  "canTrickyDodgeEnemies",
```
osse101, line 1134: Its unclear that this means kill.

region/norfair/east/Bubble Mountain.json
```diff
@@ -1110,6 +1110,194 @@
...
+              "minHeight": 1,
+              "maxHeight": 1,
+              "minTiles": 13,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [
+                {"or": [
+                  "HiJump",
+                  {"and": [
+                    "canTrickyDashJump",
```
osse101, line 1129: Dboost doesn't need a trickydashjump but maybe that is covered somewhere already.

region/norfair/east/Bubble Mountain.json
```diff
@@ -1110,6 +1110,194 @@
...
+                  ]},
+                  {"and": [
+                    "canTrickyDashJump",
+                    "canTrickySpringBallJump"
+                  ]}
+                ]}
+              ],
+              "note": ["This applies to Ridley Tank Room, Halfie Climb Room, and Dust Torizo Room."],
+              "detailNote": [
+                "In the case of the Spring Ball mid-air jump, main forward momentum while entering the pause menu,",
```
osse101, line 1161:
```suggestion
                "In the case of the Spring Ball mid-air jump, maintain forward momentum while entering the pause menu,",
```

region/norfair/east/Bubble Mountain.json
```diff
@@ -1110,6 +1110,194 @@
...
+            {
+              "minHeight": 2,
+              "maxHeight": 2,
+              "minTiles": 11.4375,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [
+                {"or": [
+                  {"and": [
+                    "HiJump",
```
osse101, line 1149: Leaving ridley tank room feels like Collision Oscillator.  Something is making it harder than it should be.

region/norfair/east/Bubble Mountain.json
blkerby: Makes sense, restructured to add a dboost option now.

region/norfair/east/Bubble Mountain.json
blkerby: Key things:
1) Don't back into the corner; just press against the wall, turn around, and run.
2) Aim down immediately after jumping.

With this, there's a two-frame window for the jump, with extra run speeds $2.B or $2.C. If you back into the corner, then $2.B won't work because you'll bonk the ceiling of the door frame. And if you don't aim down fast enough, $2.C won't work because you'll catch on the floor of the door frame. So if you were to back into the corner and try to spin jump out (without aiming down), then it just wouldn't work at all regardless of the jump timing.

Added some notes to the Ridley Tank Room setup now.

PR #2034, opened 2025-04-25
G-Mode: Big Pink Overhaul

kjbranch:
- Add missing remote acquire strats (still missing remote acquiring both items simultaneously, but not planning on adding that)
- Add missing strats that leave at 4 while still in G-mode
- Add missing Spring Flings to the top item and from the middle item going up
- Room cleanup

### DIFF
--- a/region/brinstar/pink/Big Pink.json
+++ b/region/brinstar/pink/Big Pink.json
@@ -3529,6 +3529,25 @@
       ]
     },
     {
+      "link": [15, 10],
+      "name": "G-mode Morph with HiJump Spring Fling",
+      "requires": [
+        "canEnterGMode",
+        "HiJump",
+        "h_canArtificialMorphSpringFling",
+        {"or": [
+          "canTrickyDodgeEnemies",
+          "h_canArtificialMorphPowerBomb",
+          {"enemyDamage": {"enemy": "Sm. Sidehopper", "type": "contact", "hits": 1}}
+        ]}
+      ],
+      "flashSuitChecked": true,
+      "note": [
+        "Perform a Spring Fling from the right pillar by jumping immediately before the pause finishes and turning off Spring Ball,",
+        "quickly pausing again and turning Spring Ball back on can help as well."
+      ]
+    },
+    {
       "id": 130,
       "link": [15, 12],
       "name": "G-mode Morph Overload PLMs",
--- a/region/brinstar/pink/Big Pink.json
+++ b/region/brinstar/pink/Big Pink.json
@@ -324,7 +324,7 @@
         [1, 0, 0, 0, 0],
         [1, 0, 0, 0, 0]
       ],
-      "note": "Represents being at the top item in direct g-mode with artificial morph. This is used to remote acquire the item at another location."
+      "note": "Represents being at the top item in direct g-mode with Morph or artificial morph. This is used to remote acquire the item at another location."
     },
     {
       "id": 17,
@@ -441,7 +441,8 @@
         {"id": 12},
         {"id": 13},
         {"id": 15},
-        {"id": 16}
+        {"id": 16},
+        {"id": 17}
       ]
     },
     {
@@ -453,7 +454,8 @@
         {"id": 10},
         {"id": 13},
         {"id": 15},
-        {"id": 16}
+        {"id": 16},
+        {"id": 17}
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


region/brinstar/pink/Big Pink.json
```diff
@@ -1318,6 +1542,51 @@
         "A Power Bomb also works, but will clear B, Samus can just go 4->10->13."
       ]
     },
+    {
+      "link": [4, 15],
+      "name": "G-Mode Morph, Avoid Crumbles, Bomb Power Bomb Blocks",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "direct",
```
osse101, line 1550: Add a copy strat for indirect?

region/brinstar/pink/Big Pink.json
```diff
@@ -1798,7 +2151,8 @@
       "note": [
         "While using multiple Power Bomb boost Spring Ball jumps, to climb the room,",
         "overload PLMs by moving through the camera scroll blocks in front of the passageway leading to the exit of Spore Spawn Supers.",
-        "Note that one Power Bomb can be saved by carefully jumping out of this morph passageway.",
+        "Note that one Power Bomb can be saved with a Spring Fling; pause and just before the screen is fully black, jump, disable Spring ball, then buffer right on the unpause.",
```
osse101, line 2154: If you are using the item pedestal and in direct, this would force collection of the item.

region/brinstar/pink/Big Pink.json
```diff
@@ -2036,6 +2343,43 @@
...
+          "h_canArtificialMorphPowerBomb",
+          {"enemyDamage": {"enemy": "Sm. Sidehopper", "type": "contact", "hits": 1}}
+        ]}
+      ],
+      "clearsObstacles": ["C"],
+      "flashSuitChecked": true,
+      "note": [
+        "While using multiple Power Bomb boost Spring Ball jumps, to climb the room,",
+        "overload PLMs by moving through the camera scroll blocks in front of the passageway leading to the exit of Spore Spawn Supers.",
+        "Note that one Power Bomb can be saved with a Spring Fling; pause and just before the screen is fully black, jump, disable Spring ball, then buffer right on the unpause.",
```
osse101, line 2376: It is vague where to do this

region/brinstar/pink/Big Pink.json
```diff
@@ -3019,160 +3500,409 @@
...
+      ],
+      "collectsItems": [11],
+      "flashSuitChecked": true
+    },
+    {
+      "link": [17, 12],
+      "name": "G-mode Morph",
+      "requires": [
+        "canEnterGMode",
+        "h_canArtificialMorphMovement"
```
osse101, line 3902: This strat is a good option for when you don't have artificialMorphMovement.

osse101 (commented):
I think 16 & 17 can make you double kill hoppers with PowerBombs.  But the PR looks good to me.

region/brinstar/pink/Big Pink.json
kjbranch: It could maybe be added - it would require coming in with a jump or bomb boost, because otherwise you will fall onto the crumble.

kjbranch: I'll put this one in as a fixme, because it if was a water room, I dont think it would work. (without gravity, spring, or morph)

kjbranch:
Thanks! Added a lot of FIXMEs for the feedback, unfortunately many of them would be hard to implement.

region/brinstar/pink/Big Pink.json
```diff
@@ -3019,160 +3524,408 @@
...
+      ],
+      "collectsItems": [11],
+      "flashSuitChecked": true
+    },
+    {
+      "link": [17, 12],
+      "name": "G-mode Morph",
+      "requires": [
+        "canEnterGMode"
```
osse101, line 3925: This is unsound.  PLMs need to be overloaded and you need the item present to use this strat to accomplish that.
So i guess you have add back the `artificialMorphMovement` anyway, but it can be OR'd against `lossOfAccess`.

That's how it plays out in my seeds anyway 😿 I get the item and then need to GMode down to charge.

kjbranch: Oh whoops. Glad you caught that. 
`artificialMorphMovement` wouldn't be quite enough without a spring fling, but it wouldn't be too bad. But, anyways, using the camera scroll block will use a different strat.

PR #2035, opened 2025-04-27
Croc Escape spring ball strats

blkerby:
This adds a couple notable strats based on videos that Sam recently shared:
- Ceiling Mockball Spring Fling:
  - original video by Sam: https://videos.maprando.com/video/6221
  - video from room entry, for measuring frames: https://videos.maprando.com/video/6228
- Spring Ball Jump into Damage Boost:
  - original video by Sam: https://videos.maprando.com/video/6061
  - video from room entry, for measuring frames: https://videos.maprando.com/video/6229

There's more work that can be done in the room. I didn't look into the frozen Geruta walljump variant, and Kyle also mentioned a G-mode setup with the Geruta.

### DIFF
--- a/region/norfair/west/Crocomire Escape.json
+++ b/region/norfair/west/Crocomire Escape.json
@@ -849,7 +849,7 @@
       "name": "Spring Ball Jump into Damage Boost",
       "requires": [
         {"notable": "Spring Ball Jump into Damage Boost"},
-        {"heatFrames": 1600},
+        {"heatFrames": 1920},
         "canTrickyDodgeEnemies",
         "canTrickySpringBallJump",
         "canNeutralDamageBoost",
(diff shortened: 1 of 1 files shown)


region/norfair/west/Crocomire Escape.json
```diff
@@ -844,6 +844,22 @@
         "Also requires a crouchjump and downgrab to complete the maneuver."
       ]
     },
+    {
+      "link": [2, 3],
+      "name": "Spring Ball Jump into Damage Boost",
+      "requires": [
+        {"notable": "Spring Ball Jump into Damage Boost"},
+        {"heatFrames": 1600},
```
osse101, line 852: You need to add RNG protection.  2 more springball jumps makes it better, but not guaranteed still, at 32 seconds=1920frames.

blkerby: Makes sense. I didn't practice it enough to understand the patterns, just assumed I had messed it up when it didn't work sometimes. Increased the heat frames to 1920 now.

PR #2037, opened 2025-04-29
Big Pink side platform jumps

blkerby:
A few months ago I thought this side platform jump into "Super Block Speedball" seemed Insane-level and wasn't able to get it without a save state in the door transition. Trying it again, I found that doing a slower shortcharge can make it more lenient (ideally you want not much more than $1.2 extra run speed, which is where a drop-off in jump height happens) . I think it still seems Insane if the setup is heated?

This also adds a `comeInBlueSpaceJumping` variant of the strat.

### DIFF
--- a/region/brinstar/pink/Big Pink.json
+++ b/region/brinstar/pink/Big Pink.json
@@ -1731,6 +1731,98 @@
       "devNote": "The runway here has two unusable tiles."
     },
     {
+      "link": [7, 13],
+      "name": "Side Platform Jump into Super Block Speedball",
+      "entranceCondition": {
+        "comeInWithSidePlatform": {
+          "platforms": [
+            {
+              "minHeight": 1,
+              "maxHeight": 1,
+              "minTiles": 16,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [
+                {"or": [
+                  "h_heatProof",
+                  "canBeVeryPatient"
+                ]}
+              ],
+              "note": ["This applies to Lava Dive Room."]
+            },
+            {
+              "minHeight": 2,
+              "maxHeight": 2,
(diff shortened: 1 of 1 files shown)


osse101:
I'm better at shortcharging now, so these feel better than the last time I tried the strat.  In summary, you can do an insane level shortcharge to make the super shot kinda easy or you can do a bad shortcharge and have an insane time landing the super.  
So I visualize it as `or:[insaneJump, 14TileShortcharge]`.  

The heat has the same effect as on all other shortcharges and makes it a notch harder.  I can get it right away from lava dive, but its a tough shortcharge to get every time or even much of the time.

blkerby:
> I'm better at shortcharging now, so these feel better than the last time I tried the strat. In summary, you can do an insane level shortcharge to make the super shot kinda easy or you can do a bad shortcharge and have an insane time landing the super. So I visualize it as `or:[insaneJump, 14TileShortcharge]`.

Yeah I was thinking of it like that too. Adjusted the `requires` to be written like that now, and clarified the note.

> The heat has the same effect as on all other shortcharges and makes it a notch harder. 

Right. The tricky thing is that with this entrance condition we don't have a way to model the shortcharge being done in the other room (possibly heated). I.e., the `getBlueSpeed` requirements are going to be interpreted as applying in Big Pink, so the heat difficulty adjustment is not going to automatically apply. The `{"or": ["h_heatProof", "canBeVeryPatient"]}` is a way to force that difficulty bump in the heated case. It's a little hacky. An alternative would be to introduce another entrance condition, e.g. `comeInWithBlueSidePlatformJump` and include a `maxExtraRunSpeed` property; but that would also be more complicated and require more support on the randomizer/tooling side so I'm not sure it would be worth it 🤷‍♂️

PR #2039, opened 2025-04-30
Rename helpers to not use "can"

kjbranch:
`can` is use for tech names, so I dont think it should be used for helpers. This is split into 2 PRs. All the artificial morph ones will be separate


blkerby (approved):
Looks good to me.

PR #2040, opened 2025-04-30
Below Spazer side platform jumps

blkerby:
This is similar to the Big Pink super block speedball, except much easier. Coming in from Lava Dive appears impossible since you can't get enough run speed to make it to the ledge while also getting blue speed.


osse101:
I think Its Very Hard if you also provide some item to help, like spazer for the block shot.  Most ways of running can work.

PR #2044, opened 2025-04-30
Golden Torizo side platform jumps

blkerby:
When I looked at this a few months ago, I thought some of these would be Beyond. But with a better understanding now of how to set these up, they are seeming fine in Extreme and Insane. The main things that make it easier are
1. Pause buffering the morph input. This breaks up the tight input sequence, and makes it more reliable to use a visual cue for the pause press to get a consistently timed morph.
2. Using a shorter runway (in the long setup rooms). Having too much horizontal speed makes the timing for the unmorph tighter, or even impossible depending on the morph timing. Since you have to wait for the morph animation to finish before unmorphing, in that time you could overshoot past the bomb blocks if your speed is too high. And of course, at lower speeds it is also easier to time a last-frame jump.

### DIFF
--- a/region/lowernorfair/west/Golden Torizo's Room.json
+++ b/region/lowernorfair/west/Golden Torizo's Room.json
@@ -905,6 +905,102 @@
       ]
     },
     {
+      "link": [2, 4],
+      "name": "Side Platform Cross Room Jump",
+      "entranceCondition": {
+        "comeInWithSidePlatform": {
+          "platforms": [
+            {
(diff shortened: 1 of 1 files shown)


osse101:
- I gave 1 shot at hijumpless through noob bridge, and you get the height but I couldn't unmorph in time to avoid going too far left.
- It all looks fair with the pause buffered morph
- Getting not enough height to break the last bomb block, or too much and touching a solid block feels pretty bad.

PR #2045, opened 2025-05-01
Screw Attack Room side platform and blue speed jumps

blkerby:
Along with blue speed side platform jumps from the middle right door, this also adds blue speed jumps using regular (connected) runways.

I was chatting with Kyle and he pointed out that there's more that could be done: we could add side platform strats that jump in with Screw Attack (instead of with blue speed), and also strats that use blue speed to break both sets of bomb blocks, i.e. jumping up to break the top ones then falling down to break the bottom ones and reach the item. I left these as a FIXME since I didn't want this PR to get too complicated.


region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -1314,6 +1314,242 @@
...
+      "entranceCondition": {
+        "comeInGettingBlueSpeed": {
+          "length": 0,
+          "openEnd": 1,
+          "minExtraRunSpeed": "$0.C",
+          "maxExtraRunSpeed": "$0.F"
+        }
+      },
+      "requires": [
+        "HiJump",
```
osse101, line 1390: This is probably an insane level shortcharge.

osse101 (commented):
I associate the pause buffer morph with `canInsaneJump`, but then it looks more like a positioning based trick like canTrickyDashJump.

region/lowernorfair/west/Screw Attack Room.json
blkerby: Currently a speed of $0.F or lower would be a Beyond shortcharge, as $1.0 is the lowest speed assumed possible in Insane: https://github.com/vg-json-data/sm-json-data/blob/master/strats.md#blue-run-speed-table

Makes sense to reduce the $1.0 in that table down to $0.F, to make this logical for Insane. Updated that now.

blkerby:
> I associate the pause buffer morph with `canInsaneJump`, but then it looks more like a positioning based trick like canTrickyDashJump.

Yeah some of the cases felt close to the border between Expert and Extreme. I think adding `canInsaneJump` to them can be fine, just let me know which ones

region/lowernorfair/west/Screw Attack Room.json
```diff
@@ -1314,6 +1314,242 @@
...
+              "note": ["This applies to Metal Pirates Room."]
+            },
+            {
+              "minHeight": 3,
+              "maxHeight": 3,
+              "obstructions": [[5, 2]],
+              "minTiles": 31.4375,
+              "speedBooster": true,
+              "requires": [
+                "HiJump",
```
osse101, line 1527:
```suggestion
                "HiJump",
                "canInsaneJump",
```

osse101 (commented):
Is there interest in breaking both sets of bomb blocks, and going item to top door?

blkerby:
> Is there interest in breaking both sets of bomb blocks, and going item to top door?

Yeah, I put it as a FIXME

PR #2046, opened 2025-05-01
Maridia Elevator Room right-to-top side platform jumps

blkerby:
These ones are pretty dumb.


osse101 (dismissed):
You can Springball Bounce and turn off springball before the door, and its a little easier.

blkerby:
> You can Springball Bounce and turn off springball before the door, and its a little easier.

Oh yeah, makes sense, and we already have a cross-room "Blue Spring Ball Bounce" strat here. Added a note about it in the new strat.

PR #2047, opened 2025-05-01
Add Green Hill Zone side platform setup, extend Halfie Climb setup

blkerby:
Kyle pointed out this Green Hill Zone setup. It has a 4-tile-long obstruction, which makes it have very limited use. It only seems to be usable with the gate open, with the only application being jumping up the bottom left of Maridia Elevator Room.

The Halfie Climb Room setup was too short because I forgot about being able to shoot out the shot blocks. In theory there could be situations where the Oums get in your way, but it didn't seem like it's really a problem. If needed we could add a `comeInNormally` entrance condition.

This should wrap up the first pass at side platforms.

### DIFF
--- a/region/brinstar/green/Green Brinstar Main Shaft.json
+++ b/region/brinstar/green/Green Brinstar Main Shaft.json
@@ -4709,13 +4709,27 @@
         "comeInWithSidePlatform": {
           "platforms": [
             {
+              "minHeight": 2,
+              "maxHeight": 2,
+              "minTiles": 45,
+              "speedBooster": true,
+              "obstructions": [[1, 0]],
+              "requires": [],
+              "note": "Applies to Halfie Climb Room."
+            },
+            {
               "minHeight": 3,
               "maxHeight": 3,
               "minTiles": 39.4375,
               "speedBooster": true,
               "obstructions": [[3, 2]],
               "requires": [],
-              "note": "Applies to Metal Pirates Room."
+              "note": "Applies to Metal Pirates Room.",
+              "detailNote": [
+                "This has a 2-frame window for the jump,",
+                "then a 1-frame or 2-frame window for the morph (after the transition), depending on the jump,",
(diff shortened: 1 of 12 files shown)


region/brinstar/green/Green Brinstar Main Shaft.json
```diff
@@ -4708,14 +4708,28 @@
...
+            },
             {
               "minHeight": 3,
               "maxHeight": 3,
               "minTiles": 39.4375,
               "speedBooster": true,
               "obstructions": [[3, 2]],
               "requires": [],
-              "note": "Applies to Metal Pirates Room."
+              "note": "Applies to Metal Pirates Room.",
```
osse101, line 4727: Springball bouncing is easier for both.  

In metal pirates you have to do a trickydashjump from left wall->~8 tiles from the right wall.

blkerby: I started looking into adding a spring ball bounce strat now and then realized that there's an in-room tricky dash jump into spring ball jump strat that's much easier than either of these :man_facepalming: Added a `devNote` now to mention it. Should we keep the side platform strats with that, or just remove them?

osse101: Springball Bounce halfie climb is a fine alternative to trickydashjump.  
in room trickydashjump is simpler than the out of room trickydashjump.  I could see disabling trickyDashJump but not canInsaneJump, but the in room jump is very reasonable.

blkerby: Sorry for the delay on this, been distracted with ALTTP editor stuff. Just added a cross-room spring ball bounce strat, because yeah it is a reasonable alternative to the tricky dash jump. I was going to put `canTrickySpringBallBounce` (Expert) on it, but testing it out it seems simple/lenient enough that this doesn't seem needed; so it's just a `canTrickyJump`, putting it into Very Hard if that sounds good?

PR #2048, opened 2025-05-02
Refine Botwoon ETank right-side shinecharges

blkerby:
- The existing strats required 2 runway tiles in order to stutter water shinecharge here. This PR adds 1-tile options, with the `canPreciseStutterWaterShineCharge` tech. With 1 runway tile, it requires a 1 or 2 frame release of forward, and a last-frame forward repress, which seems ok for Expert.
- Options are added to go back out through the right door with temporary blue or shinecharged. For the "Leave Shinecharged" variants, I just copied the strats from Aqueduct. In theory, the slopes will slow down your horizontal movement while running into the room, making it so you cover relatively more ground while spin jumping back out. So for some of these it should be possible to leave with slightly more frames remaining here compared to Aqueduct, but I don't think it's significant enough to bother changing the strats. 
- I added a junction node (7), since like in other rooms, there are several ways to gain the blue speed and also several ways to use it.

### DIFF
--- a/region/maridia/inner-pink/Botwoon Energy Tank Room.json
+++ b/region/maridia/inner-pink/Botwoon Energy Tank Room.json
@@ -2124,7 +2124,6 @@
       "requires": [
         "h_waterGetBlueSpeed"
       ],
-      "clearsObstacles": ["A"],
       "devNote": [
         "This is just a rough estimate of the minimum number of tiles that this runway can represent.",
         "It is designed for a skill level that won't be able to use canStutterWaterShineCharge."
@@ -2141,8 +2140,7 @@
       },
       "requires": [
         "h_stutterWaterGetBlueSpeed"
-      ],
-      "clearsObstacles": ["A"]
+      ]
     },
     {
       "link": [4, 7],
@@ -2155,7 +2153,6 @@
       "requires": [
         "h_stutterWaterGetBlueSpeed"
       ],
-      "clearsObstacles": ["A"],
       "note": [
         "With only 1 tile of runway, this requires releasing forward for 1 or 2 frames",
         "and repressing forward on the last possible frame."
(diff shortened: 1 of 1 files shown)


region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -1998,6 +2111,90 @@
...
+      "link": [4, 7],
+      "name": "Blue Spring Ball Bounce",
+      "entranceCondition": {
+        "comeInWithBlueSpringBallBounce": {
+          "movementType": "controlled"
+        }
+      },
+      "requires": [],
+      "devNote": [
+        "With a mid-air spring ball jump it would be possible to go through the morph tunnel.",
```
osse101, line 2194: I don't get this note.
I think it still bounces through the speed blocks.

osse101 (commented):
There is some inconsistency with 7 being right of the sand while also breaking the speed blocks when you enter the node.  It'd be nice if that could be cleared up somehow, like by saying 7 is a longer stretch within the speed blocks.  But if not, the obstacles are overall cleared how they are supposed to be.

blkerby:
> There is some inconsistency with 7 being right of the sand while also breaking the speed blocks when you enter the node. It'd be nice if that could be cleared up somehow, like by saying 7 is a longer stretch within the speed blocks. But if not, the obstacles are overall cleared how they are supposed to be.

Ah, right. This got mixed up because I initially defined 7 as being by the left door, then changed it to be right of the sand because it makes more sense for the 7->4 leaveWithTemporaryBlue. I updated it now so that the `obstaclesCleared` happens in the 7->1 strat instead of in the strats that end at 7.

region/maridia/inner-pink/Botwoon Energy Tank Room.json
blkerby: Updated the note to try to make it more clear now. This strat does bounce through the speed blocks. The note is just saying that with these items you could already traverse the room in a simpler way by spring ball jumping up to the morph tunnel, rather than doing a blue spring ball bounce. The only logical advantage of the blue spring ball bounce is to circumvent the item in case you want to save it to be collected later.

PR #2050, opened 2025-05-04
Refine Crab Shaft shinecharges

blkerby:
- Added a junction node 5 representing getting a shinecharge after running in from the left door.
- Added temporary blue chain strats from there to all 3 doors.
- Refined the shinespark frames and excessFrames for the strat that shinesparks up. The old value of 6 excess frames was unsound for getting onto the top platform.

### DIFF
--- a/region/maridia/inner-pink/Crab Shaft.json
+++ b/region/maridia/inner-pink/Crab Shaft.json
@@ -71,6 +71,21 @@
         [1, 0],
         [1, 1]
       ]
+    },
+    {
+      "id": 5,
+      "name": "Middle Right Shinecharged",
+      "nodeType": "junction",
+      "nodeSubType": "junction",
+      "mapTileMask": [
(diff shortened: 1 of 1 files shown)


region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -1660,6 +1910,102 @@
...
       "flashSuitChecked": true
+    },
+    {
+      "link": [5, 1],
+      "name": "Leave With Spark",
+      "startsWithShineCharge": true,
+      "requires": [
+        "canShinechargeMovementComplex",
+        {"shineChargeFrames": 120},
+        {"shinespark": {"frames": 9}}
```
osse101, line 1921: - There isn't a simple way to shinespark back out the door.
- Are shineChargeFrames lenienced at this step?  I don't want scaling to make a return shinespark impossible.

PR #2055, opened 2025-05-07
Refine G-Mode: Kraid Brinstar

kjbranch:


region/brinstar/kraid/Warehouse Kihunter Room.json
```diff
@@ -282,12 +282,17 @@
...
+          "canTrickyGMode",
           {"enemyDamage": {"enemy": "Kihunter (green)", "type": "contact", "hits": 1}}
         ]}
       ],
       "flashSuitChecked": true,
-      "note": "Overload the PLMs by rolling through the camera scroll blocks which are just above the shot blocks.",
-      "devNote": "In immobile, the Kihunters are in a worse location, but Samus starts with i-frames."
+      "note": [
+        "Overload the PLMs by rolling through the camera scroll blocks which are just above the shot blocks.",
+        "It is possible but tricky to do this damageless without a Power Bomb. Get the first Kihunter to swoop left twice,",
```
osse101, line 292: I rushed to the shot blocks and backed up when swooped at.
Move in front of the blocks 3 times and retreat 3 times, then the PLMs are overloaded too.
But it feels like there is some ki movement RNG because they sometimes move differently and then it doesn't work. -direct

PR #2056, opened 2025-05-07
G-Mode: Etecoon E-Tank Room

kjbranch:


region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -1304,6 +1567,115 @@
...
+              "ScrewAttack",
+              "Ice",
+              {"ammo": {"type": "Missile", "count": 5}},
+              {"ammo": {"type": "Super", "count": 5}}
+            ]}
+          ]}
+        ]},
+        {"or": [
+          "canBeVeryPatient",
+          "h_artificialMorphCeilingBombJump",
```
osse101, line 1614: Options where you have morph
- walljump instant morph
- hijump

osse101 (commented):
You have to fall in the pit before PLMs are overloaded, to touch the camera scroll blocks so that the camera can reset.
This problem only applies to PBs, or no Morph movement.  And you can plan your tile touches to overload using the scroll blocks while falling, but the strat needs to more strongly call out the need to do that.

PR #2059, opened 2025-05-19
Crab Maze leave shinecharged

blkerby:
There's already a "Leave with Spark" water shinecharge strat here in Very Hard that can only spark out in bottom position. The new strat can make it out with frames remaining. Without canInsaneJump, the extra lenience is designed to make it so that in Expert you would only be expected to spark at the doorway, not actually leave with frames remaining.

Video: https://videos.maprando.com/video/6401

### DIFF
--- a/region/crateria/east/Crab Maze.json
+++ b/region/crateria/east/Crab Maze.json
@@ -456,6 +456,27 @@
       }
     },
     {
+      "link": [2, 2],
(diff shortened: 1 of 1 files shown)


osse101 (dismissed):
My best left the room with 7 frames remaining.  But all of these PreciseStutterWaterShineCharges have been too hard for me.

PR #2068, opened 2025-05-26
Warehouse E-Tank tricky evade beetoms

blkerby:
Videos:
- From an air room: https://videos.maprando.com/video/6476
- From a water room: https://videos.maprando.com/video/6475

### DIFF
--- a/region/brinstar/kraid/Warehouse Energy Tank Room.json
+++ b/region/brinstar/kraid/Warehouse Energy Tank Room.json
@@ -395,6 +395,17 @@
       ]
     },
     {
+      "link": [1, 2],
+      "name": "Very Tricky Evade",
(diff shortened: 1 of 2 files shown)


blkerby:
Added requirements for the door to be unlocked, since both evade strats require landing in the door frame. For example, they wouldn't work with the vanilla gray door.

osse101:
When I tried your setup, the leftmost beetom wouldn't jump forward when it was supposed to.  Getting grabbed in the transition didn't seem safe either.  I found my own sequence of movements, so there is a way I'm satisfied with.

This strat involves exactly copying a setup so it's more like a notable strat.  Maybe it's sufficient to be in extreme and have a guide video.

PR #2070, opened 2025-05-26
Add mid-air morph damage boost option to escape beta PBs

blkerby:
This is an easier option (if Morph is available) so seems good to include in logic.

Video by Sam: https://videos.maprando.com/video/5804

### DIFF
--- a/region/brinstar/red/Beta Power Bomb Room.json
+++ b/region/brinstar/red/Beta Power Bomb Room.json
@@ -522,7 +522,13 @@
       "name": "Damage Boost Escape",
       "requires": [
         {"obstaclesCleared": ["B"]},
-        "canHorizontalDamageBoost",
+        {"or": [
+          {"and": [
+            "canMidAirMorph",
+            "canNeutralDamageBoost"
+          ]},
+          "canHorizontalDamageBoost"
+        ]},
         {"thornHits": 1}
       ]
     },


osse101 (dismissed):
Looking at this, I'm thinking the `canHorizontalDamageBoost` is unreasonable.  Its more intuitive than there being a safe spot, but I don't think you can reliably escape until you discover the edge of the thorn and probably find the safe spot.
It does still function as an alternate tech with some damage, compared to the trickyJump.

region/brinstar/red/Beta Power Bomb Room.json
```diff
@@ -522,7 +522,11 @@
       "name": "Damage Boost Escape",
       "requires": [
         {"obstaclesCleared": ["B"]},
-        "canHorizontalDamageBoost",
+        "canNeutralDamageBoost",
+        {"or": [
+          "canMidAirMorph",
+          "canTrickyJump"
```
osse101, line 528: With trickyjump, you would land on the safe spot using #21

blkerby: Right, with this PR the morphless damage boost option doesn't add anything logically since it is technically superseded by the safe spot strat; having it as an option just helps document that it's a possible thing to do. This is assuming VH is the right place for both the morphless damage boost and safe spot strats?

kjbranch: We typically havent been including strats that are strictly harder or have more requirements than another strat (This seems to be a similar difficulty but has more requirements). Generally, when we add strats like that, we will try to combine them together, so that we can mention them both in the same note - which could help provide an alternative if someone is struggling with a strat. Is there a good way to combine the morphless tricky jump variant with the safe spot jump?

PR #2073, opened 2025-05-26
Add reqs to deal with Pirates in Escape Room 4 L->R

blkerby:
With nothing, killing all the Pirates here takes quite a long time, which doesn't seem fair for Basic. And even for Hard, if you were in a situation where you really had to kill them (e.g. walljumpless with IBJ) and had no weapons, I think it would feel pretty bad. So this splits off "Evade Pirates" and "Tank Damage" strats from the "Base" kill strat. There could possibly be more options here, so let me know if I missed something.

### DIFF
--- a/region/tourian/main/Tourian Escape Room 4.json
+++ b/region/tourian/main/Tourian Escape Room 4.json
@@ -269,11 +269,70 @@
           "HiJump",
           "SpaceJump",
           "canWalljump",
-          "canIBJ",
+          "canJumpIntoIBJ",
+          {"and": [
+            "canIBJ",
+            "canBePatient"
+          ]},
           "canSpringBallJumpMidAir"
+        ]},
+        {"or": [
+          "canBePatient",
+          {"enemyKill": {
+            "enemies": [
+              ["Tourian Space Pirate (all)"],
+              ["Tourian Space Pirate (all)"],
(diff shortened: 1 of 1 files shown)


region/tourian/main/Tourian Escape Room 4.json
```diff
@@ -269,11 +269,70 @@
...
+      "requires": [
+        "canCarefulJump",
+        {"enemyDamage": {
+          "enemy": "Tourian Space Pirate (all)",
+          "type": "contact",
+          "hits": 2
+        }},
+        {"or": [
+          "HiJump",
+          "canWalljump",
```
osse101, line 326: I think, `canConsecutiveWalljump` for difficulty placement.  Or it would take more pirate hits to climb up, at medium.

blkerby: `canConsecutiveWalljump` is a Medium tech so it wouldn't change the difficulty. I think you're right it's a problem. Medium players will probably get knocked down almost indefinitely while trying to walljump past the Pirates. It needs some non-trivial enemy dodging to avoid that outcome, and this fits more in Hard. So seems best to just remove the `canWalljump` option here. The Hard-level Pirate evade with walljump is covered already in the other strat.

Issue #1462, opened 2024-04-12
Add timings to strats (e.g. for escape timings)


blkerby:
Currently, escape timings in Map Rando are defined in a JSON file that mostly just gives one timing for each pair of doors in every room: https://github.com/blkerby/MapRandomizer/blob/main/rust/data/escape_timings.json. Recently there has been interest in supporting more refined escape timings, ones that don't assume all movement items are collected, and possibly taking into account cross-room strats (e.g. to shinespark through Croc Speedway in reverse). This is a much bigger undertaking than what we've done so far with escape timings, and I think it would make sense to be part of the sm-json-data. If this is done, a main benefit would be that in Desolate seeds it would no longer be needed for all movement items to be given for the escape, and escapes could require doing trickier strats. It would also allow us to support more options for "Escape start" seeds, where you would get only a subset of the items and again might have to do trickier strats.

Kyle and I were discussing a couple possible strategies:

1. Include timing data on strats in sm-json-data; e.g. this could be a new strat property `frames`. Ideally every strat could have a timing, but incomplete coverage can also work; strats that don't have a timing just wouldn't be logically usable in the escape.
2. Create a new project/repository specifically for escape timings. It would probably follow a similar structure to sm-json-data and share the same schemas, tech, and helper definitions but could have its own room data.

If possible, the first option seems most appealing, since having timing data on strats could have many applications, making it nice if it could stay integrated with the rest of the data. Examples of other things that could be done with it include finding optimized routes through a seed (minimizing total time) and supporting randomizer options that modify which rooms are heated. 

There are a number of challenges:

- In many non-heated rooms, junctions are currently defined in the way that doesn't make sense for measuring timings. For example, in Business Center the main junction has a large area covering most of the room, so there's no reasonable way to define timings for strats that move to or from that junction. I think that for timings, we would often need new strats that go directly from door to door, bypassing the junction. In some cases defining new junctions could be helpful; similar to what we're already doing in heated rooms, junctions would basically need to be defined as specific points in the room rather than large areas. The new strats that bypass the junctions (or use new ones) could clutter up the data, which was part of the motivation for considering using a separate project. On the other hand, we already have lots of strats that will seem like clutter to many players (e.g. G-mode strats, grapple teleports, runways, Crystal Flash refills, etc.); we could just add an attribute to mark strats that are only useful for "speed" purposes, i.e. where the timing matters, so then these strats could be filtered out if people aren't interested in them.
[...]

PR #2078, opened 2025-05-30
Mention escape Hero shot in Big Pink.json

cconnett:
I and at least one other person in Discord were unaware of how to navigate this graph edge during escape and sought help. Explaining the technique in the `note` field should help future players.

### DIFF
--- a/region/brinstar/pink/Big Pink.json
+++ b/region/brinstar/pink/Big Pink.json
@@ -2735,7 +2735,8 @@
           {"obstaclesCleared": ["E"]}
         ]}
       ],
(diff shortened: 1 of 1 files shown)


blkerby:
Currently the sm-json-data isn't the right place for escape-specific logic. This could change someday (e.g. if the approach in https://github.com/vg-json-data/sm-json-data/issues/1462 ever happens). For now, the Map Rando escape logic is defined in https://github.com/blkerby/MapRandomizer/blob/main/rust/data/escape_timings.json. It doesn't currently have the capability to add notes to edges but that would be a useful feature.

PR #2079, opened 2025-05-30
Kraid X-mode leave shinecharged

blkerby:
We had a note about this one, suggesting it was left out because canRiskPermanentLossOfAccess would be required, but it's not really because gaining a shinecharge is also possible with Kraid dead.

Videos:
- the one Sam recently shared: https://videos.maprando.com/video/6505
- variation with a neutral damage boost: https://videos.maprando.com/video/6506

### DIFF
--- a/region/brinstar/kraid/Kraid Room.json
+++ b/region/brinstar/kraid/Kraid Room.json
@@ -231,7 +231,7 @@
     {
       "id": 2,
       "link": [1, 1],
-      "name": "Leave Shinecharged",
+      "name": "Leave Shinecharged (Kraid Defeated)",
       "requires": [
         {"canShineCharge": {"usedTiles": 22, "openEnd": 0}},
         "canShinechargeMovement",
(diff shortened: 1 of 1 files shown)


region/brinstar/kraid/Kraid Room.json
```diff
@@ -231,18 +231,56 @@
...
+        ]},
+        "canShinechargeMovement",
+        {"shineChargeFrames": 55},
+        {"or": [
+          {"and": [
+            "canNeutralDamageBoost",
+            "canBeVeryPatient"
+          ]},
+          {"shineChargeFrames": 10}
+        ]}
```
kjbranch, line 269: I don't follow this `or`, but the rest looks good

kjbranch (dismissed):

region/brinstar/kraid/Kraid Room.json
blkerby: That `or` just makes it expect 10 more shinecharge frames for the case where you don't do the neutral damage boost to get up to the ledge quicker. My thinking is the damage boost version fits more in Insane, and the 5 lenience shinecharge frames in Extreme isn't enough to make up the difference so you want a little extra lenience here.

PR #2085, opened 2025-06-05
Refine some Metroid Room 2 shinecharges

blkerby:
I started going through the backlog of Incomplete videos by Sam and saw several strat options were missing for going up the left side of Metroid Room 2. These are mostly walljump-avoid strats, but not strictly since they also avoid `canMidairWiggle` tech. 

There's also a two-tile squeeze strat for top-to-bottom which surprisingly appears slightly faster than the moonfall and Morph variants. The difference seemed not quite enough to justify tightening the shinecharge frames though. It feels more difficult than the moonfall/Morph options overall but easier to avoid Metroid damage so it can be a viable strat. The Metroid damage scaling for this strat in general (across all the variants) seemed a bit unreasonable, because if you don't do it just right then you take like 10x more Metroid damage; so I split off a "canInsaneJump" option.

Videos for all the new strats/variants on the video site.

### DIFF
--- a/region/tourian/main/Metroid Room 2.json
+++ b/region/tourian/main/Metroid Room 2.json
@@ -414,8 +414,9 @@
       "note": [
         "If Metroids are alive, minimizing damage from them requires tricky movement to manipulate them:",
         "the moonfalling option must be done starting from almost a tile to the left of the opening;",
-        "likewise, the morphing option should soft-morph about a tile to the left of the opening. "
-      ]
+        "likewise, the morphing option should soft-morph about a tile to the left of the opening."
+      ],
+      "devNote": ["FIXME: Unmorphing and downbacking through the gap can leave with more frames remaining but is more difficult."]
     },
     {
       "id": 20,
--- a/region/tourian/main/Metroid Room 2.json
+++ b/region/tourian/main/Metroid Room 2.json
@@ -805,8 +805,8 @@
         "Go up the left side of the room to make it out the top door with a few shinecharge frames remaining."
       ],
       "detailNote": [
-        "Do a big initial jump to land in the middle of the lower platform,",
-        "and jump before running more than 2.75 tiles, to avoid a Speed Booster dropoff in jump height."
+        "Avoid getting more than 2.75 tiles of run speed on the lower platform (or Speed Booster will shorten Samus' jump):",
+        "either release run early, or do a big initial jump to land in the middle of the lower platform before starting to run."
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


region/tourian/main/Metroid Room 2.json
```diff
@@ -754,7 +765,49 @@
...
+          ]}
+        ]},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],
+      "flashSuitChecked": true,
+      "note": [
+        "Go up the left side of the room to make it out the top door with a few shinecharge frames remaining."
+      ],
+      "detailNote": [
+        "Do a big initial jump to land in the middle of the lower platform,",
```
osse101, line 808: I preferred jumping normally and tapping run.

region/tourian/main/Metroid Room 2.json
```diff
@@ -931,6 +984,63 @@
...
+          {"and": [
+            {"or": [
+              "f_KilledMetroidRoom2",
+              "ScrewAttack",
+              {"and": [
+                {"metroidFrames": 145},
+                {"or": [
+                  "Plasma",
+                  "Wave",
+                  {"ammo": {"type": "Missile", "count": 1}}
```
osse101, line 1007: By the looks of it, these don't work with a super locked door.  Or would be adding extra requirements.

osse101: Actually, extra requirements would be correct for strat repeatability.

region/tourian/main/Metroid Room 2.json
blkerby: Makes sense. Updated the note to include that as an option.

region/tourian/main/Metroid Room 2.json
blkerby: Yeah the `unlocksDoors` here already requires `f_KilledMetroidRoom2` or `ScrewAttack`, which makes the Plasma/Wave/Missile option irrelevant.

region/tourian/main/Metroid Room 2.json
```diff
@@ -389,11 +389,16 @@
...
+          "Morph",
+          "canTwoTileSqueeze"
         ]},
         {"or": [
           "f_KilledMetroidRoom2",
-          {"metroidFrames": 25}
+          {"metroidFrames": 100},
+          {"and": [
+            "canInsaneJump",
+            {"metroidFrames": 10}
```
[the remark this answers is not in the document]
osse101: Like this https://videos.maprando.com/video/6631
And it would be the hardest method to execute.

I feel I would always trust TwoTileSqueeze over moonfall here.  But they are both low level tech in context here, and not driving the logic.

PR #2093, opened 2025-06-12
minor fixes to Big Pink

internet-account2:
As I recently learned, those enemies are not called Zebbos but Zebs.

Since the devnote was part of a 2->12 strat, I think `1->15->12` is probably an oversight?

No idea if this pr is problematic since there's another one open for the same file right now.


kjbranch (approved):
Good catch. In general, having two PRs in the same room is not going to cause conflicts, unless theyre changing a major portion of the room or significant changes to adjacent strats.

PR #2094, opened 2025-06-23
Fixes to strats in my plando

kjbranch:


region/wreckedship/main/Bowling Alley.json
```diff
@@ -1318,6 +1319,52 @@
...
+        {"or": [
+          "HiJump",
+          {"and": [
+            "canTrickySpringBallJump",
+            "canInsaneJump"
+          ]},
+          {"and": [
+            "canTrickySpringBallJump",
+            {"thornHits": 1}
+          ]},
```
osse101, line 1339: With trickyjump, you shouldn't be taking a spike hit. Its pretty safe with a turnaround spinjump.
An extra tech to consider adding would be `canBePatient` since it would be slower to reattempt with springball jumps.

> "spikeHits": 1

region/wreckedship/main/Bowling Alley.json
```diff
@@ -1318,6 +1319,52 @@
...
+        "h_bombThings",
+        "h_CrystalFlash",
+        "canXRayClimb",
+        {"or": [
+          "h_fourTileJumpMorph",
+          "h_useSpringBall",
+          "canIBJ",
+          "SpeedBooster",
+          {"and": [
+            {"not": "f_DefeatedPhantoon"},
```
osse101, line 1355: Is this to climb farther left in the room?

osse101 (commented):
https://videos.maprando.com/video/6753
The bowling part using an IBJ.  It should be possible to not take a spike hit but I couldn't get it -- the morph is very tight.
Using speed looks like it would work too but is also very difficult.  You can Jump into the spikes for enough runway to jump high enough, I think.

region/wreckedship/main/Bowling Alley.json
kjbranch: Yep. I wasnt sure if it was worth mentioning, since the midair morph isnt bad. There could also be strats that take spike hits while using the left crumble blocks, but it seems like overkill.

kjbranch:
I couldnt get the IBJ to work, it is too precise for me.
I almost got the speedy jump to work with the power off, but even then, it's still very difficult.

region/wreckedship/main/Bowling Alley.json
```diff
@@ -1318,6 +1319,86 @@
...
+        "canJumpIntoIBJ",
+        "canDoubleBombJump",
+        {"or": [
+          {"spikeHits": 1},
+          {"and": [
+            {"not": "f_DefeatedPhantoon"},
+            "canRiskPermanentLossOfAccess"
+          ]}
+        ]},
+        "h_bombThings",
```
osse101, line 1379:
```suggestion
```

region/wreckedship/main/Bowling Alley.json
```diff
@@ -1318,6 +1319,86 @@
...
+        {"or": [
+          "h_fourTileJumpMorph",
+          "h_useSpringBall",
+          "canIBJ",
+          "SpeedBooster",
+          {"and": [
+            {"not": "f_DefeatedPhantoon"},
+            "canRiskPermanentLossOfAccess"
+          ]}
+        ]}
```
osse101, line 1391:
```suggestion
        "canXRayClimb"
```

osse101 (dismissed):

PR #2095, opened 2025-06-24
Direct G-mode open blue gates

kjbranch:
Also add missing indirect strats in Double Chamber

### DIFF
--- a/region/brinstar/green/Green Hill Zone.json
+++ b/region/brinstar/green/Green Hill Zone.json
@@ -1171,6 +1171,22 @@
       "flashSuitChecked": true
     },
     {
+      "link": [3, 1],
+      "name": "G-Mode Direct, Open Gate",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "direct",
+          "morphed": false
+        }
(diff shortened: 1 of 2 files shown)


kjbranch:
Here's an example video: https://videos.maprando.com/video/5169
Crouching increases the frame window from 1f to 5f

osse101 (commented):
It looks wrong to write gateglitching without any tech requirements but this is probably a case of spreading the word that this is something you can do in GMode with overloaded PLMs.  It resembles the reverse Botwoon beam(not wave) fight, but still different.

The heated versions require varia or the item being Etank/Varia to not instantly die.

kjbranch:
I was thinking the same thing. I would have at least put the gate glitch tech on it if it wouldnt require me to duplicate all of those strats in Double Chamber for a direct/indirect option. I think the Double Chamber ones will be very uncommon, and I think the Green Hill Zone wont ever appear (it could with mirrored rooms, if it's connected to the top of Halfie Climb or Pink Brin PBs, since the door is instantly closed in G-mode, but that's the only use I could think of)

PR #2098, opened 2025-06-24
Second pass at upward G-mode setups

kjbranch:


region/norfair/crocomire/Post Crocomire Jump Room.json
```diff
@@ -840,6 +840,31 @@
...
+        ]}
+      ],
+      "exitCondition": {
+        "leaveWithGModeSetup": {}
+      },
+      "flashSuitChecked": true,
+      "note": [
+        "It can help to moonwalk towards a Mella that has never been on screen, so that it starts swooping immediately as it first enters the screen.",
+        "This will make it so that when Samus is below it, it will always swoop and go up slightly compared to the previous swoop.",
+        "Once it is swooping high enough, freeze the Mella while it is both high enough and still moving upwards.",
```
osse101, line 863: This is why I prefer using a low mella.  
Jump and aim down to freeze it.  Your hitbox will be small and the mella moving fast and up.

I didn't think having morph was a big deal; with the high mella method.  For the same reason that the airball is forgiving, so is the spinjump.

Avoiding damage, so your energy is set appropriately, is the hardest part.  Freezing at the right time may take a few tries but isn't too bad.

region/norfair/crocomire/Post Crocomire Shaft.json
```diff
@@ -660,7 +660,7 @@
       "requires": [
         "canTrickyUseFrozenEnemies",
         {"ammo": {"type": "Super", "count": 1}},
-        "h_useSpringBall",
+        "Morph",
```
osse101, line 663: Its not obvious but you can jump onto the viola without morph if you aim down.  Otherwise you trip and touch the transition.

PR #2100, opened 2025-06-25
Indirect upward G-mode, more Everest strats

kjbranch:


kjbranch:
I updated the tech description in #2103

osse101 (dismissed):

PR #2109, opened 2025-07-05
Refine Seaweed Room shinecharges

blkerby:
- Bugfix: the 3->1 strats were missing a Wave requirement.
- Add strats based on Sam's method to carry a shinecharge from bottom-right to bottom-left without Wave.
- Likewise, add strats to carry shinecharge from bottom-left to bottom-right without Wave.
- Refine/tighten shinecharge frames throughout (and add videos to document them).
- Include comeInShinecharging+leaveShinecharged variants that were skipped before.

### DIFF
--- a/region/tourian/main/Seaweed Room.json
+++ b/region/tourian/main/Seaweed Room.json
@@ -711,7 +711,7 @@
     },
     {
       "link": [3, 1],
-      "name": "Come in Shinecharging, Leave Shinecharged",
+      "name": "Come in Shinecharging, Leave Shinecharged (Wave)",
       "entranceCondition": {
         "comeInShinecharging": {
           "length": 0,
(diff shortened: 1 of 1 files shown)


region/tourian/main/Seaweed Room.json
```diff
@@ -680,18 +772,27 @@
...
-        "HiJump",
-        "canWalljump",
-        "canShinechargeMovementComplex",
-        {"shinespark": {"frames": 5}}
+        "Wave",
+        {"or": [
+          {"and": [
+            "HiJump",
+            "canWalljump",
+            {"shineChargeFrames": 100},
```
osse101, line 786: I can't get 100 frames with hijump+walljump.  And my best Hijump only is 1 frame slower than my best hi+wall, but hijump only is a good deal harder to do fast.

osse101 (commented):
There is a [3, 1] come in shinecharging, leave sparking without wave.
You shoot 4 upshots and then follow them up.
With just walljump its barely possible.  Hijump makes it more reasonable.

region/tourian/main/Seaweed Room.json
blkerby: Here's a video showing getting 93 frames with HiJump+walljump: https://videos.maprando.com/video/6938. My best is 91.

blkerby:
> There is a [3, 1] come in shinecharging, leave sparking without wave. You shoot 4 upshots and then follow them up. With just walljump its barely possible. Hijump makes it more reasonable.

Nice find. Added strats for that now. With Hi-Jump it's possible to leave with frames remaining. Bootless I think sparking out bottom position fits in Extreme, top position feels Insane.

region/tourian/main/Seaweed Room.json
osse101: That may be the problem.  I was doing "Carry Shinecharge", above this.

osse101 (approved):

PR #2110, opened 2025-07-06
More WS ETank speedy jump strats

blkerby:


region/wreckedship/main/Wrecked Ship Energy Tank Room.json
```diff
@@ -370,7 +370,118 @@
...
+          "speedBooster": true,
+          "minTiles": 23
+        }
+      },
+      "requires": [
+        "canTrickyJump",
+        "canMomentumConservingMorph",
+        {"or": [
+          "canInsaneJump",
+          {"thornHits": 5}
```
osse101, line 435: What does 5 thorn hits mean?

region/wreckedship/main/Wrecked Ship Energy Tank Room.json
```diff
@@ -370,7 +370,118 @@
...
+        "Using speed from the adjacent room, perform a ceiling mockball and equip or unequip Spring Ball to make it across the gap."
+      ]
+    },
+    {
+      "link": [1, 2],
+      "name": "Spring Ball Bounce Spring Fling (Power Off)",
+      "entranceCondition": {
+        "comeInWithMockball": {
+          "speedBooster": true,
+          "remoteAndLandingMinTiles": [[19, 0]]
```
osse101, line 472: Does this include this rooms runway?  Because the slope is hard to land on.

region/wreckedship/main/Wrecked Ship Energy Tank Room.json
osse101: - An airball version should be added, without `canMomentumConservingMorph`, with about 3 more tiles of runway.
- A full speed jump version, with no items but speed, seems like it should be added as well.  You have to time a pose change somewhat well to make it.
- A dev note should be added to explain `thornHits` are listed for leniency.

UN farm looked approximately frame perfect on the jump.  If you rush everything you can make it back to the farm on no tanks, which is nice for retries and escaping with the item, and good that expert gets an extra etank.
I don't know how many runways have 23 tiles but not 26 for an easier jump.

PR #2111, opened 2025-07-07
Wrecked Ship G-Mode: Second Pass

kjbranch:


region/wreckedship/main/Bowling Alley.json
```diff
@@ -800,6 +820,112 @@
...
+        }
+      },
+      "requires": [
+        "f_DefeatedPhantoon",
+        "Morph",
+        {"or": [
+          "SpaceJump",
+          {"and": [
+            "canUseIFrames",
+            {"spikeHits": 3}
```
osse101, line 839: Damageboosting into the ground makes 2 spike hits fully reasonable at ~expert.
You can do it without the damageboost but that has only a small number of frames lenience.

region/wreckedship/main/Bowling Alley.json
```diff
@@ -1399,6 +1525,40 @@
...
+        ]}
+      ],
+      "flashSuitChecked": true,
+      "note": "Quickly unmorph and jump to the left edge to take as few spike hits as possible."
+    },
+    {
+      "link": [7, 3],
+      "name": "G-Mode, Bowling Then Go Down",
+      "requires": [
+        {"spikeHits": 1},
```
osse101, line 1546: You can place a bomb before getting grabbed to bounce upward after the release, avoiding the first spike hit.

region/wreckedship/main/Sponge Bath.json
```diff
@@ -401,6 +401,25 @@
...
+      "link": [1, 2],
+      "name": "G-Mode Morph Spring Ball IBJ",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "any",
+          "morphed": true
+        }
+      },
+      "requires": [
+        "h_artificialMorphJumpIntoIBJ"
```
osse101, line 414:
```suggestion
        "h_artificialMorphJumpIntoIBJ",
        {"tech": "canBombJumpWaterEscape"}
```

region/wreckedship/main/Bowling Alley.json
kjbranch: Very cool, I never would have guessed.

I added a horizontal boost or insane jump to the normal 2-spike strat, it is still in VH, but we could make it a notable if you think it should be moved to expert.

PR #2112, opened 2025-07-07
Bootless two-wall UWJ

### LINKED ISSUE #711: Add underwater wall-jumping tech and strats

kjbranch:
This addresses Bobbob's Level 5 underwater wall jumps, described in #711. Level 5 looks to be similar or easier to level 4, but more annoying. It also has a small number of uses. This should be enough to close #711, as the remaining levels are unreasonable to implement or expect.

I'm expecting tech to be in `Beyond` and the Crab Shaft notable to be `Ignored`.

### DIFF
--- a/region/maridia/inner-pink/Aqueduct.json
+++ b/region/maridia/inner-pink/Aqueduct.json
@@ -1638,6 +1638,35 @@
       "devNote": "This does not include canRiskPermanentLossOfAccess, as it is only worth doing this strat if the items are there."
     },
     {
+      "link": [2, 7],
+      "name": "G-Mode Overload PLMs - Power Bomb the Items, Bootless UWJ",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "direct",
+          "morphed": false
+        }
+      },
+      "requires": [
+        {"notable": "G-Mode Overload PLMs - Power Bomb the Items"},
+        {"or": [
+          {"itemNotCollectedAtNode": 7},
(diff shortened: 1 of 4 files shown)


region/maridia/inner-pink/Aqueduct.json
```diff
@@ -2502,6 +2531,83 @@
...
+        "Climb one snail to place the Power Bomb, then another to reduce the distance of the wall jump climb."
+      ],
+      "devNote": "This does not include canRiskPermanentLossOfAccess, as it is only worth doing this strat if the items are there."
+    },
+    {
+      "link": [5, 7],
+      "name": "G-Mode Morph Overload PLMs - Bomb the Speed Blocks, Bootless UWJ",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "direct",
```
osse101, line 2598: Why is this special, and direct?

kjbranch: This one is separate because its artificial morphed from the right door - if it went down to the G-mode morph junction, it would need Morph, HiJump, or Gravity to snail climb back up.

Fixed the issue with direct, and removed a 10->7 strat, as it could just go 10->11->7 with those same items

PR #2124, opened 2025-07-12
Refine Volcano Room

blkerby:
- Handle heated runway at top door
- Handle locked doors at bottom
- Tighten lava frames for top-to-bottom suitless strat
- Remove stray `heatFrames` on the bottom leaveWithRunway strat.

### DIFF
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -242,6 +242,22 @@ __Example:__
 }}
 ```
 
+#### lavaFramesWithEnergyDrops object
+
+A `lavaFramesWithEnergyDrops` object represents the need for Samus to spend time in lava, but with the possibility of offsetting some of the lava damage using energy drops from enemies. Any lava damage is logically applied before the energy gain, so Samus must be able to survive the heat before picking up the drops. Any energy gain is logically capped to not exceed the heat damage, so this logical requirement cannot result in a net energy gain.
+
+The actual amount of energy gained typically depends on RNG and also on which ammo types are completely full. The drop probabilities for each enemy type is given in the [enemies](enemies/main.json) file. Because of the randomness involved, the logical amount of energy gain is open to various interpretations. For example, the mean, the median, or a lower confidence limit could be used.
+
+__Example:__
+```json
+{"lavaFramesWithEnergyDrops": {
+  "frames": 200,
+  "drops": [
+    {"enemy": "Fune", "count": 1}
+  ]
+}}
+```
+
 #### gravitylessHeatFrames object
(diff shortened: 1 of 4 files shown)


region/norfair/east/Volcano Room.json
```diff
@@ -160,17 +175,25 @@
...
           ]},
           {"lavaFrames": 150}
         ]}
+      ],
+      "exitCondition": {
+        "leaveNormally": {}
+      },
+      "unlocksDoors": [
+        {"types": ["missiles"], "requires": [{"lavaFrames": 80}]},
+        {"types": ["super"], "requires": []},
```
osse101, line 195: This added lava frames when spacejumping.  The fastest way I saw with a blue door was to do an aim down shot to open the door.

region/norfair/east/Volcano Room.json
```diff
@@ -238,7 +269,24 @@
           {"heatFrames": 35}
         ]},
         {"heatFrames": 405}
```
osse101, line 271: SpaceJump + Come in Running + Mockball can be done without an etank.

region/norfair/east/Volcano Room.json
```diff
@@ -147,6 +154,14 @@
             {"lavaFrames": 60}
           ]}
         ]}
+      ],
+      "exitCondition": {
+        "leaveNormally": {}
```
osse101, line 159: FIXME: LeaveWithRunway using gravity and no Speedbooster.

osse101 (commented):
https://videos.maprando.com/video/7025
Without the PB I was able to make it out after 94 damage.  Avoiding damage while space jumping is key.

Most of my comments aren't about the PR.  PR changes look fine.  Lava Frames no gravity  don't scale appropriately with heat multiplier.

logicalRequirements.md
```diff
@@ -242,6 +242,22 @@ __Example:__
 }}
 ```
 
+#### lavaFramesWithEnergyDrops object
+
+A `lavaFramesWithEnergyDrops` object represents the need for Samus to spend time in lava, but with the possibility of offsetting some of the lava damage using energy drops from enemies. Any lava damage is logically applied before the energy gain, so Samus must be able to survive the heat before picking up the drops. Any energy gain is logically capped to not exceed the heat damage, so this logical requirement cannot result in a net energy gain.
```
osse101, line 247: > Heat damage

Every room except this one has heat+lava damage simultaneously.  In SMZ3, Gravity grants full protection from lava and Varia full protection from heat.  So I think the usage would be running both lavaFramesWithEnergyDrops and HeatFrameWithEnergyDrops in parallel so the randomizer picks which is better for the situation.

PR #2126, opened 2025-07-13
Below Spazer CWJ jump shot speedball

blkerby:
This is based on a recent Bobbob video:

https://www.youtube.com/watch?v=_ynzaEpkBEs

I tested it out and found there's a decent range of speeds that work.


osse101 (commented):
This is so bad.  I have to learn the approximate speed and height for each transition.  And then its still a small frame window jump.
- I'd add canBePatient
- Metal Pirates was much harder than Metroids 1
- With enough speed you can skip the CWJ and directly jump for the Speedball

PR #2127, opened 2025-07-13
Pink Brinstar farms

blkerby:
Aside from the farm migration, this PR reworks the "Doorway Sidehopper Kill" strat in Mission Impossible Room and turns it into a notable strat. This is moving it up from Very Hard to Expert as I found it was trickier to do reliably than I had expected; the trick I found was to delay shooting the Hopper until after it hops 4 times, which is not very intuitive. For the doorway strat, I don't think additional beams or Morph really affect the difficulty, as the main thing is knowing how to manipulate the Hopper RNG to avoid getting hit in the doorway. And of course, if you have strong enough weapons then it's already covered by other strats that don't involve getting in the doorway.

### DIFF
--- a/region/brinstar/pink/Waterway Energy Tank Room.json
+++ b/region/brinstar/pink/Waterway Energy Tank Room.json
@@ -178,11 +178,11 @@
                   "Spazer",
                   "Plasma"
                 ]},
-                {"cycleFrames": 360}    
+                {"cycleFrames": 340}
               ]},
               {"and": [
                 "ScrewAttack",
-                {"cycleFrames": 385}  
+                {"cycleFrames": 370}
               ]},
               {"and": [
                 "Grapple",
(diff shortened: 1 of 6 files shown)


region/brinstar/pink/Waterway Energy Tank Room.json
```diff
@@ -167,19 +167,75 @@
...
+                "Grapple",
+                {"cycleFrames": 400}
+              ]},
+              {"cycleFrames": 530}
+            ]}
           ]},
-          {"partialRefill": {"type": "PowerBomb", "limit": 6}}
+          {"and": [
+            {"or": [
+              "Wave",
```
osse101, line 196:
```suggestion
            {"or": [
              "Charge",
              "Wave",
```

blkerby:
I'm curious how you got your 780 frames with Power Beam suitless? I was trying for a while with running past the first Zero's drop but it always despawned before I could collect it on the way back. Eventually I found you could fire two horizontal shots by jumping like this and then collect the drop while going left: https://videos.maprando.com/video/7046 . But that's more like 710 frames so I'm guessing it's not what you had in mind?

PR #2131, opened 2025-07-16
Warehouse Brinstar farms

blkerby:
For Warehouse Kihunter Room, I did separate farm strats for each of the three doors where you can reset the room, because the way you interact with the Kihunters differs so much based on which direction you're coming from.

### DIFF
--- a/region/brinstar/kraid/Warehouse Kihunter Room.json
+++ b/region/brinstar/kraid/Warehouse Kihunter Room.json
@@ -248,6 +248,50 @@
       ]
     },
     {
+      "link": [1, 1],
+      "name": "Kihunter Farm",
+      "requires": [
+        {"resetRoom": {"nodes": [1]}},
+        {"or": [
+          {"and": [
+            "ScrewAttack",
+            {"cycleFrames": 360}
+          ]},
+          {"and": [
+            "Plasma",
+            {"cycleFrames": 360}
+          ]},
+          {"and": [
+            "Wave",
+            {"cycleFrames": 390}
+          ]},
+          {"and": [
+            "Spazer",
+            {"cycleFrames": 390}
+          ]},
+          {"and": [
+            "Ice",
+            "canDodgeWhileShooting",
+            {"cycleFrames": 600}
+          ]},
+          {"and": [
+            "canPseudoScrew",
+            "canTrickyDodgeEnemies",
+            {"cycleFrames": 600}
+          ]},
+          {"and": [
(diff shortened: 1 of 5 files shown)


region/brinstar/kraid/Warehouse Kihunter Room.json
```diff
@@ -373,6 +417,52 @@
...
+          ]},
+          {"and": [
+            "Charge",
+            {"cycleFrames": 780}
+          ]},
+          {"and": [
+            "Ice",
+            {"cycleFrames": 840}
+          ]},
+          {"cycleFrames": 1200}
```
osse101, line 457: - I Think these kill from below strats want canPrepareForNextRoom to manip the first kihunter to not fall right on your head.
- Collecting all of the drops with a slow killing weapon is hard at best.
- canCameraManip helps with hitting the kihunters for some of these

region/brinstar/kraid/Warehouse Kihunter Room.json
```diff
@@ -481,6 +571,49 @@
         "Then continue chaining temporary blue through the door."
       ]
     },
+    {
+      "link": [3, 3],
+      "name": "Kihunter Farm",
+      "requires": [
+        {"resetRoom": {"nodes": [1]}},
```
osse101, line 578:
```suggestion
        {"resetRoom": {"nodes": [3]}},
```

region/brinstar/kraid/Warehouse Kihunter Room.json
blkerby: I didn't have any trouble with the first Kihunter falling down, probably because I wasn't shooting out the shot block fast enough for it to happen. Testing it again, I can see that if you rush to shoot the block as fast as possible, then it does happen. I'm not sure "canPrepareForNextRoom" is needed, since if the player has fast enough movement for it to be a problem, then they can probably be expected to already know or be able to figure out how to avoid it. Added a note about it.

For collecting the drops, I try to spread out the damage rather than focusing on one Kihunter at a time. This way they can all be killed at closer to the same time. It only really seemed to be an issue with Power Beam, added a stronger tech requirement now for this case (canTrickyDodgeEnemies) since it does seem a lot harder to avoid damage while feeling pressure to collect the drops.

I wasn't sure what canCameraManip would mean here?

osse101: I was planting and shooting up through the hole.  Jumping and shooting works better than I thought.

PR #2132, opened 2025-07-16
simpleHeatFrames and simpleCycleFrames

blkerby:
This PR just adds the schema, docs, and a few examples, mostly `simpleHeatFrames` in Speed Booster Hall.

Been wanting to add this for a while. The immediate motivation is that when we get to converting heated farms (e.g. Gamets) into the farm cycle schema, we're going to need this `simpleHeatFrames`, because otherwise the heat multiplier would mess things up too much with being able to get a net positive energy gain.

### DIFF
--- a/region/norfair/east/Speed Booster Hall.json
+++ b/region/norfair/east/Speed Booster Hall.json
@@ -237,6 +237,26 @@
       "clearsObstacles": ["A"]
     },
     {
+      "id": 8,
+      "link": [1, 2],
+      "name": "Speed Run, Leave Shinecharged",
+      "requires": [
+        {"obstaclesNotCleared": ["A"]},
+        "h_shinechargeMaxRunway",
+        {"simpleHeatFrames": 310},
+        {"heatFrames": 40},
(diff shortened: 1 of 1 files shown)


osse101 (commented):
I don't agree with removing `Speed Run and Leave Shinecharged` because leave with runway is a tier higher in difficulty.  The hero shot is probably very hard, but feels harder because it doesn't come up often.  On the other hand storing the shinespark without falling through the crumbles is harder than it seems as well.

blkerby:
> I don't agree with removing `Speed Run and Leave Shinecharged` because leave with runway is a tier higher in difficulty. The hero shot is probably very hard, but feels harder because it doesn't come up often. On the other hand storing the shinespark without falling through the crumbles is harder than it seems as well.

Makes sense, I didn't see the purpose of it and assumed it was just a holdover from an older schema or something. It's true that the leaveShinecharged doesn't require hero shot to open the door, so yeah it can make sense to keep it as a separate strat.

PR #2133, opened 2025-07-17
Aqueduct G-Mode: Second Pass

kjbranch:


osse101 (commented):
13 can freely link to 10
10 can freely link to 12
Everything should already be covered, but they are noticibly similar GMode junctions.

I call the damageless snail moonfalls freefall clips because its not a normalized timing.  These are pretty lenient so with practice it may be extreme level like you had it.

kjbranch:
I'm not sure if 13->10, 10->12 will allow any more strats, but it is nice to add for clarity and in case more strats are found and added in the future. 

I think the freefall clip sounds nice. I hadn't ever worked with that tech before, but it looks like exactly the place for it. The snails are much easier to position and there is no precise freeze timing component, so it is likely easier than other moonfalls, so I would also feel fine if it does get lowered back to Extreme.

PR #2135, opened 2025-07-18
Add canLongXRayClimb tech

blkerby:
The new tech `canLongXRayClimb` distinguishes climbs of length ~2 screens or more. For Map Rando this will help us to put only ~1 screen climbs into Very Hard logic. The idea is that when first learning to X-Ray climb it is nicer to only have to start out with relatively short ones.

For some grapple teleport X-Ray climb strats, this required splitting off a second strat for the 2-screen variant. This is helpful anyway because now we can make the note more specific, telling the player how far to climb.

In Green Hill Zone, there was one case where the 2-screen grapple teleport X-ray climb variant was missing, so that is added here.

I also took the opportunity to clean up some of the older strat names, removing "Left-Side"/"Right-Side" since that information is redundant with the node that you're entering through.

### DIFF
--- a/region/crateria/central/Parlor and Alcatraz.json
+++ b/region/crateria/central/Parlor and Alcatraz.json
@@ -613,11 +613,11 @@
         }
       },
       "requires": [
-        "canXRayClimb"
+        "canLongXRayClimb"
       ],
       "note": [
         "Fall into the first Samus Eater in Hellway or Alpha Power Bomb Room.",
-        "After teleporting and passing through the transition, X-Ray climb to reach the space behind the bomb blocks at the top of the room.",
+        "After teleporting and passing through the transition, X-Ray climb 2 screens to reach the space behind the bomb blocks at the top of the room.",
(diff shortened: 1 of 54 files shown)


region/crateria/central/Parlor and Alcatraz.json
```diff
@@ -1236,10 +1236,10 @@
     {
       "id": 42,
       "link": [4, 1],
-      "name": "Grapple Teleport X-Ray Climb",
+      "name": "Grapple Teleport X-Ray Climb (1 Screen)",
       "entranceCondition": {
         "comeInWithGrappleTeleport": {
-          "blockPositions": [[2, 18], [2, 19], [2, 28], [2, 29], [2, 34]]
```
osse101, line 1242: Was [2 34] missed on 5 and 6 before?

osse101 (approved):

region/crateria/central/Parlor and Alcatraz.json
blkerby: Yeah, looks like it.

Issue #1788, opened 2024-11-22
Model spin-jump through doors


blkerby:
Some strats require spin jumping into the room. Examples include needing to wall jump immediately when entering a room, or needing to activate a shinespark mid-air. 

For horizontal doors, the `comeInJumping` entrance condition could be used to express this requirement. It's currently not used very consistently. This probably has minimal significance, since by spin-jumping immediately before the transition, there should be almost never any effect on the amount of frames spent in the previous room (e.g. heat damage or shinecharge frames) compared to running through the transition. Still, for clarity, it may be good to start being more consistent about it.

For vertical doors the difference could matter a little more. Quick drop can be used to fall more quickly through a vertical door, but this leaves Samus in a non-spinning state in the next room. As a result, some combinations are probably unsound currently. And currently we don't have any way to express a spin-jump entry through a vertical door. To be effective, we need a way to also express this in combination with a `comeInShinecharged` condition; so this is related to issue #1578.

PR #2147, opened 2025-07-21
Refine Crumble Shaft

blkerby:
- Many heat frames are tightened.
- More lenience is added to the wall-jump climbs on low difficulty (especially Medium)
- Some apparently non-useful strats are removed:
    - HiJump Climb: The instruction to wall jump directly off the top-right crumble block seemed questionable; using the floating crumble platform is much easier and also seems faster. I combined this with the bootless version since there doesn't seem to be a reason for it to be a separate strat.
    - WallJump and SpringBall: The problem is the Sova gets in your way where you want to bounce with Spring Ball. You can wait for it with canStaggeredWalljump, or bounce in a very precise place on the edge of the crumbles; but these both seem more difficult than just wall jumping the regular way off the side of the crumble block.
    - Frozen Sova Climb: With the left-side climb, a Sova gets directly in your way (assuming a bottom-door entry), so you either have to staggered wall jump to wait for it, or carefully wall jump around it. The whole thing seems significantly more difficult than just walljumping up the right side the regular way.

Wall jumping up either side of the room is quite tricky for Medium: I think it can be ok but we have to allow for the fact that the player is expected to fall down many times, either from missing the precise wall jumps or from Sovas getting in the way. So that's the reason for adding the extra lenience heat frames in the absence of `canDodgeWhileShooting` tech. And those will stack on top of the heat damage multiplier, which I think is necessary; at that level, the player is going to want to have a ton of tanks here.

### DIFF
--- a/region/norfair/west/Crumble Shaft.json
+++ b/region/norfair/west/Crumble Shaft.json
@@ -214,7 +214,14 @@
       "link": [1, 2],
       "name": "Base",
       "requires": [
-        {"heatFrames": 200}
+        {"simpleHeatFrames": 150},
+        {"or": [
+          {"and": [
+            "canMoonfall",
+            {"heatFrames": 20}
+          ]},
+          {"heatFrames": 45}
+        ]}
       ]
     },
     {
@@ -396,24 +403,27 @@
       "note": "Shoot the shot block, then Spring Ball on the middle platform."
     },
     {
-      "id": 11,
-      "link": [2, 1],
-      "name": "HiJump Climb",
-      "requires": [
-        "HiJump",
-        "canConsecutiveWalljump",
-        {"heatFrames": 500}
-      ],
-      "note": "Walljump up the right, then setup a walljump on the top-right crumble platform directly from the right wall."
-    },
-    {
       "id": 12,
       "link": [2, 1],
-      "name": "HiJumpless Climb",
+      "name": "Wall Jump Climb",
       "requires": [
         "canPreciseWalljump",
         "canConsecutiveWalljump",
-        {"heatFrames": 500}
+        {"or": [
+          {"and": [
+            "HiJump",
+            {"heatFrames": 270}
+          ]},
+          {"heatFrames": 340}
+        ]},
+        {"or": [
+          "canDodgeWhileShooting",
+          {"heatFrames": 300}
+        ]},
+        {"or": [
+          "canTrickyJump",
+          {"heatFrames": 300}
+        ]}
       ],
       "note": "Walljump up the right, then walljump off the top-middle crumble platform then off the top-right one."
     },
@@ -424,21 +434,20 @@
       "requires": [
         "canCrumbleJump",
         "canTrickyJump",
-        {"heatFrames": 800}
+        {"or": [
+          {"and": [
+            "HiJump",
+            "canTrickyDodgeEnemies",
+            {"heatFrames": 370}
+          ]},
+          {"heatFrames": 660}
+        ]}
       ],
-      "note": "Do 9 successive crumble jumps up the platforms.",
-      "devNote": "This is only really useful without wall jumps."
-    },
-    {
-      "id": 14,
-      "link": [2, 1],
-      "name": "Wall Jump Climb Crumble Jump",
-      "requires": [
-        "canCrumbleJump",
-        "canConsecutiveWalljump",
-        {"heatFrames": 600}
-      ],
-      "note": "Walljump up the right then do one crumble jump on the top middle platform."
+      "wallJumpAvoid": true,
+      "note": [
+        "Do 9 successive crumble jumps up the platforms.",
+        "If Hi-Jump is available, then it is possible to skip 4 of the crumble platforms."
+      ]
     },
     {
       "id": 15,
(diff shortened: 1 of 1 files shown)


region/norfair/west/Crumble Shaft.json
```diff
@@ -214,7 +214,14 @@
       "link": [1, 2],
       "name": "Base",
       "requires": [
-        {"heatFrames": 200}
+        {"simpleHeatFrames": 150},
+        {"or": [
+          {"and": [
+            "canMoonfall",
```
osse101, line 220: - The moonfall is hard to pull off most of the time because of the crumble block, so not quite a basic moonfall.
- Landing on the doorframe while moonfalling takes some moonfall delaying actions that aren't really part of basic moonfall.

Covered by heat scaling though.

region/norfair/west/Crumble Shaft.json
```diff
@@ -395,25 +402,28 @@
...
       "requires": [
         "canPreciseWalljump",
         "canConsecutiveWalljump",
-        {"heatFrames": 500}
+        {"or": [
+          {"and": [
+            "HiJump",
+            {"heatFrames": 270}
+          ]},
+          {"heatFrames": 340}
```
osse101, line 417: I can't get under 6 seconds unless I enter with a spinjump

region/norfair/west/Crumble Shaft.json
```diff
@@ -424,29 +434,28 @@
...
-      "name": "Wall Jump Climb Crumble Jump",
-      "requires": [
-        "canCrumbleJump",
-        "canConsecutiveWalljump",
-        {"heatFrames": 600}
+        {"or": [
+          {"and": [
+            "HiJump",
+            "canTrickyDodgeEnemies",
+            {"heatFrames": 370}
```
osse101, line 441: For this time I had to enter with a spinjump

region/norfair/west/Crumble Shaft.json
```diff
@@ -424,29 +434,28 @@
...
+          ]},
+          {"heatFrames": 660}
```
osse101, line 443: I think 9 successive crumble jumps should have canInsaneJump.
Getting a 630 time took some aggressive sova dodging.  Did not enter with spinjump

region/norfair/west/Crumble Shaft.json
blkerby: I was entering with a spinjump from a water room to simulate the worst-case scenario. We don't have an entrance condition to specifically represent this (issue https://github.com/vg-json-data/sm-json-data/issues/1788) though I know we sometimes use `canPrepareForNextRoom`. In this case, since it only saves a few heat frames, which can be accounted for in the multiplier, I was thinking it's better not to clutter up the requirements. Added a mention in the `notes` now though, because it's useful to document the expectations to be able to reproduce the heat frames.

region/norfair/west/Crumble Shaft.json
blkerby: https://videos.maprando.com/video/7143

region/norfair/west/Crumble Shaft.json
blkerby: For 660 I was killing some of the Sovas as I went. Trying it a little more, with aggressive dodging 612 was my best. Lowering it to 630 seems reasonable for Insane.

And yeah, 9 crumble jumps in a row does seem like too much for Very Hard. I think it can be ok in Expert though. At the Expert level, with the heat frame lenience (1.15 * 630 = 725 frames) you can almost just do full-height jumps, and dodging the Sovas is not bad in this case. 

It took me a while to get comfortable with the crumble jump window, but it's feeling pretty reliable. I find whenever I fail a crumble jump it's because I'm pressing jump too late. The 8-frame window is so large that it feels almost impossible to press it too early. I'm aiming now to press jump when Samus' feet are about 1.5 tiles above the block, which is much higher than what I used to aim for; of course it would depend on fall speed and on the amount of input delay (I'm playing on emulator so I'm sure I have more delay than some).

region/norfair/west/Crumble Shaft.json
blkerby: Oh yeah, Ice+HiJump is a good one, added that now.

The crumble jumping is interesting. After playing with it more, I kind of find it safer now than the wall jump strat. With the wall jumps, I sometimes miss getting into one of the 1-tile spaces or accidentally break spin doing the wiggle at the top. The crumble jumps are simpler and there's less that can go wrong. If I fall it is usually recoverable by crumble jumping off the platform below; I don't think freezing enemies would make it easier as it kind of just complicates things. I'm doing regular crumble jumps here though. Crumble spin jumping like in your video seems crazy difficult to me; I tried for a while but couldn't even make it to the third platform that way, haha. Those are 2-frame windows for each jump in that case, compared to 8-frame window for regular crumble jumps.

Video of doing a bunch of attempts of the crumble jump vs. wall jumps: https://videos.maprando.com/video/7200 (not a fair comparison because I was practicing the crumble jumps more today :shrug: )

I don't actually think the crumble jump should be placed lower than the wall jumps, but Extreme seems too high I think. I'm not really sure what to do with the crumble spin jump numbers because that seems so difficult I'm not sure what the requirements would look like.

osse101: Extreme has to be too high.  
The heat frame counts are complicated.  Reckless movement saves seconds and falling doubles your damage taken.

Issue #2149, opened 2025-07-21
Bobbob's Feedback for Crateria, Blue Brin


kjbranch:
Green Pirates Shaft:
- [ ] 3 to 2 and 4 to 2:  You can X-Ray climb to avoid taking damage. i think you have it written that this still takes projectile damage...so not sure how notable this is.
- [ ] 9 to 1:  The ice x-ray climb requires morph, which i disagree with. heck, with tunnel crawl and turn-around aim cancel, getting the pixel should be fairly free. so wrap that in an 'or' with morph, and be like "your own fault for enabling tunnel crawl, this is what you get". i'm only mentioning this, since i like as many morphless strats as possible to be in:  i find very late morph to be interesting.
- [x] 9 to 2:  ice moonfall. this has since been added. there are many ways to do it, and it is fairly free (especially since the beetoms don't die to ice so you can re-freeze one of them while falling).
- [x] 2,3,4:  G-mode setup, by bringing a Beetom down. Some notes on this:  Getting to the Beetoms requires Morph+PB, X-Ray Climb, or starting from 1; getting back down needs Morph, although you can probably bring a Beetom down with you with an ice moondance or ice moonfall; Beetoms don't provide knockback, so this always puts you in g-mode immobile; this needs some number of Beetom hits (and maybe Green Pirate hits if going to the bottom) to setup; you could X-Ray climb up to 1 with a Beetom attached, but this would be obnoxious with the Beetom damaging you as you go up and also you shaking off the Beetom sometimes...not even really sure how functional that would be, and way too lazy to test.

Statues Room:
- [x] A bomb double spring ball jump works (so hjb-less), as done in my 14% xSpring run. Not sure how often this is in the logic, so just a reminder this needs morph+bombs+spring:  a power bomb doesn't work, since you can't pause with a power bomb on screen.

Gauntlet Entrance:
- [x] devnote mentions a blue spring ball strat. There would also be temp blue strats. I am not sure how notable these really are, with the shinespark strats...they are notable in vanilla, where there is a wall in the next room, but that isn't really modeled yet? for completion, tho, note this strat in vanilla (this also shows how to handle the yapping maw, to get the full middle runway without killing/freezing it):  https://vimeo.com/535363869
- [ ] (The above is complete, but we also should check the leave with spark frames in the above video and tighten them.)

Gauntlet Energy Tank Room:
[...]

PR #2151, opened 2025-07-22
Moat CWJ bomb boost

blkerby:
This was one of the items on Bobbob's list from a couple years ago (https://github.com/vg-json-data/sm-json-data/issues/2149).

I'm thinking this is probably an Extreme notable, though maybe it could be Insane.

This bomb boost strat can be useful if the door on the left is locked since the usual CWJ strat won't work in that case.

Video: https://videos.maprando.com/video/7142

### DIFF
--- a/region/crateria/east/The Moat.json
+++ b/region/crateria/east/The Moat.json
@@ -1106,6 +1106,28 @@
       ]
     },
     {
+      "link": [3, 2],
+      "name": "CWJ Bomb Boost",
+      "requires": [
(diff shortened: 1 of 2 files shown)


osse101:
This is one I play with frequently to showboat.  I would say it feels Extreme.  There's the softlock risk but also you can probably bomb your way back out to the left.  It is faster than HBJ so if it were consistent enough for Expert, it would be the backup strat for crossing moat in speedruns.

PR #2153, opened 2025-07-22
Fish Tank and Main Street farms

blkerby:


region/maridia/outer/Main Street.json
```diff
@@ -3920,6 +4137,79 @@
...
+          {"and": [
+            "Spazer",
+            {"cycleFrames": 690}
+          ]},
+          {"and": [
+            "Ice",
+            {"cycleFrames": 710}
+          ]},
+          {"and": [
+            "Charge",
```
osse101, line 4203: I get sub 12 without charge.  Doing two shots per jump makes it easier than with gravity.
With charge I can kill the crab too fast and have to wait for the drop to move itself closer so there is no time gain.

With gravity, charge saved me 30f

region/maridia/outer/Main Street.json
```diff
@@ -3920,6 +4137,79 @@
...
+      "name": "Sciser Farm",
+      "requires": [
+        "Morph",
+        {"resetRoom": {"nodes": [5]}},
+        {"or": [
+          {"and": [
+            "Gravity",
+            {"or": [
+              {"and": [
+                "ScrewAttack",
```
osse101, line 4151: springball speeds this up by 100f
And I actually see 450f on screw and pseudoscrew

And grapple is missing from these lists

PR #2159, opened 2025-07-23
Refine Moat temp blue

blkerby:
More finds from Sam's videos:
- For the Gravity jump variant, it can be done relatively easily without any jump assist (HiJump/SpringBall).
- Blue spring ball bounce L->R is a new interesting option.

### DIFF
--- a/region/crateria/east/The Moat.json
+++ b/region/crateria/east/The Moat.json
@@ -568,7 +568,8 @@
       },
       "requires": [
         "canTrickySpringBallBounce",
-        "canChainTemporaryBlue"
+        "canChainTemporaryBlue",
+        "canSpeedball"
       ],
       "exitCondition": {
         "leaveWithTemporaryBlue": {}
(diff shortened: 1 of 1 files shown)


region/crateria/east/The Moat.json
```diff
@@ -563,6 +555,34 @@
...
+        "comeInGettingBlueSpeed": {
+          "length": 2,
+          "openEnd": 0,
+          "steepDownTiles": 1,
+          "minExtraRunSpeed": "$0.8"
+        }
+      },
+      "requires": [
+        "canTrickySpringBallBounce",
+        "canChainTemporaryBlue"
```
osse101, line 571:
```suggestion
        "canChainTemporaryBlue",
        "canSpeedball"
```

region/crateria/east/The Moat.json
```diff
@@ -563,6 +555,34 @@
...
+      ],
+      "exitCondition": {
+        "leaveWithTemporaryBlue": {}
+      },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
+      "collectsItems": [3],
+      "flashSuitChecked": true,
+      "note": [
+        "Enter with blue speed and bounce on the item pedestal to make it across, then unmorph into temporary blue."
```
osse101, line 580:
```suggestion
        "Enter with blue speed and bounce on the item pedestal to make it across, then unmorph into temporary blue.",
        "Morph on the left side of the pedestal so that there is enough space to use a controlled SpringBall bounce."
```

osse101 (commented):
I found the springball bounce easier than chaining right to left as well.

PR #2160, opened 2025-07-24
Add dodge option for Pants Room grapple jump

blkerby:
Avoiding the Menus doesn't seem bad, by just grapple jumping up all the way to the top.

Video by Sam: https://videos.maprando.com/video/6717


osse101:
Confirmed.  And I like this approach to the grapple swings more than how I was doing it.

PR #2163, opened 2025-07-24
LN G-Mode: Second Pass

kjbranch:


region/lowernorfair/east/Ridley's Room.json
```diff
@@ -622,7 +622,7 @@
           "canWalljump",
           {"and": [
             "canJumpIntoIBJ",
-            "canTrickyJump",
+            "canInsaneJump",
```
osse101, line 625: I can't avoid ridley without double IBJ or HiJump

PR #2169, opened 2025-07-26
Indiana Jones G-Mode Restructure

kjbranch:
This was done to clean up the strats and make it more clear what to do. There is a lot going on still, so maybe it would be worth adding a junction node back, but it is not immediately clear how to best do that.

Also worth noting that the some of the old artificial morph strats required "canIBJ" instead of its morphless variant.


osse101 (approved):
A GMode node looks useful but I also don't see it saving lines of code.   There are the traversal nodes that come together to IBJ to the item then fall before branching again.  And there are `Remote Acquire And` strats that would continue on to PB the blocks?
So I don't see groupings being large enough to be of much help.

PR #2182, opened 2025-07-29
Downgrade Plasma Spark precise crouch jump

blkerby:
This was supposed to be an Expert notable now, but that isn't working as intended because of the `canInsaneJump` requirement. So this downgrades it to a `canTrickyJump`.


kjbranch (dismissed):
I believe that is the same as the jump in Watering Hole. This one is currently in VH, should it be made an Expert notable as well?
https://dev.maprando.com/logic/room/181/2/1/15

PR #2186, opened 2025-07-31
Plasma Room damage-avoiding pseudo screw

blkerby:
I'm thinking Extreme for the lenient version of this (assuming 2 hits), Insane for killing all 4 Pirates damageless.

### DIFF
--- a/region/maridia/inner-yellow/Plasma Room.json
+++ b/region/maridia/inner-yellow/Plasma Room.json
@@ -357,7 +357,11 @@
                 "type": "contact",
                 "hits": 1
               }},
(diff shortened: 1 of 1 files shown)


blkerby:
Here's a video showing it damageless: https://videos.maprando.com/video/7301

I've seen people do this with catching the back of the Pirate hitbox while falling, but I find it easier to hit the Pirate from the front while spin jumping up through it.

osse101:
I find I like jumping up into the pirate more too.  Avoiding two hits happens regularly with the little practice i've done, and I got damageless once so it seems fair 😉

PR #2189, opened 2025-08-01
Pink Maridia farms part 2: Botwoon Hallway, etc.

blkerby:


region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -403,19 +403,13 @@
...
-          "Spazer",
-          "Wave",
-          "Plasma",
-          "canUseGrapple",
-          "canBePatient"
-        ]}
+        {"cycleFrames": 490}
+      ],
+      "farmCycleDrops": [
+        {"enemy": "Sciser", "count": 1}
```
osse101, line 412: This is a no item 2 farm.
No items + no tech isn't trivial to 1 farm either but maybe leniency makes that problem go away.

region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -1918,6 +1983,45 @@
...
+            "Spazer",
+            {"cycleFrames": 160}
+          ]},
+          {"and": [
+            "Grapple",
+            {"cycleFrames": 160}
+          ]}
+        ]},
+        {"or": [
+          "Grapple",
```
osse101, line 2011: I can't pull off a grapple kill and stay on the platform.

region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -1918,6 +1983,45 @@
...
+      "name": "Sciser Farm (1 Sciser, Fast Cycle)",
+      "requires": [
+        {"resetRoom": {"nodes": [3]}},
+        {"or": [
+          "canCrouchJump",
+          "h_underwaterCrouchJumpWithFlashSuit"
+        ]},
+        {"or": [
+          {"and": [
+            "Plasma",
```
osse101, line 1997: It feels rare to ever see the crab killed fast enough to not need to jump after it, so I rate this very hard.

region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -1918,6 +1983,45 @@
...
+            {"cycleFrames": 150}
+          ]},
+          {"and": [
+            "Wave",
```
osse101, line 2001: This one is expert, but since you are practicing it over and over it may be very hard, but higher than plasma

region/maridia/inner-pink/Crab Shaft.json
osse101: There aren't any instant kill beams listed, so I don't see grapple helping in any other case either.

region/maridia/inner-pink/Crab Shaft.json
osse101: ~~I don't think it works with gravity~~.  Maybe you'd get the 3 crab kills in that case anyway.
It's just unintuitive.  You can diagonal shot kill it with or without gravity.

region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -1918,6 +1983,45 @@
         "Knocking the crab off while it is on the the door or ceiling will not work."
       ]
     },
+    {
+      "link": [4, 4],
+      "name": "Sciser Farm (1 Sciser, Fast Cycle)",
+      "requires": [
+        {"resetRoom": {"nodes": [3]}},
+        {"or": [
+          "canCrouchJump",
```
osse101, line 1992: Being treated as nearly free, it looks like.

region/maridia/inner-pink/Crab Shaft.json
blkerby: > I can't pull off a grapple kill and stay on the platform.

I was doing it from a water entry above. It looks like for it to work from an air entry, it needs either a wall jump or Morph to reset fall speed. Since we don't yet have a way to represent a water entry, I can just add requirements for one of those two things.

region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -403,19 +403,14 @@
...
-        {"partialRefill": {"type": "PowerBomb", "limit": 6}},
-        {"or": [
-          "Gravity",
-          "Spazer",
-          "Wave",
-          "Plasma",
-          "canUseGrapple",
-          "canBePatient"
-        ]}
+        {"cycleFrames": 860},
```
osse101, line 409: Now that there are 2 crabs, at least grapple should be added as a farm weapon.  Jumping to the right loses a lot of time, which is why grapple is so strong.  And with Gravity you would use a different farm strat.

Issue #1911, opened 2025-01-23
Add grapple super sink tech


blkerby:
- Bypassing the bottom right door shell in Everest: https://discord.com/channels/1053421401354285129/1073621773184159764/1332057624795021393
- Also possible to get down to the Charge item in Big Pink

blkerby:
Getting stored fall speed in PCJR, clipping down many screens in the next room (Climb): https://discord.com/channels/1053421401354285129/1073621773184159764/1332436756200558703

PR #2192, opened 2025-08-03
Super sinks

blkerby:
This adds a tech for Grapple super sinks, along with some applications. So far there are just two places to set up the supersink:
- Mt. Everest: door lock bypass through either of the bottom doors
- Indiana Jones Room: exit condition to leave with a supersink through the left door, with many uses in other rooms.

This is a starting point; I'm sure there will be more later. So far I'm only considering the method with Spring Ball. I know there can be another way to do it with a soft unmorph (slopekiller) but I think that must be more precise?

I tried for a while to get the supersink in Big Pink (to clip down into the Charge Beam area or the bottom left door) but never even managed to release with enough fall speed to get a supersink. Sniq's video for it is here in case someone else wants to try to figure it out: https://www.youtube.com/watch?v=Tn-R9KQG0o0. I seem to remember someone mentioning that it was in a plando, so I guess it must be humanly viable and I'm just missing the trick to it. 

More cross-room setups may be possible too, and depending on what is found, it could happen that we need to adapt the schema to specify constraints on fall speed (and possibly allow crossing while not morphed), since this determines the frames remaining to position in the neighboring room. But for now it seemed like overkill since there's only the one cross-room setup.

### DIFF
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -4100,7 +4100,6 @@
       "name": "Super Sink Door Lock Skip",
       "requires": [
         "canSuperSink",
-        "Morph",
         "SpringBall"
       ],
       "bypassesDoorShell": true,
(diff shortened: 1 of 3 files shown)


kjbranch:
This partially addresses #1911

osse101 (commented):
Indiana Jones was harder to do, but is also more useful.

Everest felt ok for extreme so you could probably put both them both in insane.

PR #2193, opened 2025-08-04
Worst Room X-mode temporary blue & refinements

blkerby:
This adds a new notable for the X-mode strat that somerando(caauyjdp) recently shared.

While working on this, I noticed that the heat frames for descending the room (to node 4) were pretty lenient, and it looked like it mostly had to do with the assumption that you stop at node 5 along the way, and node 5 is defined kind of broadly (e.g. it can include being on top of the Namihe doing a CF) which makes those heat frames large. A fairly significant amount of heat frames can be saved by bypassing node 5 and going directly to node 4, weaving between the platforms, so I added strats for that.

Video: https://videos.maprando.com/video/7347

### DIFF
--- a/region/lowernorfair/east/The Worst Room In The Game.json
+++ b/region/lowernorfair/east/The Worst Room In The Game.json
@@ -153,6 +153,7 @@
       "to": [
         {"id": 1},
         {"id": 3},
+        {"id": 4},
         {"id": 5}
       ]
     },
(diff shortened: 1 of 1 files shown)


osse101 (dismissed):
The video makes it look like you can shinespark out 🔥 

It does feel weird.  I can't figure it out.  Is 1 try worth of lenience ok in insane?

PR #2198, opened 2025-08-05
Add canPowerBombItemOverloadPLMs

kjbranch:


region/brinstar/pink/Big Pink.json
```diff
@@ -3755,6 +3757,31 @@
         "If PLMs are already overloaded, Samus will need to quickly exit G-mode before the Power Bomb explodes."
       ]
     },
+    {
+      "link": [15, 13],
```
osse101, line 3761: 15 doesn't say you are in direct gmode
The description of 15 uses the old room name for wave gate room too...

osse101: The [4, 15] change also makes it look like we aren't watching PLM state correctly.
But it may only be making you overload PLMs twice.

kjbranch: Yikes, this was a big error, but should be fixed now. It turns out that PBing the item (from the right) is only beneficial if you dont have morph, because otherwise you can just use the camera scroll blocks, so that also simplified things a bit.

[4, 15] should be fine as is, because once at 15, you can overload PLMs with just h_artificialMorphMovement, whether going up or down.

Issue #1216, opened 2023-12-05
Heated rooms needing more refined Crystal Flash strats


blkerby:
Not a complete list, but a few I've noticed so far:

- [ ] Cathedral Entrance: logic currently doesn't think you can get through from right to left tankless with a Crystal Flash, but it's possible.
- [ ] Lava Dive: could use a strat that does a Crystal Flash in the lava
- [ ] Amphitheatre: could use strats doing a Crystal Flash in the acid
- [ ] Spiky Platforms Tunnel: could use strats for doing a Crystal Flash on the platforms in the middle of the room.

In most cases, probably the simplest way to add new Crystal Flash strats would be to define new junction nodes. But it's also possible to add strats that have a Crystal Flash in the middle of them (i.e. with heat frames before and after).

PR #2206, opened 2025-08-10
More heated Crystal Flash strats

blkerby:
This partially addresses #1216.

This also changes the `h_heatedCrystalFlash` helper to use `simpleHeatFrames` since normally you aren't doing anything while waiting for the CF to start. Speed Booster Hall was the main one motivating this change. In theory we could also have `simpleLavaFrames` and `simpleAcidFrames` for the lava/acid CF helpers, but those logical requirements haven't been defined and I'm not sure if it's worth the trouble :shrug: A little extra lenience for lava/acid CF may be fine anyway; for lava/acid there's not really a case like Speed Booster Hall where the movement is very simple.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -816,7 +816,7 @@
         {
           "name": "h_heatedCrystalFlash",
           "requires": [
-            {"heatFrames": 185},
+            {"simpleHeatFrames": 185},
             {"or": [
               {"and": [
                 "h_heatProof",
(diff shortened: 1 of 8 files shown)


osse101:
- There is a skill check in heated crystal flashes (Placing bomb at 55 energy vs 45) but including the leniency in the CF helper doesn't describe it better since your CF health is mixed with the situational requirements and that is tied to node placement.  The other heat leniencies happening in the room are good enough.

- Jumping through lava doesn't scale properly, like most lava situations.

PR #2207, opened 2025-08-10
Yellow Maridia farms part 1: Butterfly Room, etc.

blkerby:


region/maridia/inner-yellow/Kassiuz Room.json
```diff
@@ -111,15 +111,85 @@
...
+            "Wave",
+            {"cycleFrames": 1080}
+          ]},
+          {"and": [
+            "Spazer",
+            {"cycleFrames": 1140}
+          ]},
+          {"cycleFrames": 2040}
+        ]},
+        {"and": [
```
osse101, line 185: I thought tests caught top-level `and` now

region/maridia/inner-yellow/Kassiuz Room.json
```diff
@@ -111,15 +111,85 @@
...
-        {"partialRefill": {"type": "PowerBomb", "limit": 6}}
+        {"or": [
+          "canWalljump",
+          {"and": [
+            "SpaceJump",
+            {"cycleFrames": 300}
+          ]},
+          {"and": [
+            "canTrickySpringBallJump",
+            {"cycleFrames": 520},
```
osse101, line 155: A little bit high

region/maridia/inner-yellow/Maridia Elevator Room.json
```diff
@@ -237,6 +237,43 @@
...
+          ]},
+          {"and": [
+            {"resetRoom": {"nodes": [3]}},
+            {"cycleFrames": 1680},
+            "canTrickyWalljump",
+            "canTrickyJump"
+          ]}
+        ]},
+        {"or": [
+          "Plasma",
```
osse101, line 265: This should have a knowledge tech.  Otherwise it either looks unkillable, or dies randomly.
canDodgeWhileShooting maybe, but I don't think `hard` knows that it dies when moving left even.

region/maridia/inner-yellow/Kassiuz Room.json
blkerby: Good catch. Looks like the tests checked `and` inside of `and` but not `and` at the top level. Added that now. There weren't too many violations so I went through and fixed them all.

region/maridia/inner-yellow/Kassiuz Room.json
blkerby: Fair enough, this was the last case I checked and I was getting tired of the room, haha. Tightened it now.

region/maridia/inner-yellow/Maridia Elevator Room.json
blkerby: Added `canDodgeWhileShooting` and a FIXME.

Kyle and I were talking the other day and wondering if we should have more enemy-specific tech, like we have for Mellas, Metroids, Big Boy, and crabs. In the past I've thought it can make things a little cluttered to have too many tech for specific things. On the other hand, tech notes can be a good place to explain enemy-specific techniques/knowledge. And we may be able to organize the tech better in a hierarchical way which can reduce the clutter. Owtches of course are less important than those other enemies that have their own tech. Just some thoughts

osse101: Specific enemy tech could be hard to place in one difficulty level.  Like bombing off metroids, bombing in air, jumping over, grouping, and grouping without getting grabbed. (Even though Metroids have canMetroidAvoid tech).  Owtches have can kill moving left, and can jump over fast moving enemy in water with low ceiling.  Skultera can talk about rubbing enemies while frozen and how they lose height over time if there is a slope.  Pirates range from shooting at them for no damage to crouching to angle turn cancelling towards them, and include hitboxing, and their collision box.

Breaking it down into smaller actions would mean adding the ability to make tech more navigable.

My reaction is that you want a small wiki page talking about how to deal with each enemy in the context of this randomizer.  Weapon vulnerabilities is another thing that takes looking up but wouldn't be expected, leaving the strange cases as being harder to know about.

PR #2211, opened 2025-08-11
Add canComplexGMode to Crateria

kjbranch:


region/crateria/east/West Ocean.json
```diff
@@ -3132,7 +3139,6 @@
       "requires": [
         "canPowerBombItemOverloadPLMs",
         {"notable": "G-mode Overload PLMs by Power Bombing Morph Maze Item"},
-        "canEnterGMode",
```
osse101, line 3135: These two look like complexGmode

PR #2214, opened 2025-08-12
Add canComplexGMode to LN

kjbranch:


region/lowernorfair/east/Wasteland.json
```diff
@@ -563,7 +563,7 @@
         "Morph",
         "canTrickyJump",
         {"or": [
-          "canInsaneJump",
+          "canComplexGMode",
```
osse101, line 566: insaneJump; this is a movement check

region/lowernorfair/west/Golden Torizo's Room.json
```diff
@@ -644,7 +647,7 @@
             {"or": [
               "h_artificialMorphJumpIntoIBJ",
               "h_artificialMorphDoubleBombJump",
-              "canInsaneJump"
+              "canComplexGMode"
```
osse101, line 650: More of an insaneJump.  Could be both.

region/lowernorfair/west/Golden Torizo's Room.json
```diff
@@ -888,7 +891,7 @@
...
```
osse101, line 894: here too

region/lowernorfair/west/Golden Torizo's Room.json
```diff
@@ -1017,7 +1020,7 @@
         {"or": [
           "f_DefeatedGoldenTorizo",
           {"and": [
-            "canInsaneJump",
+            "canComplexGMode",
```
osse101, line 1023: here too

osse101 (commented):
Collecting drops when exiting GMode, and standing a number of pixels in a door (i.e. exit from same door) sound like complexGMode uses.  So I think it should be added as a requirement for heatedGMode

kjbranch:
> Collecting drops when exiting GMode, and standing a number of pixels in a door (i.e. exit from same door) sound like complexGMode uses. So I think it should be added as a requirement for heatedGMode

Some strats are pretty easy, such as entering and immediately leaving a heated room, so I don't think it would make sense to put it on the tech.
- Putting it on the heated pause abuse helper would cover the first one.
- Strats that exit direct g-mode just before touching the transition are remote acquire strats, which would already be complex, and collecting flags - which I think should be included as complex as well.

osse101 (approved):

PR #2223, opened 2025-08-13
Green Maridia farms part 1: Sand Halls

blkerby:


region/maridia/inner-green/East Sand Hall.json
```diff
@@ -985,6 +1069,111 @@
...
+      "link": [2, 2],
+      "name": "Evir Farm (1 Evir, Gravity, Screw Attack)",
+      "requires": [
+        {"resetRoom": {"nodes": [2]}},
+        "Gravity",
+        "ScrewAttack",
+        "canCarefulJump",
+        {"cycleFrames": 290},
+        {"or": [
+          "canWalljump",
```
osse101, line 1116: I'm not so sure Medium can handle this walljump.  Could you think about adding a spacejump option?
With HiJump, you would probably eventually happen to get a good jump.

Existing strats have canPlayInSand on these movement options.

osse101 (commented):
Pretty much all of these are uncomfortable farms
- Sand makes movement difficult (screw attack)
- Evir projectiles block beams (wave and spazer)
- Evirs are high damage and can just kill you
- East left door suitless -- I expect people to jump into the room at least on their first farm attempt and get hit
- Player movement is going to be different from prachack movement meaning random evir projectile hits will happen, in the cases where you skip the sunken evir.

Its a stressful farm, but I've seen runners take it if they need to.
I'm not sure exactly what I would want changed.  Maybe more frames for barely on camera kill, they were kinda hard to do in time and may not scale with movement.

Disable ice tech might need to be added in the wave and spazer cases too.

region/maridia/inner-green/East Sand Hall.json
blkerby: Sounds good. Added a Space Jump option and put canPlayInSand on the others. The Gravity+HiJump option feels close to the line between Medium and Hard. Makes sense not to expect more in the farm strat than is expected in the room traversal strat, probably safest to keep it in Hard for now.

PR #2234, opened 2025-08-16
Fix kihunter enemy names

blkerby:
Surprised the tests wouldn't catch this kind of thing.


blkerby:
Going to just merge now (without review) in order to get the master branch out of a broken state.

PR #2237, opened 2025-08-19
Remove R-mode Req from G-mode

kjbranch:
Of course, Github makes this simple change difficult to follow, so I'll describe the changes:

- Removed R-mode as a requirement from G-mode (no longer an extension)
- Split the tech description to fill a note and devNote for G-mode, upward setup, downward setup
- Move upward setup and downward setup to be inside the extension of g-mode, it now requires g-mode instead of r-mode, and added a FIXME devNote to them

I have more I want to do, but the changes were starting to look too messy, so I'll finish the rest later.

### DIFF
--- a/tech.json
+++ b/tech.json
@@ -2495,189 +2495,199 @@
           "note": [
             "Ability to enter R-mode.",
             "Have some energy in reserves (set to Auto), take damage while going through a door, use X-Ray while entering the next room, and release X-Ray after reserves have finished filling."
+          ]
+        },
+        {
+          "id": 162,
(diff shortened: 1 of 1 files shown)


osse101 (approved):
If there was a separation of EnterRmode and UseRMode, EnterGMode would be an extension of EnterRMode still.
But then looking at up/down GMode, that is going to be just as hard whether it is RMode or GMode so we want a way to share them.  
An upward RMode strat would be getting a blue suit in Croc farm.
A downward RMode strat I think could be clipping the crumble block in indiana jones.

PR #2239, opened 2025-08-19
Red Firefleas IBJ into HBJ, remove X-Ray Room references

kjbranch:


region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -1229,12 +1229,23 @@
...
     },
+    {
+      "link": [4, 1],
+      "name": "IBJ Into HBJ",
+      "requires": [
+        {"or": [
+          "canLongIBJ",
+          "canJumpIntoIBJ"
+        ]},
+        "canHBJ"
```
osse101, line 1240: HBJ is meant to be used from the ground, in the air is treated as a special case.
Trying it out, airborne HBJ feels like a short diagonal bomb jump.
It doesn't feel safe enough to me for expert.

kjbranch:
What do you think about this?
The other instances of HBJ that are particularly difficult which may be worth thinking about:
- Ice Beam Tutorial G-mode: has an insane jump
- Red Tower: is notable but still in `Expert`
- Pillars right to left (Bombs): It's got an insane jump and is notable in `Insane`

PR #2249, opened 2025-08-20
Ice beam crouch gate clip

blkerby:
At higher speeds the frame window for the jump is larger but for me it didn't translate into feeling easier. I'm thinking this would probably be an Insane notable. You have to reset the room between attempts, and there's a possibility of the neighboring room being heated, which would be a particularly bad scenario.

Video by Sam: https://videos.maprando.com/video/7557

### DIFF
--- a/region/norfair/west/Ice Beam Gate Room.json
+++ b/region/norfair/west/Ice Beam Gate Room.json
@@ -1208,6 +1208,28 @@
       ]
     },
     {
+      "link": [4, 2],
+      "name": "Arm Pump Crouch Gate Clip",
+      "entranceCondition": {
+        "comeInNormally": {}
+      },
+      "requires": [
+        {"notable": "Arm Pump Crouch Gate Clip"},
+        "canCrouchGateClip"
+      ],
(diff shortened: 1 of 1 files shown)


osse101 (dismissed):
I made it with 13 tiles but not with 11.  I'm sure I could eventually get it if I spend longer at it.  Insane seems right.

This may be a good time to add the CrouchGateClip from the right (no speed or morph).  I find the jump easier, but it would be a softlock to miss so its overall worse.

blkerby:
Hmm, interesting, didn't even know right-to-left could be done like that. It seems to require heavy arm pumping? With emulator slowdown (75% speed) I was able to get it a couple times out of a few hundred attempts probably. At full speed it didn't feel like I could even get close. So it seems way more difficult, more like Beyond territory. But maybe I'm missing something?

osse101:
I just got it 2/10 times. There is heavy arm pumping.  You jump pretty close to the gate.  That's why the left->right clips looked weird jumping from so far away, to me.
https://videos.maprando.com

PR #2253, opened 2025-08-22
Crateria Central check flash suits

blkerby:
I was happy to confirm that the requirements related to flash suit already seem nicely set up in the tech/helpers, so almost all strats can just be marked `"flashSuitChecked": true`. Some possible exceptions that I'm trying to watch out for:
- Strats involving Grapple. These can sometimes need the "h_midAirShootUp" helper. Grapple jump/fling strats I'm just putting `"flashSuitChecked": false` on for now because I think they could use a closer look. The tech `canGrappleJump` currently has a `noFlashSuit` requirement but that should possibly be refined later.
- Strats involving tight mid-air morphs from a stand. Without Hi-Jump, even `canJumpIntoIBJ` can be a bit tricky though I'm assuming not enough to justify adding the Insane-level `canTrickyCarryFlashSuit` as a requirement (PR #2259 puts a `canComplexCarryFlashSuit` requirement on it). For jump morphs in a 4-high space like West Sand Pit, I would think that would probably be `canTrickyCarryFlashSuit`, but there don't seem to be any jumps like that in Crateria.

### DIFF
--- a/region/crateria/central/Climb Supers Room.json
+++ b/region/crateria/central/Climb Supers Room.json
@@ -251,7 +251,7 @@
           "blocks": [{"position": [9, 2], "note": "Closest Grapple block to door"}]
         }
       },
-      "flashSuitChecked": false
+      "flashSuitChecked": true
     },
     {
       "id": 2,
--- a/region/crateria/central/Climb Supers Room.json
+++ b/region/crateria/central/Climb Supers Room.json
@@ -351,7 +351,10 @@
       "name": "Base",
       "requires": [
         {"or": [
-          "Grapple",
+          {"and": [
+            "Grapple",
+            "h_midAirShootUp"
+          ]},
           "SpaceJump"
         ]}
       ],
@@ -1535,7 +1538,10 @@
       "name": "Base",
       "requires": [
         {"or": [
-          "Grapple",
+          {"and": [
+            "Grapple",
+            "h_midAirShootUp"
+          ]},
           "SpaceJump"
         ]}
       ],
--- a/region/crateria/central/Climb Supers Room.json
+++ b/region/crateria/central/Climb Supers Room.json
@@ -238,7 +238,8 @@
           "length": 6,
           "openEnd": 1
         }
-      }
+      },
+      "flashSuitChecked": true
     },
     {
       "id": 75,
@@ -249,7 +250,8 @@
         "leaveWithGrappleSwing": {
           "blocks": [{"position": [9, 2], "note": "Closest Grapple block to door"}]
         }
-      }
+      },
+      "flashSuitChecked": false
     },
     {
       "id": 2,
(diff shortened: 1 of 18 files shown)


region/crateria/central/Climb Supers Room.json
```diff
@@ -349,10 +351,14 @@
       "name": "Base",
       "requires": [
         {"or": [
-          "Grapple",
+          {"and": [
+            "Grapple",
+            "h_midAirShootUp"
```
osse101, line 356: This can use `complexCarry` once it is added.

region/crateria/central/Climb Supers Room.json
```diff
@@ -364,7 +370,8 @@
           "canUseIFrames",
           {"spikeHits": 1}
         ]}
-      ]
+      ],
+      "flashSuitChecked": true
```
osse101, line 374: `canUseIFrames` would be another one that may need to be `complexCarry`

region/crateria/central/Climb Supers Room.json
```diff
@@ -477,6 +490,7 @@
         "canLateralMidAirMorph",
         "canSpringFling"
       ],
+      "flashSuitChecked": true,
```
osse101, line 493: It's possible `canLateralMidAirMorph` is `complexCarry` too since that is the most common place to fumble an input.  With the hard stuff, you are paying attention to your flash suit.

region/crateria/central/Climb Supers Room.json
```diff
@@ -1609,6 +1654,7 @@
           ]}
         ]}
       ],
+      "flashSuitChecked": true,
```
osse101, line 1657: This might be complex or might not.  It should be roughly as hard as the damageboost for fumbling the flash suit.

region/crateria/central/Climb.json
```diff
@@ -2216,6 +2265,7 @@
           {"obstaclesNotCleared": ["B"]}
         ]}
       ],
+      "flashSuitChecked": true,
       "note": [
         "Diagonal shinespark to break the bomb blocks to the morph tunnel on the right.",
         "Spark from the lowest platform that is only one tile from the right wall (part of the bottom right door will be on screen).",
```
osse101, line 2271: A full height jump
Or a jump into aim down with hijump
can save some energy

region/crateria/central/Climb Supers Room.json
blkerby: I don't see the interaction between `canUseIFrames` and a flash suit? Only issue here would be if the player is trying to horizontally damage boost, but there's no reason to do that

region/crateria/central/Climb Supers Room.json
blkerby: I'm not seeing the risk with `canLateralMidAirMorph`. You'll be pressing down to morph, so it wouldn't be possible to accidentally press up?

PR #2255, opened 2025-08-23
Crateria East check flash suits

blkerby:
Unrelated to the main purpose of the PR, I'm deleting obsolete "Leave Normally" strats where I see them (since these are now implicit).

~~In East Ocean, preserving a flash suit with the bomb jump water escape and Choot damage boost seemed fairly tricky. It's not as bad as gate glitching with Morph only (no spring ball) but I figured probably still worthy of `canTrickyCarryFlashSuit`?~~

I'm not sure what to do about bomb jump water escape and damage boosts in East Ocean (and in general). They might be a good candidate for the Extreme-level tech `canComplexCarryFlashSuit` which Kyle is working on adding? The damage boost doesn't seem as bad as I thought it was at first: I was trying it before with an aim-down buffered crouch jump, but angle-down jump seems easier.

### DIFF
--- a/region/crateria/east/East Ocean.json
+++ b/region/crateria/east/East Ocean.json
@@ -1035,13 +1035,7 @@
         {"notable": "Suitless Damage Boost Water Escape"},
         "canSuitlessMaridia",
         "canHorizontalDamageBoost",
-        {"or": [
-          "canCrouchJump",
-          {"and": [
-            "h_underwaterCrouchJumpWithFlashSuit",
-            "canTrickyCarryFlashSuit"
-          ]}
-        ]},
+        "h_underwaterCrouchJump",
         {"enemyDamage": {"enemy": "Choot", "type": "contact", "hits": 1}}
       ],
       "flashSuitChecked": true,


region/crateria/east/Homing Geemer Room.json
```diff
@@ -90,6 +92,7 @@
         {"cycleFrames": 150}
       ],
       "farmCycleDrops": [{"enemy": "Geemer (orange)", "count": 1}],
+      "flashSuitChecked": true,
```
osse101, line 95: Would use tricky morph tech

region/crateria/east/The Moat.json
```diff
@@ -141,6 +143,7 @@
           "blockPositions": [[7, 2]]
         }
       },
+      "flashSuitChecked": true,
```
osse101, line 146: springball would work.
Not enough space to spinjump and unmorph

region/crateria/east/The Moat.json
```diff
@@ -157,6 +160,7 @@
           "SpaceJump"
         ]}
       ],
+      "flashSuitChecked": true,
```
osse101, line 163: You wouldn't be able to avoid the item with grapple
- A really good swing can cross over

PR #2258, opened 2025-08-23
Add can(Complex)CarryFlashSuit, no uses yet

kjbranch:
What do you think of these placements? Do you think bomb jump water escape should be part of complex (Extreme) or tricky (Insane)?
I haven't added these tech to any other helpers or strats yet. 

I didn't edit the descriptions for spike suit or elevator CF.

### DIFF
--- a/tech.json
+++ b/tech.json
@@ -2141,7 +2141,9 @@
         },
         {
           "name": "canCarryFlashSuit",
-          "techRequires": [],
+          "techRequires": [
+            "canShinechargeMovementComplex"
+          ],
           "otherRequires": [],
           "note": [
             "A flash suit is a glitched state where Samus has the ability to shinespark without having a shinecharge timer.",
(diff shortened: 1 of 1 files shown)


blkerby (dismissed):

osse101:
some more

- `canUseIFrames` is a common problem one. You try and jump too late or land weird and lose your spikesuit.
- `enemyDamage` for similar reasons, unless you are morphing through the enemy.  Just running through is ok.
- `canHorizontalDamageBoost`, same thing, mistime it and lose it.
- `canLateralMidAirMorph` is often a problem because you miss the morph and then lose the spikesuit.  It may only be for cases where you are doing something in addition to morphing, like shooting a door open.
- `LeaveWhileSpaceJumping` often lands close to the door frame and that's normally a problem.  But maybe you would spin jump out of it?
- `bombjumpwaterescape` didn't feel bad.  Maybe iffy for expert, just has some quick button swapping but its very lenient.  The lava dive room isn't any harder
- Farm strats may need to go into complex.  Anything with an up-shot does.

And there needs to be more heat frame leniency because your movements are going to be stiff the entire time.

blkerby:
It's going to take some major work on the randomizer side to support a separate heat frame multiplier for flash suit. I was hoping to avoid that. But I can work on it. It would help solve some other problems with ammo/energy management (e.g. the case that comes up sometimes in Attic on Hard settings, where the randomizer currently unable to properly explore both options of tanking damage and killing the enemies with Missiles, since they are in the same strat in an `or`).

blkerby:
If we are applying a separate heat multiplier for flash suit, I assume we can apply the same to farm cycles. And then any farms that specifically needed a mid-air up shot would need that added as a `h_midAirShootUp` requirement.

`canUseIFrames` and `enemyDamage` look like ones to keep an eye out for, in case of specific instances where they might create a problem. I don't think they're ones where it would have to be baked into the general requirements?

`canLateralMidAirMorph` seems like another borderline one, in that as long as you get the mid-air morph, having a flash suit doesn't affect anything. But if you miss the morph and then also release forward while still holding jump then you lose the flash suit. To me that seems more like a failure of basic `canCarryFlashSuit` movement, because this same risk applies any time you jump/spinjump and break spin; it's not specific to `canLateralMidAirMorph`. And in the case of `canLateralMidAirMorph`, even if the morph is tight you can still keep holding forward to be safe in case you miss it. There could maybe be specific scenarios where it's more of a problem though?

PR #2259, opened 2025-08-24
Main uses of canComplexCarryFlashSuit

blkerby:
This builds on PR #2258 , which should be merged first.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -643,8 +643,7 @@
                 "canComplexCarryFlashSuit"
               ]},
               {"and": [
-                "Morph",
-                "canInsaneJump",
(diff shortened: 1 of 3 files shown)


kjbranch (dismissed):
Looks good.

I think it would be good if we had another helper, like `h_ComplexCarryFlashSuit`, defined as `canComplexCarryFlashSuit` or `noFlashSuit`. And then we could prevent `canComplexCarryFlashSuit` from being used in strats (similar to how we don't use `SpringBall` and instead use `h_useSpringBall`.)

osse101 (dismissed):

PR #2262, opened 2025-08-25
A couple more helpers for flash suits

blkerby:
We can use these to mark individual cases that are more difficult to carry a flash suit than reflected by their tech requirements alone.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -748,6 +748,24 @@
           ]
         },
         {
+          "name": "h_complexToCarryFlashSuit",
+          "requires": [
+            {"or": [
+              {"noFlashSuit": {}},
+              "canComplexCarryFlashSuit"
+            ]}
+          ]
+        },
+        {
+          "name": "h_trickyToCarryFlashSuit",
+          "requires": [
+            {"or": [
+              {"noFlashSuit": {}},
+              "canTrickyCarryFlashSuit"
+            ]}
+          ]
+        },
+        {
           "name": "h_PlasmaHitbox",
           "requires": [
             "Plasma",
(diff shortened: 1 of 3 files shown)


region/brinstar/blue/Blue Brinstar Boulder Room.json
```diff
@@ -874,10 +874,7 @@
       "requires": [
         "canBombJumpWaterEscape",
         "canJumpIntoIBJ",
-        {"or": [
-          "canTrickyCarryFlashSuit",
-          {"noFlashSuit": {}}
-        ]}
+        "h_trickyToCarryFlashSuit"
```
kjbranch, line 877: does this one need to remain? I dont know if it is harder than a normal bomb jump water escape.

helpers.json
```diff
@@ -747,6 +747,24 @@
             "canTrickySpringBallJump"
           ]
         },
+        {
+          "name": "h_complexToCarryFlashSuit",
```
kjbranch, line 751: not sure how i feel about the `To`. It feels a little odd, but i guess i am fine either way.

kjbranch (dismissed):

helpers.json
blkerby: Yeah the idea of the "To" is just to avoid implying that having/carrying a flash suit is a requirement, only that *if* you're carrying one then it is complex to do so

region/brinstar/blue/Blue Brinstar Boulder Room.json
blkerby: It's more complicated than some other water escapes, because of the tide and because of having to go into an IBJ (not just a single bomb boost). But the difficulty doesn't really have anything to do with the flash suit, nor does it seem Insane level, so probably best to remove it, yeah.

osse101 (approved):

PR #2263, opened 2025-08-27
Refine flash suits and crouch jumps

kjbranch:
Went through all the crouch jumps to add the underwater variants. Tested many and added `"flashSuitChecked": true` and a few crouch jump alternatives.

Still need to go through the max height spring ball jumps.

### DIFF
--- a/region/lowernorfair/east/Lower Norfair Fireflea Room.json
+++ b/region/lowernorfair/east/Lower Norfair Fireflea Room.json
@@ -1192,11 +1192,12 @@
       "name": "Bug Boost Escape",
       "requires": [
         "HiJump",
+        "canCameraManip",
         "canHorizontalDamageBoost",
-        "canTrickyJump",
+        "canTrickyDodgeEnemies",
         {"or": [
           "canCrouchJump",
-          "canInsaneJump"
+          "canTrickyCarryFlashSuit"
         ]},
         {"enemyDamage": {"enemy": "Fireflea", "type": "contact", "hits": 1}}
       ],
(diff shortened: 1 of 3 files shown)


region/lowernorfair/east/Lower Norfair Fireflea Room.json
```diff
@@ -1194,15 +1194,19 @@
         "HiJump",
         "canHorizontalDamageBoost",
         "canTrickyJump",
-        "canCrouchJump",
+        {"or": [
+          "canCrouchJump",
+          "canInsaneJump"
```
blkerby, line 1199: I think `h_trickyToCarryFlashSuit` may be justified here? Without a crouch jump, it's pretty precise.

region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -989,6 +989,28 @@
...
+    },
+    {
+      "link": [4, 2],
+      "name": "Jump Into Respawning Block - Blocks Mostly Intact",
+      "requires": [
+        {"obstaclesCleared": ["B"]},
+        {"heatFrames": 600},
+        "canJumpIntoRespawningBlock",
+        {"or": [
+          "canInsaneJump",
```
blkerby, line 1003: This one might be fine as is, though I wonder if it should maybe be bumped up a tier? (e.g. with `h_complexToCarryFlashSuit` for the lenient version, `h_trickyToCarryFlashSuit` for the one without lenience).

I found that it could be pretty reliable by shooting up immediately followed by 5 full-height jumps; some timing cue like this would probably be good to mention in the note.

blkerby:
I only glanced over most of the PR, but for the parts I looked at, it all seems good.

PR #2264, opened 2025-08-27
Tourian check flash suits

blkerby:
Rooms with an interesting flash suit interaction :
- Big Boy Room: flash suit makes it hard to break the top seaweed while baby skipping counter-clockwise. Doing it clockwise should be an option. Using a Power Bomb can be another alternative (saw this on your stream yesterday @osse101 ). I'm not good enough at baby skipping to estimate how these variations would impact difficulty, so I just put `h_trickyToCarryFlashSuit` to be safe. It can be adjusted if there is a case that would be ok with just `complex`.
- Mother Brain Room: Destroying the Zebs or doing Ice Zeb Skip are both a bit awkward, seemed like `complex` level.
- Tourian Escape Room 4: Freezing the Pirates while climbing the left side is tricky. Killing them also seemed non-trivial (complex level).
- Tourian Hopper Room: going left-to-right, opening the door can be awkward if the Hoppers are still chasing you. Regardless of the entry strat (Morph roll, Screw, Squeeze under, or Tank damage) it always seemed possible to get the Hoppers stuck off-camera if moving cleanly/quickly enough; or if the Hoppers do follow, there can still be enough time to shoot it open. But it affects the difficulty enough that it seemed worth putting `h_complexToCarryFlashSuit` on all the left-to-right strats that have the Hoppers alive.

This also updates the docs to say that `leaveSpaceJumping` comes with an implicit `h_trickyToCarryFlashSuit` requirement. In my testing I saw that the case that you lose the flash suit is if you are exactly 1 frame late on the Space Jump, and it still happens even if you are holding forward. Since some applications may require a last-frame jump, this did seem like high risk since being off by 1 frame will make you lose it. It could probably be refined by putting this requirement in the room with the entrance condition (comeInSpaceJumping), only in cases needing a very late/low jump; but for now it seemed safest to just put it everywhere. Of course, `leaveSpinning` strats (even ones involving Space Jump) won't be affected, since those don't require a low jump.

### DIFF
--- a/region/tourian/main/Big Boy Room.json
+++ b/region/tourian/main/Big Boy Room.json
@@ -86,6 +86,7 @@
       },
       "requires": [],
       "clearsObstacles": ["A"],
+      "flashSuitChecked": true,
       "note": "The Baby Metroid is not active when entering from the left side of the room."
     },
     {
@@ -100,7 +101,8 @@
           "length": 5,
           "openEnd": 1
         }
-      }
+      },
+      "flashSuitChecked": true
     },
     {
       "id": 36,
@@ -119,6 +121,7 @@
           "obstruction": [1, 0]
         }
       },
+      "flashSuitChecked": true,
       "devNote": ["Max extra run speed $3.1"]
     },
     {
@@ -150,7 +153,8 @@
             "openEnd": 1
           }
         }
-      }
+      },
+      "flashSuitChecked": true
     },
     {
       "id": 5,
@@ -170,7 +174,8 @@
             "openEnd": 1
           }
         }
-      }
+      },
+      "flashSuitChecked": true
     },
     {
       "id": 6,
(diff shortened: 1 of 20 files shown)


osse101 (dismissed):
- There are many not-special strats that would be hard in expert, so it could be better to start flash suits in extreme after all.  Metroid avoid and kills, walljumpless acid run, M4 itemless, M1 rinka boost, MB fight, MB glass break.

- Zeb skip with ice+grapple is slightly easier, especially if downbacking

- Ice only Escape4 is hard.  The flash suit only makes it harder to damage the pirates down in time, and maybe avoid them.

blkerby:
> * There are many not-special strats that would be hard in expert, so it could be better to start flash suits in extreme after all.  Metroid avoid and kills, walljumpless acid run, M4 itemless, M1 rinka boost, MB fight, MB glass break.

Yeah starting it in Extreme for now seems like the safer option. It could always be moved down later if players end up being comfortable with it. And having `canCarryFlashSuit` and `canComplexCarryFlashSuit` separated from each other can still be helpful in giving players a way to ease into the tech if they would like.

PR #2266, opened 2025-08-28
Crocomire Norfair check flash suits

blkerby:
- Post Croc Farm Room had the most interesting changes: using the Ripper to grapple across from top-left to top-right is always possible with a flash suit, but using it to get up from the bottom of the room requires "h_midAirShootUp". With a flash suit, damage boosting off the Ripper becomes logically significant, because you may not be able to use the Gamets. Boosting off a Gamet to the top-right also becomes a useful option.
- Heat frames in Post Croc Power Bomb Room seemed way off, so I updated them.
- Added a grapple jump strat in Grapple Beam Room which is only useful for carrying a flash suit.
- Added a "Tricky Dash Jump with Hi-Jump" strat in Indiana Jones Room to directly reach the item from the bottom-right with PB blocks cleared. It mostly only makes a difference logically if carrying a flash suit, since otherwise you could just shinespark, farming Mellas for energy.

### DIFF
--- a/region/norfair/crocomire/Indiana Jones Room.json
+++ b/region/norfair/crocomire/Indiana Jones Room.json
@@ -1296,7 +1296,8 @@
       "requires": [
         {"notable": "Indiana Jones Grapple"},
         "canUseEnemies",
-        "canPreciseGrapple"
+        "canPreciseGrapple",
+        "h_midAirShootUp"
       ],
       "flashSuitChecked": true,
       "note": "Involves grappling off several Ripper 2."
--- a/region/norfair/crocomire/Indiana Jones Room.json
+++ b/region/norfair/crocomire/Indiana Jones Room.json
@@ -194,7 +194,8 @@
           "length": 10,
           "openEnd": 1
         }
-      }
+      },
+      "flashSuitChecked": true
     },
     {
       "id": 2,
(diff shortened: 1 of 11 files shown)


region/norfair/crocomire/Indiana Jones Room.json
```diff
@@ -1246,29 +1298,33 @@
         "canUseEnemies",
         "canPreciseGrapple"
```
osse101, line 1299:
```suggestion
        "canPreciseGrapple",
        "h_midAirShootUp"
```

region/norfair/crocomire/Post Crocomire Farming Room.json
```diff
@@ -1261,6 +1318,36 @@
...
+    {
+      "link": [4, 1],
+      "name": "Ripper Damage Boost",
+      "requires": [
+        {"notable": "Damage Boost"},
+        "canTrickyJump",
+        "canHorizontalDamageBoost",
+        {"enemyDamage": {"enemy": "Ripper 2 (green)", "type": "contact", "hits": 1}}
+      ],
+      "wallJumpAvoid": true,
```
osse101, line 1338: This would show up a lot by being a lower health cost than the gamet boost

PR #2267, opened 2025-08-29
Refine flash suits and max height spring ball

kjbranch:
Removed flash suit checked from a couple max height spring ball into double reserve trigger. Not sure if those were tested properly? If so, I can add them back

### DIFF
--- a/region/maridia/inner-pink/Halfie Climb Room.json
+++ b/region/maridia/inner-pink/Halfie Climb Room.json
@@ -865,14 +865,16 @@
         "canSuitlessMaridia",
         "HiJump",
         "canTrickyJump",
-        "h_maxHeightSpringBallJump",
-        "canStationaryLateralMidAirMorph"
+        "canTrickySpringBallJump",
+        "canSpringFling"
       ],
+      "flashSuitChecked": true,
       "note": [
         "SpringBall jump just as Samus exits the water to reach the distant ledge.",
-        "A Crouch jump makes the SpringBall jump easier, but at the cost of some horizontal movement."
+        "In order to move horizontally far enough, quickly pause and morph in order to gain a Spring Fling as Samus equips Spring Ball.",
+        "This can be done from a spin jump or from a stationary lateral mid-air morph."
       ],
-      "devNote": "canStationaryLateralMidAirMorph is not strictly required but better describes the difficulty."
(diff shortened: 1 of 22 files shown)


region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -865,14 +865,16 @@
...
       ],
+      "flashSuitChecked": true,
       "note": [
         "SpringBall jump just as Samus exits the water to reach the distant ledge.",
-        "A Crouch jump makes the SpringBall jump easier, but at the cost of some horizontal movement."
+        "In order to move horizontally far enough, quickly pause and morph in order to gain a Spring Fling as Samus equips Spring Ball.",
+        "This can be done from a spin jump or from a stationary lateral mid-air morph."
       ],
-      "devNote": "canStationaryLateralMidAirMorph is not strictly required but better describes the difficulty."
+      "detailNote": "There is a two-frame window for when Samus can jump after escaping the waterline. Starting from a crouch jump does not increase this window."
```
osse101, line 877: I see 8 frames when doing the crouch jump + stationaryLateralMidAirMorph

region/maridia/outer/Main Street.json
```diff
@@ -5117,7 +5118,7 @@
         "canInsaneJump",
         "canReserveDoubleDamageBoost",
         {"or": [
-          "h_maxHeightSpringBallJump",
+          "h_underwaterMaxHeightSpringBallJump",
```
osse101, line 5121: The springballjump part isn't any harder.  I had to relearn how to do it  is all.
The blind damageboost with a spikesuit sounds scary; I didn't test that.

region/maridia/outer/Maridia Tube.json
```diff
@@ -3223,7 +3223,7 @@
         {"or": [
           "Gravity",
           "HiJump",
-          "h_maxHeightSpringBallJump",
+          "canTrickySpringBallJump",
```
osse101, line 3226: trickyJump is there because the unmorph is unintuitive for hard.  But that isn't so important for GMode strats.

region/maridia/outer/Main Street.json
kjbranch: I wasn't really sure what to do with this - i think it would look how I wrote it, but since it hasn't been tested, I just put flash suit checked = false

PR #2268, opened 2025-08-29
Norfair West check flash suits

blkerby:
While checking flash suits, I also looked at heat frames and made some adjustments (not a full pass).

- Crocomire Escape:
  - Some heat frames were very loose.
  - Later we might consider adding a junction at the top right, to factor out requirements for dealing with the Geruta.
- Crumble Shaft:
  - Some heat frames between top-right door and item were loose.
- Ice Beam Gate Room:
  - Entering from the bottom door and killing the Dessgeegas damageless with Power Beam felt like it needed canTrickyDodgeEnemies. This would put it in Expert instead of Very Hard as it is now.
- Ice Beam Snake Room:
  - The heat frames in the Base strat 3->1 seemed impossible.
  - Some tech requirement seemed needed to dodge the fireballs in the Base strat 1->3.
  - Many other heat frames were loose.
  - Added farmWhileShooting versions of killing the Sovas.
  - ~~Some variants of the "Kill Sovas" 4->2 strat seemed unsound: you need to get in the tunnel fast (with wall jump or HiJump) in order to get through in time to kill all the Sovas without tanking a hit. Either that or you wait for all the Sovas to come out, but that's only worthwhile with heat-proof.~~ The "Speedrun Ice Beam Entry" had some overlap with the "Tank Damage" and "Kill Sovas" strats. Simplified it while adding a "farmWhileShooting" variant.
- Ice Beam Trippers Room:
  - The Base heat frames could be tightened a bit. Previously it looks like it would not have been logically possible to traverse both ways tankless, and now it would.
  - The "Leave With Runway" at the door nodes had unnecessary `heatFrames`.


blkerby:
Oh, I think I was confused about the "Kill Sovas" strat in Ice Beam Snakes Room. I see now that you can kill all 3 Sovas with Wave + Plasma (or a Power Bomb) without getting into the tunnel quickly, and that's what made it different from the "Speedrun Ice Beam Entry". Let me fix it real quick

osse101 (approved):

PR #2269, opened 2025-08-31
West Maridia check flash suits

blkerby:
Some of the more difficult tricks (mostly tricky grapple jumps) I'm just leaving as `"flashSuitChecked": false` for now.

### DIFF
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -429,7 +429,8 @@
           "length": 2,
           "openEnd": 1
         }
-      }
+      },
+      "flashSuitChecked": true
     },
     {
       "id": 2,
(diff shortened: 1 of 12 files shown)


region/maridia/outer/Mt. Everest.json
```diff
@@ -4359,6 +4435,7 @@
       ],
       "resetsObstacles": ["A", "B"],
       "farmCycleDrops": [{"enemy": "Sciser", "count": 3}],
+      "flashSuitChecked": true,
```
osse101, line 4438: h_midAirShootUp
The farm strats that use Powamps to move around can't reach them from the ground

osse101 (commented):
Everest - Short Crab Climb - is fine

PR #2274, opened 2025-09-05
Miscellaneous grapple teleport updates

blkerby:
- Update the notes to use hexadecimal for X/Y positions since this is how they now appear in the Practice Hack InfoHUD.
- Add a new tech `canGrappleTeleportWallEscape` and applications.
- Add some missing Carry Grapple Teleport strats in Single Chamber (they already existed coming from the bottom door  but not the others).

The `canGrappleTeleportWallEscape` strats that are added only cover setups where it is possible to escape a locked door without opening it (using Morph or X-Ray), which there are only three: the Colosseum top-left and top-right doors, and the Halfie Climb top-right door. For the Halfie Climb left-side doors (and Draygon left door), the Grapple block is too close so you're not able to retract Grapple to pull Samus out of the door. Strats could still be added in those cases to escape the door by unlocking it; but the usefulness would be significantly lower, since if the door is unlocked, you could just bypass it (using the existing strats) and then go back through, which in most cases would either be free or cost just 10 energy if the other room is heated (though currently the randomizer doesn't even model that 10 energy for going back through the door, except in the context of `resetRoom` requirements).


region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -599,6 +599,23 @@
       },
       "requires": []
     },
+    {
+      "link": [2, 1],
+      "name": "Grapple Teleport Door Escape",
+      "entranceCondition": {
+        "comeInWithGrappleTeleport": {
+          "blockPositions": [[3, 12], [3, 13]]
```
osse101, line 607:
```suggestion
          "blockPositions": [[3, 12]]
```

region/lowernorfair/west/Golden Torizo's Room.json
```diff
@@ -520,6 +520,27 @@
         {"heatFrames": 50}
       ]
     },
+    {
+      "link": [2, 1],
+      "name": "Grapple Teleport Door Escape",
+      "entranceCondition": {
+        "comeInWithGrappleTeleport": {
+          "blockPositions": [[3, 12], [3, 13]]
```
osse101, line 528:
```suggestion
          "blockPositions": [[3, 12]]
```

region/norfair/east/Single Chamber.json
```diff
@@ -1209,6 +1209,22 @@
       "requires": [],
       "bypassesDoorShell": true
     },
+    {
+      "link": [3, 1],
+      "name": "Carry Grapple Teleport",
```
osse101, line 1214: `canGrappleTeleportWallEscape` not worth adding?

osse101 (commented):
I couldn't get the correct setup after all.  [3, 13] is easy to stop by the transition with no speed, but then I either couldn't retract in the next room or couldn't retract far enough.  I was always put in crouch and couldn't xray climb.

region/norfair/east/Single Chamber.json
blkerby: Good catch, yeah it can be useful for reaching node 4 if you are walljumpless and carrying a flash suit. Similarly with 2->1. Added those now.

PR #2276, opened 2025-09-07
grapple gate glitch flashsuit checked

internet-account2:
Grapple Gate Glitch keeps flashsuit.

Am I understanding correctly that all the listed requirements for the tech don't "kill" the flashsuit (at least I think they don't? But that's mostly just guessing on my part, to be honest) and therefore the procedure is just to add the "checkedFlashSuit": true part and it's done?

### DIFF
--- a/region/brinstar/green/Green Hill Zone.json
+++ b/region/brinstar/green/Green Hill Zone.json
@@ -1214,6 +1214,7 @@
         "canTrickyJump"
       ],
       "clearsObstacles": ["B"],
+      "flashSuitChecked": true,
       "note": [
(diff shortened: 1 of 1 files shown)


internet-account2:
I'm guessing the actual real approach would probably be to collect a whole area at once and then submit that. I hope this is not just some unhelpful noise and if it is, sorry

blkerby:
It's all good, always nice to see someone taking interest in contributing to the logic. Green Brinstar was one of the first (sub)areas checked for flash suit logic, and things have evolved a bit since then, so a second pass can be helpful. Every strat should eventually get flashSuitChecked set to true

PR #2278, opened 2025-09-09
Hard Boss Updates

kjbranch:
This is to update the expectations of some of the hardest possible boss fights for Map Rando's `Hard` logic.

In addition, some other things need to be addressed in the boss calculator logic:
- Botwoon needs a lot of work, mostly including significantly more ammo
- Draygon with Gravity needs more damage taken without Morph
- Ridley and Phantoon need to be checked still as well, I may look into those eventually, but not right now


region/lowernorfair/west/Golden Torizo's Room.json
```diff
@@ -2077,8 +2115,19 @@
...
+    {
+      "id": 8,
+      "name": "Safe Spot Kill (Lenient)",
+      "note": [
+        "The ability to get into the safe spot in the Golden Torizo fight, where Samus can attack while being safe from most of GT's attacks.",
+        "Samus is not able to shoot GT with Supers in this lenient safe spot and she must have fewer than 32 Missiles at all times.",
+        "To get into the safe spot near the door: Stand on the left side of the floor pillar for GT to jump back. Then stand to the right of the nearby background pillar.",
+        "This safe spot expects getting hit by GT before it jumps back. The space includes the less lenient spot, and also extends left 9 extra pixels."
+      ],
+      "devNote": "There is a 14 px window where Samus is safe from all of GT's attacks except the thrown Super attack and the attack that only occurs when Samus has too many Missiles."
```
osse101, line 2128: A comment about the bug spawner attack when he steps forward could fit here, since it can hit you still.

region/norfair/crocomire/Crocomire's Room.json
```diff
@@ -775,15 +789,26 @@
...
       "name": "Crocomire Fight (Missiles)",
       "requires": [
-        {"ammo": {"type": "Missile", "count": 25}},
         {"or": [
-          "canDodgeWhileShooting",
-          {"ammo": {"type": "Missile", "count": 25}}
+          {"and": [
+            "canDodgeWhileShooting",
+            {"ammo": {"type": "Missile", "count": 20}},
+            {"resourceMissingAtMost": [{"type": "Energy", "count": 50}]}
```
osse101, line 796: I still prefer 25 missiles.  20 is probably most of the time but might feel bad.

PR #2283, opened 2025-09-13
Kraid Brinstar check flash suits 

kjbranch:
Coming from the bottom of kihunter room looks like it might need some work. I just marked it as `checked = false` for now.


osse101 (commented):
Most of my comments can be part of another PR

PR #2285, opened 2025-09-19
Yellow Maridia check flash suits

blkerby:


region/maridia/inner-yellow/Watering Hole.json
```diff
@@ -332,21 +348,23 @@
       "name": "Use Flash Suit",
       "requires": [
         {"useFlashSuit": {}},
-        {"shinespark": {"frames": 32, "excessFrames": 13}}
+        {"shinespark": {"frames": 15, "excessFrames": 12}}
```
osse101, line 351: The shorter shinespark would use additional tech since `Crouch Jump Down Grab` is hard enough to not know you can make it.

blkerby: Makes sense, updated it now to have both options. It probably won't make a difference since you could farm the Zebs before and after to have plenty of energy. But it's good to get it right so that it can work correctly in case game behavior changes are made, like reducing initial energy capacity (e.g. 49 instead of 99) or Zebs no longer dropping energy.

PR #2288, opened 2025-09-24
Green Maridia check flash suits pt. 1

blkerby:
For East Sand Hall, I made a lot of changes, mostly related to adding a new node to represent the rightmost pillar. There are many ways of reaching the pillar from the right, and from there you can use a flash suit to reach the left side with less energy than if sparking from near the door. Alternatively, you can continue on to the left side of the room using frozen Evirs. The door stuck Flatley jump seemed unique in putting you on the pillar without the opportunity of freezing the Evir (except if using a Power Bomb to make it rise); I added an obstacle to represent this.

The room diagram should probably be updated before this is merged.


osse101 (approved):

region/maridia/inner-green/East Sand Hall.json
```diff
@@ -1757,7 +1882,72 @@
...
+      ],
+      "flashSuitChecked": true,
+      "devNote": [
+        "With a controlled shinespark end, starting with 74 energy, this can collect an Evir drop."
+      ]
+    },
+    {
+      "link": [5, 2],
+      "name": "Base",
+      "requires": [],
```
kjbranch, line 1903: Doesn't really matter, since you shouldnt be able to get here without one of these. But it looks like a nontrivial jump
```suggestion
      "requires": [
        "h_navigateUnderwater"
      ],
```

region/maridia/inner-green/East Sand Hall.json
```diff
@@ -1757,7 +1882,72 @@
...
+        "Use the Evir as a platform, which should have been frozen soon after entering the room."
+      ]
+    },
+    {
+      "link": [5, 4],
+      "name": "Power Bomb and Freeze Evir",
+      "requires": [
+        "h_usePowerBomb",
+        "canTrickyUseFrozenEnemies",
+        "canPlayInSand"
```
kjbranch, line 1944: Not seeing why this one has playinsand and the one above doesn't. If anything, this one I would think you can choose your freezing position.

With having to place a PB, I wonder if it needs a tricky jump or evir projectile hit

kjbranch (approved):
I can work on the room diagram

PR #2289, opened 2025-09-24
Green Maridia check flash suits pt. 2 (Oasis)

blkerby:
Oasis needed a lot of new stuff in order to take flash suits into account. The biggest thing is that node 4 needed to be split into two nodes: node 4 now refers specifically to being on the top platform, while the new node 9 means one level lower. 

Previously, we considered these to be interchangeable because any movement item that could get you up from 9 to 4 could also get you from 4 up through the door, and vice versa. However, with flash suits, you could spark up through the door from node 4 but not from node 9 (more specifically, from node 9 you could only diagonally spark through the door, but that isn't yet modeled). Also from node 4 you can always unlock the top door, where from node 9 you can only unlock it if it is a Power Bomb door (or potentially, with Wave, if it is a beam door, though this isn't modeled yet).

Most of the methods of breaking the bomb blocks from below with blue speed were missing, probably because we didn't consider it important since shinesparking is usually possible as an alternative. But when carrying a flash suit, the blue speed methods are more relevant.

### DIFF
--- a/region/maridia/inner-green/Oasis.json
+++ b/region/maridia/inner-green/Oasis.json
@@ -1061,7 +1061,7 @@
     {
       "id": 16,
       "link": [1, 3],
-      "name": "Cross Room Jump Through Top Door (Screw Attack)",
+      "name": "Cross Room Jump Boomerang Through Top Door (Screw Attack)",
       "entranceCondition": {
         "comeInJumping": {
           "speedBooster": true,
@@ -1069,7 +1069,7 @@
         }
       },
       "requires": [
-        {"notable": "Cross Room Jump Through Top Door"},
+        {"notable": "Cross Room Jump Boomerang To Top"},
         "canInsaneJump",
         "canCrossRoomJumpIntoWater",
         "ScrewAttack",
@@ -1106,7 +1106,7 @@
     {
       "id": 72,
       "link": [1, 3],
-      "name": "Cross Room Jump Through Top Door (Blue Speed)",
+      "name": "Cross Room Jump Boomerang Through Top Door (Blue Speed)",
       "entranceCondition": {
         "comeInGettingBlueSpeed": {
           "length": 0,
@@ -1116,7 +1116,7 @@
         }
       },
       "requires": [
-        {"notable": "Cross Room Jump Through Top Door"},
+        {"notable": "Cross Room Jump Boomerang To Top"},
         "canInsaneJump",
         "canCrossRoomJumpIntoWater",
         "canMomentumConservingTurnaround",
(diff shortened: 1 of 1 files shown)


osse101:
You can shoot the door at 9 on the right side with everything.

region/maridia/inner-green/Oasis.json
```diff
@@ -894,44 +1188,179 @@
...
+        {"notable": "Cross Room Jump Through Top Door"},
+        "canInsaneJump",
+        "canCrossRoomJumpIntoWater",
+        "canMomentumConservingTurnaround",
+        "h_trickyToCarryFlashSuit"
+      ],
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],
```
osse101, line 1255: Should this be tested against; or is there meaning in this?
4 doesn't have a door and there is no nodeID on the unlockDoors requirements.

region/maridia/inner-green/Oasis.json
```diff
@@ -1820,37 +2638,170 @@
...
-        "canInsaneJump",
+        {"notable": "Cross Room Jump Through Top Door"},
+        "canInsaneJump",
+        "canCrossRoomJumpIntoWater",
+        "ScrewAttack",
+        "canMomentumConservingTurnaround",
+        "h_trickyToCarryFlashSuit"
+      ],
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
```
osse101, line 2663: here too

region/maridia/inner-green/Oasis.json
```diff
@@ -2649,12 +3774,73 @@
...
+      "name": "Base",
+      "requires": [],
+      "flashSuitChecked": true
+    },
+    {
+      "link": [9, 9],
+      "name": "Unlock Door",
+      "requires": [],
+      "unlocksDoors": [
+        {"nodeId": 3, "types": ["powerbomb"], "requires": []}
```
osse101, line 3838: Everything works from the right side.

osse101 (commented):
All the trust that these side-platform jumps are possible.

region/maridia/inner-green/Oasis.json
blkerby: Oh neat, wasn't thinking about the asymmetry with left-facing shots.

region/maridia/inner-green/Oasis.json
```diff
@@ -2347,23 +3400,68 @@
...
+    {
       "link": [6, 4],
+      "name": "Use Flash Suit",
+      "requires": [
+        {"useFlashSuit": {}},
+        {"shinespark": {"frames": 17, "excessFrames": 2}}
+      ],
+      "flashSuitChecked": true,
+      "note": [
+        "Diagonal spark from the center platform, from a position hanging over the bomb blocks."
```
kjbranch, line 3430:
```suggestion
        "Diagonal spark from the center platform, from a standing position hanging over the bomb blocks."
```

kjbranch (commented):
Just looked through to see if there were changes needed based on the node change - i didnt check the cross room strats or carry flash suit stuff.

I do wonder if node 5 and even 6 could be deleted

region/maridia/inner-green/Oasis.json
```diff
@@ -894,44 +1188,171 @@
...
+      "entranceCondition": {
+        "comeInGettingBlueSpeed": {
+          "length": 0,
+          "openEnd": 1,
+          "minExtraRunSpeed": "$3.2",
+          "maxExtraRunSpeed": "$3.F"
+        }
+      },
+      "requires": [
+        {"notable": "Cross Room Jump Through Top Door"},
```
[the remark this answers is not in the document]
blkerby: I like the idea of changing the notable name, to more accurately describe what it includes. How about "Cross Room Jump Boomerang To Top"?

PR #2291, opened 2025-09-25
A few more East Sand Hall cross-room strats

blkerby:
The 1->2 "Cross-Room Tricky Spring Ball Bounce" feels like it might be significantly harder in its comeInWithSpringBallBounce form (especially if actually bouncing, and not just mockballing) compared to the comeInRunning form; so it might need additional requirements, but I wasn't sure.


blkerby:
I took a look in TAStudio and confirmed the left-to-right bounce does require more precision with a mockball/bounce entry compared to comeInRunning. With either a mockball/bounce entry it looks like it always requires a frame-perfect release of jump, regardless of the position you jump/bounce from; the exception would be if you can enter with low base speed (e.g. mockballing close to the door) but that has its own issues and we can't model it. On the other hand, with a comeInRunning entry in the scenario I tested I saw a 3-frame window for the jump release; I only TAS'd it in the one case (with speed $2.3) but it feels similarly more lenient for the other run speeds. So based on that, I added a `canBeVeryPatient` requirement to the `comeInWithSpringBallBounce` version. I'm guessing both notables will be Extreme and then this one case gets bumped up to Insane.

region/maridia/inner-green/East Sand Hall.json
```diff
@@ -999,10 +1129,55 @@
...
+          "speedBooster": true,
+          "movementType": "controlled",
+          "remoteAndLandingMinTiles": [[6.4375, 1]]
+        }
+      },
+      "requires": [
+        {"notable": "Cross-Room Tricky Spring Ball Bounce (Right to Left)"},
+        "canTrickyDashJump",
+        "canTrickySpringBallBounce",
+        "canPlayInSand"
```
osse101, line 1172: Highly recommend canSpringFling.  Getting out of the sand on the left is pretty hard.

region/maridia/inner-green/East Sand Hall.json
```diff
@@ -814,6 +870,80 @@
...
+      "requires": [
+        "canPreciseSpaceJump",
+        "canMomentumConservingMorph",
+        "canTrickyDodgeEnemies",
+        "canPlayInSand"
+      ],
+      "flashSuitChecked": true,
+      "note": [
+        "Space Jump into the room relatively low through the doorway.",
+        "Do a momentum-conserving morph on the ceiling."
```
osse101, line 913: This may make people overthink the morph.  You are aiming for the slopey ceiling trap to morph against instead of the flat ceiling, so its a regular morph as soon as you enter the room.

blkerby: Added some more detail to the note to try to clarify. From what I can tell, morphing too early will bonk the corner and fail, so it does need to be a momentum conserving morph, and you're right it's about sliding past the corner rather than the sandfall ceiling. https://videos.maprando.com/video/8138

region/maridia/inner-green/East Sand Hall.json
osse101: I couldn't make it to the end of the sand fall most of the time.  I'll have to look again

osse101: Guess I was unmorphing too late.

PR #2292, opened 2025-09-27
Norfair East check flash suits pt. 1 (Cathedral)

blkerby:
This includes significant updates to Cathedral heat frames. 

Initially I was looking at the right-to-left Cathedral traversal, to make sure killing the Sova worked with a flash suit (seems fine since you shoot it horizontally). Based on how precisely the strat modeled the extra 5 heat frames from not having a weapon, and based on the not-round-value of 176 used for the 4->1 heat frames, I gathered that this was intended to model a tankless traversal on Insane; however, the total energy used added up to 99, which wouldn't be enough to survive tankless (it would need to be 98).

The requirements here are tricky because node 4 is slightly nebulous, and because things can depend on which side you're coming from and exactly how the Gerutas happen to move. I took the opportunity to tighten things where I noticed they were loose. The tankless right-to-left traversal did seem reasonable for Insane, with 98 energy consumed. The 176 heat frames for 4->1 seemed impossibly tight (even accounting for node 4 being possibly anywhere on the two platforms it touches), but the 2->4 heat frames were loose so it worked out.

It looked to me like the Geruta obstacle "A" wasn't really useful: even with just Power Beam, it seemed better to kill the Geruta on the way down to the item rather than tank a hit from it, even accounting for the possibility that you may have Gravity (for enemy damage reduction) and that the item could be an ETank.

### DIFF
--- a/region/norfair/east/Bat Cave.json
+++ b/region/norfair/east/Bat Cave.json
@@ -109,7 +109,8 @@
           "length": 3,
           "openEnd": 1
         }
-      }
+      },
+      "flashSuitChecked": true
     },
     {
       "id": 2,
@@ -130,7 +131,8 @@
         {"cycleFrames": 300}
       ],
       "resetsObstacles": ["A"],
-      "farmCycleDrops": [{"enemy": "Skree", "count": 3}]
+      "farmCycleDrops": [{"enemy": "Skree", "count": 3}],
+      "flashSuitChecked": true
     },
     {
       "id": 3,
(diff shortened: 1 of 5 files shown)


region/norfair/east/Bat Cave.json
```diff
@@ -154,6 +157,7 @@
         "canTrickyUseFrozenEnemies",
         {"heatFrames": 260}
       ],
+      "flashSuitChecked": true,
```
osse101, line 160: This is complex carry

osse101 (commented):
Bubble Mountain `212`  32, 31, 172, 173, 174, 175, 176, 223 all seem fine

osse101 (approved):

PR #2293, opened 2025-09-27
Main Hall cross with Crystal Flash strats

blkerby:
Main Hall is one of the rooms that I skipped earlier when adding "Cross with Crystal Flash" strats. I saw some videos from Sam here which gave the motivation to add them.

With recent changes to the randomizer, we can now have multiple Crystal Flash options within the same strat and it can choose the best one, so I took advantage of that for the 1->4 strat.

Videos for all of these are on the video site.

### DIFF
--- a/region/lowernorfair/east/Main Hall.json
+++ b/region/lowernorfair/east/Main Hall.json
@@ -174,13 +174,49 @@
     {
       "id": 5,
       "link": [1, 4],
-      "name": "TrickyJumps",
+      "name": "Tricky Jumps",
       "requires": [
         "h_navigateHeatRooms",
         "canTrickyJump",
         {"heatFrames": 570}
       ],
       "note": "Take the platforms two at a time by building some run speed for each jump on a one tile runway."
+    },
+    {
+      "link": [1, 4],
+      "name": "Tricky Jumps with Crystal Flash",
+      "requires": [
+        "canTrickyJump",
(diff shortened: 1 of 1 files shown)


region/lowernorfair/east/Main Hall.json
```diff
@@ -174,14 +174,50 @@
...
       "requires": [
         "h_navigateHeatRooms",
         "canTrickyJump",
         {"heatFrames": 570}
       ],
       "note": "Take the platforms two at a time by building some run speed for each jump on a one tile runway."
     },
+    {
+      "link": [1, 4],
+      "name": "Tricky Jumps with Crystal Flash",
```
osse101, line 187: You can cross with 2 Crystal Flashes no tanks.  Its not even that bad.
The 1 CF no tanks takes some frame maxing

blkerby: Ok, added strats for double and triple CF. For double CF, I could only get it with SpaceJump: https://videos.maprando.com/video/8214.

osse101: https://videos.maprando.com/video/8222
I went for the extra jump on CF2 because I had slow elevators on, and that was unreasonable.  
I hit the elevator at 54 energy being more safe.  The first series of jumps has to be perfect though.

PR #2295, opened 2025-09-28
Tighten Lava Grapple Tunnel grapple heat frames

blkerby:
Looked into this after noticing a video by Sam that showed you could traverse Lava Grapple Tunnel tankless with Grapple. It took a while to get the hang of how to optimize the swings, but it seems not too bad.

With the new amount of heat frames, crossing tankless would be in Expert logic. I think this may be ok, but if it's not then we could add some extra heat frames in an `or` with `canInsaneJump` or something. 

Videos for this are on the video site.

### DIFF
--- a/region/norfair/east/Lava Grapple Tunnel.json
+++ b/region/norfair/east/Lava Grapple Tunnel.json
@@ -102,8 +102,17 @@
       "name": "Grapple",
       "requires": [
         "Grapple",
-        {"heatFrames": 320}
+        {"heatFrames": 320},
+        {"or": [
+          {"noFlashSuit": {}},
+          {"and": [
+            "canMidAirMorph",
+            "canComplexCarryFlashSuit",
+            {"heatFrames": 130}
+          ]}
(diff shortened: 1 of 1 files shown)


osse101 (commented):
Thats crazy; I've been telling people it was 90 damage to logically grapple across for a long time.

320 frames feels too tight, I can barely do it.

In speedruns you optimize for lag reduction and consistency so I'd take a few extra points of damage due to that.

blkerby:
> Thats crazy; I've been telling people it was 90 damage to logically grapple across for a long time.
> 
> 320 frames feels too tight, I can barely do it.
> 
> In speedruns you optimize for lag reduction and consistency so I'd take a few extra points of damage due to that.

My best is 293 frames. Here's a better attempt: https://videos.maprando.com/video/8158 . The biggest things are keeping grapple maximally extended as much as possible, and re-pressing it as quickly as possible after releasing. The lag is helpful in slowing down the timing.

PR #2307, opened 2025-10-06
Norfair East check flash suits pt. 7 (Upper Norfair Farm Room)

blkerby:
Upper Norfair Farming Room has some significant refinements:
- Added a "Delayed Gamet Farm" where you bring the Gamets up.
- Added Base strats to cross the top of the room (1 <-> 3). It this uses 125 heat frames compared to the currently expected amount of 150 by stopping on the floating platform (node 4).
- Added `canJumpIntoIBJ` and `canUnmorphBombBoost` variants for getting up.
- Tightened some other heat frames, splitting options out by movement item where they were lumped together before.

### DIFF
--- a/region/norfair/east/Speed Booster Hall.json
+++ b/region/norfair/east/Speed Booster Hall.json
@@ -114,7 +114,8 @@
           "length": 7,
           "openEnd": 1
         }
-      }
+      },
+      "flashSuitChecked": true
     },
     {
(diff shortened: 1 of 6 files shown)


blkerby:
Upper Norfair Farming Room could probably use a room diagram update to show the 1 <-> 3 edge.

PR #2329, opened 2025-10-15
add slopespark to dachora room

nn357:
https://videos.maprando.com/video/8689 (leave right door)

https://videos.maprando.com/video/8687 (leave left door)

https://videos.maprando.com/video/8077 (regular slopespark)

### DIFF
--- a/region/brinstar/pink/Dachora Room.json
+++ b/region/brinstar/pink/Dachora Room.json
@@ -1235,6 +1235,7 @@
           "position": "bottom"
         }
       },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
       "flashSuitChecked": true
     },
     {


osse101:
You can bring a zeela to the dachora door for a `canFlashSuitDoubleSpark` like how we do for GMode setups; wait at the speed blocks in temp blue and break them when it gets close.  It takes 2 minutes, but I'd put canBeVeryPatient on it because its easy to fail.

osse101:
> That's 2 minutes (assuming done correctly) to get one attempt at the slopespark? If so, would that maybe be canBeExtremelyPatient? With Supers I imagine you could speed it up to more of canBeVeryPatient level, with the ability to retry the slopespark with the Zeela down there?

That's sensible.  Its not trivial to get the Zeela down there, but should be ok at that level. And supers would let you reattempt quickly.  Then it needs leniency supers and energy for retries, at insane.

PR #2330, opened 2025-10-15
add x-mode spikesuit to kraideyedoor room.

nn357:
https://videos.maprando.com/video/8696 [leave right door]

https://videos.maprando.com/video/8693 [leave left door]

https://videos.maprando.com/video/8691 [no leave]

### DIFF
--- a/region/brinstar/kraid/Kraid Eye Door Room.json
+++ b/region/brinstar/kraid/Kraid Eye Door Room.json
@@ -780,9 +780,11 @@
       ],
       "flashSuitChecked": true,
       "note": [
-        "Requires precice movement. Down-back during the fall to clear the lower platform while shooting out the blocks.",
-        "May be possible to open a green/red door with a super missile but it is on the edge of possibility.",
+        "Requires precise movement. Down-back during the fall to clear the lower platform while shooting out the blocks.",
         "Unmorphing on the first possible frame leaves in top position; second frame leaves in middle position."
+      ],
+      "devNote": [
+        "It may be possible to open a green/red door with a super missile but it is on the edge of possibility."
       ]
     },
     {


blkerby (dismissed):
Looks good. 

Not directly related to X-mode spikesuit, but there are some more spikesuit strats that I imagine should work here (though I haven't tested most of these):
- Enter shinecharged or shinecharging from the left or top-right, and gain a spikesuit + spark out the left door.
- Enter shinecharged or shinecharging from the right door, and gain a spikesuit + spark out the right door (bottom position).
- Enter with a flash suit (or blue suit) and use it to spikesuit + spark out the left or right door. You could spark out the right door in either position in this case.

Feel free to either add them here or they can be looked at later.

region/brinstar/kraid/Kraid Eye Door Room.json
```diff
@@ -651,6 +740,40 @@
...
+        }
+      },
+      "requires": [
+        "canFlashSuitDoubleSpark",
+        {"shineChargeFrames": 175},
+        "canDownBack",
+        "canShinechargeMovementTricky",
+        {"thornHits": 1},
+        "canSpikeSuit",
+        {"shinespark": {"frames": 20}}
```
blkerby:
```suggestion
        {"shinespark": {"frames": 19}}
```
It's ok to put this as 19 the same as the others (even though in my one video I sparked from further away, using an extra energy).

region/brinstar/kraid/Kraid Eye Door Room.json
```diff
@@ -1080,6 +1203,86 @@
...
+        "canBeVeryPatient",
+        "h_ThornXModeSpikeSuit",
+        {"shinespark": {"frames": 31}}
+      ],
+      "exitCondition": {
+        "leaveWithSpark": {"position": "top"}
+      },
+      "note": [
+        "A one frame dash jump will give more leniency to land in the thorns.",
+        "Only first frame spikesuit will clear the pillar before the right door."
```
blkerby:
```suggestion
        "Use the thorns on the left side of the room.",
        "A 1-frame dash jump will give more leniency to land in the thorns.",
        "Only unmorphing on the first possible frame will work; second frame will bonk the pillar."
```

blkerby (dismissed):

blkerby (dismissed):
One other thought, instead of `{"useFlashSuit": {}}` maybe it would make sense to make a helper to cover both flash suit and blue suit since they should work equally well here?:

```
        {
          "name": "h_useStoredSpark",
          "requires": [
            {"or": [
              {"useFlashSuit": {}},
              {"useBlueSuit": {}}
            ]}
          ]
        },
```

This would be a blue-suit-to-flash-suit conversion strat, so it can also make sense to postpone to later since there would be other strats for converting too.

region/brinstar/kraid/Kraid Eye Door Room.json
```diff
@@ -651,6 +745,40 @@
...
+      },
+      "unlocksDoors": [
+        {"types": ["ammo"], "requires": ["never"]}
+      ],
+      "note": [
+        "Requires precice movement. Down-back during the fall to clear the lower platform while shooting out the blocks."
+      ],
+      "devNote": [
+        "May be possible to open a green/red door with a super missile but it is on the edge of possibility.",
+        "Unmorphing on the first possible frame leaves in top position; second frame leaves in middle position."
```
osse101: - move this line into `note`, like the others
- The frame you unmorph on effecting the height you leave at should be added to the tech description.  Only, this matters on the `canFlashSuitDoubleSpark` instead of the spikesuit, but not on every `canFlashSuitDoubleSpark` source.  I'd add something to the `canFlashSuitDoubleSpark` description.

blkerby (approved):

PR #2339, opened 2025-10-19
west sand hall reserve dboost

internet-account2:
I'm not sure if the comment about 2 energy is too wordy, could probably also be simply dropped.

### DIFF
--- a/region/maridia/inner-green/West Sand Hall.json
+++ b/region/maridia/inner-green/West Sand Hall.json
@@ -1374,11 +1374,14 @@
       "flashSuitChecked": true,
       "note": [
         "Use the projectile from a fully submerged Evir to boost Samus into the sandfall, providing just enough height to make the first and hardest jump.",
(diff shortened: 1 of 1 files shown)


region/maridia/inner-green/West Sand Hall.json
```diff
@@ -1374,6 +1374,7 @@
       "flashSuitChecked": true,
       "note": [
         "Use the projectile from a fully submerged Evir to boost Samus into the sandfall, providing just enough height to make the first and hardest jump.",
+        "With reserves, if using auto mode instead of pause abuse to tank the hit, set reserves to 1 energy to use maximum knockback frames; 2 would already make it more difficult.",
```
osse101: I'd like to see this moved to the detailNote and to be shortened.  Most practically, it should say to use Reserves set to manual and not auto.  But it could be interesting to mention 1 energy of auto is possible.
Setting reserves to exactly 1 is a high level tech too.

The tech requirements aren't impactful here but need to be looked at projectwide for other damageboosts.

PR #2350, opened 2025-10-22
Green Brinstar Early Supers Room R-Mode Spark Interrupt

aquanight:
Video Link: https://videos.maprando.com/video/3024

Notable elements:
- Left Door has a respawning enemy, so no need to CF or manage energy capacity.
- Access to top shelf copies from the the other [2, 4] strats.
- Left door can run across and do the right door strat, or as an alternative: energy setup using the respawning Zeb, then run across the bottom way and shinecharge over the hopper pit. Needs only Morph to get out, since the blue suit will clear the bomb blocks.

### DIFF
--- a/region/brinstar/green/Early Supers Room.json
+++ b/region/brinstar/green/Early Supers Room.json
@@ -517,7 +517,7 @@
         "then perform a 'slow short-charge' across the speedway: you must hold run at least until you have enough speed to get under the shutters.",
         "1-tap shinecharge is used to shinecharge over the Small Sidehopper Pit, before passing the 2nd to last shutter. Fall in, take a hit for",
         "i-frames so you can land and be ready to wind up before you get hit again.",
-        "Once you have blue suit you can simply Morph and roll right through the bomb blocks to escape."
+        "Once you have blue suit you can simply Morph and roll through the bomb blocks to escape."
       ]
     },
     {


region/brinstar/green/Early Supers Room.json
```diff
@@ -496,6 +496,31 @@
...
+      "link": [1, 3],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "Morph",
+        {"refill": ["Energy"]},
+        "canSlowShortCharge",
+        {"canShineCharge": {"usedTiles": 30, "openEnd": 1}},
```
osse101: `canSlowShortCharge` with 30 tile runway doesn't seem right...
canSlowShortCharge is releasing run as you turn blue to continue running.  And is being phased out.

region/brinstar/green/Early Supers Room.json
```diff
@@ -496,6 +496,31 @@
...
     },
+    {
+      "link": [1, 3],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "Morph",
+        {"refill": ["Energy"]},
```
osse101: Might have to `OR:[canBePatient, h_disableEtanks]` in the case of big Etank numbers.  Even then I don't know how long that takes.

region/brinstar/green/Early Supers Room.json
blkerby: A regular 1-tap shortcharge seems to work fine: hold forward and then start holding dash just a moment later.

osse101: A single stutter is safe, and any level of 1-tap works.  So its not mechanically hard but is a strat you want to practice once or twice.  That means it needs more of a knowledge check tech and `canRModeSparkInterrupt` can work for that.

region/brinstar/green/Early Supers Room.json
```diff
@@ -496,6 +496,31 @@
...
+      ],
+      "clearsObstacles": ["A"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Use the respawning Zeb to set up with reserves and enough energy to tank one Small Sidehopper hit without triggering Reserves",
+        "then perform a 'slow short-charge' across the speedway: you must hold run at least until you have enough speed to get under the shutters.",
+        "Aim to shinecharge over the Small Sidehopper Pit, or at least under the 2nd to last shutter. Fall in, take a hit for i-frames so you can",
+        "land and be ready to wind up before you get hit again.",
+        "Once you have blue suit you can simply Morph and roll right through the bomb blocks to escape."
```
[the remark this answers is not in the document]
aquanight: Done, though I wonder if this won't eventually be covered under `h_bombThings`.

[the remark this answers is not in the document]
blkerby: > I wonder if this won't eventually be covered under `h_bombThings`.

I think it may be better for blue suit strats to be kept separate. Otherwise, in map rando spoiler routes, for example, it would be hard for players to see that a blue suit was expected; e.g. it could just look like they're expected to get through bomb blocks with a Base strat, without Bombs or Power Bombs.

region/brinstar/green/Early Supers Room.json
```diff
@@ -978,6 +1030,43 @@
...
+          "canSpringBallJumpMidAir",
+          "h_crouchJumpDownGrab"
+        ]},
+        {"or": [
+          "h_CrystalFlash",
+          {"and": [
+            {"disableEquipment": "ETank"},
+            {"resourceAvailable": [{"type": "Energy", "count": 80}]},
+            {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
+            {"resourceMissingAtMost": [{"type": "Super", "count": 0}]},
```
blkerby: There's enough Wavers up there that it's not exactly necessary to rely on being full Missiles/Supers. If we put the assumed energy available to 95, that could work because it would be almost guaranteed to get at least one energy drop. On the other hand, Missiles and Supers can also be farmed here, so being full on them is also not much of a restriction. If the neighboring room involved a heated R-mode setup, the energy requirement could possibly be the more impactful one. So probably it is best done the way you have it here.

PR #2352, opened 2025-10-23
Green Brinstar Etecoons Energy Tank Room R-Mode Spark Interrupt

aquanight:
This one is a bit of a doozy. There are two options for R-Mode Spark Interrupts, partly depending on which door is entered.

It's not straight-forward because the in-room runway is too far from the enemies to carry a shinecharge and separated by a one-way barrier (Ice X-Ray Standup Beetom Clip can provide two-way access for retries).

Option 1: Usable from upper doors only:
Shinecharge on the upper runway, then drop through the crumbles and use thorns for spikesuit. With a flash suit stored, you have infinite time to farm/damage on Zebbos to prepare Energy, and then convert it to blue suit with reserve interrupt.
Possible alternative, if the shinecharge is missed or entered from bottom doors: Spike X-Mode using the thorn tunnel to shinecharge and then gain flash suit - however, in that case Option 2 is relatively easier.

Option 2: Usable from any door:
Prepare energy using a Zebbo farm, then perform Spike X-Mode on the thorns with dash speed to gain shinecharge. Time shinecharge and X-Ray release with i-frame refresh to avoid a second thorn hit, then get back to the Zebbo pipe and reserve interrupt.
For now I just have this on the bottom right door: once it's refined, it can be copied to other doors with added Morph requirement (using the left Zebbo pipe instead of the right).

### DIFF
--- a/region/brinstar/green/Etecoon Energy Tank Room.json
+++ b/region/brinstar/green/Etecoon Energy Tank Room.json
@@ -2163,7 +2163,7 @@
       "blueSuitChecked": true,
       "note": [
         "The flash suit must be gained in-room due to R-Mode. Refer to the 'Gain Flash Suit (Spikesuit)' strat for how to do this.",
-        "Set up energy as needed using the Zebbo farm. When ready, use the stored shinecharge and convert it into a blue suit.",
+        "Set up energy as needed using the Zebbo farm. When ready, use the stored shinecharge and convert it into a blue suit."
       ]
     },
     {


blkerby:
I updated the tests so that they would no longer give an error when including a `{"useFlashSuit": {}}` before `canRModeSparkInterrupt`. A `{"useFlashSuit": {}}` requirement is not necessary since you are already known to have a flash suit at this point and the `canRModeSparkInterrupt` will lose it anyway; but it wasn't right for it to be giving an error.

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -366,6 +366,39 @@
...
+      "name": "Spikesuit - R-Mode Spark Interrupt (Gain Blue Suit)",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"canShineCharge": {"usedTiles": 20, "openEnd": 1}},
+        "canShinechargeMovementTricky",
+        "canBeVeryPatient",
+        {"thornHits": 1},
+        "canTrickyCarryFlashSuit",
```
blkerby:
```suggestion
```
Seems like nothing special going on here with carrying the flash suit.

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -366,6 +366,39 @@
...
+        "canSpikeSuit",
+        "Morph",
```
blkerby:
```suggestion
```
Morph is already included in the `canSpikeSuit` requirement.

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -366,6 +366,39 @@
...
+        {"enemyKill": {
+          "enemies": [["Beetom"]],
+          "explicitWeapons": ["ScrewAttack", "Missile", "PowerBomb"]
```
blkerby: This list of weapons was used in the farming strats; it leaves out Supers though I can't remember if there is a good reason why. In any case, if we have an `enemyKill` option here it should probably just use `"excludedWeapons": ["Bombs"]`. But there would also be a `canTrickyDodgeEnemies` option that avoids the Beetoms and gets them stuck out of the way; at this level it could maybe just be expected, but should at least be an option.

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -2089,6 +2139,65 @@
       "name": "Base",
       "requires": [],
       "flashSuitChecked": true
+    },
+    {
+      "link": [7, 7],
+      "name": "Spikesuit - R-Mode Spark Interrupt (Gain Blue Suit)",
```
blkerby:
```suggestion
      "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
```
The note should also be updated to reflect that the flash suit is obtained prior to this strat, no longer as part of this strat itself.

aquanight: To follow with other flash-suit using strats, perhaps name should be "Use Flash Suit - R-Mode Spark Interrupt (Gain Blue Suit)"?

blkerby: Even better, yeah

region/brinstar/green/Etecoon Energy Tank Room.json
```diff
@@ -2089,6 +2139,62 @@
...
+        {"refill": ["Energy"]},
+        {"useFlashSuit": {}},
+        {"autoReserveTrigger": {"maxReserveEnergy": 95}},
+        "canRModeSparkInterrupt"
+      ],
+      "resetsObstacles": ["R-Mode"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "The flash suit must be gained in-room due to R-Mode. Refer to the 'Gain Flash Suit (Spikesuit)' strat for how to do this.",
```
blkerby: It's ok for now but usually I wouldn't recommend referring to another strat by name, because it's likely to become outdated

PR #2353, opened 2025-10-23
Green Brinstar Fireflea Room R-Mode Spark Interrupt

aquanight:
Reserve farming is very good here due to Firefleas guaranteeing Large Energy when full on PBs, but four is still not enough to completely fill an E-Tank.

If the runway is too tight, the thorns could be used for Spike X-Mode (will want a separate strat).

Not sure how to accurately reflect the correct energy state in the event of autoReserveTrigger, with the fireflea self-destructing and dropping a Large Energy right on top of Samus. This can only happen on the firefleam farm route, since in the CF route they will (except for a very very low chance) drop PBs instead.

Note that you should only farm one Fireflea. If you farm all four the room becomes very dark, and Samus's sprite is affected by the darkness in R-Mode, which can make it extremely difficult to see position or the magic frame when short-charging.

### DIFF
--- a/region/brinstar/green/Green Brinstar Fireflea Room.json
+++ b/region/brinstar/green/Green Brinstar Fireflea Room.json
@@ -187,6 +187,43 @@
       "flashSuitChecked": true
     },
     {
+      "link": [1, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"or": [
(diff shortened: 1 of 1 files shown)


blkerby:
> Not sure how to accurately reflect the correct energy state in the event of autoReserveTrigger, with the fireflea self-destructing and dropping a Large Energy right on top of Samus.

You could represent this as a `{"partialRefill": [{"type": "Energy", "count": 21}]}`. It would be good to include it because it can allow you to safely farm the room afterward, in case both neighboring rooms are heated.

PR #2362, opened 2025-10-26
R-Mode Spark Interrupt - Improved Reserve Refill modeling

aquanight:
Changes to the previous R-Mode Spark Interrupt strats to take a different approach to how tracking of energy collection is handling, for the non-Crystal Flash + non-respawning enemy paths.

This approach takes a few assumptions about player approach to handling energy in these rooms when preparing for R-Mode:
- That the player will have enough reserves that they can end up at full energy after entering R-Mode. This assumption is supported by using the destination room itself to farm energy, should nothing better be available, or else that logic simply won't use that room.
- That the player will farm mobs as needed for "enough" reserves to execute the strat, and will abort and re-farm and reset if they fail to get adequate drops. Any further farming can be done - and typically much more safely - once blue suit is gained.

The new helper, `h_RModeCanFillReserves` enforces the first point by:
- Ensuring minimum 99 energy when E-Tanks can be disabled.
- Ensuring no regular energy is missing otherwise, and requires `canRiskPermanentLossOfAccess` to use this path.

The `partialRefill` can now act under the assumption the player will take no further damage once they disable Energy Tanks. Any required damage would need to be placed in the requirements before invoking the helper - for example in Morph Ball Room, the `enemyDamage` from being jumped by a Sidehopper on entry is ordered this way - and this means if the player has such reserve capacity that they could tank the hit and still be able to disable down to 99, they can still proceed on this path.

In many cases, the `partialRefill` can simply assumes the player will get at least one large energy drop (thus setting the refill at 20), which now makes it logically safe to reset the room if only heated resets are available. However, in the case of Morph Ball Room - or similar rooms with poor energy drops, the assumption will be for a small energy with the baseline patience level for the strat (e.g. `canBePatient` for Morph Room), and large energy with greater patience (e.g. `canBeExtremelyPatient`). This represents the player being aware of their reserve state before committing to the R-Mode Spark Interrupt, and being able to abort if they don't get enough drops.

An option that may be on the table for later rooms in the R-Mode Spark Interrupt project is to consider the ability to abuse "low energy health drop bias" (aka health-bomb) in rooms with either large numbers of enemies or enemies with relatively high ratio of Large Energy drop compared to Small Energy, but this possibility will be addressed later.

### DIFF
--- a/region/brinstar/blue/Blue Brinstar Energy Tank Room.json
+++ b/region/brinstar/blue/Blue Brinstar Energy Tank Room.json
@@ -523,7 +523,7 @@
           {"and": [
             "h_RModeCanRefillReserves",
             {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
-            {"partialRefill": {"type": "ReserveEnergy", "limit": 10}}
+            {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
           ]}
         ]},
         {"canShineCharge": {"usedTiles": 30, "openEnd": 0}},
@@ -869,10 +869,9 @@
           {"and": [
             "Morph",
             "h_destroyBombWalls",
-            {"disableEquipment": "ETank"},
-            {"resourceAvailable": [{"type": "Energy", "count": 90}]},
+            "h_RModeCanRefillReserves",
             {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
-            {"partialRefill": {"type": "ReserveEnergy", "limit": 1}}
+            {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
           ]}
         ]},
         {"canShineCharge": {"usedTiles": 30, "openEnd": 0}},
--- a/region/brinstar/blue/Morph Ball Room.json
+++ b/region/brinstar/blue/Morph Ball Room.json
@@ -273,7 +273,7 @@
               ]},
               {"and": [
                 {"partialRefill": {"type": "ReserveEnergy", "limit": 20}},
-                "canBeVeryPatient"
+                "canBeExtremelyPatient"
               ]}
             ]}
           ]}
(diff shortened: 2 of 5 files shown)


aquanight:
> `{"resourceAvailable": [{"type": "RegularEnergy", "count": 99}]},` Won't match with rooms where you have to use some of your reserves to reach the Rmode room. I can only think of heat rooms with enemies that take a long time to reach the door that would be a problem, like pre-cathedral and ridley room, that would be a problem.

In the case of passing through heat to get to the R-Mode rooms, this would essentially require the player to have extra Energy/Reserve to be able to satisfy the ability to have the 99 (and thus be able to pick up reserve) upon entering R-Mode, or else be able to farm *somewhere*. Worst case they have to farm the R-Mode room by so far - of the rooms I've worked on already - only two have poor drop rates.

For the case of R-Mode Sparks in heated rooms (including Ridley, of all places!), they will naturally need a bit more attention. At current I am anticipating that the need to allow for very low energy states in R-Mode (to avoid permanent loss of access situations) may end up requiring `h_HeatProof` - in fact this may also be forced when using Crystal Flash at high energy capacity, but another option that could be considered is variant helpers that can aim for 199, 299, or even 399 energy (but could *only* be used with `disableEquipment: ETank`) where the player would populate the screen with a large number of drops, disable to a full tank, and then *quickly* pick them all up.
It is worth noting that it would only need heat concerns to get energy and a runway. Then you simply turn off Varia and start windup at 01 energy left.

> And you still would need a heavy farm room in the RMode room in that case. Its a similar small set of problem rooms to gaining the extra energy after the RMode to use to leave. But it would be nice if we could account for both not being full energy and healing extra after.

Due to the post-spark state, if the room's drops are bad it'll probably need to be out of logic anyway due to not being able to safely reset the room or otherwise farm back up.

osse101 (commented):
I think reworking this with something like `resourceCollect` is where we want to go with this.  And that may involve opening up the helper again, so I'm not sure that this does simplify requirements.  But it does have some nice parts.

The rooms that wouldn't work are few, and knowable.  They could stand out because standing at the door and waiting is all you need to do to wait for an enemy, and that is what wouldn't be logical.

PR #2363, opened 2025-10-27
add xmode bluesuit stuff

nn357:
not quite sure if the doublexmode tech is correct

### DIFF
--- a/tech.json
+++ b/tech.json
@@ -2514,6 +2514,26 @@
                 "Damage down to where auto-reserves will trigger after one more enemy hit.",
                 "Use a runway to gain a shinecharge, and press jump to activate shinespark wind-up near an enemy,",
                 "in such a way that the enemy hits Samus during the wind-up, triggering an R-Mode forced stand-up."
+              ],
+              "extensionTechs": [
+                {
+                  "name": "canRModePauseAbuseSparkInterrupt",
+                  "techRequires": [
+                    "canPauseAbuse",
+                    "canRModeSparkInterrupt"
+                  ],
+                  "otherRequires": [],
+                  "note": [
+                    "The ability to gain a blue suit by triggering an R-Mode forced stand-up during a shinespark wind-up,",
+                    "combined with using pause abuse to delay the interrupt, such as when the interrupting damage must be",
+                    "received with Samus not being in a shinespark windup. Execution is similar to a standard Spark Interrupt",
+                    "except that Reserves are set to Manual, and the player must press Start to pause just before taking the",
+                    "interrupting damage to reach 00 energy. The player must then initiate windup just before the pause fadeout",
+                    "completes. Once the game is paused, switching Reserve to Auto and unpausing will complete the R-Mode",
+                    "interrupt. The screen will remain completely black during this time, which can be corrected by pausing the",
+                    "game again."
+                  ]
+                }
               ]
             },
             {
@@ -2891,6 +2911,7 @@
                       ],
                       "extensionTechs": [
                         {
+                          "id": 220,
                           "name": "canRightSideDashlessDoorStuck",
                           "techRequires": [
                             "canRightSideDoorStuckFromWater"
@@ -3473,5 +3494,5 @@
       ]
     }
   ],
-  "nextTechId": 220
+  "nextTechId": 221
 }
--- a/helpers.json
+++ b/helpers.json
@@ -1421,8 +1421,16 @@
               "canInsaneJump"
             ]}
           ]
+        },
+        {
+          "name": "h_storedSpark",
+          "requires": [
+            {"or": [
+              {"useFlashSuit": {}},
+              {"blueSuitShinecharge": {}}
+            ]}
+          ]
         }
-
       ]
     },
     {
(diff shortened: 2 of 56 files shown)


tech.json
```diff
@@ -2552,6 +2552,63 @@
                 "This tech should not be used directly; instead use a helper,",
                 "in order to correctly account for the amount of Crystal Flash refill, which depends on the environment."
               ]
+            },
+            {
+              "name": "canXModeBlueSuit",
```
osse101: I think you should erase `canSuperJump` and apply it's ID here `160`
The only place it was used was in climb supers room, and it was to fall through the Speed blocks.

XMode__Suit techs are both SpeedBooster and XRay techs so just gotta pick one and where it is in SpeedBooster seems better.

Unless you have plans for upward momentum conservation that would fit a different structure better.

tech.json
```diff
@@ -2552,6 +2552,63 @@
...
+                "A blue suit will not be gained if dash is released after a shinespark has crashed against a wall or object.",
+                "Releasing dash mid shinespark will cause Samus to SuperJump and potentially go out of bounds but this can also be used to jump away from spikes or thorns.",
+                "Care must be taken around spikes or thorns after a blue suit is gained if there are active enemies as I-frames will expire quickly and escaping can be difficult."
+              ]
+            },
+            {
+              "name": "canDoubleXModeBlueSuit",
+              "techRequires": [
+               "canXModeBlueSuit",
+               "canBeVeryPatient",
```
osse101:
```suggestion
```
The maneuver isn't slow to pull of, so Patience would only mean lenience.  But with techs that is handled by tech difficulty placement. 
The usage of the tech may have lenience on each of the XModes separately, or it could focus more on the final XMode blue suit by assuming the player has practiced the trick enough to get it semi reliable.  
Then if there is a harder than normal instance of `canDoubleXModeBlueSuit`, it would either be a `notable` or gain `canBeExtremelyPatient`.  Since this would be placed in insane already.

osse101 (commented):
Making `canDoubleXModeBlueSuit` makes sense as a way to group a pair of tricks together to smartly handle them.  It would just be nice if the level of difficulty could be shared with e.g. flash suit into XModeBlueSuit.  

But "Use XMode to store a shinespark and quickly use it" is what this tech describes.
Unless it is adequate to say "Use XMode to store a shinespark" + "canShinechargeMovementTricky" + "Use a second frame perfect trick" + "Which uses the tech `canXModeBlueSuit` encloses this tech.  It is a little novel that you can use the xmode shinecharge right away in another xmode, but the clues exist.  

A tech that just combines frame precise tricks may be too vague, however.  But there are shared instances of this strat structure so it could work; flash suit + Xmode blue suit, blue suit + superjump, xmode run + Xmode blue suit.

How special does `canDoubleXModeBlueSuit` feel?

blkerby:
Using a separate helper for each combination sounds appealing as a way to streamline the strat logic and keep our options open about how many individual tech to add. E.g. we could have "h_spikeDoubleXModeBlueSuit", "h_thornDoubleXModeBlueSuit", "h_spikeFlashSuitIntoXModeBlueSuit", '"h_thornFlashSuitIntoXModeBlueSuit". I think regular canXModeBlueSuit feels like Extreme, i.e. where you use a runway near the spikes (in-room or cross-room) to gain the shinecharge. Doing it with a flash suit is mechanically easier but potentially more punishing, depending on if you are able to save with the flash suit and how close the save is. Because of the worst-case scenario being more bad, I imagine flash-suit into X-mode blue suit goes in Insane, but there could be value in it being a separate tech since I could see someone wanting to toggle it on before some of the more mechanically demanding difficult ones (like the double X-mode). 

At least for me, the double X-mode seems quite a bit harder to learn. I'm at a point now when I attempt double X-mode that after I'm warmed up I can almost always get the first X-mode but then usually mess up the second one; the extra time pressure makes it feel different and harder to get right. Especially when trying to minimize spike hits (to get a total of 3), there's a fairly narrow window to jump high enough to get the bounce for the second X-mode before i-frame runs out, assuming you have to bounce on the spikes for the second X-mode; depending on the room there could be something else to bounce on which would reduce time pressure since then it only comes down to the shinecharge frames. Either way, I imagine with more practice it could be pretty reliable, since in theory the most precise part comes down to "just" hitting two 2-frame windows in a row.

And yeah, using `canBeVeryPatient` to put not-slow strats into Insane has always been a bit of a dirty trick to put things into the right difficulty when we don't yet have a specific tech for the difficult thing, so we try to avoid that especially when there's a more specific tech to use instead. Probably all those uses will eventually be eliminated once the tech gets more fleshed out.

osse101:
`h_spikeDoubleXModeBlueSuit` and the thorns version can be added and have everything bundled together since it is in the same room.  But converting a flash suit or blue suit would have to be split up.  It would still be useful to write lenience into `h_spikeFlashSuitIntoXModeBlueSuit`, but it would also allow the randomizer to overwrite the helper with something like `never`, or `HasSaved & SaveFlashSuitQOLEnabled`.  Maybe we'll see how that side develops.
Keeping DoubleXMode as a tech makes sense to me.

tech.json
```diff
@@ -2552,6 +2552,62 @@
...
+                "Build the intial shinecharge by entering X-Mode with dash speed, once shinecharged release dash to exit X-Mode.",
+                "Before the shinecharged window expires, quickly re-enter X-Mode and peform a shinespark. Due to the short window, this second X-Mode has to be done first try.",
+                "Release dash during the shinespark wind-up animation to exit X-Mode, interrupting the shinespark and gaining a blue suit.",
+                "A blue suit will not be gained if dash is released after a shinespark has crashed against a wall or object.",
+                "Releasing dash mid shinespark will cause Samus to SuperJump and potentially go out of bounds but this can also be used to jump away from spikes or thorns.",
+                "Care must be taken around spikes or thorns after a blue suit is gained if there are active enemies as I-frames will expire quickly and escaping can be difficult."
+              ]
+            },
+            {
+              "name": "canSlopeXModeBlueSuit",
```
osse101: I don't think we need this tech.  It can be represented as `canSlopeXMode` + `canXModeBlueSuit`.
I don't think `canSlopeXMode` can build a shinespark so the tech has limited functionality.

nn357: agreed, I was thinking ti should be a helper or something but not sure why i wrote that into tech.. i've taken it out now.

PR #2365, opened 2025-10-29
Kraid Eye Door Room R-Mode Spark Interrupt

aquanight:
Room notes:
- Respawning Zeb.
- No runway, so X-Mode provides the shinecharge.
- Equal access from all doors.


osse101 (commented):
This XMode has `canBeVeryPatient` on it for being about a 2 frame jump before the XMode.  
I think you want to describe how to pull off the RModeSparkInterrupt too.  I imagine you spawn the bug before XModing, or can you run to the bug spawner after?

PR #2367, opened 2025-10-29
Kraid's Room R-Mode Spark Interrupt

aquanight:
Room Notes:
- Only possible while Kraid is alive, so hard `canRiskPermanentLossOfAccess`.
- Requires X-Mode for shinecharge.
- Crystal Flash can be used to gain energy without causing Kraid to rise (allowing for KQK after).
- Otherwise, farming nails requires Kraid standup.
- There's a supposed case where the one (and only) nail that can yield drops can be despawned, thereby preventing any further drops from being farmed. Details about what can cause this, and what kind of tech may be needed to avoid it, may need research.
  - If nothing else, it may be necessary to treat the nail like a non-respawning enemy.

### DIFF
--- a/region/brinstar/kraid/Kraid Room.json
+++ b/region/brinstar/kraid/Kraid Room.json
@@ -288,7 +288,10 @@
       "requires": [
         {"not": "f_DefeatedKraid"},
         "canRiskPermanentLossOfAccess",
-        "canBeVeryPatient",
+        {"or": [
+          "canBeVeryPatient",
+          {"disableEquipment": "ETank"}
+        ]},
         {"refill": ["Energy"]},
         {"or": [
           "Charge",
@@ -519,7 +522,10 @@
         {"or": [
           "h_CrystalFlash",
           {"and": [
-            "canBeVeryPatient",
+            {"or": [
+              "canBeVeryPatient",
+              {"disableEquipment": "ETank"}
+            ]},
             {"refill": ["Energy"]},
             {"or": [
               "Charge",


region/brinstar/kraid/Kraid Room.json
```diff
@@ -253,6 +253,40 @@
...
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"not": "f_DefeatedKraid"},
+        "canRiskPermanentLossOfAccess",
+        {"or": [
+          "h_CrystalFlash",
+          {"and": [
+            {"refill": ["Energy"]},
```
osse101: This needs DisableEtanks or canbeExtremely?Patient.  Or another stack of canRiskPermanentLossOfAccess to not collect etanks.

region/brinstar/kraid/Kraid Room.json
```diff
@@ -427,6 +461,50 @@
...
+              "canPrepareForNextRoom",
+              "h_KraidCameraFix"
+            ]},
+            "SpaceJump"
+          ]}
+        ]},
+        {"or": [
+          "h_CrystalFlash",
+          {"and": [
+            {"refill": ["Energy"]},
```
osse101: this too

aquanight:
Regarding the nail despawn issue: I asked around on the metconst discoard. PJBoy provided this video as a reference:
https://www.youtube.com/watch?v=R6nkgGb265Q

I believe the root cause is an enemy-projectile overload being caused by excessive dud-shot projectiles being spawned during the contact routines associated with Kraid's head and body.

If this is correct, it would allow that the despawn issue can be avoided by refraining from firing stray shots that strike Kraid while farming. I don't know if this would need special tech to represent this knowledge or the higher degree of shot accuracy, but it would at the very least be in an `or` against `ScrewAttack`.

region/brinstar/kraid/Kraid Room.json
aquanight: Comparing to farming bugs:

Nail drop rate is 0% Nothing and slightly more than 3-to-1 biased towards Large Energy when full on ammo. In terms of drop rate it is superior to Zebs (2% Nothing and even split between small/large) and worse than Zebbo (0% Nothing and 0% Small Energy).

Farm cycle is notably worse than both: only one nail can be farmed (the other must be avoided or destroyed), and this must be done while evading Kraid and his spike missiles.

I'm thinking maybe somewhere between `canBePatient` and `canBeVeryPatient`? Definitely `canDodgeWhileShooting`, maybe `canTrickyDodgeEnemies`?

PR #2369, opened 2025-10-31
Mission Impossible R-Mode Spark Interrupt

aquanight:
Room notes:
- Very short runway unless you take a spike hit and use i-frames.
- Sidehoppers again? And only two of them. Ew. This calls for a different approach...

R-Mode Spark can only be gained in the upper half. Bottom half would need a way to get past the crumble block (perhaps a G-Mode strat to visit later?).

In this room, the favored strat for initiating the R-Mode Spark Interrupt is to just get rid of the Sidehoppers (and hope they drop some energy or else it's Crystal Flash time) - we're doing this with spikes!

As explained in the note: the method to spark interrupt off of an immobile damage source (such as spikes, thorns, heat, lava, or acid) is a little different than typical:
1. Set reserves to Manual
2. Get shinecharge.
3. Press Start and immediately take the last hit for 00 energy.
4. At the last moment before the game pauses, begin the windup. (Note for future rooms using this method: be aware of any special movement needs to ensure Samus can do this - air physics and/or walking into spikes works fine, for example, and no-knockback sources like heat also work.)
5. In the pause menu, set reserves to Auto. Leave regular energy on 00 and unpause.
6. Listen to the audio for the auto reserve trigger to complete.
7. Pause and unpause the game again to restore visibility.

### DIFF
--- a/region/brinstar/blue/Blue Brinstar Boulder Room.json
+++ b/region/brinstar/blue/Blue Brinstar Boulder Room.json
@@ -150,6 +150,7 @@
       "flashSuitChecked": true
     },
     {
+      "id": 54,
       "link": [1, 1],
       "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
       "entranceCondition": {
@@ -160,8 +161,8 @@
         "h_CrystalFlash",
         {"disableEquipment": "Varia"},
         {"disableEquipment": "Gravity"},
-        {"canShineCharge": { "usedTiles": 16, "openEnd": 1 } },
-        {"autoReserveTrigger": { }},
+        {"canShineCharge": {"usedTiles": 16, "openEnd": 1}},
+        {"autoReserveTrigger": {}},
         "canRModeSparkInterrupt"
       ],
       "flashSuitChecked": true,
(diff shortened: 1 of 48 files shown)


region/brinstar/pink/Mission Impossible Room.json
```diff
@@ -929,6 +945,53 @@
...
+      ],
+      "resetsObstacles": ["R-Mode"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Kill both Sidehoppers for energy, or else Crystal Flash (kill both Sidehoppers anyway so you have the runway free).",
+        "To R-Mode Spark Interrupt from spikes, damage down to within 1 more spike hit of health and set reserves to Manual.",
+        "Gain shinecharge, press Start and immediately step on spikes during the fadeout. At the last moment before the game",
+        "pauses, begin spark windup. In the pause menu, set reserves back to Auto but do not refill any energy (unpause with",
+        "00 energy). As soon as you hear the reserve fill complete, pause and unpause again to restore visibility."
```
osse101: I'd like you to write this description into the `canRModeSparkInterrupt` tech description instead.  I don't think it is different enough to be an extension tech and we can attach movement type tech where needed. (`canRModeSparkInterrupt` is easy enough to do at lower difficulties, but there are so many additional requirements, like `canCarryBlueSuit`, that it may need to be placed high)

This probably works with any kind of pause abuse too, enemy hit/heat/spike.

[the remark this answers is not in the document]
osse101: > As soon as you hear the reserve fill complete

I mash start when the unpause is done

osse101: This can be a follow-up PR, I believe

osse101 (commented):
Adding an obstacle to every room that you can take damage in is going to be very computationally expensive.

aquanight:
> Adding an obstacle to every room that you can take damage in is going to be very computationally expensive.

It's not my plan to do this every time this comes up, but when there are many options to logically deal with a particular obstacle (in this case, the two Sidehopper kill had many options due to some being safe/less safe depending on whether the door could be opened), it does avoid the need to duplicate 4-5 strats worth of logic into a single strat.

And in this particular case, it potentially benefits from a G-Mode Direct entry -> back into bottom door -> X-Ray climb to enable access from the bottom door.

PR #2372, opened 2025-10-31
Waterway Energy Tank Room R-Mode Spark Interrupt

aquanight:
Room Notes:

- Two long runways: one needs some logic to get to, the other needs Gravity to use.
- Zeros are easily accessed for energy setup. The rest of the enemies can only be accessed for farming with Gravity (water physics speed means not all of the energy that drops could be collected). Alternatively they can be farmed after blue suit.

### DIFF
--- a/region/brinstar/pink/Waterway Energy Tank Room.json
+++ b/region/brinstar/pink/Waterway Energy Tank Room.json
@@ -141,8 +141,10 @@
           ]}
         ]},
         {"autoReserveTrigger": {"maxReserveEnergy": 95}},
-        "canRModeSparkInterrupt"
+        "canRModeSparkInterrupt",
+        {"partialRefill": {"type": "Energy", "limit": 25}}
       ],
+      "clearsObstacles": ["A"],
       "flashSuitChecked": true,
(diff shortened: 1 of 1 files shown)


region/brinstar/pink/Waterway Energy Tank Room.json
```diff
@@ -97,6 +97,60 @@
...
+                    "canTrickyUseFrozenEnemies",
+                    {"ammo": {"type": "Super", "count": 1}}
+                  ]}
+                ]}
+              ]}
+            ]},
+            {"canShineCharge": {"usedTiles": 32, "openEnd": 1}}
+          ]}
+        ]},
+        {"autoReserveTrigger": {"maxReserveEnergy": 95}},
```
osse101: You probably want to give some dummy real energy, from farming the fish, so you can reload the room incase of heat on the other side.

aquanight: Added 20 energy for 3 Puyos. I don't trust the Skulltera's drop rate enough.

PR #2374, opened 2025-10-31
Dachora Room R-Mode Spark Interrupt

aquanight:
Room Notes:

- Fairly standard room with non-respawning enemies and full-length runway.
- Access from the bottom door needs a shinespark, which incurs mandatory damage before farming/CF.
- Speed block pit may trip players up in execution, but logically has no requirements to avoid. Consider lenience options?

### DIFF
--- a/region/brinstar/pink/Dachora Room.json
+++ b/region/brinstar/pink/Dachora Room.json
@@ -746,7 +746,7 @@
           {"shinespark": {"frames": 110, "excessFrames": 3}}
         ]},
         {"or": [
-          {"resourceAvailable": {"type": "ReserveEnergy", "count": 1}},
+          {"resourceAvailable": [{"type": "ReserveEnergy", "count": 1}]},
           {"and": [
             "h_RModeCanRefillReserves",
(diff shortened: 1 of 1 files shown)


region/brinstar/pink/Dachora Room.json
```diff
@@ -879,6 +961,33 @@
...
+        "h_shinechargeMaxRunway",
+        {"autoReserveTrigger": {}},
+        "canRModeSparkInterrupt"
+      ],
+      "clearsObstacles": ["A"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Crystal Flash after getting past the Metarees, or else run through the bomb wall and farm Reos and Zeelas. Leave",
+        "one Reo or Zeela alive. Free full runway to shinecharge then get hit by the remaining enemy to interrupt."
```
osse101:
```suggestion
        "one Reo or Zeela alive. Build a Shinespark then get hit by the remaining enemy to interrupt."
```

PR #2377, opened 2025-10-31
Climb R-Mode Spark Interrupt

aquanight:
Room Notes:

- Requires Zebes Awake for the standard version.
- Wall pirate drop rate is pretty good when full on Missiles.
- The bottom most pirate is best to lead towards the bottom of the room for interrupt. The player can start the windup anywhere in the line of sight to trigger a shot and jump.
- All standard strats are turned off during `f_ZebesSetAblaze` due to lack of wall pirates (and need for special attention due to the rising acid) - since this is the escape flag, it is *not* additionally flagged as permanent loss of access.
- Gaining bluesuit will logically clear the bomb wall, even if it didn't have to be cleared for the shinecharge. But this is fine since blue suit will clear it anyway.

Special Notes for `f_ZebesSetAblaze`:
- Need enemyKill requirements for the special pirates. (Plasma Beam? Ammo? Hyper Beam obviously works too.)
- Two strats are added for this condition, from each of the bottom doors. The theory of the strat is the same: clear the special pirates (they drop nothing) and then climb up the platforms for a Crystal Flash safely away from the acid.
- From there we get to use an acid-dip R-Mode Spark Interrupt. The expected reserve content after CF needed special attention due to the effects of high e-tanks. Without disableable e-tank, the auto-reserve trigger is capped at 50 (49 before CF + 1500 CF energy - 1499 energy cap = 50 reserves). With disabled e-tanks, the player can expect to get the full 400. I expect to see more use of Disable-ETank-aware CF planning when heated rooms start coming up.
- By this point, the acid has already risen considerably. Under limited reserves, the best option is to escape out the bottom right door and reset the room (and the acid) to carry the blue suit up Climb. With 400 reserves, it may be possible to get back above the acid if the goal is to carry it upwards. Either way, blue suit doesn't impede vertical movement in Climb.
- Acid frame counts need review.

### DIFF
--- a/region/crateria/central/Climb.json
+++ b/region/crateria/central/Climb.json
@@ -935,7 +935,7 @@
           {"enemyDamage": {"enemy": "Custom Climb Pirate", "type": "contact", "hits": 1}},
           {"enemyKill": {
             "enemies": [ ["Custom Climb Pirate", "Custom Climb Pirate"]],
-            "explicitWeapons": ["Charge+Plasma", "Super", "PowerBomb"]
+            "explicitWeapons": ["Plasma", "Super", "PowerBomb"]
           }}
         ]},
         "h_CrystalFlash",


region/crateria/central/Climb.json
```diff
@@ -881,6 +922,48 @@
...
       ]
     },
+    {
+      "link": [2, 5],
+      "name": "Zebes Ablaze - R-Mode Spark Interrupt (Gain Blue Suit)",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "Gravity",
```
osse101: The enemyDamage event is a little more important than a way to kill them.  But you need way past the pirates before you can CF.

region/crateria/central/Climb.json
```diff
@@ -881,6 +922,48 @@
...
+        {"or": [
+          {"and": [
+            {"disableEquipment": "ETank"},
+            {"refill": [ "ReserveEnergy" ]},
+            {"autoReserveTrigger": {}}
+          ]},
+          {"autoReserveTrigger": {"maxReserveEnergy": 50}}
+        ]},
+        "canRModeSparkInterrupt",
+        {"acidFrames": 80}
```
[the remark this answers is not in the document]
aquanight: I'd considered if it might be viable to, say, aim for trying to quickly get up the shaft if you wanted to get out the top door, but after thinking about it, it'd probably be less acid exposure to just bail and reset than to try to go up, so I added the reset requirement.

While this works for randomizers like MapRando where all doors are effectively blue (but maprando doesn't use this for escape logic... yet), it may need to be revisited in the vanilla condition where all doors except the top are grey doors.

region/crateria/central/Climb.json
aquanight: What I added will assume taking a hit from the pirates before CF if you don't have some weapons that can kill them fast (supers, any kind of Plasma, or power bomb). Charge+IWS can also quickly kill one but has to recharge to kill the other, which may be too much on top of the extra shot cooldown due to X-Ray.

PR #2386, opened 2025-11-06
add x-mode spikesuit and bluesuit to etecoon etank

nn357:
I removed the shinespark requirement from the double x-mode helpers.. there is already a shinespark requirement in the single helper which is always referenced in the double helper]

https://videos.maprando.com/video/9038 [xmode spikesuit]

https://videos.maprando.com/video/9053 [bluesuit - come in shinecharged]

https://videos.maprando.com/video/8851 pbluesuit - double xmode]

### DIFF
--- a/region/brinstar/green/Etecoon Energy Tank Room.json
+++ b/region/brinstar/green/Etecoon Energy Tank Room.json
@@ -981,8 +981,8 @@
         {"shineChargeFrames": 135},
         "h_XModeThornHit",
         "canSuperJump",
-        "canXModeBlueSuit",
-        {"shinespark": {"frames": 7}}
+        {"shinespark": {"frames": 7}},
+        "canXModeBlueSuit"
       ],
       "flashSuitChecked": true
     },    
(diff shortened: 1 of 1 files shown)


blkerby:
Looks nice. Some more that could be added:
- `canSuperJump` variants that clip up through the crumbles could be added (both comeInShinecharged and double X-mode versions). 
- A right-side comeInShinecharged blue suit strat could be added.

In theory using the runway at the top of the room is also possible, but it's a first-try trick and seems not really useful since it's probably easier/better to get a spikesuit, farm up, and convert to blue suit (once we add the tech/helper for that).

nn357:
> Looks nice. Some more that could be added:
> 
> * `canSuperJump` variants that clip up through the crumbles could be added (both comeInShinecharged and double X-mode versions).
> * A right-side comeInShinecharged blue suit strat.
> 
> In theory using the runway at the top of the room is also possible, but it's a first-try trick and seems not really useful since it's probably easier/better to get a spikesuit, farm up, and convert to blue suit (once we add the tech/helper for that).

Done.

I gave 7 shinespark frames for the super jump (there is a leniency window of what will clip and what will crash into the wall) between 4-7 frames of spark is the sweet spot, anything more and you risk crashing into the wall of going oob. So I set it at 7.

PR #2392, opened 2025-11-07
Gauntlet Entrance R-Mode Spark Interrupt

aquanight:
Room Notes:

- Bomb barriers block access to the runway. Traversal strats for Screw Attack, Bombs, and Power Bombs (with the Crystal Flash option mixed in - the Crystal Flash should be used *last*) were incorporated with a slight adjustment to account for farming strategies.
- Specifically: options involving killing Yapping Maws was split into a separate strat, that in turn assumes you will just kill all four Yapping Maws. Doing so already provides likely enough energy without having to farm Wavers too.
- The runway may need to be double-checked for acid constraints.

### DIFF
--- a/region/crateria/west/Gauntlet Entrance.json
+++ b/region/crateria/west/Gauntlet Entrance.json
@@ -123,7 +123,13 @@
           ]}
         ]},
         {"autoReserveTrigger": {"maxReserveEnergy": 95}},
-        "canRModeSparkInterrupt"
+        {"or": [
+          {"and": [
+            "canInsaneJump",
+            "canRModeSparkInterrupt"
+          ]},
+          "canRModePauseAbuseSparkInterrupt"
+        ]}
       ],
       "clearsObstacles": ["A"],
       "flashSuitChecked": true,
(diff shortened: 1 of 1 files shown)


region/crateria/west/Gauntlet Entrance.json
```diff
@@ -80,6 +80,103 @@
...
+      "blueSuitChecked": true,
+      "note": [
+        "Clear the bomb barriers and farm two Wavers. You can keep one to interrupt, but if you need to",
+        "farm it, you can still use the acid to interrupt.",
+        "If you mess up the spark position, remember to keep at least any button held to prevent a Yapping Maw crash."
+      ]
+    },
+    {
+      "link": [1, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit) - Yapping Maw Kill",
```
osse101: This isn't logically unique enough from the other strat so there is no reason to include it.

region/crateria/west/Gauntlet Entrance.json
```diff
@@ -80,6 +80,103 @@
...
   ],
   "strats": [
+    {
+      "link": [1, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "canTrickyJump",
```
osse101:
```suggestion
        "canInsaneJump",
```

PR #2395, opened 2025-11-07
Terminator Room R-Mode Spark Interrupt

aquanight:
Room notes:

- Perhaps the simplest room in West Crateria: just six Geemers and a full-length runway. Wavers make for better interrupting than farming.

Oh also there's an item in the way, so it gets collected.

### DIFF
--- a/region/crateria/west/Terminator Room.json
+++ b/region/crateria/west/Terminator Room.json
@@ -107,7 +107,7 @@
           "h_CrystalFlash",
           {"and": [
             "h_RModeCanRefillReserves",
-            {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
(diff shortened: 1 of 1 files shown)


region/crateria/west/Terminator Room.json
```diff
@@ -96,6 +96,32 @@
...
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"or": [
+          "h_CrystalFlash",
+          {"and": [
+            "h_RModeCanRefillReserves",
+            {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
+            {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
```
osse101: At full missiles you'll get a pile of energy.  You could reach 20 reserve without being full on missiles.

PR #2396, opened 2025-11-07
Wasteland R-Mode Spark Interrupt

aquanight:
The first of many heated rooms!

And because the pitiful 20 energy you can get from Disgeegas is nowhere near enough to escape the room, it requires `h_heatProof`.

Possibility to evaluate:
Disable E-Tanks + CF to force full reserves -> Will this be enough to escape heated?

### DIFF
--- a/region/lowernorfair/east/Wasteland.json
+++ b/region/lowernorfair/east/Wasteland.json
@@ -674,46 +674,31 @@
       "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
       "requires": [
         {"obstaclesCleared": ["R-Mode"]},
-        {"obstaclesNotCleared": ["C"]},
-        "h_heatProof",
+        {"obstaclesCleared": ["C", "D"]},
         {"or": [
+          "h_heatedCrystalFlashForReserveEnergy",
           {"and": [
-            "h_heatedCrystalFlash",
-            {"obstaclesCleared": ["D"]}
-          ]},
-          {"and": [
-            {"obstaclesNotCleared": ["D"]},
             "h_RModeCanRefillReserves",
             {"or": [
+              {"partialRefill": {"type": "ReserveEnergy", "limit": 5}},
               {"and": [
-                "canBePatient",
-                {"partialRefill": {"type": "ReserveEnergy", "limit": 20}},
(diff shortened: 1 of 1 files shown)


region/lowernorfair/east/Wasteland.json
```diff
@@ -641,6 +669,44 @@
...
+        "h_heatProof",
+        {"or": [
+          {"and": [
+            "h_heatedCrystalFlash",
+            {"obstaclesCleared": ["D"]}
+          ]},
+          {"and": [
+            {"obstaclesNotCleared": ["D"]},
+            "h_RModeCanRefillReserves",
+            {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
```
osse101: I was killing them with missiles and could get health drops fine

region/lowernorfair/east/Wasteland.json
```diff
@@ -641,6 +669,57 @@
...
+              ]},
+              {"and": [
+                {"or": [
+                  {"partialRefill": {"type": "ReserveEnergy", "limit": 5}},
+                  {"and": [
+                    "canBeVeryPatient",
+                    {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
+                  ]}
+                ]},
+                {"enemyKill": {"enemies": [["Dessgeega"], ["Dessgeega", "Dessgeega", "Dessgeega"]]}}
```
osse101: Charge and Morph Bombs probably need patience.
Pseudo screw is out.

aquanight: Hopefully excluding Bombs/Charge from the "not patient" route works. I think this only excludes Charge Power Beam so if I need to come back and get Charge + (some mix of Ice/Wave/Spazer) I could add those.

Plasma (with or without Charge) definitely should not require patience.

region/lowernorfair/east/Wasteland.json
```diff
@@ -641,6 +669,57 @@
       "clearsObstacles": ["A", "B"],
       "flashSuitChecked": true
     },
+    {
+      "link": [4, 4],
```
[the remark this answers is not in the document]
osse101: - You could fake it.  If the player is at 4 with all desgeegas dead and at full energy, they probably put a little energy in reserves and had the option of performing the strat before killing the last one.  
- Or use heat to trigger reserves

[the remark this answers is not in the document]
aquanight: Let's see if that works.

(I may need to create a helper for heat triggering.)

PR #2399, opened 2025-11-09
Add canRModeCrystalFlashInterrupt

kjbranch:


tech.json
```diff
@@ -2387,6 +2387,44 @@
...
+                {"resourceAvailable": [{"type": "Super", "count": 10}]},
+                {"resourceAvailable": [{"type": "PowerBomb", "count": 10}]},
+                {"gainFlashSuit": {}}
+              ],
+              "note": [
+                "The ability to gain a flash suit by interrupting a Crystal Flash with an R-mode forced standup.",
+                "This can be done by taking enough damage to force an auto reserve trigger after the Crystal Flash has at least partially filled the Reserve Energy."
+              ],
+              "detailNote": [
+                "It is typically required to have a small number of E-Tanks in order to be able to be able to take enough damage to trigger reserves.",
```
blkerby:
```suggestion
                "It is typically required to have a small number of E-Tanks in order to be able to take enough damage to trigger reserves.",
```

tech.json
```diff
@@ -2387,6 +2387,44 @@
...
+                "canRMode",
+                {"tech": "canCrystalFlash"},
+                "canCarryFlashSuit"
+              ],
+              "otherRequires": [
+                {"resourceAtMost": [{"type": "RegularEnergy", "count": 50}]},
+                {"resourceAtMost": [{"type": "ReserveEnergy", "count": 0}]},
+                {"or": [
+                  {"disableEquipment": "ETank"},
+                  "canRiskPermanentLossOfAccess"
```
blkerby: I think the `canRiskPermanentLossOfAccess` option would need additional requirements to express the low amount of E-Tanks. Might be better to drop it for now?

PR #2406, opened 2025-11-11
Add canTrickyRModeCrystalFlashInterrupt, add some uses in Maridia

kjbranch:
This relies on #2399 and should not be merged before it

We were thinking that this would probably be in Map Rando's `Extreme`, along with the non-tricky variant. The non-tricky variant is probably the easiest way to obtain a flash suit, and could have been lowered to Expert, if it weren't so difficult to traverse rooms with a flash suit.

### DIFF
--- a/region/wreckedship/main/Spiky Death Room.json
+++ b/region/wreckedship/main/Spiky Death Room.json
@@ -223,7 +223,7 @@
         "comeInWithRMode": {}
       },
       "requires": [
-        "canRModeCrystalFlashInterrupt",
+        "canTrickyRModeCrystalFlashInterrupt",
         {"ammo": {"type": "Missile", "count": 2}}
       ],
       "flashSuitChecked": true,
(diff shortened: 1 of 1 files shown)


region/maridia/inner-green/Shaktool Room.json
```diff
@@ -273,6 +273,93 @@
...
+      "requires": [
+        "canRModeCrystalFlashInterrupt",
+        {"or": [
+          {"and": [
+            {"ammo": {"type": "Missile", "count": 10}},
+            {"or": [
+              "h_partialEnemyDamageReduction",
+              {"and": [
+                {"resourceCapacity": [{"type": "RegularEnergy", "count": 149}]},
+                {"resourceCapacity": [{"type": "ReserveEnergy", "count": 149}]}
```
osse101: Energy would be set to 39 in both of these cases

PR #2415, opened 2025-11-15
Add implicitHeatFrames property to autoReserveTrigger

### LINKED ISSUE #2403: R-Mode Standup : Implicit heat drain is avoidable
Steps to Reproduce:

1. Be in R-Mode in a heated room. Turn off all suits.
2. Have X-Ray selected and be standing on normal ground.
3. Wait for the auto-reserve to activate.
4. Buffer X-Ray activation.
5. When the refill completes, release X-Ray.

X-Ray will generally need to be already selected. Investigation of the HUD handler suggests that it is possible a frame-perfect switch to X-Ray (having previously selected the item before it - or nothing if there are no items) on the frame reserve finishes can still allow execution of this technique, but once the reserve auto-trigger has engaged, the Item Select button will not respond until the trigger completes.

If done in unprotected heat (or acid/lava), the immediate activation of X-Ray will prevent the loss of energy that accumulates during the reserve trigger. Once X-Ray is released, Samus will resume play with the full reserve complement added to energy.

The following situations were tested to confirm this works:
- Standing on the ground.
- While having a flash suit, jumping at low energy remaining (performing R-Mode Spark Interrupt).

Tested and have not yet successfully created:
- R-Mode standup out of a spinjump.

Suspect to be possible:
- Standing on the ground or any other valid X-Ray pose while not in R-Mode - no forced standup (e.g. avoiding the need to manually draw reserves while running through heat at the cost of losing momentum).

For strats that use an `autoReserveTrigger` while inside a periodic damage hazard (such as heat, acid, or, lava), they need to be able to individually evaluate whether an X-Ray usage can be buffered to prevent the heat loss. Only triggers found in rooms containing heat, acid, lava, or Samus Eaters will need this.

Obviously, being able to do so is dependent on actually having X-Ray.

**Technical Details**

The variable at $7E:0A50 (unit) and $7E:0A4E (subunit) track the accumulation of "periodic damage to Samus", which come from heat, acid, lava, Samus Eaters, Grapple Beam contact with a broken turret in Draygon's room, and touching spikes.

The "Samus new state" handler executed as part of the "main gameplay" state normally calls a routine at $90:E9CE to deal with this variable. It applies suit divisor, subtracts the result from Samus energy, flooring the result at 00 if it would be negative, then zeroing out the periodic damage variables.

All but the last step are skipped if the "time is frozen" flag is set. X-Ray, of course, is one such way to set that flag, and indeed, Samus incurs no damage from heat while using X-Ray. However, auto-reserves are slightly different.

Auto-reserve does indeed also set the time-is-frozen flag, but it also sets the Samus new state handler to the "Samus is locked" function. This handler does... very little. Of interest to this issue, it does *not* call $90:E9CE. Thus the periodic damage values accumulated by the heat, lava, and acid routines (run separately) are not zero'd out every frame as they would be with X-Ray.

When auto-reserve resumes time it also unlocks Samus, restoring the normal "new state handler". Both of these things happen simultaneously, so next frame the $90:E9CE gets to apply all of the built-up periodic damage. Ouch!

Except.

The *current state handler* was also restored to normal, and it gets one frame to process before the new state handler does. And as part of the current state handler is the call to $90:DCDD and eventually something called the "HUD selection handler". This is the procedure that gets to see that if X-Ray is selected, if the X-Ray pose is valid, and we have dash held, it begins X-Ray activation. The first thing X-Ray activation does is freeze time, and thus the $90:E9CE routine will simply zero out the built-up periodic damage while applying none of it to your E-Tanks.

The HUD selection handler is also responsible for handling the Item Select/Item Cancel button inputs, but these perform their function only when they are "newly pressed". If so, the handler will advance/clear the selection as appropriate before calling the current selection's handler. Thus if the selection is one step before X-Ray, it is possible to buffer the run button, but the Item Select input to advance to X-Ray is frame perfect: it must occur on the exact frame that time has been unfrozen. This is only relevant in scenarios where X-Ray cannot simply be pre-selected.

aquanight:
Adds a property `implicitHeatFrames` to control this behavior of `autoReserveTrigger`. The implied `heatFrames` can be suppressed (no) or converted to `suitlessHeatFrames` (suitless).

Also adds a tech `canReserveTriggerBufferXRay` which should be paired with triggers that apply a value of "no".

Resolves #2403

### DIFF
--- a/logicalRequirements.md
+++ b/logicalRequirements.md
@@ -201,7 +201,7 @@ An `autoReserveTrigger` object represents a logical requirement for "auto" reser
 
 * _minReserveEnergy_: The minimum amount of energy in reserves which will satisfy this requirement (default: 1).
 * _maxReserveEnergy_: The maximum amount of energy in reserves which will satisfy this requirement (default: 400).
-* _implicitHeatFrames_: Default is yes, which applies heatFrames equal to the amount of reserve transferred if the room is heated. No prevents this application, and the value "suitless" applies suitlessHeatFrames instead.
+* _implicitHeatFrames_: This takes one of three possible values, "yes", "no", or "suitless". The default is "yes", which applies `heatFrames` equal to the amount of reserve transferred if the room is heated. This application is prevented by the value "no", while the value "suitless" applies `suitlessHeatFrames` instead.
 
 __Example:__
 ```json


blkerby:
The canReserveTriggerBufferXRay tech is interesting. I wonder if we'll see it used in races, since it would be faster than pausing and manually transfering reserves; just it's less flexible since there's not the same kind of control over when the transfer happens.

aquanight:
Video demonstration for non-R-Mode use of X-Ray buffering: https://videos.maprando.com/video/9246

Perhaps could use this as the tech video?

blkerby (approved):

PR #2418, opened 2025-11-15
add loptr gmode setup to mickey mouse

nn357:
https://videos.maprando.com/video/7611
https://videos.maprando.com/video/7610

### DIFF
--- a/region/lowernorfair/east/Mickey Mouse Room.json
+++ b/region/lowernorfair/east/Mickey Mouse Room.json
@@ -1936,7 +1936,7 @@
         {"notable": "Multiviola Ice Clip"},
         "canBeVeryPatient",
         "canTrickyGMode",
-        "h_heatproof",
+        "h_heatProof",
         "h_XRayMorphIceClip",
         "canTrickyJump",
         "ScrewAttack",


region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -1929,6 +1929,32 @@
...
+      "unlocksDoors": [
+        {"types": ["super", "missiles"], "requires": []},
+        {"types": ["powerbomb"], "requires": ["never"]}
+      ],
+      "exitCondition": {
+        "leaveWithGModeSetup": {}
+      },
+      "flashSuitChecked": true,
+      "note": [
+        "The Multiviola must pass through the crumble blocks on the first cycle it has the opportunity to do so,",
```
osse101: Letting the viola spend a significant amount of time below or above the crumble blocks still works.
I did need to break the two left sets of bomb blocks, and not break the right set of bomb blocks.

region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -1929,6 +1929,32 @@
...
     },
+    {
+      "link": [9, 2],
+      "name": "G-Mode Setup - Get Hit By Multiviola",
+      "requires": [
+        {"notable": "Multiviola Ice Clip"},
+        "canBeVeryPatient",
+        "canTrickyGMode",
+        "h_heatProof",
+        "h_XRayMorphIceClip",
```
osse101: You can do it without xray standup, but you already have the items so why not.
Or you can use a flash suit as your method of breaking the bomb blocks.

nn357: Yeah! I can always put the other methods in, I did try them, but since you have x-ray already (although it would clear a flashsuit).. i'll put in some other options.

region/lowernorfair/east/Mickey Mouse Room.json
osse101: I didn't wait long.  Just saying that I didn't see any signs that they wouldn't pass through the crumbles after the first cycle.
My setup on the freeze is to get hit while against one of the walls and shoot once knockback ends.  And you can keep retrying that.

nn357: yeah it will pass through the crumbles on any cycle but if you dont take it through the crumbles on the first opportunity it never goes through the top shot block (at least according to loptr), i tried it on later cycles and it does go through the crumbles just never gets through the top shot block

PR #2422, opened 2025-11-16
A few more R-mode standup clips

blkerby:
- Adds "canRModeStandupClip" as a requirement to existing strats in Parlor and Golden Torizo's Room.
- Clarifies the notes in Botwoon Hallway to say to release down during the refill and re-press it afterward. Holding down during the refill is bad because it usually makes you bonk, reducing the frame window from 8 frames to 2 frames.
- Adds strats in Lava Farm Tunnel, Purple Shaft, and Lava Dive.

The new strats are based on videos that Sam shared a while ago:
- Lava Farm Tunnel: https://videos.maprando.com/video/5745
- Purple Shaft: https://videos.maprando.com/video/5743
- Lava Dive: https://videos.maprando.com/video/5739

There are more R-mode standup clip strats that could still be added: for example, I skipped the Wasteland one to avoid a conflict with the open PR #2396, and skipped the one in Fast Pillars Setup Room because it is more complex to write up. Using a forced standup with Mochtroids in Halfie Climb and Colosseum to clip into the air pockets may also be possible (for door lock skips using Grapple).

### DIFF
--- a/tech.json
+++ b/tech.json
@@ -3117,7 +3117,7 @@
                 "This is commonly done with a jump into a mid-air morph to reach a height just below the ceiling when reserves trigger.",
                 "Jumping and aiming down is also possible, in case Morph is unavailable or if there is not enough space to mid-air morph;",
                 "aiming down does not reduce Samus' hitbox as much as morphing, so it requires more precise timing.",
-                "In any case, it is generally important to not be holding down when the reserve trigger finishes (typically it is best to release all inputs),",
+                "In any case, it is generally important to not be holding down when the reserve trigger finishes,",
                 "as this would be likely to cause Samus to bonk instead of clipping through."
               ]
             }


region/norfair/east/Lava Farm Tunnel.json
```diff
@@ -1192,6 +1233,40 @@
...
     },
+    {
+      "link": [4, 3],
+      "name": "R-Mode Forced Standup Door Lock Skip",
+      "requires": [
+        {"obstaclesCleared": ["R-Mode"]},
+        "canRModeStandupClip",
+        "canInsaneJump",
+        {"or": [
+          "canMidAirMorph",
```
osse101: I felt a difference with springball but not with jump-morph.
Pause abuse + visual cue was as good as anything other than spring ball for me.

blkerby:
> Why is lava farm tunnel written a little different?

I used an R-mode obstacle in Lava Farm Tunnel because there are several doors you can enter through but you will be using the Gamet farm regardless. An R-mode obstacle will be wanted here anyway for R-mode spark interrupt strats later. Because of the Gamet farm there's no need for a Crystal Flash here, unlike the other rooms, and this also means there's a reason to consider a variant without Morph.

More variants could probably be added with a G-mode artificial morph Crystal Flash but I didn't want to look at those on this pass.

osse101:
> I wasn't sure which room(s) this is referring to; is there some change you're looking for?

In lava dive, lava farm tunnel, and purple shaft I found jumping and midair morph to be equivalently difficult to using manual reserve & shot to create a visual cue.  And jumping up with just springball had a noticeable benefit.  So I would have it say `OR:[h_canUseSPringball, canBeVeryPatient]`  
Did you have a different testing experience?  I felt my jump morphs were poor.

osse101 (approved):
I don't really want anything moved it's just that there is a gap in relative difficulty.  RMode door clips aren't fit for Expert and are consistent enough to not really be insane.  So it's ok how it is.

PR #2425, opened 2025-11-17
add tricky upper gmode setup to mickey mouse room.

nn357:
sorry for the re-do.. had to delete the fork.

Where were we on this? Loptr mentioned if you don't get the viola through the crumbles on the first opportunity it goes into a pattern where it will still go through but never go through the top shot block..  I tried letting it cycle in the clip area and it wouldnt go through the shot blocks.

I think Osse had some success with bombs instead of screw attack? (can add an or for those if so)

https://videos.maprando.com/video/7611
https://videos.maprando.com/video/7610

### DIFF
--- a/region/lowernorfair/east/Mickey Mouse Room.json
+++ b/region/lowernorfair/east/Mickey Mouse Room.json
@@ -1965,6 +1965,7 @@
               "Bombs",
               "ScrewAttack",
               {"ammo": { "type": "PowerBomb", "count": 1}},
+              {"obstaclesCleared": ["A"]},
               {"obstaclesCleared": ["B"]}
             ]}
           ]}


region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -1929,6 +1929,60 @@
...
+            "h_preciseIceClip",
+            {"useFlashSuit": {}},
+            {"shinespark": {"frames": 11, "excessFrames": 5}}
+          ]},
+          {"and": [
+            "h_XRayMorphIceClip",
+            {"or": [
+              "Bombs",
+              "ScrewAttack",
+              {"ammo": { "type": "PowerBomb", "count": 1}},
```
osse101:
```suggestion
              {"ammo": { "type": "PowerBomb", "count": 1}},
              {"obstaclesCleared": ["A"]},
```

PR #2426, opened 2025-11-18
Mickey Mouse Room R-Mode Spark Interrupt

aquanight:
Room Notes:

From the bottom left door:
- Screw Attack is the only way to avoid getting hit by a Multiviola by entry.
- Once you clear the first wave of Dessgeegas and the Multiviolas, you have a clear shot for blue speed for your choice of: run up to the last Dessgeegas and use them to interrupt, run through to shinecharge and hop up to the Multiviola box, or shinecharge heat interrupt.

From the top right door:
- Four different strats depending on when - or even if - you use a Crystal Flash. CF above the bomb blocks, CF while killing the Multiviolas, or CF after the Dessgeegas, and then no CF at all.
- `canBePatient` is mandated on the non-CF route due to the crumble block one-way preventing a direct reset.

### DIFF
--- a/region/lowernorfair/east/Mickey Mouse Room.json
+++ b/region/lowernorfair/east/Mickey Mouse Room.json
@@ -1169,52 +1169,12 @@
                 {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "explicitWeapons": ["PowerBomb"]}},
                 {"heatFrames": 500}
               ]},
-              {"and": ["Ice", "Wave", "Spazer", {"heatFrames": 750}]},
-              {"and": [
-                {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "excludedWeapons": ["Bombs", "PseudoScrew", "PowerBomb"]}},
-                {"heatFrames": 3000}
-              ]}
+              {"and": ["Ice", "Wave", "Spazer", {"heatFrames": 750}]}
             ]}
           ]},
           {"and": [
             "h_heatProof",
-            {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "excludedWeapons": ["Bombs", "PseudoScrew", "PowerBomb"]}}
-          ]}
-        ]},
-        {"or": [
-          {"and": ["Wave", "Plasma", {"heatFrames": 450}]},
-          {"and": ["Plasma", {"heatFrames": 750}]},
-          {"and": ["Ice", "Wave", "Spazer", {"heatFrames": 750}]},
-          {"and": ["Wave", {"heatFrames": 1350}]}
-        ]},
-        {"or": [
-          {"and": [
-            {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "excludedWeapons": ["Bombs", "PseudoScrew"]}},
-            {"heatFrames": 3000}
-          ]},
-          {"and": [
-            "Plasma",
-            {"heatFrames": 250}
-          ]},
-          {"and": [
-            "Ice", "Wave", "Spazer",
-            {"heatFrames": 600}
-          ]},
-          {"and": [
-            "ScrewAttack",
-            {"heatFrames": 200}
-          ]},
-          {"and": [
-            {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "explicitWeapons": ["Missile"]}},
-            {"heatFrames": 670}
-          ]},
-          {"and": [
-            {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "explicitWeapons": ["Super"]}},
-            {"heatFrames": 400}
-          ]},
-          {"and": [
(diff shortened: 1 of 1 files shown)


region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -798,6 +866,499 @@
...
+          ]}
+        ]},
+        {"or": [
+          {"and": [
+            {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "excludedWeapons": ["Bombs", "PseudoScrew"]}},
+            {"heatFrames": 3000}
+          ]},
+          {"and": [
+            "Plasma",
+            {"heatFrames": 650}
```
osse101: "Beam Kill" Says this is 250 + 150 frames.

region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -798,6 +866,499 @@
...
+            {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "explicitWeapons": ["Missile"]}},
+            {"heatFrames": 670}
+          ]},
+          {"and": [
+            {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "explicitWeapons": ["Super"]}},
+            {"heatFrames": 400}
+          ]},
+          {"and": [
+            {"enemyKill": {"enemies": [["Dessgeega", "Dessgeega"]], "explicitWeapons": ["PowerBomb"]}},
+            {"heatFrames": 350}
```
osse101: Ammo Kill says this is 500

region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -798,6 +866,499 @@
...
+                "h_useMorphBombs",
+                {"heatFrames": 60}
+              ]}
+            ]}
+          ]}
+        ]},
+        {"or": [
+          {"and": [
+            {"heatFrames": 400},
+            {"enemyDamage": {"enemy": "Multiviola", "type": "contact", "hits": 3}}
```
osse101: The full energy requirement in `h_RModeCanRefillReserves` makes it so you can't tank the damage.

region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -798,6 +866,499 @@
...
+            },
+            {"heatFrames": 400}
+          ]},
+          {"and": [
+            "ScrewAttack",
+            {"heatFrames": 320}
+          ]}
+        ]},
+        "h_RModeCanRefillReserves",
+        {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
```
osse101: missile kill options should be removed.  But you are killing enough things to reach 20 in health drops easy.

region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -798,6 +866,499 @@
...
+        {"partialRefill": {"type": "ReserveEnergy", "limit": 20}},
+        "h_shinechargeMaxRunway",
+        {"autoReserveTrigger": {"maxReserveEnergy": 95}},
+        "canRModeSparkInterrupt",
+        {"heatFrames": 30}
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
```
osse101: This needs to clear most of the obstacles.  Mainly CDEF so you can't ice clip.  And A to be safe.

region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -798,6 +866,499 @@
...
+            {"heatFrames": 320}
+          ]}
+        ]},
+        {"or": [
+          {"resourceAvailable": [{"type": "ReserveEnergy", "count": 1}]},
+          {"and": [
+            "h_RModeCanRefillReserves",
+            {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
+            {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
+          ]}
```
osse101: Do we need this branch?  You crystal flashed for energy.  There are no enemy damage events so if you are out of reserves to heat damage you cannot refill them.

And you spent missiles on the CF and won't be full.

region/lowernorfair/east/Mickey Mouse Room.json
```diff
@@ -798,6 +866,499 @@
...
     },
+    {
+      "link": [2, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit) - Bomb Blocks Crystal Flash",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "h_heatedCrystalFlashForReserveEnergy",
+        {"heatFrames": 15},
```
osse101: This should come first?

region/lowernorfair/east/Mickey Mouse Room.json
aquanight: It can still work on the `disable: ETank` path as long as you still have 99 energy.

osse101 (approved):

PR #2428, opened 2025-11-19
R-mode CF interrupt: Red Fish, Crab Shaft

kjbranch:


region/maridia/outer/Red Fish Room.json
```diff
@@ -164,6 +171,38 @@
...
+      "requires": [
+        {"obstaclesCleared": ["R-Mode"]},
+        "canRModeCrystalFlashInterrupt",
+        {"ammo": {"type": "Missile", "count": 10}},
+        {"ammo": {"type": "Super", "count": 10}}
+      ],
+      "resetsObstacles": ["R-Mode"],
+      "flashSuitChecked": true,
+      "note": [
+        "Kill the Zebbos and have their drops spawn out of the way.",
```
osse101: This looks RNG with the oscillation of samus up and down.
But you get three chances and it seems to work most of the time.

Supering a crab in crab shaft was RNG whether I took 0 hits 1 hit or more before the crab died.

osse101 (approved):

region/maridia/outer/Red Fish Room.json
kjbranch: I havent been trying the super method. I wonder if there are very many places it would be feasible (like here). I'll keep my eye out.

PR #2429, opened 2025-11-19
add 10pb cf to spongebath room

nn357:
this one is really easy and doesnt require tanking any damage.. let the bull get close to samus, shoot it once and morph / lay pb... they have 100% drop rate for pbs so as long as Samus isn't in healthbomb range its always a pb..

https://videos.maprando.com/video/9280
https://videos.maprando.com/video/9277

strat is a bit basic, maybe some options could be added for farming if healthbomb flag is set.

### DIFF
--- a/region/wreckedship/main/Sponge Bath.json
+++ b/region/wreckedship/main/Sponge Bath.json
@@ -242,15 +242,19 @@
       "name": "10 Power Bomb Crystal Flash",
       "requires": [
         "f_DefeatedPhantoon",
-        {"resourceAvailable": [{"type": "RegularEnergy", "count": 50}]},
-          {"or": [
-            "ScrewAttack",
-            "canPseudoScrew",
-            "Plasma",
-            "Grapple",
-            {"resetRoom": {"nodes": [2]}},
-            {"ammo": {"type": "Missile", "count": 3}},
-            {"ammo": {"type": "Super", "count": 3}}
+        {"or":[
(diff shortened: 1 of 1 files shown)


nn357:
added a few things to it but i still dont think its quite right.. I did have a look at the other "can 10pb cf" but i couldnt find any that have anything in them like we are trying here.

region/wreckedship/main/Sponge Bath.json
```diff
@@ -242,6 +242,16 @@
       "name": "10 Power Bomb Crystal Flash",
       "requires": [
         "f_DefeatedPhantoon",
+          {"or": [
+            {"resourceAvailable": [{"type": "RegularEnergy", "count": 50}]},
+            "ScrewAttack",
+            "canPseudoScrew",
+            "Plasma",
+            "Grapple",
+            {"resetRoom": {"nodes": [2]}},
```
osse101: reset room would be anded with everything except resourceavailable

region/wreckedship/main/Sponge Bath.json
```diff
@@ -242,8 +242,8 @@
       "name": "10 Power Bomb Crystal Flash",
       "requires": [
         "f_DefeatedPhantoon",
+        {"resourceAvailable": [{"type": "RegularEnergy", "count": 50}]},
```
osse101: If you have 50 energy, you are not in health bomb and can kill the bull with the PB performing the 10PB CF.
resetRoom is for farming energy to ensure you are not in healthbomb.

PR #2433, opened 2025-11-20
R-mode CF Interrupt - Below Botwoon, Halfie Climb

kjbranch:
This finishes off the "easy" variants in Maridia. I'll probably make note of the rest and then move on to another region. 
Removed the easily-avoidable G-mode strat to nowhere in Below Botwoon

### DIFF
--- a/region/maridia/inner-pink/Halfie Climb Room.json
+++ b/region/maridia/inner-pink/Halfie Climb Room.json
@@ -3697,8 +3697,7 @@
       "flashSuitChecked": true,
       "note": [
         "Time the Crystal Flash so that the Oum can hit Samus.",
-        "This is fairly lenient, but one method is to Crystal Flash near the shot block pillar while the Oum is to the right of Samus moving right.",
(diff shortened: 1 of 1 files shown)


region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -3622,8 +3656,106 @@
...
+            {"resourceCapacity": [{"type": "ReserveEnergy", "count": 199}]}
+          ]}
+        ]}
+      ],
+      "resetsObstacles": ["R-Mode"],
+      "flashSuitChecked": true,
+      "note": [
+        "Time the Crystal Flash so that the Oum can hit Samus.",
+        "This is fairly lenient, but one method is to Crystal Flash near the shot block pillar while the Oum is to the right of Samus moving right.",
+        "With some testing, it seemed like the Oum was more likely to hit Samus when it is moving to the left."
```
osse101:
```suggestion
        "This is fairly lenient, but one method is to Crystal Flash near the shot block pillar while the Oum is to the right of Samus, near the door."
```
Like that?

osse101 (approved):

PR #2439, opened 2025-11-21
Lower Norfair Fireflea Room R-Mode Spark Interrupt

aquanight:
Room Notes:

- The only R-Mode Spark in Lower Norfair to not be heated, so no heat interrupt option.
- Power Bombs are safely usable after Crystal Flash in this room, as long as Firefleas haven't been killed yet.
- Two different runways. One requires killing the Fune in the way, the other (longer) requires being able to traverse the vertical section.
- Crystal Flash strats from the upper doors don't require going down at all.
- Spike Pause Abuse strat from the lower door doesn't require going up, but does need to farm ~~all five Firefleas~~ one Fireflea and have full damage reduction to avoid a Crystal Flash.
- All other farming strats go for the Firefleas and farm for one energy drop (always large). ~~Missing power bomb lenience based on how many firefleas can be accessed.~~
- Farming Funes wasn't included because Firefleas are way better for it.

### DIFF
--- a/region/lowernorfair/east/Lower Norfair Fireflea Room.json
+++ b/region/lowernorfair/east/Lower Norfair Fireflea Room.json
@@ -784,7 +784,7 @@
       "flashSuitChecked": true
     },
     {
-      "link": [2, 3],
+      "link": [2, 2],
       "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
       "entranceCondition": {
         "comeInWithRMode": {}
@@ -804,8 +804,7 @@
         ]},
         {"canShineCharge": {"usedTiles": 20, "gentleDownTiles": 2, "openEnd": 1}},
         {"autoReserveTrigger": {"maxReserveEnergy": 95}},
-        "canRModeSparkInterrupt",
-        {"spikeHits": 1}
+        "canRModeSparkInterrupt"
       ],
       "clearsObstacles": ["A"],
       "flashSuitChecked": true,


aquanight:
In R-Mode, the "fireflea darkening effect" is also applied to Samus sprite and all tiles that would otherwise be exempt (doors, spikes, etc), though oddly not the Firefleas themselves. This effect ends after the R-Mode standup, although the room remains darkened. As usual, X-Ray can also provide temporary relief.

Killing more than one fireflea results in poor visibility that can only be recovered by exiting R-Mode. As such, I've restricted Fireflea farming to just one. May be worth revisiting with the addition of a tech flag to allow for blind movement (perhaps `canOffScreenMovement` is sufficient).

aquanight:
Light Pillar may be of value to mitigate the visibility issue of farming multiple Firefleas while still in R-Mode. I am looking into how this might be logically incorporated.

Notable, however, in that once it is time to shinecharge, the light pillar will necessarily be lost due to the release of run button while short-charging.

region/lowernorfair/east/Lower Norfair Fireflea Room.json
```diff
@@ -294,6 +294,116 @@
...
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Crystal Flash, then pick up a shinecharge.",
+        "Drop into the top Fune's fireball path for the interupt."
+      ]
+    },
+    {
+      "link": [1, 3],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit) - Farm Firefleas",
```
osse101: I'd like if this were cleaner.
- If Supers or PBs -> farm a fune
- Tricky jump and can jump over bottom fune -> farm a fireflea

Getting past the top rock is a thing, but rolling over it is free.  With trickyJump as knowledge check.
And if you don't have morph, pseudoscrew it, with trickyjump as a knowledge check.

region/lowernorfair/east/Lower Norfair Fireflea Room.json
```diff
@@ -586,6 +696,96 @@
...
+            "ScrewAttack",
+            {"or": [
+              "canCarefulJump",
+              "HiJump",
+              "SpaceJump",
+              {"enemyKill": {"enemies": [["Fune"]]}}
+            ]}
+          ]}
+        ]},
+        {"or": [
```
osse101: This OR block and the one for climbing the room should be reversed.
You're allowed to take damage climbing the room in this direction.
But then the shinespark out of fireflea pit matters

PR #2446, opened 2025-11-22
Ridley grab door lock skip

blkerby:
The setup seems reliable enough that this could probably be Extreme?

Videos:
- With Varia, by Sam: https://videos.maprando.com/video/8769
- With Varia, a bit simplified version: https://videos.maprando.com/video/9313
- With Gravity: https://videos.maprando.com/video/9314


osse101:
Close to reliable enough for Expert.
Except it fails for me sometimes, and is a bizarre interaction.

PR #2454, opened 2025-11-23
Fast Rippers R-Mode Spark Interrupt

aquanight:
Room notes:
- Six Rippers, with guaranteed Large Energy while full on Supers.
  - For clarity, I will refer to them as numbered 1 through 6, starting from the top. Their patterns are noteworthy in this strat:
  - Ripper 1 and 2 reach the left door, but don't reach the green gate.
  - Ripper 3, 4, and 5 bounce between both doors, but only 4 and 5 are low enough to interfere with the runway.
  - Ripper 6 is trapped between two platforms.
- Killing with supers just refunds your supers. Screw Attack and Power Bomb kills only. (In theory, health bias can force energy drops and allow for Super kills combined with heatProof, but they can't be grouped quickly enough to accumulate drops.)
- Special note: once blue suit is gained, Screw Attack ceases to be effective, 
- The green gate doesn't have to be opened if 18 tiles is an acceptable runway. When opened from the left, one Ripper could be sacrificed to refund the super. When opened from the right, green gate glitch leniency will determine how many Rippers must be sacrificed before you can get energy again.
- [1, 1] strat is presumed to be Super-less (can't open the gate), so no Crystal Flash option. If you can CF, you can open the gate and use the [1, 2] strat.

Farming:
- Keep Ripper 3 alive, for the best interrupt.
- Kill 4 and 5 first, to de-complicate movement.
- Interrupt close to the green gate (if you can't open it) or the right door, and with Ripper 3 approaching you from the left to then bounce of the gate/door, pass back through you while you still have i-frames. If it bounced off the gate, it'll come back in time for a second hit when you had more than 300 reserves. If it bounced off the door, a full reserve fill-up is safe.

Crystal Flash (Left Door):
- Morph quickly to avoid getting hit by Rippers. Stay by the door.
- Wait an extra second or two for Ripper 3 to start going off screen, then drop your Power Bomb.
- During the Crystal Flash, 4 and 5 will start to come back and will be destroyed by the R-Mode light orb. They are almost certain to drop Supers, which ensures you can open the gate.
- Open the gate for a longer runway.
- Interrupt near the door with Ripper 3 moving left.

Crystal Flash (Right Door):
- CF before trying to open the gate.
- Dodge Ripper 4 and 5 once they come flying through, then follow behind them and watch them explode on the light orb. Pick up the Supers they drop if you want.
- From here it's identical to [1, 2].

### DIFF
--- a/region/lowernorfair/west/Fast Ripper Room.json
+++ b/region/lowernorfair/west/Fast Ripper Room.json
@@ -99,6 +99,40 @@
   ],
   "strats": [
     {
+      "link": [1, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit) - Timed Heated Crystal Flash",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"resourceAvailable": [{"type": "RegularEnergy", "count": 150}]},
+        {"resourceAtMost": [{"type": "RegularEnergy", "count": 180}]},
+        {"resourceMissingAtMost": [{"type": "Super", "count": 0}]},
+        {"disableEquipment": "HiJump"},
+        {"heatFrames": 388},
+        "h_heatedCrystalFlash",
+        {"partialRefill": {"type": "ReserveEnergy", "limit": 30}},
+        {"canShineCharge": {"usedTiles": 18, "openEnd": 0}},
+        "h_heatTriggerRModeSparkInterrupt",
+        "canTrickyJump",
+        {"heatFrames": 92}
+      ],
+      "clearsObstacles": ["B"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Requires about 150-180 Reserve at R-Mode entry to be able to correctly time the Crystal Flash.",
+        "The correct timing to place the power bomb is on taking 97 (53-83 energy left) energy worth of heat.",
+        "The Power Bomb explosion kills all six Rippers near the door, placing their",
+        "energy drops in easy reach to grab them and fill Reserve Energy.",
+        "Run to the green gate and wait until 69 energy is left and start running. Shinecharge at the end",
+        "of the runway, then use quick, speedy jumps back across the platforms towards the left door. You should",
+        "be able to make it to the left high platform (2nd from the door) before you need to windup and interrupt.",
+        "Shoot the door open and you can make it out."
+      ],
+      "devNote": "This strat makes a suitless CF with max E-Tanks viable without needing disable E-Tanks."
+    },
+    {
       "id": 43,
       "link": [1, 1],
       "name": "Open Door",
@@ -703,34 +737,6 @@
         "If any rippers remain, you can farm them with Power Bombs (or Supers at low health) but Screw Attack will no",
         "longer destroy them while you have blue suit."
       ]
-    },
-    {
-      "link": [1, 3],
-      "name": "R-Mode Spark Interrupt (Gain Blue Suit) - Timed Heated Crystal Flash",
-      "entranceCondition": {
-        "comeInWithRMode": {}
-      },
-      "requires": [
-        {"resourceAvailable": [{"type": "RegularEnergy", "count": 150}]},
-        {"resourceAtMost": [{"type": "RegularEnergy", "count": 180}]},
-        {"resourceMissingAtMost": [{"type": "Super", "count": 0}]},
-        {"heatFrames": 388},
-        "h_heatedCrystalFlash",
-        {"partialRefill": {"type": "ReserveEnergy", "limit": 30}},
-        {"canShineCharge": {"usedTiles": 18, "openEnd": 0}},
-        "h_heatTriggerRModeSparkInterrupt"
-      ],
-      "clearsObstacles": ["B"],
-      "flashSuitChecked": true,
-      "blueSuitChecked": true,
-      "note": [
-        "Requires about 150-180 Reserve at R-Mode entry to be able to correctly time the Crystal Flash.",
-        "The correct timing to place the power bomb is on taking 97 energy worth of heat.",
-        "The Power Bomb explosion kills all six Rippers near the door, placing their",
-        "energy drops in easy reach to grab them and fill Reserve Energy.",
-        "Heat interrupt after getting shinecharge on the green gate runway."
-      ],
-      "devNote": "This strat makes a suitless CF with max E-Tanks viable without needing disable E-Tanks."
     },
     {
       "id": 5,


aquanight:
> Things to investigate:
> 
> * Left door, Fast heated CF (1440 energy) enabling a six Ripper farm (120 - requires full on supers _before_ the CF) might clear the 1500 threshold for a suitless heated CF strat without a loss of access risk. Requires heat interrupt since it will hit Ripper 3 as well.
> * Two super-tank right door strat by opening the gate first within six supers, then farm Rippers to refill supers and CF after.
> * Combine Direct+Indirect G-Mode for an artificial morph strat from the right door.

The first two have been added. The third has a video ( https://videos.maprando.com/video/9337 ) but it is not yet clear how to express this specific combined G-Mode state in logic.

region/lowernorfair/west/Fast Ripper Room.json
```diff
@@ -98,6 +98,103 @@
     }
   ],
   "strats": [
+    {
+      "link": [1, 1],
```
osse101: 1 to 3.  The shinespark happens at 3.

region/lowernorfair/west/Fast Ripper Room.json
```diff
@@ -98,6 +98,103 @@
...
+        ]},
+        {"or": [
+          {"and": [
+            {"or": [
+              "ScrewAttack",
+              {"enemyKill": {"enemies": [["Ripper 2 (red)"]], "explicitWeapons": ["PowerBomb"]}}
+            ]},
+            "h_RModeCanRefillReserves",
+            {"resourceMissingAtMost": [{"type": "Super", "count": 0}]},
+            {"partialRefill": {"type": "ReserveEnergy", "limit": 120}}
```
osse101: The order of events either has you kill the rippers early, and disable the etanks before taking heatFrames.  Or have a harder time crossing the room, requiring more tech or enemyDamage.

region/lowernorfair/west/Fast Ripper Room.json
```diff
@@ -98,6 +98,103 @@
...
+            {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
+          ]},
+          {"and": [
+            {"or": [
+              "ScrewAttack",
+              {"enemyKill": {"enemies": [["Ripper 2 (red)"]], "explicitWeapons": ["PowerBomb"]}}
+            ]},
+            "h_heatProof",
+            {"disableEquipment": "ETank"},
+            {"resourceMissingAtMost": [{"type": "Super", "count": 0}]},
```
osse101:
```suggestion
            {"resourceMissingAtMost": [{"type": "Super", "count": 4}]},
```
?

region/lowernorfair/west/Fast Ripper Room.json
```diff
@@ -1082,6 +1255,73 @@
...
+          {"and": [
+            {"or": [
+              {"resourceAvailable": [{"type": "Super", "count": 10}]},
+              {"and": [
+                {"or": [
+                  "ScrewAttack",
+                  {"enemyKill": {"enemies": [["Ripper 2 (red)", "Ripper 2 (red)"]], "explicitWeapons": ["PowerBomb"]}}
+                ]},
+                {"heatFrames": 600},
+                {"partialRefill": {"type": "Super", "limit": 10}}
```
osse101: You can't refill to 10 if you take 10 tries on the gate glitch.
It must be hard to refill 6.
And the powerbomb says you kill 2, so this sequence is a bit misleading and looks like it would want `First-Try Gate Glitch`

region/lowernorfair/west/Fast Ripper Room.json
```diff
@@ -98,6 +98,103 @@
...
+        ]},
+        "canRModeSparkInterrupt"
+      ],
+      "clearsObstacles": ["B"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Clear the bottom two rippers first to clear the runway by the green gate for getting speed.",
+        "Keep the third Ripper from the top alive for the best interrupt.",
+        "With high reserve energy, take the interrupt close to the gate and with the Ripper moving",
```
osse101: Is it hard to time the shinespark with a ripper being in the right position?

region/lowernorfair/west/Fast Ripper Room.json
aquanight: There's no energy minimum (whereas `h_RModeCanRefillReserves` has a 99 minimum to get reserve from just one drop): this is a "farm a lot of energy from low energy" path. To get to "reserve energy" from 01 needs 5 energy drops and ideally you want all 6 for 21 Reserve, so it must be that none of the Rippers drop Supers.

Actually... if you're below 30, your Super count doesn't matter anymore. You'll want to group the Rippers for PB killing if that's the way, as Screw Attack risks picking up individual drops and you're back to getting Supers once you pass 50.

So I think the way here might be to drop the Super limit for PB kills but keep it for Screw Attack. Super kill might also be viable for accumulating on-screen energy drops.

Killing all 6 does require a PB for the 6th (can't screw attack kill with blue suit) after interrupt, or kill it before and heat interrupt.

region/lowernorfair/west/Fast Ripper Room.json
aquanight: Running right from the end of the platform you can start running when the ripper comes on screen, and a 2-tap shinecharge should give you a small window to shinecharge and windup. It's a small window, hard to tell if it's too small for Expert.

Another timing option is to use autoshot. Turn off bombs and charge beam, dodge the ripper as needed, moonwalk into the starting position, hanging just off the edge of the runway. Crouch, hold shot, then morph (keep the shoot button held!). Wait for the ripper to fly from right to left, and unmorph (still holding shoot!) right as it disappears off the left edge. Count 8 shots. On the 9th shot exactly is when it comes back on screen, so on the _8th_ shot you should start running. This is a _much_ nicer window too. The ripper will appear behind you right around the 3rd or 4th tap and start to catch up, but you have plenty of time to shinecharge and jump into windup before it gets there.

region/lowernorfair/west/Fast Ripper Room.json
```diff
@@ -98,6 +98,103 @@
...
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"or": [
+          {"and": [
+            "ScrewAttack",
+            {"heatFrames": 350}
+          ]},
+          {"and": [
+            {"enemyKill": {"enemies": [["Ripper 2 (red)", "Ripper 2 (red)"]], "explicitWeapons": ["PowerBomb"]}},
```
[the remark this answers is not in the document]
aquanight: Delayed PB would require being able to disable an E-Tank before pickup.

PR #2455, opened 2025-11-23
add spikesuits to spiky death room

nn357:
also tidied up the xmode shinecharge strat by using the helper.

https://videos.maprando.com/video/9010 [right door] https://videos.maprando.com/video/9009 [left door]

I tried to come in normally shinecharged and fall to the ground but unless you come in jumping the kzan hits you before you have time to gain flashsuit,

### DIFF
--- a/region/wreckedship/main/Spiky Death Room.json
+++ b/region/wreckedship/main/Spiky Death Room.json
@@ -902,8 +902,6 @@
       },
       "requires": [
         {"shineChargeFrames": 35},
-        "canTrickyJump",
-        "canComplexCarryFlashSuit",
         "f_DefeatedPhantoon",
         "h_spikeSuitSpikeHitLeniency",
         {"spikeHits": 1},


nn357:
> > I tried to come in normally shinecharged and fall to the ground but unless you come in jumping the kzan hits you before you have time to gain flashsuit,
> 
> My experience has been you float onto the kzan and then jump into the side of the spikes. You can walljump morph off the door, jump into the doorframe, or have gravity.

would the kzan floating thing require any extra tech on it like trickyjumps or such?

osse101:
> > > I tried to come in normally shinecharged and fall to the ground but unless you come in jumping the kzan hits you before you have time to gain flashsuit,
> > 
> > 
> > My experience has been you float onto the kzan and then jump into the side of the spikes. You can walljump morph off the door, jump into the doorframe, or have gravity.
> 
> would the kzan floating thing require any extra tech on it like trickyjumps or such?

no ... I just fall onto it.  If you are realistically entering the room you probably have some speed so canTrickyJump probably does fit after all.  As part of knowing to kill your speed.  
You have to move faster compared to the other options afterwards, so there could be a reason to add canShingechargeMovementTricky or canComplexCarryFlashSuit.

PR #2460, opened 2025-11-26
Hellway R-Mode Spark Interrupt

aquanight:
Room Notes:

- Respawning Zebs provide energy.
- Samus Eater Shinecharge. The best energy value is 20/10/5 to enter the plant and leave with just enough to get hit by a Zeb and interrupt.

### DIFF
--- a/region/brinstar/red/Hellway.json
+++ b/region/brinstar/red/Hellway.json
@@ -185,10 +185,10 @@
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
-        "Fill up on the respawning Zebs. Drain energy down to 20/10/5 and jump into the",
-        "floor plant with forward momentum to shinecharge while getting eaten. At 8/4/2 energy",
-        "left, shinecharge and get ready to jump out. Get hit by the Zeb to complete the",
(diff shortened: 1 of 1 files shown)


aquanight:
For left door, I found you can use the second plant when coming from the left, and for the Zeb coming from the first you can kill it and leave a drop hanging which will keep it from respawning for a bit.

That might also be possible from the right door.

But either way, I also find that at $84:ACB8, the floor Samus Eater PLM is enforcing a minimum of 16 i-frames on Samus with every frame of contact. Once you turn blue, i-frames become zero'd by blue suit, but then blue suit also makes you invulnerable in its own way and destroys the Zeb.

osse101:
>  destroys the Zeb

But then you get a big energy and won't get interrupted.  You would have to pause or jump into spikes.
We try to describe an easy way of executing the strat, since finding one is part of testing.

aquanight:
Added notes about leaving the Zebbo drop hanging, and which plant works best - there's two that don't spawn Zebbos and either of them work, but since crossing Hellway is logically free, the right door note is basically using the same plant as from the left (the 2nd from the left).

PR #2462, opened 2025-11-26
Beta Power Bombs R-Mode Spark Interrupt

aquanight:
Room Notes:

- Sidehoppers again. Three is not great for farming.
- Two runway options: Power Bomb to clear the blocks (which can also be used to kill hoppers) and jump into a Samus Eater, or avoid clearing the blocks and have a normal runway.

### DIFF
--- a/region/brinstar/red/Beta Power Bomb Room.json
+++ b/region/brinstar/red/Beta Power Bomb Room.json
@@ -150,22 +150,18 @@
         {"samusEaterFrames": 160},
         "h_shinechargeMaxRunway",
         {"autoReserveTrigger": {}},
+        "canHorizontalDamageBoost",
         "canRModePauseAbuseSparkInterrupt"
       ],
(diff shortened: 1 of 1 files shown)


region/brinstar/red/Beta Power Bomb Room.json
```diff
@@ -99,6 +99,82 @@
...
+    },
+    {
+      "link": [1, 1],
+      "name": "R-Mode Pause Abuse Spark Interrupt (Gain Blue Suit) - Samus Eater Shinecharge",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"or": [
+          "h_CrystalFlashForReserveEnergy",
```
osse101: I think you should remove the other branch and make this strat be about Crystal Flashing.  Without the CF, its the same as the above strat but harder.

PR #2466, opened 2025-11-29
Shaktool Room R-Mode Spark Interrupt

aquanight:
Room Notes:

- Farming Shaktool normally has 1 drop if you use a beam or missile, but if you use a well-timed Shinespark, you can get as many as five. Take care not to leave the room while Shaktool is exploding, as there is a crash-the-game risk. But why would you when you want the drops anyway?
- If you can't free Shaktool from the right side, you're left with Yards. You need to clear the sand via Snail Soccer to use the runway, so you can use this to separate them for a Power Bomb kill.
- Both Shaktool and Yards hit hard, but they don't stay to bother Samus during the Reserve autofill.

### DIFF
--- a/region/maridia/inner-green/Shaktool Room.json
+++ b/region/maridia/inner-green/Shaktool Room.json
@@ -721,6 +721,10 @@
           ]}
         ]},
         "h_shinechargeMaxRunway",
+        {"or":[
+          "canWaterShineCharge",
+          "Gravity"
+        ]},
         {"autoReserveTrigger": {}},
         "canRModeSparkInterrupt"
       ],


region/maridia/inner-green/Shaktool Room.json
```diff
@@ -82,6 +87,31 @@
...
+        "f_ShaktoolDoneDigging"
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "clearsObstacles": ["A", "B", "R-Mode"],
+      "devNote": "Entering the room with the flag 'f_ShaktoolDoneDigging' set means that the sand blocks and Power Bomb blocks will not spawn."
+    },
     {
       "id": 1,
       "link": [1, 1],
```
kjbranch: Not sure why this needs an entrance condition. I wonder if we could just delete the entrance condition and the "R-mode done digging" one above

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -96,6 +126,72 @@
...
+                {"or": [
+                  {"and": [
+                    "canWaterShineCharge",
+                    {"canShineCharge": {"usedTiles": 35, "openEnd": 2}}
+                  ]},
+                  {"and": [
+                    "Gravity",
+                    {"canShineCharge": {"usedTiles": 42, "openEnd": 2}}
+                  ]}
+                ]},
```
kjbranch: Long runways are mostly only needed for cross room strats and maybe very low-level strats.
```suggestion
                "h_shinechargeMaxRunway",
```

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -96,6 +126,72 @@
...
+        {"or": [
+          {"and": [
+            "canWaterShineCharge",
+            {"canShineCharge": {"usedTiles": 35, "openEnd": 2}}
+          ]},
+          {"and": [
+            "Gravity",
+            {"canShineCharge": {"usedTiles": 42, "openEnd": 2}}
+          ]}
+        ]},
```
kjbranch:
```suggestion
        "h_shinechargeMaxRunway",
```

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -82,6 +87,31 @@
...
+      "link": [1, 1],
+      "name": "R-Mode Entry",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "clearsObstacles": ["R-Mode"]
+    },
```
kjbranch:
```suggestion
```

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -646,6 +676,87 @@
...
+          "h_CrystalFlashForReserveEnergy",
+          {"and": [
+            "h_RModeCanRefillReserves",
+            {"or": [
+              {"and": [
+                {"obstaclesCleared": ["A"]},
+                {"or": [
+                  "canWaterShineCharge",
+                  "Gravity"
+                ]},
```
kjbranch:
```suggestion
```

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -646,6 +676,87 @@
...
+                  {"and": [
+                    {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
+                    {"partialRefill": {"type": "ReserveEnergy", "limit": 60}}
+                  ]},
+                  {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
+                ]}
+              ]},
+              {"and": [
+                "canBePatient",
+                {"enemyKill": {"enemies": [["Yard"]]}},
```
osse101: As is, this doesn't include killing Yards with Speed.  And that would need a explanation as well.

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -646,6 +676,87 @@
...
+                {"or": [
+                  {"and": [
+                    {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
+                    {"partialRefill": {"type": "ReserveEnergy", "limit": 60}}
+                  ]},
+                  {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
+                ]}
+              ]},
+              {"and": [
+                "canBePatient",
```
osse101: Is canBePatient on clearing the sand or the 1/3 on big energy?

kjbranch: It's also worth noting that you'll get a small or big energy 2/3 of the time, and then you can get blue and kill the second snail before leaving to get another guaranteed energy drop (50/50 small/big).

So you'd have a 1/3 chance of failure (no drop on the first snail)
a 1/6 chance of ending with just 9 energy (small, small)
and a 1/2 chance of ending with 20+

region/maridia/inner-green/Shaktool Room.json
aquanight: This entry would allow collecting the digging flag and getting bluesuit in a single room entry.

As an example for MapRando, it would allow collecting the `f_ShaktoolDoneDigging` flag in a situation where a bluesuit is needed to return from doing so. Without this, the return path can't re-enter to get blue suit (because it'd be trying to use the digging flag when it hasn't confirmed it is reachable yet), another way to return would have to open up before the flag could be collected and then the room used for bluesuits.

region/maridia/inner-green/Shaktool Room.json
aquanight: The number of times I've accidentally run through and killed a snail in Aqueduct may have made me think it was more obvious than it actually is. I'll check this some more.

Using Bombs + Charge Beam might also help for a safer bomb kill.

region/maridia/inner-green/Shaktool Room.json
kjbranch: I wonder if we should have a `canComplexRModeSparkInterrupt`, which could mention nontrivial farming and a chance for bad drops instead of patience. 

I'm fine if patience is removed here or if it's kept with a fixme to put in a different tech. I don't think it really makes sense to have an OR for 10 vs 20 E remaining, as that's not going to change much, but I am also fine either way.

aquanight: It helps that, for example in a MapRando context, `canBePatient` keeps these less reliable strats out of Expert.

If it can do the same (or even if it doesn't) `canRelyOnLuck` would be perfect for this role in R-Mode Sparks and make it more clear that the issue is one of randomness, rather than simply taking a long time.

region/maridia/inner-green/Shaktool Room.json
```diff
@@ -646,6 +665,79 @@
...
+              ]},
+              {"partialRefill": {"type": "ReserveEnergy", "limit": 10}},
+              {"and": [
+                "canBePatient",
+                {"partialRefill": {"type": "ReserveEnergy", "limit": 25}}
+              ]}
+            ]}
+          ]}
+        ]},
+        "h_shinechargeMaxRunway",
```
osse101:
```suggestion
        "h_shinechargeMaxRunway",
        {"or":[
          "canWaterShineCharge",
          "Gravity"
        ]},
```

PR #2467, opened 2025-11-29
add spikesuit difficulty levels / refine noob bridge spikesuit setup

nn357:


nn357:
the room reset option sucked and its a difficult spikesuit anyway so i took it away, i gave a helper to ensure 3 pbs for attemps as you can farm in room to get 3 attempts

blkerby:
The way I had been thinking of handling it in the randomizer, the amount of attempts would be consistent between `h_spikeSuitPowerBombLeniency` and other leniencies `h_spikeSuitSpikeHitLeniency`, etc. On the other hand, the Power Bomb setup is a lot more niche so maybe it could be reasonable to add 50% more lenience to it like this PR does.

blkerby:
Yeah it might make sense to just have them be consistent. Most likely the quick drop option would be the preferred one anyway (e.g. canInsaneJump will probably be in an equal or lower difficulty than canTrickySpikeSuit)

region/brinstar/green/Noob Bridge.json
```diff
@@ -312,26 +312,30 @@
...
-        "canSpikeSuit",
+        "h_spikeSuitSpikeHitLeniency",
+        "canTrickySpikeSuit",
         {"shinespark": {"frames": 1, "excessFrames": 1}}
       ],
       "flashSuitChecked": true,
-      "note": "If the enemies are still alive, Power Bomb lag can delay the spikesuit timing."
+      "note": [
+        "If the enemies are still alive, Power Bomb lag can delay the spikesuit timing.",
+        "The quickdrop setup requires spin jumping for 3 frames or less,",
```
osse101: I use angle down to jump up into the spikes, if that's what this means.

nn357: I took a little bit of it out, just leaving in the note that jump above the crumbles is very small.. it is kind of obvious but some people were saying it wasnt obvious or too difficult etc.

PR #2468, opened 2025-11-29
Aqueduct R-Mode Spark Interrupt

aquanight:
Room Notes:
- Five Yards, three of which are global.
- The bottom runway is full-length, and you can free and anger the bottom Yard to interrupt.
- For the walljumpless, [5, 5] has an alternative with a shorter runway (and fewer Yards usable) that doesn't need to get back up from 2.

### DIFF
--- a/region/maridia/inner-pink/Aqueduct.json
+++ b/region/maridia/inner-pink/Aqueduct.json
@@ -1554,7 +1554,6 @@
       "link": [2, 5],
       "name": "Gravity Jump",
       "requires": [
-        "Gravity",
         "canGravityJump"
       ],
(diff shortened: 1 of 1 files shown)


osse101:
The speedy jump was the main one I had in mind, and you're right that you can't run with the blue suit.

But there is still Gravity Jump and we should assume they know how to do that plus they have Speed for the blue suit.  
Snail climb isn't any different from without a blue suit, but it's not a trivial tech.  And Gary is never on the wall when you need him.

aquanight:
I'm not seeing a Gravity Jump in current logic for [2, 5]? I'll check if it actually works.

PR #2469, opened 2025-11-29
Below Botwoon Energy Tank R-Mode Spark Interrupt

aquanight:
Room Notes:
- Multiple respawning Zoa points.
- An Owtch that has to be removed for runway access.
- Take the interrupt away from sand, ideally.

### DIFF
--- a/region/maridia/inner-pink/Below Botwoon Energy Tank.json
+++ b/region/maridia/inner-pink/Below Botwoon Energy Tank.json
@@ -118,7 +118,7 @@
       "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
       "requires": [
         "Gravity",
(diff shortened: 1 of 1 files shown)


region/maridia/inner-pink/Below Botwoon Energy Tank.json
```diff
@@ -113,6 +113,44 @@
     }
   ],
   "strats": [
+    {
+      "link": [1, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
+      "entranceCondition": {
+        "comeInWithRMode": {}
```
osse101: There is a "R-Mode" obstacle you should use.
And then you can use "A" too.

PR #2470, opened 2025-11-29
add x-mode bluesuit to boulder room.

nn357:
https://videos.maprando.com/video/9381

I put open end to (2) becuase its ok to run off the end while getting the shinecharge, you can just bounceball in the water into the spikes.

### DIFF
--- a/region/brinstar/blue/Blue Brinstar Boulder Room.json
+++ b/region/brinstar/blue/Blue Brinstar Boulder Room.json
@@ -150,17 +150,6 @@
       "flashSuitChecked": true
     },
     {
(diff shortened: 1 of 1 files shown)


region/brinstar/blue/Blue Brinstar Boulder Room.json
```diff
@@ -149,6 +149,17 @@
       },
       "flashSuitChecked": true
     },
+    {
+      "link": [1, 1],
```
osse101: Can you move this to [3, 3] so that it is grouped with the spikesuit.

PR #2472, opened 2025-11-29
Botwoon Energy Tank Room R-Mode Spark Interrupt

aquanight:
Room Notes:
- Respawning Zoa Farm.
- The full length runway is nice to have, but the 22-tile shinecharge by the Zoa pit is also right there. If for some reason 22-tile shinecharges aren't enabled, you have to get back across to the left side so you can use the long-run again.
- Morph tunnel crossing for [4, 4] is an option, but since it's optional I didn't put a `collectsItem` on it. In practice if you can do it you can collect that item separately anyway.

### DIFF
--- a/region/maridia/inner-pink/Botwoon Energy Tank Room.json
+++ b/region/maridia/inner-pink/Botwoon Energy Tank Room.json
@@ -1866,9 +1866,14 @@
       "requires": [
         "Gravity",
         {"refill": ["Energy"]},
-        "canSuitlessMaridia",
-        "canMidAirMorph",
-        {"disableEquipment": "Gravity"},
+        {"or": [
+          {"and": [
+            "canSuitlessMaridia",
+            "canMidAirMorph",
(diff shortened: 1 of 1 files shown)


region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -1836,6 +1868,36 @@
...
+    {
+      "link": [4, 4],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "Gravity",
+        {"refill": ["Energy"]},
+        {"or": [
```
osse101: Disable Gravity and jump morph would be assumed possible, given Morph.
Adding requirements would be just to avoid collecting the item because you would have to drop into the sand, go under a low wall, and fight a puyo.

PR #2476, opened 2025-11-30
Crab Shaft R-Mode Spark Interrupt

aquanight:
Room notes:

- Already had an obstacle from R-Mode CF so I used it.
- Four Scisers to farm. All four can be reached (the top one is global so you don't need to climb the shaft to get to it).
- Best runway is at the bottom. Movement is free between this one and the upper level runway.

### DIFF
--- a/region/maridia/inner-pink/Crab Shaft.json
+++ b/region/maridia/inner-pink/Crab Shaft.json
@@ -1468,6 +1468,40 @@
       "flashSuitChecked": true
     },
     {
+      "link": [2, 2],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
+      "requires": [
+        {"obstaclesCleared": ["R-Mode"]},
+        "Gravity",
+        {"or": [
+          "h_CrystalFlashForReserveEnergy",
+          {"and": [
+            "h_RModeCanRefillReserves",
+            {"or": [
+              {"and": [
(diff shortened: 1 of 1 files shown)


region/maridia/inner-pink/Crab Shaft.json
```diff
@@ -1467,6 +1467,40 @@
...
+          {"and": [
+            "h_RModeCanRefillReserves",
+            {"or": [
+              {"and": [
+                {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 2}]},
+                {"partialRefill": {"type": "ReserveEnergy", "limit": 40}}
+              ]},
+              {"and": [
+                {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 0}]},
+                {"partialRefill": {"type": "ReserveEnergy", "limit": 80}}
```
osse101: Does this mean, 
- put 60 energy in reserves
- activate reserves
- put 20 energy in main energy?

aquanight: Scisers are 0% Nothing so the "interrupt crab" can be included in the farming output (but not ammo leniency).

Supposed to be:
- If you're full on PBs you will very likely get large energy from the first three Scisers. (Small/large ratio is highly weighted toward large. Similar to Crab Maze.)
- If you aren't you can afford to be missing 2 PBs - get them back from Scisers - and still have one left to get energy from.

Then after you trigger, you can kill the last crab for (likely) another large energy.

PR #2479, opened 2025-12-02
Draygon R-Mode Spark Interrupt

aquanight:
The boss of Maridia has actually a few options for getting bluesuit.

- You can farm goops for energy and then perform a normal R-Mode Spark. Farming goops can take some patience, especially with lower level beams.
- This is one of the [select few rooms](https://wiki.supermetroid.run/G-Mode_Technical_Details#X-Ray_HDMA_object) in which a Crystal Flash does not corrupt the HDMA pile, if the player entered with Draygon alive, and thus does not create a light orb. This also means, should one be needed, the player can still use Power Bombs and X-Ray (especially for Microwave) normally.
  - If Draygon is already defeated, Crystal Flashes create light orbs as usual. Power Bomb can be used after CF to restore game audio and X-Ray usage.
- The Shinespark Kill can be used to gain a blue suit, as Draygon's death cutscene forces a standup and will give a blue suit.
- As usual, all Draygon-alive strats are loss-of-access.

### DIFF
--- a/region/maridia/inner-pink/Draygon's Room.json
+++ b/region/maridia/inner-pink/Draygon's Room.json
@@ -148,10 +148,6 @@
           "h_breakThreeDraygonTurrets"
         ]},
         "Gravity",
-        {"or": [
-          "canWalljump",
-          "SpaceJump"
-        ]},
         "h_CrystalFlashForReserveEnergy",
         {"canShineCharge": {"usedTiles": 22, "openEnd": 0}},
         "canOffScreenMovement",


region/maridia/inner-pink/Draygon's Room.json
```diff
@@ -109,6 +109,37 @@
...
+      "requires": [
+        {"not": "f_DefeatedDraygon"},
+        "canRiskPermanentLossOfAccess",
+        "Gravity",
+        "Morph",
+        {"or": [
+          "canBeVeryPatient",
+          "Wave",
+          "Plasma"
+        ]},
```
kjbranch: I wasnt needing any beams to farm. Or is this to kill the goops while shinecharged - maybe that was why I couldnt get it to work; there were a lot of goops and i just jumped over them.

kjbranch (commented):
I feel like the detailed PR comment is not needed, and is likely not going to be read during the review. I would keep it to very concise details that are particularly important and not clear when reviewing the code (usually no PR comment is needed). 

This PR comment had a lot of interesting details which I almost overlooked because there was too much text - however, a lot of this info should be documented in a different location (maybe an issue or brief devNote)

region/maridia/inner-pink/Draygon's Room.json
```diff
@@ -109,6 +109,37 @@
...
+      "link": [1, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit) - Draygon Alive",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"not": "f_DefeatedDraygon"},
+        "canRiskPermanentLossOfAccess",
+        "Gravity",
+        "Morph",
```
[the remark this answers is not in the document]
aquanight: In theory I don't think you have to break turrets. But I think you'd need some extra evasion work to be able to farm and keep ahead.

aquanight:
> Getting interrupted by a grab is a lot simpler. Its not the most reliable thing, but its going to be safer to try again.

Grab interrupt is added. Now the second time I have used `canRModeSparkInterrupt` for a non-R-Mode strat (the first was Ceres Ridley), but it's the same concept (interrupting spark windup) so it works as a knowledge check and it satisfies the speed-state checker.

If there's too many more of these, it's possible we may need `canSparkInterrupt` to move the non R-Mode variants to. Then `canRModeSparkInterrupt` can add `canRMode` to its `techRequires` that I was surprised it didn't have (technically doesn't need if it was only used in R-Mode strats).

region/maridia/inner-pink/Draygon's Room.json
```diff
@@ -1163,6 +1228,44 @@
       "requires": [],
       "flashSuitChecked": true
     },
+    {
+      "link": [2, 1],
+      "name": "Spike Suit, R-Mode Pause Abuse Spark Interrupt (Gain Blue Suit)",
```
osse101:
```suggestion
      "name": "R-Mode Pause Abuse Spark Interrupt (Gain Blue Suit)",
```

PR #2480, opened 2025-12-02
UN Farm: gate glitch with only reserves

kjbranch:
This recently came up in a Hard seed with only a reserve tank; should that require an E-Tank?
This was the best method I could think of to bump it up.


nn357:
https://www.twitch.tv/eddie/clip/LovelySplendidPangolinOSfrog-LDJSCzDu9tMEJjLZ

i meant to post in logic, i clipped Eddie talking to Oats about it

PR #2483, opened 2025-12-03
Add two new 10 PB CF tech

kjbranch:


helpers.json
```diff
@@ -1117,7 +1117,7 @@
             "can10PowerBombCrystalFlash",
             {"or": [
               "h_heatProof",
-              {"tech": "canHeatedCrystalFlash"}
+              "canHeated10PowerBombCrystalFlash"
```
osse101: This would take your ammo twice I believe.

PR #2484, opened 2025-12-03
Split out canRemoteAcquire from canComplexGMode

kjbranch:
This will help if/when `canComplexGMode` is a higher difficulty tech than `canRemoteAcquire`.

I went through all the strats with `canComplexGMode`, but there are likely some that were missing that (this PR found a few). Probably the strats missing it are currently in Map Rando's Extreme difficulty already, so it might not often be an error players notice.

### DIFF
--- a/region/lowernorfair/east/Lower Norfair Spring Ball Maze.json
+++ b/region/lowernorfair/east/Lower Norfair Spring Ball Maze.json
@@ -356,6 +356,7 @@
         }
       },
       "requires": [
+        "canRemoteAcquire",
         "canOffScreenMovement",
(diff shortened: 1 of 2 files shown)


region/lowernorfair/east/Lower Norfair Spring Ball Maze.json
```diff
@@ -312,7 +312,7 @@
         }
       },
       "requires": [
-        "canComplexGMode",
+        "canRemoteAcquire",
```
osse101: [3, 3] has remoteAcquire under canTrickyGMode

PR #2487, opened 2025-12-04
Plasma Room R-Mode Spark Interrupt

aquanight:
Room notes:
- A lot of pirates makes it easy to get 20 energy regardless of Missiles.
- Standing pirates need Screw Attack or Plasma Beam to kill. Pseudo-Screw is also possible using the damage-free notable.
  - ~~No Shinespark kill option due to energy cost.~~
- Killing any remaining pirates is free with blue suit.

### DIFF
--- a/region/maridia/inner-yellow/Plasma Room.json
+++ b/region/maridia/inner-yellow/Plasma Room.json
@@ -401,7 +401,10 @@
             {"or": [
               {"and": [
                 "canUseIFrames",
-                {"enemyDamage": {"enemy": "Pink Space Pirate (standing)", "type": "laser", "hits": 1}}
+                {"or": [
+                  {"enemyDamage": {"enemy": "Pink Space Pirate (standing)", "type": "laser", "hits": 1}},
+                  "h_pauseAbuseMinimalReserveRefill"
+                ]}
               ]},
               "ScrewAttack",
               "Plasma",


region/maridia/inner-yellow/Plasma Room.json
```diff
@@ -388,6 +389,61 @@
       "requires": [],
       "flashSuitChecked": true
     },
+    {
+      "link": [1, 3],
```
kjbranch: I think I would rather this go to node 2. Took me a bit to decide, really, but here my weak reasons for why:
- It will follow the room diagram (though that could be edited)
- It technically won't kill the top pirate (though you could do that on room exit)
- There's no [3, 2] link, so this wouldn't let you also collect the item (though it could be added)

Honestly, node 3 should probably be deleted, but if you want to keep this as [1, 3], let me know

region/maridia/inner-yellow/Plasma Room.json
```diff
@@ -388,6 +389,61 @@
...
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"or": [
+          "h_CrystalFlashForReserveEnergy",
+          {"and": [
+            {"or": [
+              "ScrewAttack",
+              "Plasma"
```
kjbranch: I think the primary way that people will farm here is by shinecharging on the bottom with short enough runway to kill the bottom pirate (looks like thats a VH runway). You can then do 0-1 energy-loss sparks to kill one of the other two pirates and/or the wall pirates if you didn't get an energy drop.
```suggestion
              "Plasma",
              {"getBlueSpeed": {"usedTiles": 19, "openEnd": 0}}
```

aquanight: The math says even if you use all 35 frames for the shinespark kill you have 92% chance to end with 20 energy. In it goes.

region/maridia/inner-yellow/Plasma Room.json
```diff
@@ -388,6 +388,37 @@
...
     },
+    {
+      "link": [1, 2],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit)",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"or": [
+          "h_CrystalFlashForReserveEnergy",
```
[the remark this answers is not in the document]
aquanight: If 19-vs-21 wasn't crossing a difficulty tier (21 is in Hard but 19 isn't), I'd have just gone with the 19.
But since it does, I think I have to at least have the option to get 21 by either tanking a hit (tank, suit, or pause abuse) or killing the pirate.

osse101 (approved):

region/maridia/inner-yellow/Plasma Room.json
```diff
@@ -388,6 +388,51 @@
...
+              {"and": [
+                "canUseIFrames",
+                {"enemyDamage": {"enemy": "Pink Space Pirate (standing)", "type": "lazer", "hits": 1}},
+                {"canShineCharge": {"usedTiles": 21, "openEnd": 0}}
+              ]},
+              {"and": [
+                {"or": ["ScrewAttack", "Plasma"]},
+                {"canShineCharge": {"usedTiles": 21, "openEnd": 0}}
+              ]},
+              {"canShineCharge": {"usedTiles": 19, "openEnd": 0}}
```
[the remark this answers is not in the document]
aquanight: The 19 tile shinecharge was for if you couldn't kill the bottom pirate. But as it turns out, if you can 19 tile shinecharge, you can 19-tile blue speed. This kills the pirate.

So I just added getblue to the pirate kill block and removed the 19 tile shinecharge.

kjbranch (dismissed):
Thanks, this structure is much easier for me to read

PR #2489, opened 2025-12-04
Watering Hole R-Mode Spark Interrupt

aquanight:
Room notes:

- Respawning Zebs for energy.
- Requires using frozen Zebs to get a runway. One by itself is a short runway, but two is a notable strat.
- With the Zebs frozen, you can either drop into the chasm and get hit by the Choot, or disable Ice Beam to force-unfreeze the Zebs.

### DIFF
--- a/region/maridia/inner-yellow/Watering Hole.json
+++ b/region/maridia/inner-yellow/Watering Hole.json
@@ -259,7 +259,7 @@
       "blueSuitChecked": true,
       "note": [
         "Freeze the two Zebs to create a runway to shinecharge with. Pixel perfect freezing of both Zebs",
-        "provides the longest runway, but a less perfect position of the second Zeb is still usable."
+        "provides the longest runway, but a slightly higher position of the second Zeb is still usable."
       ]
     },
     {


region/maridia/inner-yellow/Watering Hole.json
```diff
@@ -142,6 +142,57 @@
         "Requires a pixel-perfect freeze in order to run onto and off of the frozen Zebs."
       ]
     },
+    {
+      "link": [1, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit) - Frozen Zeb",
```
kjbranch: We are planning on removing the gain blue / flash suit in strat titles soon. The first part of the name makes it clear, and too long of strat names causes problems with the spoiler log
```suggestion
      "name": "R-Mode Spark Interrupt - Frozen Zeb Runway",
```

region/maridia/inner-yellow/Watering Hole.json
```diff
@@ -142,6 +142,57 @@
...
+      "requires": [
+        "h_frozenEnemyRunway",
+        {"canShineCharge": {"usedTiles": 13, "openEnd": 0}},
+        {"refill": ["Energy"]},
+        {"autoReserveTrigger": {"maxReserveEnergy": 95}},
+        "canRModeSparkInterrupt",
+        {"or": [
+          "Gravity",
+          {"disableEquipment": "Ice"}
+        ]}
```
kjbranch: with the choot, you dont need to disable ice, and you can get out of the water without gravity, with ice you dont need to wall jump or anything to get all the way out
```suggestion
        "canRModeSparkInterrupt"
```

region/maridia/inner-yellow/Watering Hole.json
```diff
@@ -142,6 +142,57 @@
...
+      "blueSuitChecked": true,
+      "note": [
+        "Freeze the left Zeb at a pixel perfect position to extend a runway across the top. With the shinecharge,",
+        "drop into the left chasm then quickly jump under the Choot to get hit by it and interrupt.",
+        "Freeze the Choot to help get back out, if needed."
+      ]
+    },
+    {
+      "link": [1, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit) - Double Frozen Zeb Runway",
```
osse101: You can remove this and combine it with the other strat

region/maridia/inner-yellow/Watering Hole.json
```diff
@@ -142,6 +142,57 @@
...
     },
+    {
+      "link": [1, 1],
+      "name": "R-Mode Spark Interrupt (Gain Blue Suit) - Frozen Zeb",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "h_frozenEnemyRunway",
+        {"canShineCharge": {"usedTiles": 13, "openEnd": 0}},
```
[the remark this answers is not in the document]
kjbranch: I was wondering if that would be helpful. I was thinking it would only add 1 tile, but yeah it'd be closer to 2.

region/maridia/inner-yellow/Watering Hole.json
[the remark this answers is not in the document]
aquanight: Gravity was because I mistakenly thought you needed Gravity to get up from the Choot (you do - or some other way to get up - if you fall all the way into the item hole). But when I looked at again, where you freeze it from and jump off of it are out of the water, so yes Gravity does nothing.

Disable Ice was intended as an alternative, since disabling ice forces the Zebs to unfreeze. But you can also just shinecharge closer to when they thaw out on their own.

(Also kyleb suggested this same change so it's already in.)

region/maridia/inner-yellow/Watering Hole.json
aquanight: This might be another thing that should be addressed on MapRando side, but I have concerns about 'hiding" notable strats (such as the Double Frozen Zeb Runway) when they get packaged into another strat like R-Mode Spark.

E.g. if this were displayed on MapRando it might look like this:
> Caterpillar Room
>   Elevator
>   Upper Left Door: G-Mode Setup, Get Hit By Zero
> Watering Hole
>   Right Door: R-Mode Spark Interrupt - Double Frozen Zeb Runway

If the strat was combined with the single zeb, it would not be known that the notable strat was being called for to use a longer runway.

That being said, if there's lenience in the double-Zeb freeze then the notable can be taken out as a factor.

kjbranch: I think it's ok, as it will still ensure that the player toggled that notable on ... there can definitely be exceptions, but i think its fine here

region/maridia/inner-yellow/Watering Hole.json
```diff
@@ -236,6 +236,34 @@
...
+        {"autoReserveTrigger": {"maxReserveEnergy": 95}},
+        "canRModeSparkInterrupt"
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Freeze the two Zebs to create a runway to shinecharge with. Pixel perfect freezing of both Zebs",
+        "provides the longest runway, but a less perfect position of the second Zeb is still usable.",
+        "Once shinecharged, drop into the left chasm and quickly jump under the Choot to interrupt.",
+        "Freeze the Choot to help get back out, if needed."
```
osse101:
```suggestion
        "provides the longest runway, but a less perfect position of the second Zeb is still usable."
```

kjbranch (approved):

PR #2490, opened 2025-12-05
Boyon Gate Hall R-Mode Spark Interrupt

aquanight:
Room notes:
- Respawning Zebbo farm.
- Green gate runway is Expert-usable at least, otherwise 1-tile runway between the bottom doors.
- Convenient Boyon right as you walljump up from the doors. From the runway... have your pick.
- [1, 4] movement to get up from the shinecharge is covered by starting the spark strat from [4].
- Candidate for Combined Direct + Indirect G-Mode to despawn the gate when entering from below and still use an R-Mode interrupt.

### DIFF
--- a/region/maridia/outer/Boyon Gate Hall.json
+++ b/region/maridia/outer/Boyon Gate Hall.json
@@ -1457,10 +1457,11 @@
             {"canShineCharge": {"usedTiles": 13, "openEnd": 1}}
           ]},
           {"and": [
-            {"or": [
-              {"doorUnlockedAtNode": 1},
-              {"doorUnlockedAtNode": 2}
-            ]},
+            {"doorUnlockedAtNode": 1},
+            {"canShineCharge": {"usedTiles": 12, "openEnd": 1}}
+          ]},
+          {"and": [
+            {"doorUnlockedAtNode": 2},
             {"canShineCharge": {"usedTiles": 13, "openEnd": 0}}
           ]},
           {"canShineCharge": {"usedTiles": 12, "openEnd": 0}}


region/maridia/outer/Boyon Gate Hall.json
```diff
@@ -1398,6 +1398,47 @@
...
+        {"nodeId": 1, "types": ["ammo"], "requires": []},
+        {"nodeId": 2, "types": ["ammo"], "requires": []}
+      ],
+      "resetsObstacles": ["R-Mode"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Fill up on the respawning Zebbos.",
+        "Open the gate to use that runway, or else run between the bottom doors and quickly walljump",
+        "back up to the Boyon."
```
kjbranch: You forgot to include ways to get up in the code. Probably hijump spring would also work.
```suggestion
        "Fill up on the respawning Zebbos. Use the runway with the gate or at the bottom of the room."
```

aquanight:
> > The strat starts at [4] so you already had to get up. (If you came form the upper right door, you had to open the gate, so you have that alternative.)
> 
> This would need to be mentioned, because that logic is pretty fragile. But in this case, it also won't work. You could get up with an IBJ, tricky speedy spring ball jump, maybe an x-ray climb or grapple teleport.

Yeah, IBJ wouldn't get up fast enough. I didn't see any X-Ray climbs and Grapple Teleport isn't (yet known to be?) compatible with R-Mode, so I didn't even think about those.

I didn't see a tricky speedy springball jump, but the HJB tricky dash jump video looks like it fits within shinecharge timer so I kept that one.

aquanight:
> With the Gate Runway and Double Door Runway combined in one strat, its awkward to describe how Morph provides more ShineCharge tiles in the note. Or to suggest using a frozen enemy.

Would it be worth splitting the strat so we can have this?

region/maridia/outer/Boyon Gate Hall.json
```diff
@@ -1398,6 +1398,87 @@
...
+        ]},
+        {"or": [
+          {"and": [
+            {"doorUnlockedAtNode": 1},
+            {"doorUnlockedAtNode": 2},
+            {"canShineCharge": {"usedTiles": 13, "openEnd": 1}}
+          ]},
+          {"and": [
+            {"or": [
+              {"doorUnlockedAtNode": 1},
```
osse101: With [1] open, I think you get something closer to 12+1.  You still need to shortcharge right to left, into the open door now.

PR #2492, opened 2025-12-05
Fish Tank R-Mode Spark Interrupt

aquanight:
Room notes:
- Four Skultera to farm if you can't kill Pirates (Plasma and Screw Attack only).
- As tempting as the "free Speed kill" bottom Pirate is, you need that one for the interrupt.
- Movement to get to the farming points (and back) is somewhat non-trivial, but all requirement options are covered by the `Gravity` options for getting to [6], so the strat starts there.

### DIFF
--- a/region/maridia/outer/Fish Tank.json
+++ b/region/maridia/outer/Fish Tank.json
@@ -2065,6 +2065,31 @@
       "blueSuitChecked": true
     },
     {
+      "link": [2, 2],
+      "name": "R-Mode Spark Interrupt",
+      "requires": [
+        "Gravity",
+        {"obstaclesCleared": ["R-Mode"]},
+        {"or": [
+          "h_CrystalFlashForReserveEnergy",
+          {"and": [
+            "h_RModeCanRefillReserves",
+            {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
+            {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
+          ]}
+        ]},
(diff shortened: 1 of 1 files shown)


region/maridia/outer/Fish Tank.json
```diff
@@ -3568,6 +3568,47 @@
...
+                  "Plasma",
+                  "ScrewAttack",
+                  {"and": [
+                    "canWalljump",
+                    {"canShineCharge": {"usedTiles": 33, "openEnd": 0}},
+                    {"shinespark": {"frames": 2, "excessFrames": 2}}
+                  ]}
+                ]}
+              ]}
+            ]}
```
kjbranch: with full missiles and killing the 2 available skulteras, you have almost a 90% chance of gaining some reserve energy (and you could likely farm the other 2 with a gravity jump or wall jump). After the strat, you're blue and can farm the 4 pirates. Maybe this isnt the best way to be representing these, but I don't think this strat needs all these extra options
```suggestion
            {"partialRefill": {"type": "ReserveEnergy", "limit": 20}},
```

region/maridia/outer/Fish Tank.json
```diff
@@ -3568,6 +3568,47 @@
...
+          ]}
+        ]},
+        {"canShineCharge": {"usedTiles": 33, "openEnd": 0}},
```
kjbranch:
```suggestion
        {"canShineCharge": {"usedTiles": 32, "openEnd": 1}},
```

PR #2494, opened 2025-12-05
Main Street R-Mode Spark Interrupt

aquanight:
Room notes:
- Two Scisers accessible in the platforms above,  plus the global Sciser used to interrupt.
  - You can get to three more with gravity jumps, but two are already enough so I didn't include them.
  - Crab Supers Sciser is considered inaccessible.
- Assumed use of one Super to knock down the global Sciser in case it starts climbing above the runway - if it passes the top corner of that wall, you're stuck waiting until it loops near the Speed block to try to knock it down again - or else wait for it to lap all the way around.

### DIFF
--- a/region/maridia/outer/Main Street.json
+++ b/region/maridia/outer/Main Street.json
@@ -3052,16 +3052,7 @@
           "h_CrystalFlashForReserveEnergy",
           {"and": [
             "h_RModeCanRefillReserves",
-            {"or": [
-              {"and": [
-                {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 0}]},
-                {"partialRefill": {"type": "ReserveEnergy", "limit": 60}}
-              ]},
-              {"and": [
-                {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 1}]},
-                {"partialRefill": {"type": "ReserveEnergy", "limit": 40}}
-              ]}
-            ]}
+            {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
           ]}
         ]},
         {"or": [


region/maridia/outer/Main Street.json
```diff
@@ -3042,6 +3042,50 @@
...
+            "h_RModeCanRefillReserves",
+            {"or": [
+              {"and": [
+                {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 0}]},
+                {"partialRefill": {"type": "ReserveEnergy", "limit": 60}}
+              ]},
+              {"and": [
+                {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 1}]},
+                {"partialRefill": {"type": "ReserveEnergy", "limit": 40}}
+              ]}
```
kjbranch: Is this right? You can farm 3 fish + 2 crabs with gravity and a crouch jump. Tho realistically, you'll have gravity jump as well and that gives even more enemies

region/maridia/outer/Main Street.json
```diff
@@ -3042,6 +3042,50 @@
...
+            ]}
+          ]}
+        ]},
+        {"or": [
+          {"ammo": {"type": "Super", "count": 1}},
+          "canBePatient"
+        ]},
```
kjbranch:
```suggestion
```

region/maridia/outer/Main Street.json
```diff
@@ -3042,6 +3042,50 @@
...
+        {"or": [
+          {"canShineCharge": {"usedTiles": 24, "gentleUpTiles": 3, "gentleDownTiles": 3, "openEnd": 1}},
+          {"and": [
+            {"doorUnlockedAtNode": 3},
+            {"canShineCharge": {"usedTiles": 25, "gentleUpTiles": 3, "gentleDownTiles": 3, "openEnd": 1}}
```
kjbranch: no open ends because you dont want to fall off
```suggestion
          {"canShineCharge": {"usedTiles": 24, "gentleUpTiles": 3, "gentleDownTiles": 3, "openEnd": 0}},
          {"and": [
            {"doorUnlockedAtNode": 3},
            {"canShineCharge": {"usedTiles": 25, "gentleUpTiles": 3, "gentleDownTiles": 3, "openEnd": 0}}
```

kjbranch: It's because you can fall off for the in-room spark .... but yeah, I guess you could run left to right, so this can be kept how you had it.

region/maridia/outer/Main Street.json
aquanight: 20E with not-full-ammo is 86% with 3 Sciser + 3 Skultera and 96% with all Scisers.

PR #2496, opened 2025-12-06
Botwoon E-Tank artificial morph bomb Spring Ball remote acquire

blkerby:
Videos:
- right side (by Sam): https://videos.maprando.com/video/7042
- left side: https://videos.maprando.com/video/9405

### DIFF
--- a/region/maridia/inner-pink/Botwoon Energy Tank Room.json
+++ b/region/maridia/inner-pink/Botwoon Energy Tank Room.json
@@ -2306,10 +2306,11 @@
         "h_artificialMorphUnderwaterBombIntoSpringBallJump",
         "canInsaneJump"
       ],
+      "clearsObstacles": ["B"],
       "collectsItems": [5],
       "flashSuitChecked": true,
       "note": [
-        "Use Bombs to kill the Puyos. Then gain speed by rolling from right to left.",
+        "Use Bombs to kill the Zoas. Then gain speed by rolling from right to left.",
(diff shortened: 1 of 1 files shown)


blkerby:
> doesn't this have to be in direct g-mode to remote acquire the item?

Good catch, I forgot to add the direct G-mode versions. There's still versions that doesn't require direct, but in this case there's no remote acquire and it wouldn't apply to Map Rando logic (but could still be useful if the item is Morph, or if items can be kept after reload).

region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -2238,6 +2293,50 @@
...
+        }
+      },
+      "requires": [
+        "h_artificialMorphUnderwaterBombIntoSpringBallJump",
+        "canInsaneJump"
+      ],
+      "collectsItems": [5],
+      "flashSuitChecked": true,
+      "note": [
+        "Use Bombs to kill the Puyos. Then gain speed by rolling from right to left.",
```
osse101:
```suggestion
        "Use Bombs to kill the Zoas. Then gain speed by rolling from right to left.",
```

region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -2238,6 +2293,50 @@
...
```
osse101:
```suggestion
        "Use Bombs to kill the Zoas. Then gain speed by rolling from right to left.",
```

PR #2499, opened 2025-12-06
add stationary morph turnaround as item collection method for west sand pit top right item.

nn357:
https://videos.maprando.com/video/9413

was bought up in the logic channel today by "quick10draw"

" I was surprised to find that I had gotten them early because my usual strategy for getting the right item is not included in the logic page.  The strat is rather simple:  Roll against the item stand while morphed, turn around in place while morphed, unmorph, then turn around to grab the item."

Not sure what, if anything to put in the "requires" section.

The stationary turnaround in morph has a 2 frame leniency for the direction change.

### DIFF
--- a/region/maridia/inner-pink/West Sand Pit.json
+++ b/region/maridia/inner-pink/West Sand Pit.json
@@ -607,7 +607,7 @@
       "link": [6, 4],
       "name": "Turnaround Aim Cancel",
       "requires": [
-        {"notable": "Right Item Collection"},
+        {"notable": "Right Item Turn Around Collection"},
         "canTurnaroundAimCancel"
       ],
       "note": [
@@ -619,7 +619,7 @@
       "link": [6, 4],
       "name": "Stationary Morph Turn Around",
       "requires": [
-        {"notable": "Right Item Collection"},
+        {"notable": "Right Item Turn Around Collection"},
         "canMorphTurnaround"
       ],
       "note": [
@@ -631,7 +631,7 @@
       "link": [6, 4],
       "name": "X-Ray Standup",
       "requires": [
-        {"notable": "Right Item Collection"},
+        {"notable": "Right Item Turn Around Collection"},
         "canXRayStandUp"
       ],
        "note": [
@@ -644,7 +644,7 @@
       "link": [6, 4],
       "name": "X-Ray Turnaround",
       "requires": [
-        {"notable": "Right Item Collection"},
+        {"notable": "Right Item Turn Around Collection"},
         "canXRayTurnaround"
       ],
        "note": [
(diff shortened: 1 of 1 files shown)


blkerby:
This would need a canMorphTurnaround requirement.

This would be a fairly impactful change for Hard logic so we'll want to think about it carefully. I think most places where we expect a morph turnaround (e.g. Pink Brinstar Wave Gate Room) moonwalk is an easier alternative. I think the other cases are underwater (like in Botwoon Hallway) where the turnaround is less precise, though those cases are more complex overall and slower to retry.

I think the main concern is if a 1-2 frame press may be too mechanically demanding for this difficulty. It seems more difficult than the turnaround aim cancel method (though a bit more intuitive) so that makes things awkward. We could consider bumping canMorphTurnaround up to Very Hard while possibly keeping a new underwater version (e.g. canUnderwaterMorphTurnaround) in Hard?

blkerby:
> it was complaining about the notable not having an id.... is it not autogenerated?

The notable IDs are manually generated: use the value from the "nextNotableId" at the end of the JSON, and then increment the "nextNotableId".

region/maridia/inner-pink/West Sand Pit.json
```diff
@@ -757,8 +771,15 @@
         "HiJump with a good jump from the sand can reach the Solid Rock Maze region.  Use the sandfall if Samus gets stuck in the sand.",
         "Perform several wall jumps to climb up to the water level, then precise wall jumps to break free."
       ]
+    },
+    {
+      "id": 3,
+      "name": "Right Item Collection",
+      "note": [
+        "Collecting the right side Item without Springball or explosives can be achieved with a turnaround aim cancel, stationary Morph turnaround or an X-Ray turnaround."
```
osse101:
```suggestion
        "Collecting the right side Item is possible by aligning Samus on the closest pixel to the item (while facing left) and then performing a turnaround towards the right.",
        "This can be achieved with a Morph turnaround, an X-Ray turnaround, or a small number of turnaround aim cancels."
```

region/maridia/inner-pink/West Sand Pit.json
```diff
@@ -607,15 +607,29 @@
...
     },
+    {
+      "link": [6, 4],
+      "name": "Stationary Morph Turn Around",
+      "requires": [
+        {"notable": "Right Item Collection"},
+        "canMorphTurnaround"
+      ],
+      "note": [
+        "Roll against the pedestal while morphed. Press left for 1 or 2 frames to stationary turn around, unmorph and press right to collect the item"
```
osse101:
```suggestion
        "Roll against the pedestal while morphed. Press left for 1 or 2 frames to turn around without moving a pixel, unmorph and press right to collect the item"
```

osse101 (commented):
It would help to add notes to the other two existing strats too.

The notable helps with visibility, but the spoiler log would point to a single method so the alternatives could get lost that way. 🤷

kjbranch (dismissed):

PR #2504, opened 2025-12-09
Croc Speedway: new node

nn357:
-wip-

### DIFF
--- a/region/norfair/west/Crocomire Speedway.json
+++ b/region/norfair/west/Crocomire Speedway.json
@@ -3371,6 +3371,7 @@
           {"heatFrames": 30}
         ]}
       ],
+      "devNote": "FIXME: Could have separate strats with tighter heat frames for going down or right.",
       "flashSuitChecked": true
     }
   ],


osse101:
Looks hard to work with and fine-tune.
- You either have speed and kill the pirate naturally
- End a shinespark right in front of it and probably have Varia to make use of the new node
- Reverse spark with energy puts you on the other side of the spike pit, and then there is a lot of running around.
- GMode junction node?  And have Varia if you want camera back.
But you likely want Varia anyway for spikesuits.

osse101 (commented):
- Needs the useFlashSuit / GMode connections to see the value of the node.  Right now you reach 9 by having already killed everything with speed.
- canDash to most strats
- Speedless can be confused with Dashless, when we need to cross the crumble bridge later.  Especially where it mentions damage boosts which is one of the ways you would get through the spike pit when dashless.

osse101 (approved):
I don't think the heatframes used to reach/leave 9 put it at the location marked on the diagram.

region/norfair/west/Crocomire Speedway.json
```diff
@@ -3147,6 +3172,206 @@
...
+    {
+      "link": [9, 6],
+      "name": "Base",
+      "requires": [
+        {"obstaclesCleared": ["A"]},
+        {"heatFrames": 90},
+        {"or": [
+          "SpeedBooster",
+          {"heatFrames": 60}
+        ]}
```
blkerby:
```suggestion
        {"heatFrames": 120},
        {"or": [
          "SpeedBooster",
          {"heatFrames": 30}
        ]}
```

PR #2512, opened 2025-12-11
Indiana Jones Room R-Mode Spark Interrupt

aquanight:
Two strats from each door: one has no Power Bombs, the other can clear the PB blocks.

Must be separate due to tracking the "A" PB blocks obstacle.

### DIFF
--- a/region/norfair/crocomire/Indiana Jones Room.json
+++ b/region/norfair/crocomire/Indiana Jones Room.json
@@ -463,20 +463,14 @@
         {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
         {"partialRefill": {"type": "ReserveEnergy", "limit": 20}},
         {"canShineCharge": {"usedTiles": 17, "openEnd": 0}},
-        {"or": [
-          {"autoReserveTrigger": {"maxReserveEnergy": 95}},
-          {"and": [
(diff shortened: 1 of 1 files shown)


region/norfair/crocomire/Indiana Jones Room.json
```diff
@@ -1522,6 +1583,57 @@
...
+        "comesThroughToilet": "no"
+      },
+      "requires": [
+        "h_RModeCanRefillReserves",
+        {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
+        {"partialRefill": {"type": "ReserveEnergy", "limit": 20}},
+        {"canShineCharge": {"usedTiles": 17, "openEnd": 0}},
+        {"autoReserveTrigger": {}},
+        "canRModeSparkInterrupt",
+        {"acidFrames": 30}
```
kjbranch: of the 4 strats, this is the only one with acidFrames for interrupting, though, the Mellas sound easier to use

aquanight: Should all have the option of Mella or acid now. Mella may be better at single reserve tank (depending on acid leniency), but there's still a case for acid (4 R-tanks + CF)

PR #2514, opened 2025-12-11
Cathedral Entrance R-Mode Spark Interrupt

aquanight:
Getting to [4] clears all the enemies, so should cover Sm. Dessgeega kill requirements.

### DIFF
--- a/region/norfair/east/Cathedral Entrance.json
+++ b/region/norfair/east/Cathedral Entrance.json
@@ -1156,7 +1156,7 @@
         {"canShineCharge": {"usedTiles": 43, "openEnd": 0}},
         {"or": [
           "canPreciseWalljump",
-          "canTrickyJump",
+          "canInsaneJump",
           "SpaceJump"
         ]},
         {"or": [


region/norfair/east/Cathedral Entrance.json
```diff
@@ -1217,6 +1253,32 @@
...
+      "name": "R-Mode Spark Interrupt",
+      "requires": [
+        {"obstaclesCleared": ["R-Mode"]},
+        {"or": [
+          "h_heatedCrystalFlashForReserveEnergy",
+          {"and": [
+            "h_heatProof",
+            "h_RModeCanRefillReserves",
+            {"partialRefill": {"type": "ReserveEnergy", "limit": 40}},
+            {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 0}]}
```
kjbranch: I still think lower reqs > ending with a bit more energy
```suggestion
            {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
```

osse101:
I think there should be some serious tech on everything that goes into heat interrupting at a door; at the left door in particular.  Is it reasonable to be on the ground for the xray trick?

aquanight:
> I think there should be some serious tech on everything that goes into heat interrupting at a door; at the left door in particular. Is it reasonable to be on the ground for the xray trick?

These strats are based on the "Leave Shinecharged" strats. In theory, you deviate from that strat in two ways:
- Stopping at the door (e.g. perhaps we should require `canStopOnADime`) rather than going through.
- Starting the strat when you have the right amount of energy to "run out" during windup.

The videos for them show leaving grounded, so X-Ray buffering should be free (and you can arm X-Ray during the jump since you aren't running - or at least arm whatever is before it and be able to quickly switch before windup). The only catch is timing to be able to windup just before 00 energy but this is generally true of heat interrupt. 

If using heat throughout, timing the start to try to align the 00 energy moment for for the point at the door where you interrupt is pretty narrow. The energy windows are for gaining the shinecharge are:
- Left door using hero shot: 39 - 44
- Left door, not hero shot: 37 - 44
- Right door: 31 - 44
(44 is the highest energy you can shinecharge and interrupt from heat with.)

The actual amount of energy you'd need to start with will depend on how good your execution is (the heatFrames provided can give a basic idea of what logic would need).

There's theoretically enough time to toggle Varia mid-strat, if it was needed to do this on 99 energy, but I'm not sure if `canDisableEquipment` is enough to cover disabling in the middle of technical execution.

PR #2525, opened 2025-12-12
Lower Norfair Elevator R-Mode Spark Interrupt

aquanight:
Due to the requirement to CF and heat interrupt, it is free to complete the strat basically anywhere in the room.

Thus each door gets 3 variants, ending at either itself, the other door, or going down the elevator.

### DIFF
--- a/region/norfair/east/Lower Norfair Elevator.json
+++ b/region/norfair/east/Lower Norfair Elevator.json
@@ -497,10 +497,8 @@
         "h_heatTriggerRModeSparkInterrupt"
       ],
       "unlocksDoors": [
-        {"nodeId": 1, "types": ["missiles", "super"], "requires": []},
-        {"nodeId": 1, "types": ["powerbomb"], "requires": [], "useImplicitRequires": false},
-        {"nodeId": 2, "types": ["missiles", "super"], "requires": []},
-        {"nodeId": 2, "types": ["powerbomb"], "requires": [], "useImplicitRequires": false}
+        {"nodeId": 1, "types": ["ammo"], "requires": []},
+        {"nodeId": 2, "types": ["ammo"], "requires": []}
       ],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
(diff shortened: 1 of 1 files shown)


blkerby:
Yeah I don't think we've been entirely consistent about how different types of runway ends are handled. We've generally been subtracting a half tile for temp blue where you're not allowed to touch a wall; and I guess not being able to touch a door transition should be effectively the same thing. At some point I would like to update the `canShineCharge` and `getBlueSpeed` schemas to add an enum field to describe what happens at the end of the runway; that way the actual runway geometry could be encoded exactly, with any adjustments being computed appropriately on top of it.

aquanight:
> Talked to Maddo about this and we think that it's best not to represent strats this tightly. Getting this short of a shortcharge, first try, with heat damage, is already Beyond difficulty. Trying to then squeeze out a minimum number of heat frames is not ideal, especially when it adds this much complexity.

With heat protection you don't have to do it first try. There's enough shinecharge frames to pause, disable Varia, unpause, then pause again for pause abuse interrupt if that's the method.

Absolutely true about 12 tile being the most cursed thing ever though.

kjbranch:
> With heat protection you don't have to do it first try. 

With heat protection, the ending node of the strat doesn't matter, so the extra ending nodes are only relevant when taking damage

kjbranch:
One other thing I was talking to Maddo about... We are thinking that it doesnt make sense to add useImplicitRequires = false for these types of strats yet. In general, we have not been using them for CF strats, and it would make sense to not have these types of strats be the only ones in logic, so we should wait to add this type of tightening until we are ready to make a full pass of it.

I think some of the reason that CF strats dont currently unlock doors is also because it can get pretty complex. Think of a room like Parlor, where you need to CF. You could open one of many doors, so we would need a duplicate variant at each door. And some of them could be placed in unusual spots to open multiple doors. Sounds like it might need a tech for that. Fortunately, when doing a CF, you will typically have 4 extra PBs, so this type of precision is extremely unlikely to make a difference.

PR #2527, opened 2025-12-12
add spikesuits / xmode bluesuits to lava grapple tunnel

nn357:
https://videos.maprando.com/video/9454 - come in shinecharged bluesuit 
https://videos.maprando.com/video/9453 - come in shinecharged flashsuit


osse101 (commented):
The clip showed you can use the initial spikesuit spark to go through the door.  That didn't look very useful, but you could use it to cross the room instead.

PR #2537, opened 2025-12-13
Ice Beam Snake Room R-Mode Spark Interrupt

aquanight:
Couldn't get the Fune fireball timing to work consistently so I went with heat interrupt.

### DIFF
--- a/region/norfair/west/Ice Beam Snake Room.json
+++ b/region/norfair/west/Ice Beam Snake Room.json
@@ -599,14 +599,6 @@
             {"heatFrames": 170}
           ]},
           {"and": [
-            {"or": [
-              "Morph",
-              "canDodgeWhileShooting",
-              "ScrewAttack",
-              "canPseudoScrew",
-              {"and": ["Wave", "Ice"]},
-              {"enemyDamage": {"enemy": "Fune", "type": "fireball", "hits": 2}}
-            ]},
             "h_heatProof",
             "h_RModeCanRefillReserves",
             {"enemyKill": {"enemies": [["Fune", "Fune"], ["Fune"]]}},


region/norfair/west/Ice Beam Snake Room.json
```diff
@@ -950,6 +986,43 @@
...
+      "link": [2, 3],
+      "name": "R-Mode Spark Interrupt",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "Morph",
+        {"enemyKill": {"enemies": [["Fune"]]}},
+        {"or": [
+          "h_heatedCrystalFlashForReserveEnergy",
```
osse101: here too

region/norfair/west/Ice Beam Snake Room.json
```diff
@@ -585,6 +585,49 @@
...
     },
+    {
+      "link": [1, 3],
+      "name": "R-Mode Spark Interrupt",
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"or": [
+          {"and": [ "Morph", {"heatFrames": 200} ]},
```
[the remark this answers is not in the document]
aquanight: [1, 3] should be improved now. There are still some heatFrames before the CF because you aren't entering at a position that's able to immediately CF (you can buffer/prepare for the aim down -> shoot -> morph while holding X-Ray for R-Mode to then drop on the ledge and CF), but the remainder of the 200 is after.

region/norfair/west/Ice Beam Snake Room.json
```diff
@@ -950,6 +998,46 @@
...
+        "h_heatTriggerRModeSparkInterrupt",
+        {"heatFrames": 10}
+      ],
+      "unlocksDoors": [
+        {"nodeId": 3, "types": ["ammo"], "requires": []}
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Kill the Funes first, then the Sovas. Shinecharge at the bottom and use heat damage to interrupt."
```
osse101: Why that order?  It sounds wrong since it may increase the chances of ammo drops.

PR #2543, opened 2025-12-16
Add more notables for Mt. Everest

blkerby:
I haven't spent a lot of time with these but thought these various crab climbs probably needed some shifting around in difficulty, and making more separate notables is a way to give us that flexibility. Possible difficulty placement:
- Tricky Grapple Swing (into Beach): Expert
- Right Crab Climb with Supers and HiJump: Expert+
- Right Crab Climb with Supers and Spring Ball: Expert+
- Right Crab Climb with Only HiJump: Extreme
- Right Crab Climb with Only Spring Ball: Extreme+
This also adds a `canTrickyJump` requirement to the "Spring Ball Jump with HiJump off a Crab" strat to climb halfway up the right side (to node 10). This was in Hard but reaching the top-right door from there requires something in Very Hard. The jump also seemed too tricky for Hard, especially considering it's fairly slow to set up an attempt.

### DIFF
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -4995,7 +4995,6 @@
       "requires": [
         {"notable": "Right Crab Climb with Supers and HiJump"},
         "canSuitlessMaridia",
-        "canCrazyCrabClimb",
         "HiJump",
         {"ammo": {"type": "Super", "count": 1}}
       ],


osse101:
- I'd put Right Crab Climb with Supers and HiJump: Expert, but like the position of the rest
- Only Hijump maybe doesn't need to be notable since the concept of the crab climb isn't hard and canCrazyCrabClimb covers the movement expectation.  But then its the odd one out by not being notable.
- The 3 crab strats can be notable for being 3 crabs.  2 crabs are a maybe.  But control over the placement is what we want with adding notables.
- For another time; are we gaining anything by having them split into lower part and upper part?

Did you want to keep Only Hijump as a notable?

blkerby:
> * I'd put Right Crab Climb with Supers and HiJump: Expert, but like the position of the rest

Sounds good.

>     * Only Hijump maybe doesn't need to be notable since the concept of the crab climb isn't hard and canCrazyCrabClimb covers the movement expectation.  But then its the odd one out by not being notable.

I think we have to move down canCrazyCrabClimb tech to Expert in order to put the Supers + HiJump strat into Expert. And in this case the Only Hijump has to stay notable I think. 

I just updated the remaining canCrazyCrabClimb strats (left-side Everest, and Main Street with Spring Ball) to add a canBePatient requirement to keep them in Extreme, if that sounds good?

osse101 (approved):
I dunno if `Right Crab Climb with Supers and HiJump` truly is a crazycrabclimb. crazycrabclimb to me is getting close enough to the slopes that you can accidently be pushed through the crab.  The springball counterpart probably is though.  On both you can misjudge your positioning.  Doesn't make much of a difference with all the notables.

PR #2546, opened 2025-12-17
Rinka Shaft R-Mode Spark Interrupt

aquanight:


osse101:
The `Extend runway with an open door` change would apply to the other two strats too

PR #2549, opened 2025-12-17
Mother Brain Room R-Mode Spark Interrupt

aquanight:
Rainbow beam didn't pan out for a "classic" spark interrupt, but the tank can be used for Spike X-Mode and then leave with an R-Mode blue suit.

Requires all four Zebetites open. Need to evaluate if Frozen Rinka Zebetite Un-skip is possible.

### DIFF
--- a/region/tourian/main/Mother Brain Room.json
+++ b/region/tourian/main/Mother Brain Room.json
@@ -323,6 +323,7 @@
       "flashSuitChecked": true
     },
     {
+      "id": 47,
       "link": [1, 1],
       "name": "R-Mode Crystal Flash Interrupt",
       "requires": [
@@ -526,6 +527,7 @@
       "flashSuitChecked": true
     },
     {
+      "id": 48,
       "link": [2, 2],
       "name": "R-Mode Crystal Flash Interrupt",
       "requires": [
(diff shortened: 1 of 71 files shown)


region/tourian/main/Mother Brain Room.json
```diff
@@ -656,6 +656,34 @@
...
+        "canXMode",
+        {"spikeHits": 1},
+        "h_shinechargeMaxRunway",
+        {"autoReserveTrigger": {"maxReserveEnergy": 95}},
+        "canRModeSparkInterrupt"
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Bounce off of a frozen Rinka and onto the corner to enter Spike X-Mode. Gain shinecharge, then use any Rinka to interrupt."
```
osse101: - Quickly freeze all active rinkas at the same time to CF without dying.
- Mention the MB platform should be treated as a spike.

osse101 (commented):
Make a notable for gaining XMode on MBs platform.

I had a ton of trouble getting a good rinka position, and you have to do all that while keeping your energy 1 rinka hit from death after a spike hit.  So it's like insane+

PR #2565, opened 2025-12-20
Wrecked Ship East Super R-Mode Spark Interrupt

aquanight:
This also corrects the R-Mode Stand-up Clip use of obstacle, and correctly calls an `autoReserveTrigger` for energy tracking.

For the spark interrupt itself: it takes some manipulation, but it is possible to Crystal Flash and kill only one Bull, avoiding the need to make the strat a Pause Abuse one.

### DIFF
--- a/region/wreckedship/main/Wrecked Ship East Super Room.json
+++ b/region/wreckedship/main/Wrecked Ship East Super Room.json
@@ -714,10 +714,7 @@
       "requires": [
         {"obstaclesCleared": ["R-Mode"]},
         {"or": [
-          {"and": [
-            "h_CrystalFlashForReserveEnergy",
-            "canTrickyDodgeEnemies"
-          ]},
+          "h_CrystalFlashForReserveEnergy",
           {"and": [
             "h_RModeCanRefillReserves",
             {"enemyKill": {"enemies": [["Bull", "Bull"]], "excludedWeapons": ["PowerBomb"]}},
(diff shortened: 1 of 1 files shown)


region/wreckedship/main/Wrecked Ship East Super Room.json
```diff
@@ -707,6 +708,41 @@
...
     },
+    {
+      "link": [1, 1],
+      "name": "R-Mode Pause Abuse Spark Interrupt",
+      "requires": [
+        {"obstaclesCleared": ["R-Mode"]},
+        {"or": [
+          {"and": [
+            "h_CrystalFlashForReserveEnergy",
+            "canTrickyDodgeEnemies"
```
osse101: I don't see how you could avoid killing the bulls

region/wreckedship/main/Wrecked Ship East Super Room.json
```diff
@@ -707,6 +708,41 @@
...
+        ]},
+        {"or": [
+          {"canShineCharge": {"usedTiles": 14, "openEnd": 0}},
+          {"and": [
+            "f_DefeatedPhantoon",
+            {"canShineCharge": {"usedTiles": 20, "openEnd": 2}}
+          ]}
+        ]},
+        {"autoReserveTrigger": {"maxReserveEnergy": 95}},
+        "canRModeSparkInterrupt"
```
osse101: - The bulls didn't feel easier to avoid than just using the spikes.  You have to keep them away long enough to shortcharge & keep them close enough to interrupt on.  I was messing up on them running into me during the windup; they kept missing.
- The Robot lasers only seemed to shoot 1 usable shot, and I found that to be too unreliable.

So imo it should require the pause abuse interrupt.

region/wreckedship/main/Wrecked Ship East Super Room.json
aquanight: I had at one point in testing managed to get a CF that killed only one Bull. But since pause abuse is easier anyway...

PR #2569, opened 2025-12-21
climb supers speedkeep spikesuit

nn357:
https://videos.maprando.com/video/9509


osse101:
Two speedkeeps and a spikesuit is unreasonable.  
Another strat for inroom using the notable and some leniency stuff and placement tech is more or less ok.

PR #2570, opened 2025-12-22
Caterpillar Room R-Mode Spark Interrupt

aquanight:


region/brinstar/red/Caterpillar Room.json
```diff
@@ -2369,6 +2369,43 @@
...
+                {"partialRefill": {"type": "ReserveEnergy", "limit": 20}}
+              ]},
+              {"and": [
+                {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 0}]},
+                {"partialRefill": {"type": "ReserveEnergy", "limit": 15}}
+              ]}
+            ]}
+          ]}
+        ]},
+        {"canShineCharge": {"usedTiles": 17, "openEnd": 0}},
```
osse101: - can be patient
- movement item or walljump to climb up with a zero

nn357: you could go from 2-4 with cac spikeboost if you leave that one alive?

PR #2574, opened 2025-12-23
Purple Farming Room R-Mode Spark Interrupt

aquanight:
With Wave shot running past you'll use a heat interrupt but the difference in energy can be made up with some extra post-blue farming.

### DIFF
--- a/region/norfair/east/Purple Farming Room.json
+++ b/region/norfair/east/Purple Farming Room.json
@@ -60,28 +60,6 @@
   ],
   "strats": [
     {
-      "link": [1, 1],
-      "name": "R-Mode Spark Interrupt",
-      "entranceCondition": {
-        "comeInWithRMode": {}
-      },
-      "requires": [
-        {"heatFrames": 60},
-        {"refill": ["Energy"]},
-        "h_runOverRespawningEnemies",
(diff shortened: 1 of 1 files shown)


region/norfair/east/Purple Farming Room.json
```diff
@@ -59,6 +59,28 @@
...
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        {"heatFrames": 60},
+        {"refill": ["Energy"]},
+        "h_runOverRespawningEnemies",
+        {"canShineCharge": {"usedTiles": 13, "gentleDownTiles": 1, "gentleUpTiles": 3, "openEnd": 0}},
+        {"autoReserveTrigger": {}},
+        "canRModeSparkInterrupt"
```
osse101: This ends at 1, so it should have some heat frames.  But it also says you don't farm after proccing reserves so that may be taken care of already by the presumed post strat farm.

osse101 (commented):
This is surprisingly frustrating on 0+1.  To keep drops out of the way you need to wait for the bugs to fly high, and then you kill and collect the lower bugs drops.  Essentially it turns into manually using reserves and carefully positioning for the shortcharge every attempt.

So it should require canBeVeryPatient or heatproof.

region/norfair/east/Purple Farming Room.json
aquanight: Moved the end to [2]

PR #2576, opened 2025-12-23
Worst Room In The Game R-Mode Spark Interrupt

aquanight:
No bottom door strat due to bomb blocks not respawning.

### DIFF
--- a/region/lowernorfair/east/The Worst Room In The Game.json
+++ b/region/lowernorfair/east/The Worst Room In The Game.json
@@ -255,20 +255,6 @@
       "clearsObstacles": ["R-Mode"]
     },
     {
-      "link": [1, 1],
-      "name": "R-Mode Entry, Crystal Flash",
-      "entranceCondition": {
-        "comeInWithRMode": {}
-      },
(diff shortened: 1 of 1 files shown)


region/lowernorfair/east/The Worst Room In The Game.json
```diff
@@ -1685,11 +1778,78 @@
         "and use it to break the bomb blocks below while falling through them."
       ]
     },
+    {
+      "link": [4, 4],
+      "name": "X-Mode Shinecharge, R-Mode Spark Interrupt",
```
osse101: Probably don't add this.  CF in front of a pirate is close enough to the same thing.

aquanight:
For the X-Mode strat, a Crystal Flash can't be used - it's not possible to re-enable X-Ray in this room.

For now I've stuck a "full on Power Bombs" requirement to ensure logic doesn't try to use it with an R-Mode CF entry (which are still valid to use with the 12-tile shortcharge strat). I'll need to investigate a better method to ensure this.

PR #2578, opened 2025-12-23
add slopespark to parlour

nn357:
https://videos.maprando.com/video/9226
https://videos.maprando.com/video/9225
https://videos.maprando.com/video/9224

### DIFF
--- a/region/crateria/central/Parlor and Alcatraz.json
+++ b/region/crateria/central/Parlor and Alcatraz.json
@@ -3183,25 +3183,6 @@
       "flashSuitChecked": true
     },
     {
-      "link": [8, 8],
-      "name": "Gain Flash Suit (Slopespark)",
-      "requires": [
-        {"canShineCharge": {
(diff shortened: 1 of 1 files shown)


region/crateria/central/Parlor and Alcatraz.json
```diff
@@ -3137,6 +3182,25 @@
       "unlocksDoors": [{"types": ["ammo"], "requires": []}],
       "flashSuitChecked": true
     },
+    {
+      "link": [8, 8],
+      "name": "Gain Flash Suit (Slopespark)",
```
osse101: This should exit out the left door.  We have to expect they can 18 tile shinecharge at this level.

osse101: I'm saying this [8,8] is equivalent to the [4, 4] so it should be deleted.
But there is also a missing [8, 1] that could leave through an open door.

PR #2580, opened 2025-12-24
early supers room - add gmode overload plms to top juction strats.

nn357:
Maybe a candidate for can be lucky? I noticed Kyle left a comment on Sams videos to say the global wavers might hit you (i tried it half a dozen times without issue).

The strats could also be useful for carrying a bluesuit across this room where you cannot dash.

https://videos.maprando.com/video/9531 [bluesuit with walljump]
https://videos.maprando.com/video/9530 [bluesuit gmode morph blind ibj]

https://videos.maprando.com/video/5205 [with walljump]
https://videos.maprando.com/video/5206 [with ice and hijump]

### DIFF
--- a/region/brinstar/green/Early Supers Room.json
+++ b/region/brinstar/green/Early Supers Room.json
@@ -163,6 +163,7 @@
       "requires": [
         {"or": [
           "canWalljump",
+          "SpaceJump",
           {"and": [
             "HiJump",
             "canTrickyUseFrozenEnemies"


blkerby:
Ending at node 4 implies that you could reach the right side of the room without taking damage. If there is a way to reliably do this without a Waver hit, I think it would need a note/video to explain how? Having a blue suit would probably be a way, and an option with a Waver hit could also be useful.

nn357:
> Ending at node 4 implies that you could reach the right side of the room without taking damage. If there is a way to reliably do this without a Waver hit, I think it would need a note/video to explain how? Having a blue suit would probably be a way, and an option with a Waver hit could also be useful.

yeah maybe this should end at [1] or if you can lay a pb to kill the wavers, have screw attack or tank a hit it would probably be possible! I'll test out some options and see what is possible.

region/brinstar/green/Early Supers Room.json
```diff
@@ -586,6 +586,43 @@
...
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "any",
+          "morphed": false
+        }
+      },
+      "requires": [
+        {"or": [
+          "canWalljump",
+          {"and": [
```
osse101: Hijump + speed
shinespark... you can run on the crumbles
HiJump springball
speed springball

region/brinstar/green/Early Supers Room.json
```diff
@@ -586,6 +586,43 @@
...
+      "link": [1, 4],
+      "name": "G-Mode Morph Overload PLMS with Camera Scroll Blocks - Blind IBJ to top",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "any",
+          "morphed": true
+        }
+      },
+      "requires": [
+        "h_artificialMorphBombs"
```
osse101:
```suggestion
        "h_artificialMorphBombs",
        "canOffScreenMovement",
```

osse101 (commented):
Yeah I get nowhere before the waver finds me.

region/brinstar/green/Early Supers Room.json
osse101: But then you don't need GMode.

kjbranch (commented):
I think I would want these to end back at 1 with a `"collectsItems": [4]` or take a waver hit or luck or a PB

nn357:
> I think I would want these to end back at 1 with a `"collectsItems": [4]` or take a waver hit or luck or a PB

I will put the 1-1... could add duplicates that end at 4 with a h_usepowerbomb ?

nn357:
> There are lots of ways mentioned to go [1, 4] they could be added or you could add a fixme and I can add them later.

bluesuit / waverhit (with double? suit?) / pb? Was there anything I missed maybe screw attack?

PR #2581, opened 2025-12-24
Colosseum R-Mode Spark Interrupt

aquanight:
Frozen mochtroid is so far the only reliable setup for the X-Mode shinecharge. All variants need to get back out of the water, so traversing the room (and thus farming all 8 Mochtroids) is free.

Left door needs a fast beam or a quick evade to not get instantly eaten by a Mochtroid due to X-Ray cooldown.

### DIFF
--- a/region/maridia/inner-pink/Colosseum.json
+++ b/region/maridia/inner-pink/Colosseum.json
@@ -228,10 +228,8 @@
       "requires": [
         "canTrickyUseFrozenEnemies",
         {"or": [
-          "Spazer",
-          "Wave",
-          "Plasma",
-          "canTrickyDodgeEnemies"
+          {"enemyKill": {"enemies": [["Mochtroid", "Mochtroid"]], "excludedWeapons": ["PowerBeam", "Ice", "Bombs"]}},
+          "canInsaneJump"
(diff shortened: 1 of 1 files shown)


blkerby:
I think this needs a stronger requirement than "canXMode". Maybe a new tech (e.g. "canTrickyXMode") or a notable?

region/maridia/inner-pink/Colosseum.json
```diff
@@ -219,6 +219,59 @@
...
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "canTrickyUseFrozenEnemies",
+        {"or": [
+          "Spazer",
+          "Wave",
+          "Plasma",
+          "canTrickyDodgeEnemies"
+        ]},
```
osse101: I don't know why these beams are here.
But avoiding mochtroid damage is canInsaneJump

aquanight: I found Ice+Spazer (at least) to be fast enough to clear the first Mochtroid before it can do any damage (accounting for X-Ray cooldown).

Probably could alternatively use a Missile for the CF path.

PR #2582, opened 2025-12-24
Check R-Mode obstacle on reserve-tank damage boosts

aquanight:
These three rooms have auto-reserve damage boosts in rooms that have R-Mode obstacles.

This PR adds an `obstaclesNotCleared` check to keep them from being combined with R-Mode strats.

I'm not entirely sure what logical scenario might call for it, but as an alternative to "not R-Mode", these could be looked at as a logical use of Light Pillar: this would allow using the auto-reserve trigger without ending R-Mode.

### DIFF
--- a/region/norfair/east/Cathedral Entrance.json
+++ b/region/norfair/east/Cathedral Entrance.json
@@ -1530,7 +1530,7 @@
       "name": "Sova Boost (Reserve Delay Bomb Boost and Sova Boost)",
       "requires": [
         {"notable": "Sova Boost"},
-        {"obstaclesCleared": ["R-Mode"]},
+        {"obstaclesNotCleared": ["R-Mode"]},
         "h_bombThings",
         "canMidAirMorph",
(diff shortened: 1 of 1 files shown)


osse101 (commented):
Also cathedral entrance Reserve Delay Bomb Boost
and in everest ReserveDoubleDamageBoost
and you can do it in croc escape but I don't think there's a strat for it, Reserve Delay Bomb Boost

aquanight:
> Also cathedral entrance Reserve Delay Bomb Boost

Should have that one now.

> and in everest ReserveDoubleDamageBoost 

For now Mt. Everest doesn't have an R-Mode obstacle (I didn't add one for spark interrupt). If that changes this will need to be covered then.

> and you can do it in croc escape but I don't think there's a strat for it, Reserve Delay Bomb Boost

Same as Mt. Everest, and as you say there's not currently a strat for that.

PR #2583, opened 2025-12-24
East Cactus Alley R-Mode Spark Interrupt

aquanight:
All X-Mode shinecharges are `canBeVeryPatient` due to being 1-try.

### DIFF
--- a/region/maridia/inner-pink/East Cactus Alley.json
+++ b/region/maridia/inner-pink/East Cactus Alley.json
@@ -794,16 +794,16 @@
         "canXMode",
         "h_XModeSpikeHit",
         "h_shinechargeMaxRunway",
-        "canNeutralDamageBoost",
         {"autoReserveTrigger": {}},
-        "canRModePauseAbuseSparkInterrupt"
+        "canRModePauseAbuseSparkInterrupt",
+        {"spikeHits": 1}
       ],
       "resetsObstacles": ["R-Mode"],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
         "Crystal Flash at the door, and enter X-Mode on the spikes and shinecharge.",
-        "Pause abuse on the spikes: land on the spike three tiles from the ledge and face left to neutral damage boost back to safety before the pause."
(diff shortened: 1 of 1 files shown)


blkerby:
Is there a purpose for the strats at nodes 3 and 5? It seems like with this equipment you could reach node 4 and use the strat there instead?

aquanight:
Regarding [3]:
This one was the first strat I wrote as I started working on this room, before I got to [4]. It has the dubious advantage of not needing to traverse [3, 6] but even that's just a disable + midair morph if for some reason you don't have bombs/springball.

Pending bluesuit movement, it does also have the advantage of not having to get back across the room to get back to [1]. That can be important since crossing spikes with blue can be... messy.

Regarding [5]:
I questioned this one too. I thought it might have value in not needing to be a pause-abuse, largely depending on whether or not `canRModeSparkInterrupt` and `canRModePauseAbuseSparkInterrupt` end up in different difficulty tiers (granted the 1-try X-Mode pushes it into Insane through `canBeVeryPatient` no matter what, but still).

blkerby:
There are 2 Cacatacs up at node 4, so I don't think there's a need to pause abuse? And in Very Hard you can traverse right-to-left without needing to touch the spikes: https://dev.maprando.com/logic/room/196/5/6/55.

region/maridia/inner-pink/East Cactus Alley.json
```diff
@@ -747,6 +781,31 @@
...
+        "canNeutralDamageBoost",
+        {"autoReserveTrigger": {}},
+        "canRModePauseAbuseSparkInterrupt"
+      ],
+      "resetsObstacles": ["R-Mode"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Crystal Flash at the door, and enter X-Mode on the spikes and shinecharge.",
+        "Pause abuse on the spikes: land on the spike three tiles from the ledge and face left to neutral damage boost back to safety before the pause."
```
blkerby: For me, damage boosting (neutral or horizontal) doesn't seem to work to get onto the ledge. So I have to take a spike hit after the reserve trigger.

PR #2584, opened 2025-12-24
Spiky Platforms Tunnel R-Mode Spark Interrupt

aquanight:
Room-crossing variants force leaving to reset the room and the trippers.

Crystal Flash isn't an option due to HDMA overload.

### DIFF
--- a/region/norfair/east/Spiky Platforms Tunnel.json
+++ b/region/norfair/east/Spiky Platforms Tunnel.json
@@ -145,6 +145,7 @@
       },
       "requires": [
         "h_heatProof",
+        "Morph",
         "h_RModeCanRefillReserves",
         {"enemyKill": {"enemies": [["Tripper"]]}},
         {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
@@ -332,6 +333,7 @@
       },
       "requires": [
         "h_heatProof",
+        "Morph",
         "h_RModeCanRefillReserves",
         {"enemyKill": {"enemies": [["Tripper"]]}},
         {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
@@ -632,6 +634,7 @@
       },
       "requires": [
         "h_heatProof",
+        "Morph",
         "h_RModeCanRefillReserves",
         {"enemyKill": {"enemies": [["Tripper"]]}},
         {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
@@ -859,6 +862,7 @@
       },
       "requires": [
         "h_heatProof",
+        "Morph",
         "h_RModeCanRefillReserves",
         {"enemyKill": {"enemies": [["Tripper"]]}},
         {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},


region/norfair/east/Spiky Platforms Tunnel.json
```diff
@@ -138,6 +138,52 @@
...
+        "canBeVeryPatient",
+        "canXMode",
+        "h_XModeSpikeHit",
+        "h_shinechargeMaxRunway",
+        "h_heatTriggerRModeSparkInterrupt",
+        {"or": [
+          {"lavaFrames": 200},
+          {"and": [
+            "Gravity",
+            {"lavaFrames": 130}
```
osse101: The movement is weird.  Jumping probably has slightly lower frame count because I get 60 with running.

region/norfair/east/Spiky Platforms Tunnel.json
```diff
@@ -138,6 +138,52 @@
...
+              "canBeLucky"
+            ]}
+          ]}
+        ]},
+        "canBeVeryPatient",
+        "canXMode",
+        "h_XModeSpikeHit",
+        "h_shinechargeMaxRunway",
+        "h_heatTriggerRModeSparkInterrupt",
+        {"or": [
```
osse101: You could ride/kill the opposite tripper and kill it once you are about to leave the room.

region/norfair/east/Spiky Platforms Tunnel.json
```diff
@@ -138,6 +138,52 @@
...
+      "entranceCondition": {
+        "comeInWithRMode": {}
+      },
+      "requires": [
+        "h_heatProof",
+        "h_RModeCanRefillReserves",
+        {"enemyKill": {"enemies": [["Tripper"]]}},
+        {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
+        {"or": [
+          {"partialRefill": {"type": "ReserveEnergy", "limit": 5}},
```
osse101: HeatTriggered interrupt is dicey on 5 energy.

aquanight: Since heatproof is already required, the pause abuse method should cover that. I should make that the primary path.

region/norfair/east/Spiky Platforms Tunnel.json
aquanight: That's included in the kill logic (as an alternative to `canBeLucky`). With two trippers it's about 75% to get at least one large energy.

region/norfair/east/Spiky Platforms Tunnel.json
```diff
@@ -529,6 +633,55 @@
...
+          {"and": [
+            {"or": [
+              "h_lavaProof",
+              "Grapple",
+              {"and": [
+                "canWalljump",
+                {"or": ["Super", "Ice"]}
```
osse101: What is this for, timing your powerbomb?

aquanight:
> I think you need Morph to cross to and from [2] while avoiding lava damage.

Morph is auto-required by `canXMode`. Probably wouldn't hurt to list it though.

PR #2585, opened 2025-12-25
Green Brinstar Main Shaft G-Mode Spark Interrupt

aquanight:
In theory, it's safe to drop down and CF next to the animals since the Zeela will go up.

### DIFF
--- a/region/brinstar/green/Green Brinstar Main Shaft.json
+++ b/region/brinstar/green/Green Brinstar Main Shaft.json
@@ -4325,7 +4325,7 @@
     },
     {
       "link": [8, 9],
-      "name": "G-Mode Spark Interrupt",
+      "name": "Direct G-Mode Spark Interrupt",
       "entranceCondition": {
         "comeInWithGMode": {
           "mode": "direct",


region/brinstar/green/Green Brinstar Main Shaft.json
```diff
@@ -4323,6 +4323,56 @@
...
+          "morphed": false
+        }
+      },
+      "requires": [
+        "Morph",
+        "h_usePowerBomb",
+        {"or": [
+          "canIBJ",
+          "h_useSpringBall",
+          "canMidAirMorph"
```
osse101: just require the midairmorph

region/brinstar/green/Green Brinstar Main Shaft.json
```diff
@@ -4323,6 +4323,56 @@
...
+        "h_CrystalFlashForReserveEnergy",
+        {"or": [
+          "canConsecutiveWalljump",
+          "canLongIBJ",
+          "SpaceJump"
+        ]},
+        {"or": [
+          {"canShineCharge": {"usedTiles": 17, "openEnd": 1}},
+          {"and": [
+            "h_frozenEnemyRunway",
```
osse101: Frozen runway and LongIBJ look like they add another lap of the zeela around the room.
Same with damaging down in the worst case.

aquanight: Mid-air CF didn't pan out - there's just nowhere with enough space, and it takes as many as 3 laps to damage down from 1499 energy, so that's how many Supers it'll take if E-Tanks can't be disabled.

PR #2587, opened 2025-12-25
Thread The Needle R-Mode Spark Interrupt

aquanight:


region/maridia/inner-yellow/Thread The Needle Room.json
```diff
@@ -488,6 +488,45 @@
...
+          "Plasma",
+          "Ice",
+          "canUseGrapple",
+          "ScrewAttack",
+          "canDodgeWhileShooting",
+          {"and": [
+            "h_usePowerBomb",
+            {"ammo": {"type": "PowerBomb", "count": 1}}
+          ]}
+        ]},
```
blkerby:
```suggestion
```
At this level, this whole `or` block is just noise.

PR #2590, opened 2025-12-27
Some Norfair blue suit carry logic: Bubble Mountain, etc.

blkerby:


region/norfair/east/Bubble Mountain.json
```diff
@@ -2992,11 +3217,32 @@
...
+      "requires": [
+        {"haveBlueSuit": {}},
+        {"or": [
+          "h_artificialMorphIBJ",
+          "h_artificialMorphSpringBall"
+        ]}
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true
+    },
```
kjbranch:
```suggestion
```
this wont work in direct, as you'll fall through. It won't work in indirect, because there's no downward setup here and no way to doorlock bypass while morphed and no up door is missing a door cap ... unless we were wanting to support asymmetric door connections or the possibility of coming from a non-vanilla room (with no doorcap)

region/norfair/east/Bubble Mountain.json
```diff
@@ -645,6 +677,7 @@
         ]}
       ],
       "flashSuitChecked": true,
+      "blueSuitChecked": true,
```
osse101: canDash

osse101 (dismissed):

region/norfair/east/Bubble Mountain.json
```diff
@@ -4555,7 +4920,8 @@
       ],
       "resetsObstacles": ["A"],
       "farmCycleDrops": [{"enemy": "Sova", "count": 3}],
-      "flashSuitChecked": true
+      "flashSuitChecked": true,
```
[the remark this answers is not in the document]
blkerby: Added canDash requirements now. If you're having to use the notable it would probably need more cycleFrames and I figure realistically you would probably just use a different farm instead that doesn't require falling down in that case.

PR #2599, opened 2026-01-02
Crab Maze G-Mode Spark Interrupt

aquanight:
Artificial morph from top door.

For CF, you have to overload through the shortcut tunnel and CF on the left side. Farming after the fact will only guarantee farming out of health-bomb, after that they drop most likely power bombs.

### DIFF
--- a/region/crateria/east/Crab Maze.json
+++ b/region/crateria/east/Crab Maze.json
@@ -416,8 +416,7 @@
         "Without Springball or Gravity, it is possible to overload PLMs with the crumble blocks with bombs.",
         "Take the first left above the waterline. Place bombs in the single tile nook on the left.",
         "The Sciser that starts inside the Morph maze area will move off-screen: wait for it to come back out.",
-        "The second one, however, does not and so should be prioritized to kill before it slips inside the Morph maze.",
-        "Once those are dealt with, use X-Ray until the beam reaches full width to exit G-Mode and remain in R-Mode, then collect drops.",
+        "Once outside of the Morph tunnel area, use X-Ray until the beam reaches full width to exit G-Mode and remain in R-Mode, then collect drops.",
         "Farm the rest of the Scisers, keeping the slower local Sciser at the bottom left corner to use for the interrupt.",
         "Position it in the ceiling cubby hole to safely perform the Shinecharge and jump into the Sciser as it walks back down."
       ]
@@ -459,7 +458,8 @@
         "Crystal Flash under the bomb block; Samus can then no longer use X-Ray.",
         "Jump through, and drop through to the other side. Avoid the Scisers, and manipulate one to the far right.",
         "Shinecharge and windup near the Sciser to interrupt."
-      ]
+      ],
+      "devNote": "Also possible to escape with Space Jump + frame perfect downgrab"
     },
     {
       "id": 34,


region/crateria/east/Crab Maze.json
```diff
@@ -366,6 +366,88 @@
...
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Without Springball or Gravity, it is possible to overload PLMs with the crumble blocks with bombs.",
+        "Take the first left above the waterline. Place bombs in the single tile nook on the left.",
+        "The Sciser that starts inside the Morph maze area will move off-screen: wait for it to come back out.",
+        "The second one, however, does not and so should be prioritized to kill before it slips inside the Morph maze.",
+        "Once those are dealt with, use X-Ray until the beam reaches full width and collect drops. Farm the rest of the Scisers, keeping",
+        "the slower local Sciser at the bottom left corner to use for the interrupt.",
```
kjbranch:
```suggestion
        "Once those are dealt with, use X-Ray until the beam reaches full width to exit G-Mode and remain in R-Mode, then collect drops.",
        "Farm the rest of the Scisers, keeping the slower local Sciser at the bottom left corner to use for the interrupt.",
```
- The reasoning for holding X-ray is going to need a description like this every time. It's not intuitive and isn't listed in any tech description.
- In notes, please end lines near natural breaks, like at the end of sentences.

region/crateria/east/Crab Maze.json
```diff
@@ -366,6 +366,88 @@
...
+        {"autoReserveTrigger": {}},
+        "canRModeSparkInterrupt",
+        {"partialRefill": {"type": "Energy", "limit": 65}}
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Navigate to the shortcut tunnel, evading the first Sciser, and overload the crumble block, then jump through to the",
+        "other side. Avoid the looping Sciser, and Crystal Flash near the door. Samus can no longer use X-Ray.",
+        "Shinecharge (disable Gravity if needed) and windup under the looping Sciser while it drops on top of Samus to interrupt."
```
kjbranch: - how are you overloading PLMs?
- describe which "shortcut tunnel", because its not obvious
- no need to disable gravity if you have it, thats a long runway
- I CF'd in the tunnel, it felt a little easier if your only movement is bombs, but the described location is also fine.

region/crateria/east/Crab Maze.json
```diff
@@ -430,9 +433,19 @@
...
       "requires": [
-        "Gravity",
+        "h_artificialMorphBombs",
         "h_artificialMorphMovement",
         "h_artificialMorphCrystalFlash",
+        {"or": [
+          {"and": [
+            {"or": [
+              "HiJump",
+              "canSunkenTileWideWallClimb"
```
kjbranch: can you add a FIXME that this is possible with a space jump + down grab, its probably beyond territory (need several good blind space jumps and you can barely make it with a close to frame perfect down grab)

region/crateria/east/Crab Maze.json
```diff
@@ -366,6 +366,101 @@
...
+        {"autoReserveTrigger": {}},
+        "canRModeSparkInterrupt",
+        {"partialRefill": {"type": "Energy", "limit": 65}}
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Take the first left above the waterline. Place bombs in the single tile nook on the left to overload PLMs.",
+        "Evade the Sciser and either go through the rest of the Morph maze or an overloaded crumble block, and navigate to the bomb block shortcut tunnel.",
+        "Crystal Flash under the bomb block: Samus can no longer use X-Ray.",
```
kjbranch:
```suggestion
        "Crystal Flash under the bomb block; Samus can then no longer use X-Ray.",
```

kjbranch (approved):

PR #2602, opened 2026-01-03
Fast Pillars Setup G-Mode Spark Interrupt

aquanight:
Artificial Morph from [1] to PB the barrier wall.

Only works with Immobile because it needs an X-Ray activation to release PLMs and remove the barrier wall. (Mobile would require CF which prevents X-Ray use.)

High placed PB could avoid killing the Viola and thus not require heat interrupt but I'm concerned about the pirate lasers overloading and preventing energy drops. It could work with an ammo kill.

Could also investigate psuedo-Varia, but the equivalent R-Mode strat is full-heatproof also so it didn't make sense to relax that here.

### DIFF
--- a/region/lowernorfair/east/Fast Pillars Setup Room.json
+++ b/region/lowernorfair/east/Fast Pillars Setup Room.json
@@ -81,6 +81,17 @@
         [1],
         [2]
       ]
+    },
+    {
+      "id": 6,
+      "name": "R-Mode, Bottom Junction",
+      "nodeType": "junction",
+      "nodeSubType": "r-mode",
+      "mapTileMask": [
+        [1],
+        [1],
+        [2]
+      ]
     }
   ],
   "obstacles": [
@@ -1351,7 +1362,30 @@
       ]
     },
     {
+      "id": 99,
       "link": [1, 5],
+      "name": "G-Mode Morph, Power Bomb the Blocks",
+      "entranceCondition": {
+        "comeInWithGMode": {
+          "mode": "any",
+          "morphed": true
+        }
+      },
+      "requires": [
+        "h_heatedGMode",
+        "h_artificialMorphPowerBomb",
+        {"heatFrames": 120}
+      ],
+      "clearsObstacles": ["A"],
+      "flashSuitChecked": true,
+      "note": [
+        "Place a Power Bomb then exit G-mode to break the blocks.",
+        "If entering in indirect, without Morph, quickly place the Power Bomb and unmorph and start shooting the pirate,",
+        "in order to prevent it from moving forward or placing invisible, stationary lasers over the way down."
+      ]
+    },
+    {
+      "link": [1, 6],
       "name": "Direct G-Mode Morph Spark Interrupt",
       "entranceCondition": {
         "comeInWithGMode": {
@@ -1377,42 +1411,7 @@
           {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
           "canBeLucky"
         ]},
-        {"partialRefill": {"type": "ReserveEnergy", "limit": 5}},
-        {"or": [
-          {"canShineCharge": {
-            "usedTiles": 12,
-            "gentleDownTiles": 2,
-            "gentleUpTiles": 2,
-            "openEnd": 0
-          }},
-          {"and": [
-            {"or": [
-              {"doorUnlockedAtNode": 2},
-              {"doorUnlockedAtNode": 3}
-            ]},
-            {"canShineCharge": {
-              "usedTiles": 13,
-              "gentleDownTiles": 2,
-              "gentleUpTiles": 2,
-              "openEnd": 0
-            }}
-          ]},
-          {"and": [
-            {"doorUnlockedAtNode": 2},
-            {"doorUnlockedAtNode": 3},
-            {"canShineCharge": {
-              "usedTiles": 14,
-              "gentleDownTiles": 2,
-              "gentleUpTiles": 2,
-              "openEnd": 0
-            }}
-          ]}
-        ]},
-        "h_heatTriggerRModeSparkInterrupt"
-      ],
-      "unlocksDoors": [
-        {"nodeId": 2, "types": ["ammo"], "requires": []},
-        {"nodeId": 3, "types": ["ammo"], "requires": []}
+        {"partialRefill": {"type": "ReserveEnergy", "limit": 5}}
       ],
       "clearsObstacles": ["A"],
       "flashSuitChecked": true,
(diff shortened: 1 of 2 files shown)


region/lowernorfair/east/Fast Pillars Setup Room.json
```diff
@@ -1350,6 +1350,101 @@
...
+              "explicitWeapons": ["Charge"]
+            }},
+            "canBePatient"
+          ]}
+        ]},
+        {"or": [
+          {"resourceMissingAtMost": [{"type": "Missile", "count": 0}]},
+          "canBeLucky"
+        ]},
+        {"partialRefill": {"type": "ReserveEnergy", "limit": 20}},
```
kjbranch: seems like a good chance you only get 5

region/lowernorfair/east/Fast Pillars Setup Room.json
```diff
@@ -1350,6 +1350,101 @@
...
+      "unlocksDoors": [
+        {"nodeId": 2, "types": ["ammo"], "requires": []},
+        {"nodeId": 3, "types": ["ammo"], "requires": []}
+      ],
+      "clearsObstacles": ["A"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Enter G-Mode with a full reserve tank. Wait for the wall pirate to jump on Samus to regain mobility and enter artificial morph.",
+        "Immediately lay a power bomb. Once it explodes, either kill the standing pirate, using two more powerbombs or else remain crouched while shooting it.",
```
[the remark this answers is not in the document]
aquanight: I think the only question I still have for charge-slowbeam kill is if it still needs `canBePatient`

aquanight:
I've split the different kill options into four - the better to have a more tailored description for each approach, but if nothing else, it'll at least make it simpler to refine each option and then squish it all back together after:
- Option 1 is the beam/ammo kill. After trying the Charge-only method described, I've found it to be much easier compared to the R-Mode version (indeed, perhaps this method could even be applied back to the R-Mode version and relax the charge-only kill there?) - I'm not convinced `canBePatient` is still needed for it, since Samus can use jump-shots to kill from relative safety.
- Option 2 is the 3 Power Bomb kill, fairly simple enough. With the pirate out of the way, Samus can hug the right wall to avoid lasers instead of having to drop in the left pit, but either one works.
- Option 3 is the Bomb kill, as @kjbranch demonstrated with a video. As suggested, the preferred option is to damage the pirate to between 200-399 hp so that it can die on the _second_ powerbomb hit: this is equivalent to between 17 and 23 bomb hits. I debated on if this needs `canCount`. I'm also not sure if it needs `canBePatient` - the video was less than 90 seconds so it might be a matter of how efficient the player is with the bomb usage.
- Option 4 is another video @kjbranch showed me, using a high-placed PB to clear the barrier without killing Violas, then farming them individually to get the PB back, then energy. In theory, this could be applied to the other methods, yielding an extra 5 energy if starting full on PBs, but I didn't implement this yet to keep the strats from getting too complex.

region/lowernorfair/east/Fast Pillars Setup Room.json
```diff
@@ -1350,6 +1350,284 @@
...
+          "morphed": true,
+          "mobility": "immobile"
+        }
+      },
+      "requires": [
+        "h_heatProof",
+        {"disableEquipment": "ETank"},
+        {"resourceAvailable": [{"type": "RegularEnergy", "count": 99}]},
+        "h_artificialMorphBombs",
+        "h_artificialMorphPowerBomb",
```
kjbranch:
```suggestion
        "h_artificialMorphPowerBomb",
        "canComplexGMode",
```

region/lowernorfair/east/Fast Pillars Setup Room.json
```diff
@@ -1350,6 +1350,284 @@
...
+      "unlocksDoors": [
+        {"nodeId": 2, "types": ["ammo"], "requires": []},
+        {"nodeId": 3, "types": ["ammo"], "requires": []}
+      ],
+      "clearsObstacles": ["A"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Enter G-Mode with a full reserve tank. Wait for the wall pirate to jump on Samus to regain mobility and enter artificial morph.",
+        "Lay three power bombs to kill the standing pirate. Unmorph, go all the way to the right, then activate and hold X-Ray until the beam is at full width to exit G-Mode and remain in R-Mode.",
```
kjbranch: i think it could help to avoid getting boosted up to prevent shots from getting overloaded. 
(Iirc theyve each shot once before regaining mobility, so if you trigger the wall one one more time and the ground one three more times itll be overloaded)

region/lowernorfair/east/Fast Pillars Setup Room.json
```diff
@@ -1350,6 +1350,284 @@
...
+      "requires": [
+        "h_heatProof",
+        {"disableEquipment": "ETank"},
+        {"resourceAvailable": [{"type": "RegularEnergy", "count": 99}]},
+        {"or": [
+          "h_artificialMorphSpringBall",
+          "h_artificialMorphBombs"
+        ]},
+        "h_artificialMorphPowerBomb",
+        {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 1}]},
```
kjbranch:
```suggestion
        {"resourceMissingAtMost": [{"type": "PowerBomb", "count": 0}]},
        "h_artificialMorphPowerBomb",
```
Should be equivalent, I just think itll be a little easier to read maybe

kjbranch (commented):
Seems good. A little verbose being 4 separate strats; I could see all but the bombs kill being combined (so 2 total) not sure what is best, @blkerby might have some thoughts on if more or less strats is preferrable.

region/lowernorfair/east/Fast Pillars Setup Room.json
```diff
@@ -1350,6 +1350,284 @@
...
+          "morphed": true,
+          "mobility": "immobile"
+        }
+      },
+      "requires": [
+        "h_heatProof",
+        {"disableEquipment": "ETank"},
+        {"resourceAvailable": [{"type": "RegularEnergy", "count": 99}]},
+        "h_artificialMorphPowerBomb",
+        {"enemyKill": {
```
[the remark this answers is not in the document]
aquanight: Put Bombs into excludedWeapons. Everything else should be fine.

osse101:
The extra node reads a lot cleaner.  I like it.
What's the updated scope of the PR?

> Adding Violas-alive would be a simple matter of an obstacle, and if this R-Mode junction works I can also point all the R-Mode strats at it.

If everything is duplicated for `ClearsViolaObstacle` and not, it'd be more performant to make another node with RMode + Violas alive.  Not that the current schema handling obstacles is ideal here.

PR #2604, opened 2026-01-04
LN Springball Maze G-Mode Spark Interrupt

aquanight:
G-Mode for heat from 1 and 3. No overload option from 2 to support Crystal Flash Clip.

### DIFF
--- a/region/lowernorfair/east/Lower Norfair Spring Ball Maze.json
+++ b/region/lowernorfair/east/Lower Norfair Spring Ball Maze.json
@@ -1348,8 +1348,8 @@
       "blueSuitChecked": true,
       "note": [
         "Kill the spike Alcoons before climbing to the upper runway. Be sure to jump over the invisible fireball projectiles.",
-        "Crystal Flash at the top right, just before Samus goes offscreen, to clear the runway Alcoon.",
-        "Run right until Samus stops, then start a 1-tap shinecharge to the left. Weave down and into the Alcoon to interrupt."
+        "Crystal Flash at the top right, just before Samus goes off-screen, to clear the runway Alcoon.",
+        "Shinecharge to the left, then weave down and into the Alcoon to interrupt."
       ]
     },
     {


aquanight:
Power beam shots are fast enough to kill the spike Alcoons from [3] without overloading fireballs, but they do still have to be jumped over.

kjbranch:
> G-Mode for heat from 1 and 3.

Does this mean that these strats are only using g-mode to give heat protection? Because I see heatproof options in the strat.

aquanight:
> > G-Mode for heat from 1 and 3.
> 
> Does this mean that these strats are only using g-mode to give heat protection? Because I see heatproof options in the strat.

Correct. Ending with 5 energy would require heatproof to be able to escape after the spark interrupt, but I could just let the "how do you get out through however-many heatFrames" situation handle that.

region/lowernorfair/east/Lower Norfair Spring Ball Maze.json
```diff
@@ -1243,6 +1283,67 @@
...
+          "canSpringBallBombJump",
+          {"and": [
+            "canIBJ",
+            "canBombHorizontally",
+            "canJumpIntoIBJ"
+          ]},
+          {"and": [
+            "canTrickyDashJump",
+            "canInsaneJump"
+          ]},
```
kjbranch:
```suggestion
```
I think this would require the Alcoon to be killed. (If not, it would need to be done first try to prevent it from filling the runway with fireballs. First try tricky dash jumps would be very difficult.)

region/lowernorfair/east/Lower Norfair Spring Ball Maze.json
```diff
@@ -488,6 +488,46 @@
...
+          ]}
+        ]}
+      ],
+      "clearsObstacles": ["A"],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Crystal Flash near the edge - just before Samus goes offscreen - to clear the runway Alcoon. Then run right until Samus stops.",
+        "Start a 1-tap shinecharge to the left - Samus will return on-screen just as blue speed is gained.",
+        "Weave down to the bottom and interrupt on the lower Alcoon."
```
kjbranch:
```suggestion
        "Crystal Flash near the edge - just before Samus goes off-screen - to clear the runway Alcoon.",
        "Shinecharge to the left, then weave down and into the Alcoon to interrupt."
```

region/lowernorfair/east/Lower Norfair Spring Ball Maze.json
```diff
@@ -488,6 +488,46 @@
...
+        {"or": [
+          "canDownBack",
+          "canTrickyJump"
+        ]},
+        {"or": [
+          {"and": [
+            {"disableEquipment": "ETank"},
+            {"autoReserveTrigger": {}},
+            "canRModeSparkInterrupt",
+            {"enemyDamage": {"enemy": "Alcoon", "type": "contact", "hits": 1}}
```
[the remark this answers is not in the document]
aquanight: Yeah, from what I'm finding the best strat for [7, 3] for example is still 120 heatFrames (30 hp), and it'll likely be more under bluesuit movement restrictions.

In testing I've not been able to find a solution that avoids taking a hit from the alcoon and has enough energy to escape, other than the alternate path listed here of using a pause-abuse + xray buffer.

kjbranch (approved):

PR #2605, opened 2026-01-04
Cathedral Entrance blue suit logic

blkerby:
- Refined the existing heat frames a fair bit
- Removed an obsolete old-style farm strat


region/norfair/east/Cathedral Entrance.json
```diff
@@ -1411,6 +1500,7 @@
         {"heatFrames": 215}
       ],
       "flashSuitChecked": true,
+      "blueSuitChecked": true,
```
osse101: Want to put heatproof on this?

PR #2609, opened 2026-01-08
Green Bubbles (etc.) blue suit logic

blkerby:


osse101 (commented):
I think logic could skip the Green Bubbles Tunnel item while visiting 3, but most of the time it is forced.

PR #2611, opened 2026-01-09
G-mode blue suit uses - Crateria

kjbranch:
There were already a few in logic: Landing site, final missile bombway, and maybe one or two more.

### DIFF
--- a/helpers.json
+++ b/helpers.json
@@ -1956,6 +1956,15 @@
             "canBlueSuitGModeSetup",
             {"haveBlueSuit": {}}
           ]
+        },
+        {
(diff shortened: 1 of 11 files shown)


region/wreckedship/main/Basement.json
```diff
@@ -456,7 +456,7 @@
       "name": "Immobile G-Mode Morph Workrobot Kago",
       "entranceCondition": {
         "comeInWithGMode": {
-          "mode": "any",
+          "mode": "direct",
           "morphed": true,
           "mobility": "immobile"
```
kjbranch: I randomly noticed this was the only "immobile" strat that didn't have "direct", so I edited it as well

PR #2618, opened 2026-01-11
Lava Dive blue suit logic

blkerby:


region/norfair/east/Lava Dive Room.json
```diff
@@ -669,6 +732,7 @@
         "h_heatedGModeOpenDifferentDoor"
```
osse101: This doesn't have gravity on it.  Missing canGravityJump

osse101 (commented):
canDisableEquipment needs to be updated for the blue suit has:Speedbooster check

blkerby:
> canDisableEquipment needs to be updated for the blue suit has:Speedbooster check

Oh yeah the canDisableEquipment coverage is pretty sparse right now; I think most uses are missing, probably hundreds of strats need updating. I haven't really been trying to add them all because I figure that will a separate pass, for SpeedBoosterless blue suit.

PR #2805, opened 2026-03-30
Add Crystal Sparks in Croc's Lair

kjbranch:


osse101 (commented):
Grapple 2 is a little danagerous because if you miss you can spark into a fireball.  But you can also avoid risking that if you set it up without jumping.

PR #2829, opened 2026-04-06
Green Maridia pt. 1 (Pants Room): blue suit logic

blkerby:
Pants Room:
- 4->1 and 4->3 requirements didn't seem to make sense (seemingly confused with 4->5 requirements?).
- Moved the 4->5 Base strat (Gravity + Grapple + SpaceJump) to be 1->5, since you don't need to go into the sand.
- Similarly, moved 4->5 "Suitless Flatley Turnaround Climb" to be 1->5.

East Pants Room:
- New strat 1->3 "Gravity Flatley Jump", to go up with a blue suit. There can be other options that land on the sand first but this didn't seem important.

### DIFF
--- a/region/maridia/inner-green/East Pants Room.json
+++ b/region/maridia/inner-green/East Pants Room.json
@@ -91,7 +91,8 @@
           "openEnd": 1
         }
       },
-      "flashSuitChecked": true
+      "flashSuitChecked": true,
+      "blueSuitChecked": true
     },
     {
       "id": 2,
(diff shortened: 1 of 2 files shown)


osse101 (commented):
The odd looking requirements are working around canPlayInSand, since Gravity alone wasn't enough when it was first added.  Springball is there to bounce on the sand instead of sinking into it.

And 4 must have been acting as a junction covering both doors and the sand.

PR #2833, opened 2026-04-07
Add lenience to Screw Attack Room speedy jump

kjbranch:
This felt pretty difficult to get first-try. Though, maybe it is reliable to get first-try if you have more breathing room and you dont need a full second attempt. What do you think? 

In the video, it looks like it takes about 240 frames (~240 to 180E). At those counts, it's 50% more than the strat says, which would put the video noticeably below VH-level execution. 
https://dev.maprando.com/logic/room/151/2/3/18

### DIFF
--- a/region/lowernorfair/west/Screw Attack Room.json
+++ b/region/lowernorfair/west/Screw Attack Room.json
@@ -1519,7 +1519,17 @@
           "h_backIntoCorner"
         ]},
         {"doorUnlockedAtNode": 2},
-        {"heatFrames": 160}
+        {"or": [
+          {"and": [
+            "canInsaneJump",
+            {"heatFrames": 160}
+          ]},
+          {"and": [
+            "canTrickyJump",
+            {"heatFrames": 200}
+          ]},
+          {"heatFrames": 445}
+        ]}
       ],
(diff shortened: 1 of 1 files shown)


osse101:
I think you can work heatFrames into the h_backIntoCorner block.  Moonwalk is what is costing extra time, but doing it without is real risky for trickyJump.  That's the situation you might jump and not reach, most of the time in hard you would jump late and fall.  Or touch the transition.  So I'd also do less heat leniency, modelling just the fall and second setup.

kjbranch:
I think a failed jump is pretty common, seems fair to only expect that in the extra heat frames.
This is why I did it the way I did, but I'm very open to change:
- you can either back into a corner, or you can not and do a last frame jump
- getting into position is very risky without can insanejump, so it takes it a bit extra slower without
- so the Hard variant is now 400 instead of 445 heat frames. This is how I calculated that (I'm fine putting whatever you think here, since there's the heat lenience which some steps wont need at all): 120 for a missed jump to fall down [2->5] + 45 for jumping back up [5->2] + 40 for backing into the corner a second time (and then rounded down to 200)

PR #2837, opened 2026-04-14
Rest of blue suit carry logic

blkerby:
These are the random strats here and there that were missed on the first pass. It looks like almost all of them were either already checked but just forgot to set blueSuitChecked, or they're the gray doors and other node-level strats that I usually overlooked (but mostly have no interaction with blue suit anyway).

Set "blueSuitChecked" to be required now. Also fixed a thing in `validate.py` that was sometimes making the validation errors blow up in size.

### DIFF
--- a/region/maridia/inner-pink/East Cactus Alley.json
+++ b/region/maridia/inner-pink/East Cactus Alley.json
@@ -1863,12 +1863,22 @@
           "canInsaneJump",
           "canStationaryLateralMidAirMorph"
         ]},
-        "canNeutralDamageBoost",
+        {"or": [
+          "canNeutralDamageBoost",
+          {"and": [
+            "canSpringFling",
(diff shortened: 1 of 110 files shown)


region/maridia/inner-pink/East Cactus Alley.json
```diff
@@ -1867,6 +1867,7 @@
         {"enemyDamage": {"enemy": "Cacatac", "type": "spike", "hits": 2}}
       ],
       "flashSuitChecked": true,
+      "blueSuitChecked": true,
```
osse101, line 1876: Possible with BlueSuit by turning off springball when colliding with the spike

osse101 (commented):
Moat dashless
Lower West Ocean dashless (HBJ, springfling, walljump into PB bounce (Hard))

osse101 (dismissed):

PR #2844, opened 2026-04-15
WRITG - move some things out of basic

kjbranch:
This came up as too hard in the mentor tournament discord; let me know what you think. Also includes a bugfix that loptr noticed.

### DIFF
--- a/region/lowernorfair/east/The Worst Room In The Game.json
+++ b/region/lowernorfair/east/The Worst Room In The Game.json
@@ -1179,7 +1179,6 @@
             "type": "contact",
             "hits": 1
           }},
-          "ScrewAttack",
           {"and": [
             "Ice",
             "Plasma"
@@ -1195,10 +1194,12 @@
       ],
       "flashSuitChecked": false,
       "blueSuitChecked": true,
-      "note": "Being heatproof allows slower kill methods.",
+      "note": [
+        "Being heatproof allows slower kill methods.",
+        "Note that it is possible to freeze the Namihe on entry with Plasma and Ice."
+      ],
       "devNote": [
-        "Bombs are still excluded because they take 30 bombs each and that's ridiculous.",
-        "FIXME: This probably needs some enemy-dodging tech or the ability to freeze the Namihe."
+        "Bombs are still excluded because they take 30 bombs each and that's ridiculous."
       ]
     },
     {
@@ -1256,7 +1257,6 @@
             "type": "contact",
             "hits": 1
           }},
-          "ScrewAttack",
           "Ice"
         ]},
         {"enemyKill": {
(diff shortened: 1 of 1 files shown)


region/lowernorfair/east/The Worst Room In The Game.json
```diff
@@ -1297,6 +1345,18 @@
...
       "requires": [
+        {"or": [
+          "canDodgeWhileShooting",
+          {"enemyDamage": {
+            "enemy": "Yellow Space Pirate (wall)",
+            "type": "contact",
+            "hits": 1
+          }},
+          {"and": [
+            "Ice",
```
osse101: has ice already

region/lowernorfair/east/The Worst Room In The Game.json
```diff
@@ -1172,6 +1172,19 @@
...
+            "enemy": "Yellow Space Pirate (wall)",
+            "type": "contact",
+            "hits": 1
+          }},
+          "ScrewAttack",
+          {"and": [
+            "Ice",
+            "Plasma"
+          ]}
+        ]},
```
osse101, line 1186: There's a fixme on this strat that can be removed

osse101 (commented):
ScrewAttack isn't listed on some of these for no clear reason.

I'd prefer Screw not be listed when there is a strat built for screw attack, and similarly Plasma on the charge strats.  But I think shortening the requirements makes it more likely something is overlooked, and unclear if the requirements were different between strats.  It is already not obvious that iceplasma is for the namihe.

I find screw helps much more.  With iceplasma you might try to jump up expecting to be safe and then not really have anywhere to go if you need to dodge.  And you have to trust the wall pirate doesn't jump into you if you stay by the door.  So I don't know that its really useful.

kjbranch:
When the dedicated screw attack strat was better than another in every way, I removed screw from the option. It's definitely not easy to realize, because some of them were only slightly better, like 10 heat frames. I can look into cleaning it and the plasma up a bit more tomorrow

region/lowernorfair/east/The Worst Room In The Game.json
```diff
@@ -1236,6 +1249,16 @@
...
       "name": "Charge Plasma Pirates Kill",
       "requires": [
+        {"or": [
+          "canDodgeWhileShooting",
+          {"enemyDamage": {
+            "enemy": "Yellow Space Pirate (wall)",
+            "type": "contact",
+            "hits": 1
+          }},
+          "ScrewAttack",
```
kjbranch: while this strat does benefit from screw, it only saves 10 frames (450 v 460), so I'm going to remove it. However, on the Super kill, super + screw saves like 65 frames compared to just screw, so I will keep it there. I also removed screw from the heatproof kill strat, because you can just use the Screw kill strat, and that's equivalent

kjbranch:
All of your suggestions for simplifying were correct and I think they help a lot. The only simplification that I applied that isnt completely is that screw option on the plasma charge kill, but its only by 10 frames, so I think its worth removing there anyways.

kjbranch:
I added noted about the Namihe freeze as well. I think it's not "intuitive" but a lower-level skilled player will likely just shoot on entry anyway, especially if theres an enemy and they have ice. Dodging the single fireball also felt pretty simple. There is still a risk of just yoloing it and jumping straight into the pirate or its laser.

PR #2845, opened 2026-04-16
Misc: Damage boosts and Oum clip notes

kjbranch:


region/norfair/crocomire/Post Crocomire Farming Room.json
```diff
@@ -1604,6 +1607,10 @@
         {"notable": "Damage Boost"},
         "canTrickyJump",
         "canTrickyDodgeEnemies",
+        {"or": [
+          "canTrickyWalljump",
+          "canInsaneJump"
```
osse101: This makes the most sense as a "can't climb back up" option.  It needs `canDash`, which is a little different from noBlueSuit.

osse101 (dismissed):

region/norfair/crocomire/Post Crocomire Farming Room.json
kjbranch: Had Maddo try it and he thought it was quite difficult, and thought that the wall jump variant wasnt really easier. What do you think of this?

osse101: Kinda similarly hard, after figuring out the trick to the walljump.  A delayed walljump with spinbreak when you want to dboost.  And I didn't know you had to back into corner the dash jump; I only thought it helped.
I wouldn't say they are much harder than the gamet boost since retries are faster, but the knowledge check is a little higher.  Maybe it depends on what you're good at since classic runners still sometimes opt for ripper boost.
I like higher requirements on this trick mostly only because it saves a small amount of energy over gamet boost and  it works as a single try strat if you can't get back up to the top right.

osse101: canTrickyJump isn't needed with canInsaneJump there

PR #2849, opened 2026-04-22
Rename Super Jump as Superjump

kjbranch:
Selicre suggested this, and after some thinking on it, I think they're right. This is also how the wiki spells it, though they have the J capitalized half the time. I kept the tech name with a capital J, for readability, but maybe we should make it always capital instead, I'm not sure.

### DIFF
--- a/region/brinstar/green/Etecoon Energy Tank Room.json
+++ b/region/brinstar/green/Etecoon Energy Tank Room.json
@@ -988,7 +988,7 @@
     {
       "id": 101,
       "link": [3, 2],
-      "name": "Come in Shinecharged - Gain Blue Suit (X-Mode Super Jump)",
+      "name": "Come in Shinecharged - Gain Blue Suit (X-Mode Superjump)",
       "entranceCondition": {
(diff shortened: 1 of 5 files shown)


osse101:
We use canWalljump and canCarefulJump meaning that a kind of jump uses lowercase j with fullname Walljump wheras descriptor jump keeps the jumping word as Jump and Superjump would be a kind of jump which is different from InsaneJump.
I would say strat names should match the tech name so I'd rename it to canSuperjump.  
But I prefer canWallJump and maybe still miscase it regularly.  I also see canTurnaroundSpinJump and canFlatleyJump and canGravityJump are big J, so WallJump is the only odd one.

kjbranch:
I have also disliked canWalljump being a lowercase j. I'll fix it in another PR.

osse101 (approved):
My thoughts were pretty freeform there, but now Superjump is the only odd one with a lowercase j.

kjbranch:
I think that's how it should be. Superjump would basically be one word and wall jump would be two.

PR #2850, opened 2026-04-22
Fix some Medium-level Tourian strats

kjbranch:
These felt too hard.
- Metroids 1: This makes it so that on Basic you are better off not escaping the acid with a wall jump and on Medium it takes 2 tries.
- MB: This will now not be possible immediately after Big Boy on Medium or below.

Here's the Medium seed that came up with MB after Big Boy: https://dev.maprando.com/seed/XKkV7MRPv/data/visualizer/index.html


region/tourian/main/Metroid Room 1.json
```diff
@@ -801,10 +801,14 @@
...
           {"and": [
             {"or": [
-              "canWalljump",
+              "canTrickyWalljump",
               "HiJump"
             ]},
             {"acidFrames": 20}
+          ]},
+          {"and": [
+            "canPreciseWalljump",
```
osse101: I don't think this walljump looks that hard.  For walljump vs consecutivewalljump, the basic player could be expected to miss a simple walljump but I don't see that with preciseWalljump.  So this looks like walljump+miss leniency to me.
Not wanting basic to acid dive would be fine, but I don't think canPreciseWalljump needs the leniency then.

And avoiding Metroids with Ice is going to be a little hard to manage still.

kjbranch: I've found this jump to be annoying, maybe because of the change in physics as youre jumping out or the low visibility or the high pressure to not mess up. But I can move it to just a canPreciseWalljump.

PR #2851, opened 2026-04-23
add bluesuit gmorph remote aquire to hijump etank.

nn357:
add bluesuit gmorph remote aquire to hijump etank.
https://videos.maprando.com/video/10233
https://videos.maprando.com/video/10232

### DIFF
--- a/region/norfair/west/Hi Jump Energy Tank Room.json
+++ b/region/norfair/west/Hi Jump Energy Tank Room.json
@@ -679,7 +679,7 @@
       "blueSuitChecked": true,
       "note": [
         "Enter with direct G-mode artificial morph, avoid touching too many crumble blocks and the right item if it is there.",
-        "Either IBJ or SpringBall jump into the morph tunnel and quickly roll thorough to avoid overloading PLMS and making remote aquisition impossible.",
+        "Either IBJ or SpringBall jump into the morph tunnel and quickly roll thorough to avoid overloading PLMS and making remote acquisition impossible.",
         "Touch the item, and return to the right side of the room before exiting G-mode to collect the item."
       ]
     },
--- a/region/norfair/west/Hi Jump Energy Tank Room.json
+++ b/region/norfair/west/Hi Jump Energy Tank Room.json
@@ -627,7 +627,7 @@
       "note": [
         "Enter with direct G-mode artificial morph, avoid touching too many crumble blocks and the right item if it is there.",
         "Either IBJ or SpringBall jump into the morph tunnel and quickly roll thorough to avoid overloading PLMS and making remote acquisition impossible.",
-        "Touch the item, and return to the right side of the room before exiting G-mode to collect the item."
+        "Touch the item, and continue to the left side of the room before exiting G-mode to collect the item."
       ]
     },
     {
--- a/region/norfair/west/Hi Jump Energy Tank Room.json
+++ b/region/norfair/west/Hi Jump Energy Tank Room.json
@@ -296,7 +296,15 @@
       },
       "requires": [
         "h_blueSuitGMode",
-        "h_artificialMorphIBJ",
+        {"or": [
+          "h_artificialMorphIBJ",
+        {"and": [
+          "HiJump",
+          "h_artificialMorphSpringBall"
+        ]
+        }
+      ]
+      },
         "canRemoteAcquire"
       ],
       "collectsItems": [3, 4],
--- a/region/norfair/west/Hi Jump Energy Tank Room.json
+++ b/region/norfair/west/Hi Jump Energy Tank Room.json
@@ -305,7 +305,7 @@
       "note": [
         "Enter with direct G-mode artificial morph, IBJ to the first item, quickly passing thorough it and thruogh the morph tunnel.",
         "IBJ to the right item, moving on it a few time will overload the plms and allow Samus to return to the door.",
-        "Upon exiting G-Mode, both items will be aqquired."
+        "Upon exiting G-Mode, both items will be acquired."
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


kjbranch (commented):
looks good. 
- there could be an identical strat from [2, 1] as well.
- and a 1 to 1 that collects the right item. 

I dont think blue + g-mode were checked thoroughly, I need to go through all the rooms sometime.

nn357:
> looks good.
> 
> * there could be an identical strat from [2, 1] as well.
> * and a 1 to 1 that collects the right item.
> 
> I dont think blue + g-mode were checked thoroughly, I need to go through all the rooms sometime.
add extra strats
https://videos.maprando.com/video/10251

from the left side you can grab both items

region/norfair/west/Hi Jump Energy Tank Room.json
```diff
@@ -285,6 +285,29 @@
...
+          "mode": "direct",
+          "morphed": true
+        }
+      },
+      "requires": [
+        "h_blueSuitGMode",
+        "h_artificialMorphIBJ",
+        "canRemoteAcquire"
+      ],
+      "collectsItems": [3, 4],
```
kjbranch, line 310: Not sure if it would be better to grab both or just one. Usually we just grab one item at a time, since that's all Map Rando can do, and because the left item already has remote acquire options. I can't think of any reason not to collect the left one though, since there are camera scroll blocks to overload PLMs, unless it were an item you didn't want, like an E-tank.

Lets see what Osse and Maddo think

region/norfair/west/Hi Jump Energy Tank Room.json
```diff
@@ -285,6 +285,29 @@
...
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Enter with direct G-mode artificial morph, IBJ to the first item, quickly passing thorough it and thruogh the morph tunnel.",
```
kjbranch: hijump+spring works too

kjbranch (dismissed):

region/norfair/west/Hi Jump Energy Tank Room.json
nn357: yeah I added it now :)

region/norfair/west/Hi Jump Energy Tank Room.json
```diff
@@ -657,6 +714,32 @@
...
+          "h_artificialMorphSpringBall",
+          "h_artificialMorphIBJ"
+        ]},
+        "canRemoteAcquire"
+      ],
+      "collectsItems": [4],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Enter with direct G-mode artificial morph, avoid touching too many crumble blocks and the right item if it is there.",
```
osse101: Talking about the crumble blocks is distracting information.  The right item and scroll blocks should be the only things that matter.

region/norfair/west/Hi Jump Energy Tank Room.json
```diff
@@ -657,6 +714,32 @@
...
+        "Either IBJ or SpringBall jump into the morph tunnel and quickly roll thorough to avoid overloading PLMS and making remote acquisition impossible.",
+        "Touch the item, and return to the right side of the room before exiting G-mode to collect the item."
```
osse101: It's hard to say if more is needed for the blind segment.  I think it's short enough that it is reasonable.  However if your bombjump is failing offcamera you wont' be able to see why.  So I think a note should be added saying to not hold forward into a wall while bomb jumping.  And h_blueSuitGMode should be sufficiently high tech that you can execute the blind movement and bomb timing.

region/norfair/west/Hi Jump Energy Tank Room.json
```diff
@@ -573,6 +604,32 @@
...
```
osse101:
```suggestion
        "Touch the item, and continue to the left side of the room before exiting G-mode to collect the item."
```

region/norfair/west/Hi Jump Energy Tank Room.json
osse101: There needs to be more support for avoiding items until a particular step before we need to think too hard about it.
Collecting items one at a time would mean adding avoidance tech and notes on the [3] strat when you shouldn't care most of the time.  A player would be interested in knowing that you are able to collect both items and pass through the scroll blocks without overloadingPLMs too early, so that should be highlighted.  Overall then, the player is more interested in being informed on the double item collection case instead of the single item case, I think.

region/norfair/west/Hi Jump Energy Tank Room.json
```diff
@@ -285,6 +285,37 @@
...
+      ]
+      },
+        "canRemoteAcquire"
+      ],
+      "collectsItems": [3, 4],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Enter with direct G-mode artificial morph, IBJ to the first item, quickly passing thorough it and thruogh the morph tunnel.",
+        "IBJ to the right item, moving on it a few time will overload the plms and allow Samus to return to the door.",
```
osse101: This is a blue suit strat so you can fall through the bomb blocks without thinking about it.  Or go back the way you entered since we have good movement options too.

region/norfair/west/Hi Jump Energy Tank Room.json
nn357: yeah that makes sense. I updated the description to say you can't hold the direction until after the bomb explodes due to bluesuit effects. Also uploaded a video using the oob viewer to show the simple offscreen movement.

PR #2852, opened 2026-04-24
Capitalize J in canWallJump

kjbranch:
This also capitalizes it in canWallJumpWithCharge. There are probably more tech that could benefit from standardization, but this was just one find and replace all. Others can be done in a separate PR.


osse101:
Its nice having string validation for changes like this

PR #2860, opened 2026-04-28
Maridia Tube Spring Ball

kjbranch:
What do you think about this?


osse101:
I feel its fine to add in at this time.  The concern is that unmorphing to extend your hitbox isn't intuitive for Hard, not that its hard.  Some hitbox knowledge is good to know by Hard.

PR #2863, opened 2026-04-29
add x-mode bluesuit to attic.

nn357:
https://videos.maprando.com/video/9340

forgot this one.

### DIFF
--- a/region/wreckedship/main/Attic.json
+++ b/region/wreckedship/main/Attic.json
@@ -908,20 +908,6 @@
       "blueSuitChecked": true
     },
     {
-      "link": [2, 1],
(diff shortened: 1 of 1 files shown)


region/wreckedship/main/Attic.json
```diff
@@ -907,6 +907,20 @@
       "flashSuitChecked": true,
       "blueSuitChecked": true
     },
+    {
+      "link": [2, 1],
```
osse101: I think this looks better as [2, 2] and moved under the other blue suit strats.  That way you get state change without travelling, and that's how I see things being written nowadays.

PR #2870, opened 2026-05-06
add r-mode standup clip to gauntlet etank

nn357:
https://videos.maprando.com/video/10276
https://videos.maprando.com/video/7564

### DIFF
--- a/region/crateria/west/Gauntlet Energy Tank Room.json
+++ b/region/crateria/west/Gauntlet Energy Tank Room.json
@@ -559,6 +559,24 @@
       ]
     },
     {
+      "link": [1, 3],
+      "name": "R-Mode Forced Standup clip",
+      "requires": [
+        "Morph",
+        "ScrewAttack",
+        {"obstaclesCleared": ["R-Mode"]},
+        {"partialRefill": {"type": "ReserveEnergy", "limit": 99}},
+        {"autoReserveTrigger": {"implicitHeatFrames": "no"}},
+        "canRModeStandupClip",
+        {"acidFrames": 16}
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Buffering a jump during the reserve fill and performing a backwards spinjump onto the item pedestal will minimize lava frames"
+      ]
+    },
+    {
       "id": 13,
       "link": [1, 4],
       "name": "Base",
--- a/region/crateria/west/Gauntlet Energy Tank Room.json
+++ b/region/crateria/west/Gauntlet Energy Tank Room.json
@@ -579,7 +579,7 @@
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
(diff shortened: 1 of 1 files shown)


region/crateria/west/Gauntlet Energy Tank Room.json
```diff
@@ -558,6 +558,24 @@
...
     },
+    {
+      "link": [1, 3],
+      "name": "R-Mode Forced Standup clip",
+      "requires": [
+        "Morph",
+        "ScrewAttack",
+        {"obstaclesCleared": ["R-Mode"]},
+        {"partialRefill": {"type": "ReserveEnergy", "limit": 99}},
+        {"autoReserveTrigger": {"implicitHeatFrames": "no"}},
```
osse101:
```suggestion
        {"autoReserveTrigger": {}},
```

region/crateria/west/Gauntlet Energy Tank Room.json
```diff
@@ -558,6 +558,24 @@
...
+        "canRModeStandupClip",
```
osse101: This is a trivial RMode Standup Clip so I don't want to use that tech.  `canRMode` is sortof right but its a unique usage of base RMode.  It is closest to `{"notable": "Crystal Flash Clip"}` as that doesn't seem to be a tech.  So I think this should be a notable that requires canRMode and can go in Expert.  I'm also looking at `canRMode` and don't see any current uses in Expert.

region/crateria/west/Gauntlet Energy Tank Room.json
```diff
@@ -558,6 +558,24 @@
         "because at higher speed it would be more difficult to bounce across the room."
       ]
     },
+    {
+      "link": [1, 3],
+      "name": "R-Mode Forced Standup clip",
+      "requires": [
+        "Morph",
+        "ScrewAttack",
```
osse101: You can add a shinespark option too.  To minimize health loss to acid you would refill energy after the spark or have an additional reserve tank.  But its not obvious how acid damage will affect your energy so we can use the worst case for logic.

region/crateria/west/Gauntlet Energy Tank Room.json
```diff
@@ -558,6 +558,30 @@
...
+            {"shinespark": {"frames": 52, "excessFrames": 0}}
+          ]}
+        ]},
+        {"autoReserveTrigger": {}},
+        {"acidFrames": 16}
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Buffering a jump during the reserve fill and performing a backwards spinjump onto the item pedestal will minimize lava frames"
```
internet-account2:
```suggestion
        "Buffering a jump during the reserve fill and performing a backwards spinjump onto the item pedestal will minimize acid frames"
```

Shouldn't this say acid frames instead of lava frames in this room?

PR #2872, opened 2026-05-09
botwoon etank crouch spark ledgegrab

internet-account2:
Unless I missed something this seems to be missing from logic.

Help, I have no idea what I'm doing :-)

https://videos.maprando.com/video/10371 is my best attempt at this. Is the video good?

This probably is missing at least some form of trickyJump something? Don't know what's appropriate.

I'm also unsure about what exactly are the correct values for the frames. The numbers I inserted I read from my video, the 2 excessFrames I got from trying the spark again with reduced hp.

Especially unsure about the numbers because you can exchange shinechargeFrames with shinesparkFrames...

### DIFF
--- a/region/maridia/inner-pink/Botwoon Energy Tank Room.json
+++ b/region/maridia/inner-pink/Botwoon Energy Tank Room.json
@@ -2673,22 +2673,23 @@
         "canHorizontalMidairShinespark",
         "canShinechargeMovementComplex",
         "h_underwaterCrouchJumpDownGrab",
+        {"shineChargeFrames": 55},
         {"or": [
+          {"shinespark": {"frames": 48, "excessFrames": 2}},
           {"and": [
-            {"shineChargeFrames": 49},
-            {"shinespark": {"frames": 46, "excessFrames": 2}}
-          ]},
-          {"and": [
-            {"shineChargeFrames": 167},
-            {"shinespark": {"frames": 30, "excessFrames": 2}}
+            "canShinechargeMovementTricky",
+            {"shinespark": {"frames": 43, "excessFrames": 2}}
           ]}
         ]}
       ],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
-        "Do a crouchjump while keep holding down and horizontal spark on the highest point, with a little bit of lenience.",
-        "Samus will bonk the ledge (or stop very close to it); down-grab to get on top of it."
+        "Crouch jump on entry and spark horizontally at the peak of the jump.",
+        "Samus will bonk on the corner of the ledge, but can down grab onto it.",
+        "In order to prevent sparking during the crouch jump, hold shot or aim down until Samus leaves the ground.",
+        "It is possible to move left during the jump to potentially save some health.",
+        "In that case, letting go of left while jump is still held will activate the spark."
       ]
     },
     {
--- a/region/maridia/inner-pink/Botwoon Energy Tank Room.json
+++ b/region/maridia/inner-pink/Botwoon Energy Tank Room.json
@@ -2671,16 +2671,10 @@
       "requires": [
         "canSuitlessMaridia",
         "canHorizontalMidairShinespark",
-        "canShinechargeMovementComplex",
+        "canShinechargeMovementTricky",
         "h_underwaterCrouchJumpDownGrab",
-        {"shineChargeFrames": 55},
-        {"or": [
-          {"shinespark": {"frames": 48, "excessFrames": 2}},
-          {"and": [
-            "canShinechargeMovementTricky",
-            {"shinespark": {"frames": 43, "excessFrames": 2}}
-          ]}
-        ]}
+        {"shineChargeFrames": 45},
+        {"shinespark": {"frames": 43, "excessFrames": 0}}
       ],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
--- a/region/maridia/inner-pink/Botwoon Energy Tank Room.json
+++ b/region/maridia/inner-pink/Botwoon Energy Tank Room.json
@@ -2669,13 +2669,27 @@
         "comeInShinecharged": {}
       },
       "requires": [
-        {"shineChargeFrames": 49},
-        "Morph",
+        "canSuitlessMaridia",
         "canHorizontalMidairShinespark",
-        {"shinespark": {"frames": 46, "excessFrames": 2}}
+        "canShinechargeMovementComplex",
+        "h_underwaterCrouchJumpDownGrab",
+        {"or": [
+          {"and": [
+            {"shineChargeFrames": 49},
+            {"shinespark": {"frames": 46, "excessFrames": 2}}
+          ]},
+          {"and": [
+            {"shineChargeFrames": 167},
+            {"shinespark": {"frames": 30, "excessFrames": 2}}
+          ]}
+        ]}
       ],
       "flashSuitChecked": true,
-      "blueSuitChecked": true
+      "blueSuitChecked": true,
+      "note": [
+        "Do a crouchjump while keep holding down and horizontal spark on the highest point, with a little bit of lenience.",
+        "Samus will bonk the ledge (or stop very close to it); down-grab to get on top of it."
+      ]
     },
     {
       "id": 81,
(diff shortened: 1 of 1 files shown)


region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -2662,6 +2662,36 @@
...
+        "canShinechargeMovementComplex",
+        "h_underwaterCrouchJumpDownGrab",
+        {"shineChargeFrames": 55},
+        {"or": [
+          {"shinespark": {"frames": 48, "excessFrames": 2}},
+          {"and": [
+            "canShinechargeMovementTricky",
+            {"shinespark": {"frames": 43, "excessFrames": 2}}
+          ]}
+        ]}
```
osse101:
```suggestion
        {"shineChargeFrames": 45},
        {"shinespark": {"frames": 43, "excessFrames": 0}}
```
Optimize for shineChargeFrames.  Then the lower difficulty gets extra leniency frames to move around for a better shinespark.  However I don't think moving after the crouch jump is complicated relative to the crouch jump so I removed it as a change unrelated to making everything use canShinechargeMovementTricky.

region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -2662,6 +2662,36 @@
...
+      "link": [4, 6],
+      "name": "Crouchjump Shinespark with Ledgegrab",
+      "entranceCondition": {
+        "comeInShinecharged": {}
+      },
+      "requires": [
+        "canSuitlessMaridia",
+        "canHorizontalMidairShinespark",
+        "canShinechargeMovementComplex",
+        "h_underwaterCrouchJumpDownGrab",
```
osse101:
```suggestion
        "canShinechargeMovementTricky",
        "h_underwaterCrouchJumpDownGrab",
```
- h_underwaterCrouchJumpDownGrab is meant for flash suits since you usually don't have time or a reason to crouch jump with a shinecharge.  But I think it looks more wrong to not use this helper.
- For the same reason, underwater crouchjumps the concept are part of canCarryFlashSuit.  But it doesn't make much sense to add FlashSuit tech here.  Underwater crouch jumps are still an advanced trick so I increased the level of shinecharge movement tech.

osse101 (commented):
You can probably always either gravity jump with the shinecharge or continue running into the room but there could be a tricky situation where this applies.  Like store the spark in prehalfie climb room then use it in botwoon etank.

region/maridia/inner-pink/Botwoon Energy Tank Room.json
internet-account2: Yeah I was unsure how low exactly I should go with the shineChargeFrames. Kyleb said they should be for insane but I don't have a good enough feel how tight that would be. 45 frames is something I got a single time during all of my testing, while 55 seemed "kind of difficult but doable". I'll change it to the suggested 45 frames.

I put 2 excessFrames because using 41 and stopping slightly in front of the ledge before the bonk still made a downgrab onto it possible. Should that not be modeled?

region/maridia/inner-pink/Botwoon Energy Tank Room.json
internet-account2: h_underwaterCrouchJumpDownGrab also felt kind of "hacky" but at the same time appropriate to me as well. Kyleb also thought it fits here.

internet-account2:
I also think it's probably only relevant in niche situations, but it did come up for me in a seed: I had sparkbooster and the connection was botwoon hallway, upper crab hole, terminator. So I shinecharged in terminator, jumped through crab hole and did the strat in botwoon hallway.

region/maridia/inner-pink/Botwoon Energy Tank Room.json
osse101: I think that to fall onto the ledge you need to spend extra shinecharge frames jumping higher.  It didn't look like I could keep excessFrames while going fast.

internet-account2: Makes sense, probably not a reason to include them then, agreed.

internet-account2:
I tried to record a new video for this.

- I struggle to get it with 45 frames even with slowdown ^^. Well, skill issue on my part.
- Even with frame advance when I start with 171 frames on room entry and end with 126, so in other words 45 shinecharge frames, my best result ends with 55 hp, so I think the spark frames should be 44 actually?
- Question for my understanding: if a strat requires X amount of shinecharge frames, that means the prac hack infohud must display X+1 frames on entry minimum, because once it reaches 0 you can't spark anymore, right? I ask because my first naive assumption was that when you enter the room with X shinecharge frames a strat that requires X frames minimum would work. Maybe the confusion on my part is just an artifact of how the prac hack infohud displays things?.. Hope that makes sense.

osse101:
- I find entering without dash state causes an almost 10 frame difference( 41 vs 49) from coming to a stop after entering the room.  This isn't something we've thought about before, but our cases are 1) dash state = can do a faster strat 2) walking entry = use the fast number, but a large number of frames is lost on the other side of the door 3) jump entry = there is some time lost to falling when entering the room, before you can setup the crouch jump.  So a little more extra leniency would be appropriate.
- I still see 43 frames but its at the point where going a frame faster would be 44.  But that's against my fastest attempts and not the bit of lenience in insane timings.
- I use `Frames on entry - Frames remaining when shinesparking = Frame Cost`.  
```MD
As an example, if a strat requires entering the room with at least 5 shinecharge frames remaining, it should have a requirement of `{"shineChargeFrames": 5}`. This is in spite of the fact that the amount of shinecharge frames would decrease only by 4 (from 5 to 1) between the time of entering the room and activating the spark. Conversely, if measuring the amount that the shinecharge timer decreases when executing a strat that ends in a shinespark, a value of 1 should be added to this amount to determine the minimum possible amount of logical `shineChargeFrames` that should be required.
```

- I tried the strat from Halfie Climb Room and you don't fall in water fast enough to reach Botwoon Etank Room.  Crab Hole logically requires 150frames to cross so it would need a strat update to use.

- The naming convention is "Come In Shinecharged, Crouchjump Shinespark with Ledgegrab".  That looks long so I'd drop "with Ledgegrab".  This is a key phrase used in tests matching entrance conditions to strat names.

internet-account2:
Ok, so in this case where the strat ends with a shinespark I need to add 1 frame after doing the `Frames on entry - Frames remaining when shinesparking` subtraction. That fits with my understanding then, nice.

I'm not completely sure yet if I completely understand the situation with the door entry setup.

- So my setup looks like this: I start in a wet room with a blue suit. I position myself 1px away from the door transition by moonwalking as close as possible, then xray turnaround. Then I kneel to activate the shinespark, as fast as possible enter the prac hack menu to remove the blue suit, then hold left to walk through the door, entering the next room with no momentum and 171 shinechargeFrames left
- I think your setup looks different but I'm not sure? Since you mention dash, my bet would have been that it looks like this: start in a dry room with blue suit. Position Samus as I wrote above. Hold dash, then kneel down and hold left at the same time, entering the door at 174 shinechargeFrames left.

If my guess is correct about your setup: this doesn't remove the blue suit, so doing the crouch jump restarts the shinecharge timer. That would explain to me why you get that number of shinechargeFrames, trying this myself leads to very similar results as you are writing.

osse101:
That sounds right, about my setup.  The coming to a stop bit was nonsense looking for the cause.

- The numbers aren't exactly to the frame so the +1 frame thing isn't too important, or is already rolled into the ~3or4 frame lenience.  

- I think it still needs lenciency of about 10f since its expected you don't enter the room cleanly. So 60 something.

region/maridia/inner-pink/Botwoon Energy Tank Room.json
```diff
@@ -2662,6 +2662,30 @@
...
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Crouch jump on entry and spark horizontally at the peak of the jump.",
+        "Samus will bonk on the corner of the ledge, but can down grab onto it.",
+        "In order to prevent sparking during the crouch jump, hold shot or aim down until Samus leaves the ground.",
+        "It is possible to move left during the jump to potentially save some health.",
+        "In that case, letting go of left while jump is still held will activate the spark."
+      ]
```
osse101, line 2690:
```suggestion
      ],
      "devNote": [
        "10 leniency shineChargeFrames are added for the expected poor room entry."
      ]
```

PR #2874, opened 2026-05-14
beta pbs ceiling clip iframes

internet-account2:
Would it be worth to include this note?

If the answer is yes, the second note probably would need to be slightly reworded, because it would sound weird when a previous sentence warns against partially clipping into the thorns.

### DIFF
--- a/region/brinstar/red/Beta Power Bomb Room.json
+++ b/region/brinstar/red/Beta Power Bomb Room.json
@@ -880,7 +880,8 @@
       "blueSuitChecked": true,
       "note": [
         "Use the Solid Blocks next to the Samus Eater to clip up through the Power Bomb Blocks.",
-        "Carefully jump around the thorns, use HiJump to jump over them, or walk through them using i-frames."
+        "Carefully jump around the thorns, use HiJump to jump over them, or walk through them using i-frames.",
+        "To avoid damage from sidehoppers it can help to clip into the solid block on the very edge, giving thorn I-frames for the jump up."
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


osse101:
If we expect Hoppers can be alive it should be added as a strat/requirement.  Its more likely they were killed iirc.

internet-account2:
There are currently 3 relevant strats that lead to that node, and 1 that is not currently in logic:

- [Stored Moonfall Clip](https://dev.maprando.com/logic/room/77/1/2/6) / [Stored Moonfall Clip (more speed)](https://dev.maprando.com/logic/room/77/1/2/7) (I'll count that as essentially 1 single strat)
Logically it expects to kill hoppers, but it's mentioned it can be done with avoiding them instead. In case they are avoided (requires luck, probably?) they're probably in positions that make it unlikely to get through them with the thorn I-frames, requiring luck on the way out instead again.

- [Samus Eater Teleport](https://dev.maprando.com/logic/room/77/1/2/28)
Does not expect to kill hoppers at all. Unclear how the hoppers are positioned afterwards.

- [Super Sink](https://dev.maprando.com/logic/room/77/1/2/31)
Does not expect to kill hoppers at all. Unclear how the hoppers are positioned afterwards.

- a non-logical [morphless strat](https://videos.maprando.com/video/10410) which doesn't kill the hoppers. I collected the hoppers on the left side to be able to I-frame through them. It's still not super consistent though because a hopper could do a mean jump that follows Samus and touches her as the I-frames run out. Still gives a little bit of extra chance though to avoid that hit.

For the escape, 2 strats in logic (minimally different)

- [Ceiling Clip](https://dev.maprando.com/logic/room/77/2/1/22)
- [Minimalist Ceiling Clip](https://dev.maprando.com/logic/room/77/2/1/23)

Neither requires the hoppers dead. So if you come in with samus eater tp/super sink the hoppers will be alive, logically. I think since both strats require morph it might be expected to dodge the hoppers by morphing? The jump up is still a little bit risky though and could benefit from the I-frames, probably.

Maybe there should be an additional ceiling clip strat that uses wave or thorn I-frames then? Or powerbombs, haha.

osse101:
Right, its more of an oversight/bug that we aren't handling hoppers if the PB floor is intact.
- The methods of clipping down can to standardized. There is a mix of springball only, springball+bombs, and ignored(moonfall) for clipping on the right side of the room.
- If the floor is broken, you can kill the hoppers from the solid blocks next to the hidden path to item.  
- Wave can kill through the floor, but I imagine you want to iframe clip up in every other case.  And from there can either fast kill or take damage or avoid, if the hoppers were grouped first.

osse101:
Created issue for this topic.

PR #2875, opened 2026-05-14
refine heatframes for double chamber spike ibj

nn357:
came up in a seed recently and the heatframes were so lenient it was easy to break logic.

Heat frames could maybe be tightened a little more?

single ibj

https://videos.maprando.com/video/10411

double ibj

https://videos.maprando.com/video/10412


osse101:
You wrote door to item.  
Item to door could use the same changes.

PR #2881, opened 2026-05-26
gauntlet e-tank right door g-mode setup with ice.

nn357:
https://videos.maprando.com/video/7567
https://videos.maprando.com/video/7565

### DIFF
--- a/region/crateria/west/Gauntlet Energy Tank Room.json
+++ b/region/crateria/west/Gauntlet Energy Tank Room.json
@@ -1105,10 +1105,19 @@
       "link": [2, 2],
       "name": "G-Mode Setup - Get Hit By Zebbo",
       "requires": [
-        "h_bombThings",
+        {"or": [
+          {"and": [
+            "Morph",
+            {"haveBlueSuit": {}},
+            {"acidFrames": 35}
+          ]},
+          {"and": [
+            "h_bombThings",
+            {"obstaclesCleared": ["A", "B"]}
+          ]}
+        ]},
         "canTrickyUseFrozenEnemies",
-        "canTrickyGMode",
-        {"obstaclesCleared": ["A", "B"]}
+        "canTrickyGMode"
       ],
       "exitCondition": {
         "leaveWithGModeSetup": {}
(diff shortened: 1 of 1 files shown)


region/crateria/west/Gauntlet Energy Tank Room.json
```diff
@@ -1101,6 +1101,25 @@
...
+        "canTrickyUseFrozenEnemies",
+        "canTrickyGMode",
+        {"obstaclesCleared": ["A", "B"]}
+      ],
+      "exitCondition": {
+        "leaveWithGModeSetup": {}
+      },
+      "note": [
+        "After clearing the wall blocks and the columns, lure a Zebbo over to near the door by repeatedly freezing it and moving forwards.",
+        "Powerbombs can be used to clear the final bomb block by making sure the zeb is frozen far enough away from the explosion to not be destroyed."
```
kjbranch:
```suggestion
        "After clearing the wall blocks and the columns, lure a Zebbo over to near the door by repeatedly freezing it and moving forward.",
        "Power Bombs can be used to clear the final bomb block by making sure the Zebbo is frozen far enough away from the explosion to not be killed."
```

kjbranch (commented):
I think some acid frames should be added. Do you think bombs and PBs should be in the same difficulty? You put blue suit checked - blue can break the bomb blocks, so that can be added as an OR, other than that, do you think its fine to do without dashing?

nn357:
> I think some acid frames should be added. Do you think bombs and PBs should be in the same difficulty? You put blue suit checked - blue can break the bomb blocks, so that can be added as an OR, other than that, do you think its fine to do without dashing?

Yeah acid frames sound reasonable... not sure what to put though as you have to gmode at the end and maybe less than a tanks worth unless being super lenient...

i guess it depends on what difficulty the trick is placed in.. 

same with the bombs / pbs... since you have ice anyway and you still have to time the acid level / bug freezing unfreezing / etc theres not that much difference between bombing the final block with pb or regular bombs.

https://videos.maprando.com/video/10468

i did record the bluesuit strat the other day too,  i just uploaded it.

I also realized this is from left to right door but without acid frames its possible but tricky to keep the bug on screen / avoid the acid.

kjbranch:
The reasoning for adding acid frames is because we want to add damage scaling in the future. 

I think we plan to scale enemy damage and not heat/lava/acid, but it's still worth putting something reasonable in anyways, (we may change our mind or another project may scale environmental damage).

osse101 (approved):

Issue #2878, opened 2026-05-24
Refine Warehouse E-Tank Beetom kill strat


blkerby:
Logic currently says that without any tech requirements you can kill the Beetoms with Bombs without taking damage. This seems unfair. Came up in a Medium seed recently: https://discord.com/channels/1053421401354285129/1053434704180805744/1508052094186426519

PR #2884, opened 2026-05-27
Medium updates with low Energy, non-Norfair

kjbranch:
I didn't think that Spiky Death or Mama Turtle really needed any changes. I didn't do much testing for the Beetoms, but this is how the farm strat was written.

Addresses #2878


kjbranch:
This adds the `Top Left Direct Jump` notable in Fish Tank to the rolling jump. The notable can then go down to Hard, and this will be the only use of the notable that is in Hard.

PR #2887, opened 2026-06-02
UN: Medium updates with Varia and low Energy Part 1

kjbranch:
I went through all the rooms below Bubble Mountain and left of Business Center. 

There was a complaint about Spiky Platforms being too hard at 1E+Varia, but it felt fine for Medium. I added the hit in Basic.

### DIFF
--- a/region/norfair/west/Crocomire Escape.json
+++ b/region/norfair/west/Crocomire Escape.json
@@ -1178,14 +1178,11 @@
         {"heatFrames": 360},
         {"or": [
           "canDodgeWhileShooting",
+          "Ice",
           {"enemyKill": {
             "enemies": [["Dragon"]],
             "explicitWeapons": ["Missile", "Super", "Plasma"]
           }},
-          {"and": [
-            "h_heatProof",
-            "Charge"
-          ]},
           {"and": [
             "h_lavaProof",
             {"or": [
--- a/region/norfair/west/Crocomire Escape.json
+++ b/region/norfair/west/Crocomire Escape.json
@@ -1185,6 +1185,7 @@
           }},
           {"and": [
             "h_lavaProof",
+            "h_heatProof",
             {"or": [
               "Charge",
               "ScrewAttack",
--- a/region/norfair/west/Crocomire Escape.json
+++ b/region/norfair/west/Crocomire Escape.json
@@ -411,9 +411,8 @@
           "Charge",
           "Ice",
           "Plasma",
-          {"ammo": {"type": "Missile", "count": 2}},
+          {"ammo": {"type": "Missile", "count": 6}},
           {"ammo": {"type": "Super", "count": 2}},
-          {"ammo": {"type": "PowerBomb", "count": 2}},
           {"enemyDamage": {"enemy": "Dragon", "type": "fireball", "hits": 5}},
           {"and": [
             "h_lavaProof",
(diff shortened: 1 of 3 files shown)


region/norfair/west/Crocomire Escape.json
```diff
@@ -1178,6 +1177,26 @@
...
+          {"and": [
+            "h_heatProof",
+            "Charge"
+          ]},
+          {"and": [
+            "h_lavaProof",
+            {"or": [
+              "ScrewAttack",
+              "h_usePowerBomb"
+            ]}
```
osse101, line 1196:
```suggestion
            "h_lavaProof",
            "ScrewAttack",
```

region/norfair/west/Crocomire Escape.json
```diff
@@ -1178,6 +1177,26 @@
...
         {"heatFrames": 360},
         {"or": [
+          "canDodgeWhileShooting",
+          {"enemyKill": {
+            "enemies": [["Dragon"]],
+            "explicitWeapons": ["Missile", "Super", "Plasma"]
+          }},
+          {"and": [
+            "h_heatProof",
+            "Charge"
```
osse101: This seems like `canDodgeWhileShooting` combatwise.  Is varia alone helpful enough?  Ammo will usually be an available option. And Ice can be added here.
You can keep the dragon from spawning with camera management but that's not worth adding.

region/norfair/west/Crocomire Speedway.json
```diff
@@ -2610,6 +2610,15 @@
...
           {"heatFrames": 20}
+        ]},
+        {"or": [
+          "Wave",
+          "canDodgeWhileShooting",
+          {"and": [
+            "h_usePowerBomb",
+            {"heatFrames": 120}
+          ]},
+          {"enemyDamage": {"enemy": "Cacatac", "type": "spike", "hits": 1}}
```
osse101: I don't see how a second hit happens.  And these items would prevent the first hit.

region/norfair/west/Crocomire Escape.json
kjbranch: A little surprised by this one. I was actually just wondering if lava proof+charge should be added. The platforms do respawn.

region/norfair/west/Crocomire Speedway.json
kjbranch: I think I must have zoned in and forgot there was a hit up above.

region/norfair/west/Crocomire Escape.json
osse101: I forgot they respawn.
But here's a fun mystery: Dragons, except for the first, take 2 power bombs to kill unless you are exactly 3-4 tiles away from them.

lavaproof+charge seems fine

osse101: Is it worth adding heat frames for PB and charge kill?

region/norfair/west/Crocomire Escape.json
```diff
@@ -1178,6 +1177,27 @@
...
         {"heatFrames": 360},
         {"or": [
+          "canDodgeWhileShooting",
+          "Ice",
+          {"enemyKill": {
+            "enemies": [["Dragon"]],
+            "explicitWeapons": ["Missile", "Super", "Plasma"]
+          }},
+          {"and": [
+            "h_lavaProof",
```
osse101, line 1187:
```suggestion
            "h_lavaProof",
            "h_heatProof",
```

osse101:
Screw Attack in Medium is expected to accidently break the platform after killing the dragon so heatproof was added there too.  And being slower than going directly for the speedy jump.

PR #2893, opened 2026-06-10
Move some Medium strats up in difficulty

kjbranch:
cup holder compiled some of the hardest strats they have found in Medium. A handful of them I think should be moved up. 
https://docs.google.com/document/d/1WTH7VZs5c19-Yop2O4f1f5dDE5o1eeAU9ys15A6Qv7U/edit?tab=t.nhu0ti98ciqa

### DIFF
--- a/region/brinstar/red/Red Brinstar Fireflea Room.json
+++ b/region/brinstar/red/Red Brinstar Fireflea Room.json
@@ -1979,6 +1979,10 @@
         "HiJump",
         "canSpeedyJump",
         {"or": [
+          "h_backIntoCorner",
+          "canTrickyJump"
+        ]},
+        {"or": [
           {"and": [
             "canWallJump",
             {"or": [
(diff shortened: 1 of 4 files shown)


region/wreckedship/main/Wrecked Ship Energy Tank Room.json
```diff
@@ -790,7 +794,11 @@
           {"and": [
             "canDash",
             "canCarefulJump",
-            {"disableEquipment": "HiJump"}
+            {"disableEquipment": "HiJump"},
+            {"or": [
+              "canMoonwalk",
```
osse101, line 799: This was always a somewhat unreasonable jump but the alternative gives an easy sequence break.  Not sure which is better, but putting it in hard is less unfair.

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -1978,6 +1978,10 @@
       "requires": [
         "HiJump",
         "canSpeedyJump",
+        {"or": [
+          "h_backIntoCorner",
```
osse101: This change only affects {canWallJump+Spikehits:1}.  I would instead add `canUseIFrames` to that branch.

PR #2900, opened 2026-06-23
A few incomplete videos (Low 6000s)

kjbranch:
WRITG:
https://videos.maprando.com/video/6308

Aqueduct:
https://videos.maprando.com/video/6236
https://videos.maprando.com/video/9387
https://videos.maprando.com/video/10584

### DIFF
--- a/region/lowernorfair/east/The Worst Room In The Game.json
+++ b/region/lowernorfair/east/The Worst Room In The Game.json
@@ -1821,11 +1821,29 @@
           "ScrewAttack",
           {"haveBlueSuit": {}}
         ]},
-        "h_heatedGModeOpenDifferentDoor"
+        {"or": [
+          "h_heatedGModeOpenDifferentDoor",
+          {"and": [
+            "h_heatedGModePauseAbuse",
+            "canBeLucky"
+          ]},
+          {"and": [
+            "h_heatedGModePauseAbuse",
+            "Ice",
+            "Plasma"
+          ]}
+        ]}
       ],
       "clearsObstacles": ["A"],
       "flashSuitChecked": true,
-      "blueSuitChecked": true
+      "blueSuitChecked": true,
+      "note": [
+        "To save Energy, it is possible to lure a pirate to the door and pause abuse to grab its Energy drop on G-mode exit.",
+        "This requires traversing the room quickly enough to prevent the enemies from overloading shots, or by using Ice and Plasma to freeze the Namihes on the descent.",
+        "Kill the lowest pirate on the left wall, shoot the door then return and collect the drop - which requires a large Energy drop.",
+        "Alternatively, get it to jump back to the right wall and let it walk lower, have it jump back to the left and kill it;",
+        "it is possible to escape with a small Energy drop but is very precise and either requires Ice and Plasma or requires some luck of having the pirate climb low enough."
+      ]
     },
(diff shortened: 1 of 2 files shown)


region/lowernorfair/east/The Worst Room In The Game.json
```diff
@@ -1821,11 +1821,29 @@
...
+        ]}
       ],
       "clearsObstacles": ["A"],
       "flashSuitChecked": true,
-      "blueSuitChecked": true
+      "blueSuitChecked": true,
+      "note": [
+        "To save Energy, it is possible to lure a pirate to the door and pause abuse to grab its Energy drop on G-mode exit.",
+        "This requires traversing the room quickly enough to prevent the enemies from overloading shots, or by using Ice and Plasma to freeze the Namihes on the descent.",
+        "Kill the lowest pirate on the left wall, shoot the door then return and collect the drop - which requires a large Energy drop.",
```
osse101: Something can be removed in this note to be more concise.  I think I'd say to either kill the lowest pirate on the left wall or let it climb down to the door while keeping the namihe frozen.  And the other two sentences would go in a detail note if still needed.

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -1393,6 +1354,48 @@
...
+        ]},
+        "h_shinechargeMaxRunway",
+        {"autoReserveTrigger": {}},
+        "canRModeSparkInterrupt"
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Gain Energy with a Crystal Flash or farm Yards with Bombs or Speed Booster.",
+        "Free the Yard from the bottom pit, then run across the bottom runway.",
```
osse101: Is bottom runway not the sand?

region/maridia/inner-pink/Aqueduct.json
```diff
@@ -2672,7 +2676,8 @@
           "canXRayCeilingClip",
           {"and": [
             "h_jumpIntoCrystalFlashClip",
-            "Grapple"
+            "Grapple",
+            {"obstaclesNotCleared": ["R-Mode"]}
```
osse101, line 2679: `canXRayCeilingClip` needs `"obstaclesNotCleared": ["R-Mode"]` more.  The xray snail clip to 1 doesn't accomplish anything.  This strat has an entrance condition saying we aren't in RMode.  The regular snail clip in rmode should be trivial to work around.  So I don't think there are any real conflicts.

Does CF break XRay or is it just an extra resource cost?

region/maridia/inner-pink/Aqueduct.json
kjbranch: I was trying to not change the text too much, but I dont think that's important actually - its best to freely allow overriding anything that needs edits.

region/maridia/inner-pink/Aqueduct.json
kjbranch: X-ray is fine as long as you dont CF.
CF is "fine" if you then leave the room or reserve trigger, but you cant use x-ray or PBs in the meantime.

I think we have just been restricting the CF strats, and then requiring that any R-mode use that needs a CF does so explicitly and then clears the light orb after.

PR #2905, opened 2026-06-25
Add more incomplete videos (6750-7000)

kjbranch:
Everest shinecharge:
https://videos.maprando.com/video/6768
https://videos.maprando.com/video/6769

The rest:
https://videos.maprando.com/video/6770
https://videos.maprando.com/video/6792
https://videos.maprando.com/video/6818

### DIFF
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -1411,7 +1411,7 @@
         "canMidAirMorph",
         "h_EverestMorphTunnelExpanded",
         "h_shinechargeMaxRunway",
-        "canShinechargeMovementComplex",
+        "canShinechargeMovementTricky",
         {"or": [
           {"and": [
             "canWallJump",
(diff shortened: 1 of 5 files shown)


region/maridia/outer/Mt. Everest.json
```diff
@@ -1403,6 +1403,38 @@
...
     },
+    {
+      "link": [2, 6],
+      "name": "Leave Shinecharged",
+      "requires": [
+        "Gravity",
+        "canMidAirMorph",
+        "h_EverestMorphTunnelExpanded",
+        "h_shinechargeMaxRunway",
+        "canShinechargeMovementComplex",
```
osse101: It is `canShinechargeMovementTricky` to me, with retries.

osse101 (dismissed):

region/maridia/outer/Mt. Everest.json
kjbranch: Moved this and the mountain shinecharge out the same door to `canShinechargeMovementTricky`. I just grabbed the frames from Sam's videos. Could you check if they were properly tight?

osse101: I had 140f and 150f when I rolled into the transition, so about right after unmorph.  My prac hack doesn't have expanded tunnel.  Getting the correct walljumps for tunnel entry and gravity jump are pretty hard though, and that's why tricky.

PR #2911, opened 2026-06-30
add r-mode knockback spark to green brin fireleas

nn357:
https://videos.maprando.com/video/10746 (node 2)
https://videos.maprando.com/video/10747 (node 1)

### DIFF
--- a/region/brinstar/green/Green Brinstar Fireflea Room.json
+++ b/region/brinstar/green/Green Brinstar Fireflea Room.json
@@ -249,6 +249,26 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


osse101 (commented):
They should have the same requirements

nn357:
> They should have the same requirements

the shinecharge?

osse101:
Right, the length of the shinecharge runway.  This needs to be 2 strats because there is an entrancecondition setting RMode.  RMode doesn't leave the door open so both strats would have to open it.  And there's no assumption that entering a room means that the door you entered through is blue.

PR #2913, opened 2026-07-08
Incomplete videos 7100-7250

kjbranch:
Also added a couple more requirements to a strat in Fish Tank. It already required canSuitlessMaridia or Gravity to get to this node, but I added this anyway. I dont really think its that hard. Theres a chance players will die their first time though. I'm fine moving it to Hard, damageless, if you think thats needed.

https://videos.maprando.com/video/7118 (was already a strat, just had an error)
https://videos.maprando.com/video/7238
https://videos.maprando.com/video/7246

### DIFF
--- a/region/norfair/crocomire/Post Crocomire Farming Room.json
+++ b/region/norfair/crocomire/Post Crocomire Farming Room.json
@@ -404,6 +404,31 @@
       "blueSuitChecked": true
     },
     {
+      "link": [1, 4],
+      "name": "Come In Shinecharged, Leave With Spark (Hero Shot)",
+      "entranceCondition": {
+        "comeInShinecharged": {}
+      },
+      "requires": [
+        {"shineChargeFrames": 125},
+        "canHeroShot",
+        "canShinechargeMovementTricky",
+        {"shinespark": {"frames": 24, "excessFrames": 0}}
+      ],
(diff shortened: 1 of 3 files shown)


region/norfair/crocomire/Post Crocomire Farming Room.json
```diff
@@ -1745,6 +1770,28 @@
         "canBeVeryPatient is used for difficulty placement, assuming this will require many retries."
       ]
     },
+    {
+      "link": [4, 1],
+      "name": "High Speed Wall Jump",
```
osse101: This is more effort than
```JSON
"id": 56,
"link": [3, 1],
"name": "Tricky Dash Jump",
```
so I'd lower the placement of that to very hard.  
Regarding `h_runOverRespawningEnemies`, you can trigger the gamets extra high and run under them.

kjbranch: I didnt lower the difficulty of the other one, but I can if you'd like that to be done in this PR instead of some time in the future when we make a new tech for easy tricky dash jumps

kjbranch:
It seems like that new high speed wall jump strat was not useful, so I added a slightly higher speed variant that doesnt wall jump.

PR #2916, opened 2026-07-09
Difficult enemy extend runways, startingStutterlessTiles

kjbranch:
https://videos.maprando.com/video/7351
https://videos.maprando.com/video/7369
https://videos.maprando.com/video/7371
https://videos.maprando.com/video/7372
https://videos.maprando.com/video/7375 (and https://videos.maprando.com/video/7376)
https://videos.maprando.com/video/7390

### DIFF
--- a/region/crateria/east/West Ocean.json
+++ b/region/crateria/east/West Ocean.json
@@ -2153,6 +2153,44 @@
       ]
     },
     {
+      "link": [2, 2],
+      "name": "Leave With Runway - Frozen Zeb",
+      "requires": [
+        "h_trickyFrozenEnemyRunway",
+        {"or": [
+          "h_bombThings",
+          "h_useSpringBall",
+          {"and": [
+            "canBeVeryLucky",
+            "Morph",
+            "canXRayStandUp",
+            "canPartialFloorClip",
+            "canDownGrab"
+          ]}
+        ]}
+      ],
+      "exitCondition": {
+        "leaveWithRunway": {
+          "length": 35,
+          "openEnd": 0,
+          "steepUpTiles": 4
+        }
+      },
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Freeze a Zeb while it is facing right but hasn't started moving horizontally.",
+        "Quickly traverse the morph tunnel and get above the Zeb before it thaws - it despawns if it is thawed and is horizontally off camera; it can be off camera vertically and not despawn.",
+        "Crouch on the center of the first sloped tile, just left of where the Zeb will eventually come out of the ground to position it vertically.",
+        "Follow it to the right without letting it go off camera horizontally then freeze it with a one-tile gap between it and the runway to extend the runway as far as possible."
+      ],
+      "detailNote": [
+        "it is possible to get through the morph tunnel with an X-ray standup, but it requires being very quick.",
+        "To do so, shoot the shot block, X-ray standup, partial floor clip, jump and down grab the next stair then wiggle to escape and quickly run over to where the Zeb will be.",
+        "The rest of the strat becomes one-try, as there is no way back to the Zeb with this method."
+      ]
+    },
+    {
       "id": 152,
       "link": [6, 6],
       "name": "R-Mode Entry",
--- a/region/crateria/east/West Ocean.json
+++ b/region/crateria/east/West Ocean.json
@@ -2153,7 +2153,7 @@
       ]
     },
     {
-      "link": [2, 2],
+      "link": [6, 3],
       "name": "Leave With Runway - Frozen Zeb",
       "requires": [
         "h_trickyFrozenEnemyRunway",
--- a/region/crateria/east/West Ocean.json
+++ b/region/crateria/east/West Ocean.json
@@ -2176,6 +2176,7 @@
           "steepUpTiles": 4
         }
       },
+      "unlocksDoors": [{"types": ["ammo"], "requires": []}],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
--- a/region/maridia/inner-pink/Halfie Climb Room.json
+++ b/region/maridia/inner-pink/Halfie Climb Room.json
@@ -3946,6 +3946,62 @@
       "blueSuitChecked": true
     },
     {
+      "link": [3, 3],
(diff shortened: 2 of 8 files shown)


region/crateria/east/West Ocean.json
```diff
@@ -2152,6 +2152,45 @@
         "Note that if the Zeb is unfrozen and off screen horizontally it will despawn, but if it's off screen vertically it will not."
       ]
     },
+    {
+      "link": [6, 3],
+      "name": "Leave With Runway - Frozen Zeb",
+      "requires": [
+        "h_trickyFrozenEnemyRunway",
```
osse101, line 2159: So its a little easy for extreme+, you just have to learn the height positioning cue, then do it kinda quick and following the bug isn't trivial.  
But the bigger problem is that you probably need the extra tile of runway for a hard trick.
In that way putting a high tech on the Xraystandup makes sense, since you can't get back.

region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -3945,6 +3945,62 @@
...
       "blueSuitChecked": true
     },
+    {
+      "link": [3, 3],
+      "name": "Leave With Runway - Oum Runway",
+      "requires": [
+        "Gravity",
+        "canTrickyEnemyExtendRunway",
+        {"or": [
+          "canInsaneJump",
```
osse101, line 3955: This has the same requirements as the longer runway strat

region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -3945,6 +3945,62 @@
...
+          "length": 3,
+          "openEnd": 1,
+          "startingStutterlessTiles": 2
+        }
+      },
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "Use an Oum near the door to extend the runway.",
+        "One reliable method for using the Oum is to stand near the door, press against the Oum when it is close to get it to stop,",
```
osse101, line 3970: Worth noting it should be rolling to the right when you run

region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -3945,6 +3945,62 @@
...
+        "Wait a bit for it to start its attack animation, then jump up and onto it. Depending on the timinig, Samus may take a hit."
+      ],
+      "detailNote": [
+        "It is not possible to stutter on the Oum."
```
osse101, line 3974: It really looks to me like you can stutter.  Maybe not if you get flung forward...

region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -3945,6 +3945,62 @@
...
+      "link": [3, 3],
+      "name": "Leave With Runway - Longer Oum Runway",
+      "requires": [
+        "Gravity",
+        "canTrickyEnemyExtendRunway",
+        "canInsaneJump"
+      ],
+      "exitCondition": {
+        "leaveWithRunway": {
+          "length": 4,
```
osse101, line 3987: I don't really like this longer runway strat since positioning and running on the Oums is so weird already.  But also this speedup makes shortcharging harder.

region/crateria/east/West Ocean.json
kjbranch: I don't really see this strat ever being useful. The Croc Escape gate glitch can be done without this - I think when Sam was originally determining runway length, the slope tiles werent accounted properly. I successfully got the strat without the bug extending the runway.

region/maridia/inner-pink/Halfie Climb Room.json
kjbranch: I don't really know how to test this

region/maridia/inner-pink/Halfie Climb Room.json
kjbranch: I was anticipating this not being usable for shortcharging. Because there are the `startingStutterlessTiles`, this wont be usable for shinecharging - it will count as a 4 tile runway for strats that dont require a shinecharge, and only 1 tile for strats that do require a shinecharge (which is less than what is subtracted from using a door, so it wont actually extend anything)

region/maridia/inner-pink/Halfie Climb Room.json
kjbranch: They kind of both feel like the way I wrote them. I'm definitely open to shifting one of them, deleting the longer one, or making the longer one notable, if you have a preference?

kjbranch:
Sorry this took so long.

region/maridia/inner-pink/Halfie Climb Room.json
osse101: After testing, you can stutter or running stutter before the fling but it doesn't make the shortcharge any easier.

region/maridia/inner-pink/Halfie Climb Room.json
osse101: https://videos.maprando.com/video/10950
This doorframe setup is pretty good but exactly where Oums stop appear to be dependent on their rotation cycle or something.  When I tap em and run to the door like in the clip, I active him before he reaches the door.  But when clipping through from the left I have to jump over the Oum after he bounces on the door, so he faces right.  So its still weird. 
After some practice, just running is fine as `canEnemyExtendRunway`, with the branch for avoiding damage since the run timing is kinda tight and unusual.

The longer runway is quick positioning into frame perfect movement for best results, so I'd call it insane and annoying.  And I am seeing $4.A with a bit of running against the roll, and $4.0 with flat run which is more like 5 tiles, going by the speed table.  https://videos.maprando.com/video/10951

region/maridia/inner-pink/Halfie Climb Room.json
```diff
@@ -3945,6 +3945,62 @@
       "flashSuitChecked": true,
       "blueSuitChecked": true
     },
+    {
+      "link": [3, 3],
+      "name": "Leave With Runway - Oum Runway",
+      "requires": [
+        "Gravity",
+        "canTrickyEnemyExtendRunway",
```
osse101:
```suggestion
        "canEnemyExtendRunway",
```
Takes some practice to figure out, but that's fine in Expert+

PR #2917, opened 2026-07-09
Falling Ripper avoid, Bowling CF with Space

kjbranch:
https://videos.maprando.com/video/7481
The Ripper hit killed Eddie after Big Boy.

### DIFF
--- a/region/brinstar/red/Red Tower.json
+++ b/region/brinstar/red/Red Tower.json
@@ -3180,9 +3180,15 @@
       "id": 91,
       "link": [5, 9],
       "name": "Base",
-      "requires": [],
+      "requires": [
+        {"or": [
+          "canCarefulJump",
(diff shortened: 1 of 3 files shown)


osse101 (commented):
I think the knowledge check on avoiding the rippers is Very Hard.  A number of racers have found out going down right side maridia elevator avoids damage, and speedrunners know how to moonfall down upper red tower but no one knows how to avoid those rippers unexpectedly, and its rare that it matters.

kjbranch:
What do you think about this - moving the top of Red Tower and Maridia Elevator to Hard, but keeping the bottom of Red Tower in Medium? The off-camera ones that you have to dodge never move off-camera, and I dont think you'll ever need to climb then fall right back down, so the Rippers are always going to be predictable.

PR #2924, opened 2026-08-02
Ice Beam Snake Room: Fune Press Spring Ball jump

blkerby:
This came up in a seed that Nito did tonight.

It's technically not that precise but I figure "canInsaneJump" makes sense because it's fairly weird, and you'll be under heat pressure (otherwise could just go up the left side).


blkerby:
Video by Sam: https://videos.maprando.com/video/10803

PR #2925, opened 2026-08-03
Refine X-Ray hall damage boosts

blkerby:
The "Damage Boost Speed Jump" strat had too many variants baked into it which made the note (and logic) a bit confusing. So I split off a separate strat "Patient Waver I-Frames Speed Jump" as a higher-difficulty version that minimizes damage.

- "Damage Boost Speed Jump" gets a "canTrickyDodgeEnemies" requirement, bumping it up to Expert.
- "Patient Waver I-Frames Speed Jump" avoids the initial Fireflea hit by waiting for the second Waver cycle and skips a Waver hit by doing a more precise jump. It requires waiting for the Wavers to go all the way back and forth across the room in order to get a good cycle.
- The "Waver I-Frames" from the safe block to the safe ledge (3 -> 4) reduces the amount of expected Waver hits by one in the higher-difficulty requirements.

If we bump the notable "Full Room Damage Boosts (Right to Left)" to Expert+, a lot of variants still remain in logic tankless in Expert or lower; you would just need some movement item: Ice, Bombs, Speed Booster, Hi Jump or Spring Ball. In this case, the reduction in "Waver I-Frames" hits matters for keeping the IBJ and HiJump versions in Expert logic tankless, because with 4 assumed Waver hits you would not be able to afford taking a spike hit; so Expert will now expect 3 total Waver hits to make it to the safe ledge from the right side of the room, while Extreme would expect 2 and Very Hard would expect 6.

### DIFF
--- a/region/brinstar/red/Red Brinstar Fireflea Room.json
+++ b/region/brinstar/red/Red Brinstar Fireflea Room.json
@@ -709,48 +709,36 @@
       "link": [2, 1],
       "name": "Damage Boost Speed Jump",
       "requires": [
+        "canTrickyDodgeEnemies",
         "canHorizontalDamageBoost",
         "canTrickyJump",
         "canUseIFrames",
         "h_speedJump",
-        {"or": [
-          "canInsaneJump",
-          {"and": [
-            "canTrickyDodgeEnemies",
-            {"enemyDamage": {"enemy": "Fireflea", "type": "contact", "hits": 1}}
-          ]},
-          {"enemyDamage": {"enemy": "Fireflea", "type": "contact", "hits": 2}}
-        ]},
-        {"or": [
-          {"and": [
-            "canInsaneJump",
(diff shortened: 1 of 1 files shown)


blkerby:
I guess one other thing to think about is that if you're going for the full room damage boosts and fail the last boost, then you will have taken more Waver hits than the logic would expect for getting to node 4, in which case the HiJump and IBJ strats won't necessarily work as a "backup": you wouldn't have enough energy to afford a spike hit at that point  unless you have a tank or suits. Not sure if it's something that needs to be accounted for, because the player does still have the option to just go for node 4

osse101:
https://videos.maprando.com/video/10879
leaveWithMockball is fairly easy if can already do the walljumpless speedyjump.  But I couldn't think of many uses.
The speedball is as hard as the shinespark, but technically a little easier.

PR #2926, opened 2026-08-04
Fix GT mapTileMask

kjbranch:
The bottom left tile of the room was only reachable in g-mode


dragonsinth (commented):
gt room

Does the image not need a change also, so that 2 takes up the whole bottom?

kjbranch:
It's best to keep the nodes a bit small in heated rooms, since they're not freely traversable. I think the best solution would be to create a new node, but we havent ever done that for the mapTileMask before. 

I don't think it's necessarily worth adding those kinds of strats unless we or someone using the data is wanting like map completion objectives - then we would need to add options like collecting the high map tiles in Landing Site with Space or IBJ, etc.

dragonsinth:
LGTM

PR #2936, opened 2026-08-09
Ice Beam Gate Room: Sova Ice Clip

blkerby:
The rationale for the canInsaneJump:

- The aim-down is a 2-frame window, which is fairly precise, considering how unusual this setup is and the fact that it is fairly slow to retry.
- It's only useful with a blue suit (or dashless), and not the sort of thing we'd want coming into logic on the same difficulty level as blue suit itself.

### DIFF
--- a/region/norfair/west/Ice Beam Gate Room.json
+++ b/region/norfair/west/Ice Beam Gate Room.json
@@ -1383,7 +1383,7 @@
     },
     {
       "link": [4, 2],
-      "name": "Sova Ice Clip",
+      "name": "Sova Ice Clip with X-Ray",
       "entranceCondition": {
         "comeInNormally": {}
       },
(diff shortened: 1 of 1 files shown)


region/norfair/west/Ice Beam Gate Room.json
```diff
@@ -1381,6 +1381,89 @@
...
+      "note": [
+        "Clear the shot blocks to pass through the lower morph tunnel.",
+        "Wait for the Sova to move far enough to the right,",
+        "then lay a Power Bomb to break the blocks.",
+        "After the Sova returns and passes through the closed shutter,",
+        "damage it, then fire a Super to knock it off the ceiling.",
+        "Freeze it mid-air and use X-Ray to stand up and clip into the tunnel above."
+      ],
+      "detailNote": [
+        "The freezing shot can be normalized by standing and firing the Super diagonally at the ceiling,",
```
osse101:
```suggestion
        "The freezing shot can be normalized by standing and firing the Super diagonally left at the ceiling,",
```

osse101 (dismissed):

PR #2937, opened 2026-08-11
Kraid Eye Door: frozen Zeb leave with mockball

blkerby:
The idea of this was suggested by ClaireDiviner on Joonie's stream today. Gave it a try and it worked:

https://videos.maprando.com/video/10905

### DIFF
--- a/region/brinstar/kraid/Kraid Eye Door Room.json
+++ b/region/brinstar/kraid/Kraid Eye Door Room.json
@@ -1323,6 +1323,41 @@
       ]
     },
     {
+      "link": [3, 3],
+      "name": "Leave With Mockball - Frozen Zeb",
+      "requires": [
+        "h_frozenEnemyRunway",
+        {"thornHits": 1}
+      ],
+      "exitCondition": {
+        "leaveWithMockball": {
+          "remoteRunway": {
+            "length": 5,
+            "openEnd": 1            
+          },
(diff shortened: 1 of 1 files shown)


osse101:
https://videos.maprando.com/video/10910
Your detail note about x position didn't make sense to me and trying to copy the video didn't work most of the time.  This setup worked most of the time for me but usually gave 2 spike hits.  You can also freeze the bug in its pipe, just below the correct height, and kago into it to take less damage than a spike hit.

blkerby:
> https://videos.maprando.com/video/10910 Your detail note about x position didn't make sense to me and trying to copy the video didn't work most of the time. This setup worked most of the time for me but usually gave 2 spike hits. You can also freeze the bug in its pipe, just below the correct height, and kago into it to take less damage than a spike hit.

I saw confusing things in my testing too, so thanks for forcing me to look more into it. The missing piece was that it matters if you previously killed the Zeb vs. despawning it off-camera; despawning it off-camera is required in order for the normalized setup to work. The X position is also a thing that can mess it up if it's not normalized with the thorn knockback; expanded the note a bit now to explain why: the trigger for the Zeb to rise happens before you roll off, so depending on your X alignment you could spend one more frame on the pillar after triggering the Zeb before falling off.

Video with the complete normalized setup: https://videos.maprando.com/video/10911

PR #2938, opened 2026-08-11
Big Pink: Spazer wall ice clip

blkerby:
I believe SpiffyThomas found this a couple years ago. Videos by Sam:

https://videos.maprando.com/video/2639
https://videos.maprando.com/video/2640

I was thinking maybe Extreme+ for the difficulty? I couldn't think of a tech that would cover how it differs in difficulty from the Wave version (which is Extreme), so just went with making another notable.

### DIFF
--- a/region/brinstar/pink/Big Pink.json
+++ b/region/brinstar/pink/Big Pink.json
@@ -3935,10 +3935,7 @@
         "canWallIceClip",
         {"or": [
           "Wave",
-          {"and": [
-            "Spazer",
-            {"notable": "Wall Ice Clip the Mission Impossible Crumble Blocks with Spazer"}
-          ]}
+          "Spazer"
         ]},
         {"or": [
           "h_useSpringBall",
@@ -3958,9 +3955,9 @@
         "Finally, jump with Bombs or Spring Ball to break the crumble block."
       ],
       "detailNote": [
-        "With Spazer, the trick is more precise: the shot must be fired low enough to avoid hitting the wall below the crumbles,",
-        "but high enough to hit the Zeb.",
-        "In this case, the Zeb should be positioned relatively low."
+        "With Spazer, it is easiest to freeze the Zeb with an upward shot (which also works well with Wave):",
+        "while the Zeb is frozen, the optimal positioning can be found by inching back with moonwalk",
+        "and firing test shots upward until the right-most particle of the Spazer shot touches the wall but without hitting it."
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


osse101:
I think they are the same difficulty and both should be moved to Extreme+.  The restrictions on the spazer shot aren't that impactful compared to the precision of the bugs freeze position.

blkerby:
The Spazer version feels much harder to me. I lose at least 80% of attempts to a Spazer-specific problem (bug raised too high, shot too low that it misses the bug, or shot too high that it bonks). I'm fine with Wave version moving up to Extreme+ though. Maybe the Spazer version should be Insane?

osse101:
https://videos.maprando.com/video/10953
You can look at this positioning setup that puts the bug at a good height, and is fast.  It puts the bug low in the tile so spazer has a wide window too.  Its what I use for both beam types now, after this experimentation session, until someone finds a way to time the horizontal freeze.  So it feels even more like they are the same strat to me now.

osse101:
You can also shoot straight up to freeze the bug in the crumble block with spazer, like with wave.  I can see that helping by removing the jump movement.

blkerby:
The crouch jump method is fast and it helps a lot for the Spazer setup, but it leaves the bug in a spot where you can't see it once it goes inside the big block. The timing of the jump is also fairly precise and I had a fair bit of trouble of jumping too soon (bug doesn't come high enough) or too late (bug stops going up before you jump). I can believe that if someone gets good with this setup then it can work the same with Wave, but with Wave there is more flexibility, you can see the bug more, and it's more intuitive to grasp. So I guess they still feel different to me :shrug:

osse101:
https://videos.maprando.com/video/10965 Here's the up shot.  It doesn't matter how high the bug is, but it can go above the camera where you have to jump to see it, but that height still works.  Timing the jump where the bug reaches a height where you can see it in the tunnel is very hard as you noted.  So here's a lazy shot setup from the right https://videos.maprando.com/video/10966  Wave can just shoot through the right wall for a good height.

- `"notable": "Wall Ice Clip the Mission Impossible Crumble Blocks"` only applies to wave if there is to be a spazer notable, so I suggest a rename and the full description added to the spazer notable.  Spazer needs its own strat note in some form.
- `"With Spazer, the trick is more precise: the shot must be fired low enough to avoid hitting the wall below the crumbles, but high enough to hit the Zeb."` doesn't make sense to me.  I would describe it as shooting into the wall below the crumbles such that the beam does not impact the crumble blocks.  While keeping the Zeb low enough to be hit.
It more or less looks like you can shoot above the crumble blocks but that should just be harder.

blkerby:
Ok, after seeing how you do it with the upward shot and testing it out, I agree now this is not significantly harder with Spazer than with Wave. For some reason, when I was considering the upward shot before, it didn't occur to me to freeze the Zeb closer to the left in order to have plenty of time to work out the horizontal position without the Zeb despawning. And the setup from the top-right is nice to avoid needing either the precise jump or an extra refreeze to make it climb higher.

PR #2940, opened 2026-08-11
More leaveWithMockball (Crateria, Brinstar)

blkerby:
These are some strats/variations that were missed on the first pass. Most of these are applicable to mockballing under the Ice Beam Gate shutter.

### DIFF
--- a/region/brinstar/kraid/Warehouse Energy Tank Room.json
+++ b/region/brinstar/kraid/Warehouse Energy Tank Room.json
@@ -233,7 +233,17 @@
       "link": [1, 1],
       "name": "Leave With Mockball",
       "requires": [
-        {"obstaclesCleared": ["A"]}
+        {"or": [
+          {"obstaclesCleared": ["A"]},
+          {"and": [
+            "Ice",
+            "canTrickyDodgeEnemies"
+          ]},
+          {"and": [
+            {"enemyDamage": {"enemy": "Beetom", "type": "contact", "hits": 8}},
+            "canTrickyDodgeEnemies"
+          ]}          
+        ]}
       ],
       "exitCondition": {
         "leaveWithMockball": {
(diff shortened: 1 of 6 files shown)


region/brinstar/kraid/Warehouse Energy Tank Room.json
```diff
@@ -233,7 +233,17 @@
       "link": [1, 1],
       "name": "Leave With Mockball",
       "requires": [
-        {"obstaclesCleared": ["A"]}
+        {"or": [
+          {"obstaclesCleared": ["A"]},
+          {"and": [
+            "Ice",
+            "canTrickyDodgeEnemies"
```
osse101, line 245: How does this work?  Grab the beetoms and hope to not take damage and freeze them on the rightish part of the room?  Letting them jump around and freezing when positioned doesn't seem like it'd work.  So I think we should expect damage.

blkerby: Here's a way to reliably do it damageless with Ice: https://videos.maprando.com/video/10963. I thought it was maybe ok for Expert, though we could add a canInsaneJump if you think it fits? The short runway length puts it in Extreme in Ice Beam Gate Room anyway.

Added a canDodgeWhileShooting variant that does what you describe. I think it's normally like 2 hits but put 4 to be conservative; it can maybe account for failing and needing to reset the room.

There's also a version that dodges the Beetoms with no Ice: https://videos.maprando.com/video/10964. I think this one is probably Insane.

osse101: - Damageless + Ice looks a little more specialized than typical canTrickyDodgeEnemies.  You can risk getting grabbed though and that's fine a lot of the time, so Expert is alright.  A doorcheck spinjump entry lets you freeze the right beetom on the floor, it looks like.
- Avoid with nothing should match the requirements of `Very Tricky Evade` imo.  So they should be bumped up to insane too, I think.

blkerby: I think it's a significant step up from the regular "Very Tricky Evade"; the final manipulation at the end to get the Beetoms to go left is non-trivial, and more importantly, getting the completed mockball on the 1-tile landing means you're chaining two difficult tricks together.

PR #2941, opened 2026-08-11
Add schema for Jesuswalk.

nn357:
not even sure this is correct, probably needs more work too as i dont see where enemy damage is calculated

### DIFF
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -749,9 +749,7 @@
         "comeInWithXMode": {
         }
       },
-      "requires": [
-        "canDoorTransitionXMode"
-      ],
+      "requires": [],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -743,6 +743,24 @@
       ]
     },
     {
+      "link": [1, 4],
+      "name": "Miracle Walk",
+      "entranceCondition": {
+        "comeInWithXMode": {
+        }
+      },
+      "requires": [
+        "canDoorTransitionXMode"
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "After entering the room with X-Mode, being arm-pumping. Samus will initially move several tiles to the right and then appear to be stationary waling in mid-air.",
+        "Continue to arm pump until Samus starts to move forwards again, indicating the far end of the room has been reached.",
+        "Stop arm-pumping when Samus is 1 tile away from the first ceiling grapple block and release dash to cancel X-Ray and land on the far right door platform."
+      ]
+    },
+    {
       "id": 18,
       "link": [1, 5],
       "name": "Come in Shinecharging, Leave With Temporary Blue",
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -744,7 +744,7 @@
     },
     {
       "link": [1, 4],
-      "name": "Miracle Walk",
+      "name": "X-Mode Traversal (Jesus Walk)",
       "entranceCondition": {
         "comeInWithXMode": {
         }
--- a/region/maridia/outer/Mt. Everest.json
+++ b/region/maridia/outer/Mt. Everest.json
@@ -753,7 +753,7 @@
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
-        "After entering the room with X-Mode, being arm-pumping. Samus will initially move several tiles to the right and then appear to be stationary waling in mid-air.",
+        "After entering the room with X-Mode, begin arm-pumping. Samus will initially move several tiles to the right and then appear to be stationary walking in mid-air.",
         "Continue to arm pump until Samus starts to move forwards again, indicating the far end of the room has been reached.",
         "Stop arm-pumping when Samus is 1 tile away from the first ceiling grapple block and release dash to cancel X-Ray and land on the far right door platform."
       ]
--- a/schema/m3-room.schema.json
+++ b/schema/m3-room.schema.json
@@ -1207,7 +1207,18 @@
               "description": "Represents that Samus can leave through this door while taking damage during the transition, in a pose and at a height that allow her to perform a single-frame damage boost with X-Ray buffered on the first frame after the transition. This allows Samus to reach the doorshell in the next room before the knockback frames expire, enabling her to enter X-Mode.",
               "required": [],
               "additionalProperties": false,
-              "properties": {}
+              "properties": {
+                  "enemy": {
+                  "default": "",
+                  "pattern": "^(.*)$",
+                  "title": "Name of enemy"
+                },
+                  "attack": {
+                  "default": "contact",
+                  "pattern": "^(.*)$",
+                  "title": "Type of enemy damage."
+                }
+              }
             },
             "leaveWithGMode": {
               "type": "object",
--- a/schema/m3-room.schema.json
+++ b/schema/m3-room.schema.json
@@ -1208,12 +1208,12 @@
               "required": [],
               "additionalProperties": false,
               "properties": {
-                  "enemy": {
+                "enemy": {
                   "default": "",
                   "pattern": "^(.*)$",
                   "title": "Name of enemy"
                 },
-                  "attack": {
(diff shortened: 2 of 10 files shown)


blkerby:
- Add a requirement like `{"enemyDamage": {"enemy": "Sciser", "type": "contact", "hits": 1}}` to the exit strats.
- The "canDoorTransitionXMode" tech should be implicit in `comeInWithXMode`.
- The new entrance/exit conditions should be documented in strats.md.
- "using X-Ray on the first frame after the transition": I don't think this is exactly accurate?
- "Depending on QOL settings": We avoid referencing Map Rando-specific things in this repo. Better to describe it based on the vanilla game. In the rare case that a Map Rando-specific note is important to add, the right place is in the Map Rando [fork](https://github.com/kjbranch/sm-json-data/tree/map-rando).
- "Stop arm-pumping when Samus is 1 tile away from the first ceiling grapple block and release dash to cancel X-Ray and land on the far right door platform.": Could it be worth mentioning the option to spin jump before exiting X-mode, if this makes it more lenient?
- `{"noFlashSuit": {}}"`: I think we also need `{"noBlueSuit": {}}`? Otherwise you would kill the Sciser instead of taking damage from it?
- I'm guessing more setups would be possible (e.g. Main Street nodes 2 and 3?) but I haven't tested, and they can always be added later.

nn357:
Awesome thanks for the feedback! It's probably the second frame, i'll double check.. If you goto 1 tiles before the grapple block there's no need to spin jump apart from saving time. 

Yeah I only added a couple of setups for now.

I noticed none of the gmode exit strats have enemy damage, how's that calculated, internally or is it the gmode just assumes a reserve trigger or something?

region/maridia/outer/Mt. Everest.json
```diff
@@ -742,6 +742,22 @@
...
+      "name": "Miracle Walk",
+      "entranceCondition": {
+        "comeInWithXMode": {
+        }
+      },
+      "requires": [],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "note": [
+        "After entering the room with X-Mode, being arm-pumping. Samus will initially move several tiles to the right and then appear to be stationary waling in mid-air.",
```
blkerby:
```suggestion
        "After entering the room with X-Mode, begin arm-pumping. Samus will initially move several tiles to the right and then appear to be stationary walking in mid-air.",
```

schema/m3-room.schema.json
```diff
@@ -1192,6 +1201,14 @@
                 }
               }
             },
+            "leaveWithXModeSetup": {
+              "type": "object",
+              "title": "Leave With X-Mode Setup",
+              "description": "Represents that Samus can leave through this door while taking damage through the transition, in a pose that would allow using X-Ray on the first frame after the transition. This is for setting up to enter X-mode in the next room.",
```
blkerby: "on the first frame after the transition": this should be corrected/adjusted

tech.json
```diff
@@ -3289,6 +3289,31 @@
...
+          ],
+          "detailNote": [
+            "With X-Ray selected, Samus must perform a low spin jump at the previous rooms door transition and perform a walljump check to trigger the transition while taking damage from an enemy.",
+            "During the transition hold right, dash, jump and angle up. After the door transition, on the same frame Samus changes to the damage pose, release jump and right at the same time. Continue to hold the other inputs.",
+            "Samus will now be stationary in X-Mode, partially stuck in the doorframe. Repress right to turn around and arm-pump to move forwards.",
+            "Releasing dash will exit X-Mode."
+          ],
+          "devNote": [
+            "While it is technically possible to enter this mode with a flash suit and spark to traverse while gaining a bluesuit, it requires several consecutive frame perfect inputs.",
+            "It is therefor assumed it is not humanly possible to enter with / use / retain a flashsuit with this mode."
```
blkerby:
```suggestion
            "It is therefore assumed it is not humanly viable to enter with, use, or retain a flashsuit with this mode."
```

osse101:
Reserves activating isn't mentioned anywhere... so does that mean it doesn't work without enough RegularEnergy and needs special damage handling, or is it fine?

nn357:
> Reserves activating isn't mentioned anywhere... so does that mean it doesn't work without enough RegularEnergy and needs special damage handling, or is it fine?

Oh will the sciser hit just assume it can trigger reserves if it needs to? Yeah you don't need a reserve at all for it to work but you can do with a reserve trigger i believe, i.e if you are suitless 0+1, but you need to set reserve energy to 1 (will test it).

nn357:
> Reserves activating isn't mentioned anywhere... so does that mean it doesn't work without enough RegularEnergy and needs special damage handling, or is it fine?

https://videos.maprando.com/video/10954

if you have to trigger reserves through the transition it only seems to work with 1 in reserve

osse101:
> if you have to trigger reserves through the transition it only seems to work with 1 in reserve

Then it's not so easy to handle damage inside of requirements.  So I think you want to put the enemy damage event inside the LeaveWithXMode object so the randomizer can calculate your final energy following the rules you write to strats.md.  And stopping the reserves on 1 would be `canPreciseReserveRefill` which should be fine to add.

blkerby:
> So I think you want to put the enemy damage event inside the LeaveWithXMode

Yeah, probably something like this then:

```
"leaveWithXModeSetup": {
    "enemy": "Sciser",
    "attack": "contact"
}
```

region/maridia/outer/Mt. Everest.json
```diff
@@ -742,6 +742,22 @@
         "with the same type of entry and speeds at least $5.4, it is possible do controlled bounces on both Grapple platforms, with no spring fling needed."
       ]
     },
+    {
+      "link": [1, 4],
+      "name": "Miracle Walk",
```
kjbranch: This has a standard name of Jesus Walk. I don't think it's offensive or needing to be renamed. Miracle Walk sounds like we are trying hard to avoid the word Jesus but still keep the reference. If we do care about avoiding the name, then I would rather some completely separate name like X-Mode Arm Pump Traversal

blkerby: "X-Mode Traversal" is what we have used in other places and is a bit more technically descriptive. "Jesus Walk" can also make sense because it is how this specific strat is known in the community; and the "X-Mode" aspect will be covered by the setup strat in the other room. I thought "Miracle Walk" also sounded fine but I can see how it can be read as going out of its away to avoid the word "Jesus". I don't think any of the options are offensive.

nn357:
> > So I think you want to put the enemy damage event inside the LeaveWithXMode
> 
> Yeah, probably something like this then:
> 
> ```
> "leaveWithXModeSetup": {
>     "enemy": "Sciser",
>     "attack": "contact"
> }
> ```

I added to the schema but maybe i didnt do it right

blkerby:
> I added to the schema but maybe i didnt do it right

The schema looks ok (except it looks like the indentation is messed up). I think it's just the tests having trouble with how it processes the "attack" key. To avoid the error you could add "attack" to the "badKeys" list [here](https://github.com/vg-json-data/sm-json-data/blob/96c3d12681164c1877d1a574c9ff9e548d2f122e/tests/asserts/keywords.py#L113); in theory then we should validate the "attack" value in a custom way to make sure it is contained in the list of attacks for that enemy. At some point I want to revisit the validation/testing to try to improve how it is structured

strats.md
```diff
@@ -1448,6 +1475,32 @@ __Example:__
...
+
+A `comeInWithXMode` object does not have any properties.
+
+A `comeInWithXMode` entrance condition must match with  a `leaveWithXModeSetup` entrance condition on the other side of the door:
+
+When matching with a `leaveWithXModeSetup`, a `comeInWithXMode` has implicit requirements:
+- The tech requirement `canDoorTransitionXMode`.
+- The `XRayScope` item requirement.
+- The requirements `{"and": [{"noFlashSuit": {}}, {"noBlueSuit": {}}]}`.
+- A requirement `{"disableEquipment": "Gravity"}`.
```
osse101, line 1496: There is no mention of needing water physics anywhere.  Which side even needs to be water?

nn357: both

osse101 (dismissed):

strats.md
blkerby: When I looked at this a couple years ago, I got it to work once with air-to-water. I remember TASing it out a bit and confirming that it was very terrible, though I don't remember the details and didn't make a video. So I'm ok pretending it's not a thing for now; it would probably be Beyond-level anyway.

PR #2944, opened 2026-08-14
More leaveWithMockball (Maridia, Norfair, 3-room)

blkerby:
In addition to covering more leaveWithMockball strats in Maridia/Norfair, I also did a pass of 3-room strats where you enter a room (e.g. a two-door Save room) running/spinning and then leave mockballing through an opposite door. The schema doesn't allow us to propagate the entry speed into the exit speed, but for now we can specialize to the case needed for the Ice Beam Gate Room mockball, which I think is the most significant application. 

Some of these strats are not possible to occur in Map Rando with Ice Beam Gate Room due to geometrically impossible room configurations, but I kept them anyway; some of them might have other applications, or if not, they probably will later when we extend them to general runway lengths.

### DIFF
--- a/region/brinstar/red/Skree Boost Room.json
+++ b/region/brinstar/red/Skree Boost Room.json
@@ -348,6 +348,77 @@
       ]
     },
     {
+      "link": [1, 2],
+      "name": "Come in Jumping, Leave with Mockball (Space Jump)",
+      "entranceCondition": {
+        "comeInJumping": {
+          "minTiles": 5,
+          "speedBooster": "any"
+        }
+      },
+      "requires": [
+        "canPreciseSpaceJump"
+      ],
+      "exitCondition": {
+        "leaveWithMockball": {
+          "remoteRunway": {
+            "length": 5,
+            "openEnd": 1
+          },
+          "landingRunway": {
+            "length": 1,
+            "openEnd": 1
+          }
+        }
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],      
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "devNote": [
+        "FIXME: This is a specific length needed for the Ice Beam Gate Room mockball;",
+        "when the schema allows it, represent the possible runways more generally."
+      ]
+    },    
+    {
+      "link": [1, 2],
+      "name": "Come in Spinning, Leave with Mockball (Space Jump)",
+      "entranceCondition": {
+        "comeInSpinning": {
+          "minExtraRunSpeed": "$1.A",
+          "unusableTiles": 0,
+          "speedBooster": "any"
+        }
+      },
+      "requires": [
+        "canPreciseSpaceJump"
+      ],
+      "exitCondition": {
+        "leaveWithMockball": {
+          "remoteRunway": {
+            "length": 5,
+            "openEnd": 1
+          },
+          "landingRunway": {
+            "length": 1,
+            "openEnd": 1
+          }
+        }
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "devNote": [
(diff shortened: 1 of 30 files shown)


region/brinstar/red/Skree Boost Room.json
```diff
@@ -347,6 +347,77 @@
...
+          }
+        }
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],      
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "devNote": [
```
osse101, line 380:
```suggestion
      "blueSuitChecked": true,
      "note": [
        "Optimize the mockball by landing on the edge of the runway and morphing before reaching the ground."
      ],
      "devNote": [
```

region/brinstar/red/Skree Boost Room.json
```diff
@@ -347,6 +347,77 @@
...
+            "openEnd": 1
+          }
+        }
+      },
+      "unlocksDoors": [
+        {"types": ["super"], "requires": []},
+        {"types": ["missiles", "powerbomb"], "requires": ["never"]}
+      ],
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
```
osse101, line 415:
```suggestion
      "blueSuitChecked": true,
      "note": [
        "Optimize the mockball by landing on the edge of the runway and morphing before reaching the ground."
      ],
```

region/lowernorfair/east/Three Musketeers' Room.json
```diff
@@ -476,6 +476,28 @@
         "This strat is included for completeness, though it apparently doesn't have any applications."
       ]
     },
+    {
+      "link": [2, 2],
+      "name": "Leave with Mockball",
+      "requires": [
+        "h_heatedRemoteRunway",
```
osse101, line 483:
```suggestion
        "h_heatedRemoteRunway",
        "can3HighWallMidAirMorph",
```

region/lowernorfair/east/Ridley Tank Room.json
```diff
@@ -97,6 +97,28 @@
       ],
       "devNote": "Max extra run speed $2.B, or $2.C with Hi-Jump and a quick aim-down."
     },
+    {
+      "link": [1, 1],
+      "name": "Leave with Mockball",
+      "requires": [
+        "h_heatedRemoteRunway",
```
osse101:
```suggestion
        "canTrickyJump",
```

osse101 (commented):
Most of these have a movement tech but there are two things I was looking for:
- Jumping up to a landing runway isn't Hard friendly, where you would cut your jump to mockball quickly.
- landingRunway of just the doorframe  means a mockball might fail to roll under the shutter fast enough.  So you either want extra run speed or to morph higher in the air which isn't an obvious thing.

Then `Jump-> shoot door -> mockball` generally seems at least very hard too, but I think all of those had canPreciseSpacejump.

region/brinstar/red/Skree Boost Room.json
blkerby: "morphing before reaching the ground" can be a bit ambiguous, possibly meaning "start the morph transition before reaching the ground" (always required for a mockball) or "finish the morph transition before reaching the ground" (which will result in a bounce instead of a mockball). I think we could word it more accurately, "start the morph transition early so that it completes immediately after landing"?

blkerby: Also, this kind of note would apply to every leaveWithMockball with a 1-tile landing, and there are over 100 such strats. Maybe we just put it on the "Insane Mockball" strat in Ice Beam Gate Room? I think that would cover the main application. In theory these leaveWithMockball strats could pair with cross-room controlled spring ball jumps too, though I'm not sure if there are any applications where you need the completed morph  where it wouldn't be preferable to do an uncontrolled bounce.

region/lowernorfair/east/Three Musketeers' Room.json
blkerby: "can3HighWallMidAirMorph": What is the wall jump for? Or was this supposed to be "can4HighMidAirMorph"?

osse101: I was reading over the Expert techs over and over until the walljump part went away on its own.

PR #2946, opened 2026-08-16
Add canRiskPLA to blue brinstar etank overload plms with item.

nn357:
This was mentioned in logic a few weeks ago and I recently had it come up in a seed today (on desolate with no item there anyway, but that seems like a separate issue).


region/brinstar/blue/Blue Brinstar Energy Tank Room.json
blkerby:
```suggestion
```

PR #2956, opened 2026-08-22
Crateria fixes

blkerby:
I decided to try letting AI (gpt 5.6-sol medium) do a pass through the logic and see what errors it finds, and this is what turned up for Crateria. I wrote the PR by hand, based on the subset of the AI findings that I found to be correct. 

I don't think anything here was overly broken; mostly it's minor stuff that made some logic incomplete, for example:
- You could enter Landing Site from the left with G-mode artificial morph, break the bomb blocks with PBs, exit G-mode, refill at the Ship and return, which wouldn't have been logical before because we forgot to clear the obstacle for the bomb blocks. 
- You can moonfall clip down Green Pirates Shaft with a blue suit.
- You can temp blue chain through Gauntlet Entrance then spark out.

There were a few that potentially affect soundness:
- Gauntlet E-Tank logic thought you could use Bombs with G-mode artificial morph to fully break the left bomb wall (when you can only break the first half). I think it's not too impactful because to cross the room you still have to get through more bomb blocks, which I think would need Speed. So if I'm thinking of it right, it just comes down to needing to do a shorter shortcharge than expected.
- Climb Supers: We weren't checking that the Boyons are still alive when using them as a frozen runway. I'm not sure if it actually matters, but as more stuff gets added it can get difficult to be confident that it's ok. If you did an R-mode spark interrupt you may have needed to kill the Boyons for example, though with a blue suit there wouldn't be any point to the longer runway.

There also appears to be an issue with obstacle B in West Ocean but I'll let Kyle take a look at that separately.

### DIFF
--- a/region/crateria/central/Climb Supers Room.json
+++ b/region/crateria/central/Climb Supers Room.json
@@ -162,7 +162,7 @@
     },
     {
       "id": "B",
-      "name": "Above SpeedBlocks with Temp Blue",
+      "name": "Above Speed Blocks with Temp Blue",
       "obstacleType": "abstract"
     },
     {
@@ -1336,6 +1336,7 @@
       "link": [2, 2],
       "name": "Leave with Runway, Frozen Boyon Bridge",
       "requires": [
+        {"obstaclesNotCleared": ["A"]},
         "canTrivialUseFrozenEnemies"
       ],
       "exitCondition": {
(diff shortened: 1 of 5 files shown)


kjbranch (commented):
This is very impressive.

- I knew about the Climb Supers/Boyon ones, but don't think it matters either way.
- The IBJ-PB in Landing Site is a good find, as it gives you one more PB. 
- Landing Site blue spring could let you unlock the door remotely, which is useful, but something that is missing all over the place.
- Landing Site with Screw/Blue aren't really meaningfully different.
- The Gauntlet ones all looks impactful. 

I couldnt review the Green Pirate Shaft one.

PR #2957, opened 2026-08-22
Brinstar fixes

blkerby:
Continuing through a pass of AI review, a couple noteworthy soundness issues that are fixed here:
- Big Pink: there were a couple logical ways of clearing the Super block without Supers.
- Red Tower: if your only access is through the PB blocks, you would have to kill the Beetom but could logically still use it to Ice climb.

The theme continues that most of the errors found are obstacle-related. And there are some `obstaclesNotCleared` that we could maybe rationalize as not really needed, but it seems safer to put them in.

### DIFF
--- a/region/brinstar/blue/Morph Ball Room.json
+++ b/region/brinstar/blue/Morph Ball Room.json
@@ -813,8 +813,7 @@
       "link": [1, 1],
       "name": "G-Mode Setup - Get Hit By Sidehopper",
       "requires": [
-        "h_ZebesIsAwake",
-        {"obstaclesNotCleared": ["C"]}
+        "h_ZebesIsAwake"
       ],
       "exitCondition": {
         "leaveWithGModeSetup": {}
@@ -1062,9 +1061,9 @@
             {"resourceCapacity": [{"type": "RegularEnergy", "count": 199}]},
             {"resourceCapacity": [{"type": "ReserveEnergy", "count": 199}]}
           ]}
-        ]}
+        ]},
+        {"obstaclesCleared": ["C"]}
       ],
-      "clearsObstacles": ["C"],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
@@ -1099,9 +1098,9 @@
             {"resourceCapacity": [{"type": "RegularEnergy", "count": 199}]},
             {"resourceCapacity": [{"type": "ReserveEnergy", "count": 199}]}
           ]}
-        ]}
+        ]},
+        {"obstaclesCleared": ["C"]}
       ],
-      "clearsObstacles": ["C"],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
--- a/region/brinstar/blue/Morph Ball Room.json
+++ b/region/brinstar/blue/Morph Ball Room.json
@@ -813,7 +813,8 @@
       "link": [1, 1],
       "name": "G-Mode Setup - Get Hit By Sidehopper",
       "requires": [
-        "h_ZebesIsAwake"
+        "h_ZebesIsAwake",
+        {"obstaclesNotCleared": ["C"]}
       ],
       "exitCondition": {
         "leaveWithGModeSetup": {}
@@ -1061,9 +1062,9 @@
             {"resourceCapacity": [{"type": "RegularEnergy", "count": 199}]},
             {"resourceCapacity": [{"type": "ReserveEnergy", "count": 199}]}
           ]}
-        ]},
-        {"obstaclesCleared": ["C"]}
+        ]}
       ],
+      "clearsObstacles": ["C"],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
@@ -1098,9 +1099,9 @@
             {"resourceCapacity": [{"type": "RegularEnergy", "count": 199}]},
             {"resourceCapacity": [{"type": "ReserveEnergy", "count": 199}]}
           ]}
-        ]},
-        {"obstaclesCleared": ["C"]}
+        ]}
       ],
+      "clearsObstacles": ["C"],
       "flashSuitChecked": true,
       "blueSuitChecked": true,
       "note": [
@@ -1192,6 +1193,7 @@
         ]},
         "h_bombThings"
       ],
+      "clearsObstacles": ["A"],
       "flashSuitChecked": true,
       "blueSuitChecked": true
     },
--- a/region/brinstar/green/Green Brinstar Beetom Room.json
+++ b/region/brinstar/green/Green Brinstar Beetom Room.json
@@ -39,7 +39,12 @@
     {
       "id": "A",
       "name": "Beetoms Dead or Frozen",
-      "obstacleType": "enemies"
+      "obstacleType": "enemies",
+      "devNote": [
+        "FIXME: Lumping frozen and dead together is questionable,",
+        "because some strats require the Beetoms to be alive.",
+        "The frozen state might not need to be an obstacle."
+      ]
     }
   ],
   "enemies": [
--- a/region/brinstar/green/Green Brinstar Main Shaft.json
+++ b/region/brinstar/green/Green Brinstar Main Shaft.json
@@ -5268,6 +5268,8 @@
         }
       },
       "requires": [
+        {"itemNotCollectedAtNode": 13},
+        "canRiskPermanentLossOfAccess",
         "Morph",
         {"blueSuitShinecharge": {}},
         {"or": [
--- a/region/brinstar/green/Green Brinstar Main Shaft.json
+++ b/region/brinstar/green/Green Brinstar Main Shaft.json
@@ -5287,6 +5287,11 @@
       "blueSuitChecked": true,
       "note": [
         "Spark up the shaft, then shoot the Chozo ball to overload PLMs. Then go down, through the bomb blocks and tunnel, then exit G-mode."
+      ],
+      "devNote": [
+        "FIXME: Overloading PLMs requires the item to be not collected.",
+        "The apparent unsoundness has limited impact since you could X-ray climb instead.",
+        "There are many ways to get to the top aside from blue suit shinecharge."
       ]
     },
     {
--- a/region/brinstar/green/Green Brinstar Main Shaft.json
+++ b/region/brinstar/green/Green Brinstar Main Shaft.json
@@ -5291,9 +5291,7 @@
         "Spark up the shaft, then shoot the Chozo ball to overload PLMs. Then go down, through the bomb blocks and tunnel, then exit G-mode."
       ],
       "devNote": [
-        "FIXME: Overloading PLMs requires the item to be not collected.",
-        "The apparent unsoundness has limited impact since you could X-ray climb instead.",
-        "There are many ways to get to the top aside from blue suit shinecharge."
+        "FIXME: There are many ways to get to the top aside from blue suit shinecharge."
       ]
     },
     {
--- a/region/brinstar/kraid/Warehouse Energy Tank Room.json
+++ b/region/brinstar/kraid/Warehouse Energy Tank Room.json
@@ -377,7 +377,6 @@
       "link": [1, 1],
       "name": "G-Mode Setup - Get Hit By Beetom",
       "requires": [
-        {"obstaclesNotCleared": ["A"]},        
         {"or": [
           {"noBlueSuit": {}},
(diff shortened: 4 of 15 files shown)


region/brinstar/blue/Morph Ball Room.json
```diff
@@ -813,7 +813,8 @@
       "link": [1, 1],
       "name": "G-Mode Setup - Get Hit By Sidehopper",
       "requires": [
-        "h_ZebesIsAwake"
+        "h_ZebesIsAwake",
+        {"obstaclesNotCleared": ["C"]}
```
kjbranch: I think clearing C can keep one alive if you want to. But I don't recall.

region/brinstar/blue/Morph Ball Room.json
```diff
@@ -1061,9 +1062,9 @@
             {"resourceCapacity": [{"type": "RegularEnergy", "count": 199}]},
             {"resourceCapacity": [{"type": "ReserveEnergy", "count": 199}]}
           ]}
-        ]},
-        {"obstaclesCleared": ["C"]}
```
kjbranch, line 1065: I think this one might be kind of similar. I think you kind of need the ability to clear C to use the runway. But you keep one alive. And since C was already cleared it didnt need to be cleared again.

Same in the strat below

region/brinstar/green/Green Brinstar Beetom Room.json
```diff
@@ -39,7 +39,12 @@
     {
       "id": "A",
       "name": "Beetoms Dead or Frozen",
-      "obstacleType": "enemies"
+      "obstacleType": "enemies",
+      "devNote": [
+        "FIXME: Lumping frozen and dead together is questionable,",
+        "because some strats require the Beetoms to be alive.",
+        "The frozen state might not need to be an obstacle."
```
kjbranch, line 46: If you have ice, clearing A is free. I guess the only concern is that you CF in here and need to use ice to get out. but the door should be resettable.

region/brinstar/green/Green Brinstar Main Shaft.json
```diff
@@ -5287,6 +5287,11 @@
       "blueSuitChecked": true,
       "note": [
         "Spark up the shaft, then shoot the Chozo ball to overload PLMs. Then go down, through the bomb blocks and tunnel, then exit G-mode."
+      ],
+      "devNote": [
+        "FIXME: Overloading PLMs requires the item to be not collected.",
+        "The apparent unsoundness has limited impact since you could X-ray climb instead.",
+        "There are many ways to get to the top aside from blue suit shinecharge."
       ]
     },
```
kjbranch, line 5296: This strat is only useful if you can't do a short x-ray climb. But you clearly already have high-level g-mode+blue. And the strat is unsound because the item needs to be not collected.

I propose deleting the strat, but if you want to keep it for "x-ray climb might be disabled" then add a item not collected+loss of access
```suggestion
```

region/brinstar/kraid/Warehouse Energy Tank Room.json
```diff
@@ -375,6 +377,7 @@
       "link": [1, 1],
       "name": "G-Mode Setup - Get Hit By Beetom",
       "requires": [
+        {"obstaclesNotCleared": ["A"]},
```
kjbranch: guess im not sure about these kinds of things. Do we really want to ensure every enemy is not dead in the future? because you can already reset the room since it's a g-mode setup. This would only be useful if the next room cant be reset back here but I don't think there are examples like that. And it would only be useful if you need to grab the item and you cant dodge the beetoms and the only way you can kill them to grab the item is with a PB.

region/brinstar/kraid/Warehouse Kihunter Room.json
```diff
@@ -203,6 +203,7 @@
       "link": [1, 1],
       "name": "Kihunter Ice Moonfall Door Lock Skip",
       "requires": [
+        {"obstaclesNotCleared": ["C"]},
```
kjbranch: dont really like this one either. the obstacle represents killing or dodging them. you should always be able to get here with 2 alive if that's your goal.

region/brinstar/kraid/Warehouse Kihunter Room.json
```diff
@@ -512,8 +513,9 @@
     {
       "id": 58,
       "link": [2, 2],
-      "name": "G-Mode Setup - Get Hit By KiHunter",
+      "name": "G-Mode Setup - Get Hit By Kihunter",
       "requires": [
+        {"obstaclesNotCleared": ["C"]},
```
kjbranch: same

region/brinstar/pink/Mission Impossible Room.json
```diff
@@ -372,7 +372,9 @@
       "id": 11,
       "link": [1, 1],
       "name": "G-Mode Setup - Get Hit By Sidehopper",
-      "requires": [],
+      "requires": [
+        {"obstaclesNotCleared": ["A"]}
```
kjbranch, line 376: There's not a way to do mission impossible without killing the hoppers. At least, that's what my to-do list has said for years.

region/brinstar/red/Red Tower.json
```diff
@@ -203,7 +203,7 @@
   "obstacles": [
     {
       "id": "A",
-      "name": "Power Bomb Blocks",
+      "name": "Power Bomb Blocks and Beetom",
```
kjbranch: This cant be for beetoms as well. You can move the beetoms out of the way before breaking the PB blocks

region/brinstar/red/Red Tower.json
```diff
@@ -600,6 +601,7 @@
       "link": [1, 2],
       "name": "Frozen Beetom X-Ray Climb with Morph",
       "requires": [
+        {"obstaclesNotCleared": ["A"]},
```
kjbranch: If A is broken this whole strat is pointless

region/brinstar/red/Red Tower.json
```diff
@@ -352,6 +352,7 @@
       "link": [1, 1],
       "name": "Leave With Runway - Frozen Beetom",
       "requires": [
+        {"obstaclesNotCleared": ["A"]},
```
kjbranch: you can reset the room here

region/brinstar/red/Red Tower.json
```diff
@@ -1460,6 +1462,7 @@
       "link": [1, 9],
       "name": "Frozen Beetom Ice Climb",
       "requires": [
+        {"obstaclesNotCleared": ["A"]},
```
kjbranch: This one is a problem. But i still dont like the obstacle being both objects

region/brinstar/blue/Morph Ball Room.json
blkerby: Could be true, it just might be a bit of a fragile assumption. If we want to go with that, we should probably document exactly what it means for the obstacle to be "cleared" as a note in the obstacle definition and make sure it is consistently followed everywhere that refers to the obstacle.

I think in a lot of places we're implicitly leaning on the fact that you could reset the room to respawn enemies; but this might not be valid in case the next room is not free to reset (e.g. if it is heated, or has a door that locks behind you).

region/brinstar/green/Green Brinstar Main Shaft.json
blkerby: It could be deleted. But I think there other ways of reaching the Chozo ball that could be more useful (though still canRiskPermanentLossOfAccess), for example using wall jump or Space Jump to get up there.

region/brinstar/blue/Morph Ball Room.json
kjbranch: The requirements are skill and strat dependent. If we want to make tons of obstacles in every room, that's an option. This room could have an obstacle for one hopper alive. And then maybe eventually it would need to be split into which one is alive, converting it into 3 obstacles ....

But if we do go the route of trying to "fix" all these, by adding `obstaclesNotCleared` in lots of places, we might be losing some strats that are basically always fine. For instance in this scenario, you should always be able to survive resetting the room after you killed 3 hoppers. As for asymmetric door locks, I wouldnt say that they are even close to being sound within this project right now.

region/brinstar/kraid/Warehouse Kihunter Room.json
blkerby: Having an obstacle represent "dodged" feels fragile. If we're doing that then we probably want to define it more explicitly in the obstacle definition. In my mental model, a cleared enemy obstacle means the enemy is killed unless stated otherwise. And "dodged" isn't really a well-defined state of an enemy, so I can imagine this creating problems.

region/brinstar/blue/Morph Ball Room.json
blkerby: For asymmetric door locks, I think maybe a reasonable approach for now could be to aim for soundness but not full completeness. So when it comes to the fact that normally you can reset a room at a door that you're about to exit through, we could "assume" this for the purposes of not going out of our way to create extra obstacles or strats to provide alternatives to the room reset, but maybe not assume it when it comes to leaving out requirements (e.g. an "obstaclesNotCleared" or "resetRoom" requirement.

region/brinstar/pink/Mission Impossible Room.json
blkerby: Under that assumption, if you came into the room and did Mission Impossible, it would not be sound to leave with a G-mode setup without resetting the room first, so the "obstaclesNotCleared" serves the purpose of enforcing this?

region/brinstar/red/Red Tower.json
blkerby: Good point, I'm thinking we probably need to split the Power Bomb blocks and Beetoms into separate obstacles.

region/brinstar/green/Green Brinstar Main Shaft.json
kjbranch: Those should be there. If they're missing I can work on adding them

kjbranch: Those would only be useful without a blue suit though - if you have blue, there's no reason to overload PLMs, you can just use the scroll block in the tunnel.... But i guess that kind of gets back to my first impression comment - then you could just x-ray climb.... Let me know if you want them added

region/brinstar/kraid/Warehouse Kihunter Room.json
kjbranch: a cleared enemy obstacle regularly means the enemy is dodged or moved into a safe position.

region/brinstar/pink/Mission Impossible Room.json
kjbranch: I think I worded my original comment in a confusing way.

All of the mission impossible strats require A cleared, which I don't think is necessary. 

It's not strongly related to this line of code, but something that has been on my to-do list for years.

region/brinstar/red/Red Tower.json
blkerby: Ok I pushed another commit splitting the PB blocks and Beetom into separate obstacles.

region/brinstar/blue/Morph Ball Room.json
blkerby: After thinking about it more, I would be ok dropping these new "obstaclesNotCleared" on G-mode setups where normally you could just reset the room or keep an enemy alive. It makes sense that these aren't necessarily errors since it seems like an intentional, common pattern; that kind of puts them outside the scope of what these PRs are trying to do. So maybe it's best to hold off on those and they could always be addressed later if we run into problems or decide we want to support asymmetric doors.

region/brinstar/green/Green Brinstar Main Shaft.json
blkerby: X-Ray climb requires a way to break the left bomb block though? So overloading PLMs still seems useful in case you don't have a blue suit or Bombs/PBs. X-Ray climb also requires Morph; I imagine artificial morph IBJ could be another option using the item overload?

region/brinstar/blue/Morph Ball Room.json
blkerby: Ok I dropped these "obstaclesNotCleared" from the PR.

osse101 (approved):
- Red Tower Beetom almost doesn't need to be tracked and could be a note, but entering the room from 2 isn't distinguished without the Beetom obstacle.

PR #2961, opened 2026-08-22
Norfair fixes

blkerby:
Nothing too significant here.


region/lowernorfair/east/Amphitheatre.json
```diff
@@ -330,6 +330,7 @@
         {"shinespark": {"frames": 49, "excessFrames": 8}},
         {"heatFrames": 315}
       ],
+      "clearsObstacles": ["A"],
```
osse101, line 333: Both GModes right below this also need to clear A

osse101 (dismissed):

PR #2964, opened 2026-08-29
Tourian Hopper left-to-right spin jump squeeze

blkerby:
- Adds comeInSpinning and comeInWithSidePlatform strats for avoiding Tourian Hopper damage left-to-right.
- Adds many leaveWithSidePlatform strats to match with the new Tourian Hopper strat (for cases with unusual geometry, not already covered by leaveSpinning).
- Updates docs for leaveWithSidePlatform/comeInWithSidePlatform to reflect the new intent, that these are for requirements on neighboring-room remote runway geometry, no longer necessarily requiring preserving upward momentum. In this way it becomes more parallel with leaveWithPlatformBelow/comeInWithPlatformBelow.
- Adds/updates a few other remote-runway strats where I noticed them lacking.
- Adds item requirements to "h_heatedRemoteRunwaySpaceJump" and "h_heatedRemoteRunwaySpringBall" helpers. It may have been already covered by strat-specific requirements but seems safer this way.

### DIFF
--- a/region/brinstar/pink/Mission Impossible Room.json
+++ b/region/brinstar/pink/Mission Impossible Room.json
@@ -195,6 +195,27 @@
       "devNote": "Two spike hits are expected per attempt (with any additional leniency hits being multiplied by this amount)."
     },
     {
+      "link": [1, 1],
+      "name": "Leave With Side Platform",
+      "requires": [
+        {"obstaclesCleared": ["A"]},
+        {"obstaclesNotCleared": ["B"]}        
+      ],
+      "exitCondition": {
+        "leaveWithSidePlatform": {
+          "height": 2,
+          "runway": {
+            "length": 16,
+            "openEnd": 0
+          },
+          "obstruction": [13, 2]
+        }
+      },
+      "flashSuitChecked": true,
+      "blueSuitChecked": true,
+      "devNote": ["Max extra run speed $3.A."]
+    },
+    {
       "id": 5,
       "link": [1, 1],
       "name": "Leave Spinning (Space Jump)",
--- a/region/crateria/east/Crateria Kihunter Room.json
+++ b/region/crateria/east/Crateria Kihunter Room.json
@@ -371,8 +371,10 @@
     {
       "id": 12,
       "link": [2, 2],
-      "name": "Leave Spinning",
-      "requires": [],
+      "name": "Leave Spinning (Speed Booster)",
+      "requires": [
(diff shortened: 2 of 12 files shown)


region/brinstar/pink/Mission Impossible Room.json
```diff
@@ -194,6 +194,27 @@
       "blueSuitChecked": true,
       "devNote": "Two spike hits are expected per attempt (with any additional leniency hits being multiplied by this amount)."
     },
+    {
+      "link": [1, 1],
+      "name": "Leave With Side Platform",
```
osse101, line 199: There is a spike hit + leaveSpinning here as well. But spike damage boost is a little easier

region/tourian/main/Tourian Escape Room 3.json
```diff
@@ -782,6 +782,23 @@
       "flashSuitChecked": true,
       "blueSuitChecked": true
     },
+    {
+      "link": [2, 2],
+      "name": "Leave With Side Platform",
```
osse101: You can do the simple run under strat from the speed block runway here

region/crateria/east/Crateria Kihunter Room.json
```diff
@@ -371,8 +371,10 @@
     {
       "id": 12,
       "link": [2, 2],
-      "name": "Leave Spinning",
-      "requires": [],
+      "name": "Leave Spinning (Speed Booster)",
```
osse101, line 374: Simple run under works here too

osse101 (dismissed):

region/tourian/main/Tourian Escape Room 3.json
blkerby: Good catch, removed it now.

region/brinstar/pink/Mission Impossible Room.json
blkerby: Sounds good, I am looking at cross-room i-frame/damage-boost strats in my next PR and can include the spike damage boost there.

region/crateria/east/Crateria Kihunter Room.json
blkerby: True, I think I forgot about this at one point. But also this leaveSpinning strat already existed so this is just clarifying it with the name and the explicit "h_speedDash" requirement. It could be used for purposes other than dodging the Hoppers.

PR #2963, opened 2026-08-25
Crocomire Speedway: 99 energy crumble jump reverse spark (notable)

dragonsinth:
Croc Reverse Spark with 99 energy is easier than Expert+

https://videos.maprando.com/video/10974
https://videos.maprando.com/video/10976

Notable alternative to https://github.com/vg-json-data/sm-json-data/pull/2962

https://discord.com/channels/1053421401354285129/1053436236628496454/1541859636175699998

PR #2972, opened 2026-09-09
Red Brinstar Fireflea Room: low energy spark to block

dragonsinth:
https://discord.com/channels/1053421401354285129/1053436236628496454/1546988210985631864

Similar to https://github.com/vg-json-data/sm-json-data/pull/2963 - add a strat to allow 99 energy spark.

[comeInShinecharging](https://videos.maprando.com/video/10994)

[comeInShinecharged](https://videos.maprando.com/video/10995)


kjbranch (commented):
This needs a much more detailed description. 
- where to spark from
- mention that being higher gives more wiggle room when falling

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -416,6 +416,42 @@
...
+      "name": "Low Energy Spark to Block (Come In Shinecharged)",
+      "entranceCondition": {
+        "comeInShinecharged": {}
+      },
+      "requires": [
+        {"notable": "Low Energy Spark to Block"},
+        {"shineChargeFrames": 42},
+        "canHorizontalMidairShinespark",
+        {"resourceAtMost": [{"type": "RegularEnergy", "count": 99}]},
+        {"shinespark": {"frames": 70, "excessFrames": 4}}
```
kjbranch: I think with these excess frames, you would need to start in a different position, which is then `canControllShinesparkEnd`.
```suggestion
        {"shinespark": {"frames": 70, "excessFrames": 0}}
```

dragonsinth: When I was testing this out, I was normalizing the start position to be close to the ceiling, and 2-3 tiles right of the end of the platform.  So then from that start, with 95-99 hp I was close enough to the block to fall into it by holding left or right.  So this line was me trying to reflect what I tested, but I'm not dogmatic about it, we could just say you need exactly 99.

dragonsinth: Re-testing, at 95 you need a pretty big jump forward. So I think your inclination to just normalize at 99 is probably right. With 99 energy, the startup position is incredibly lenient, anywhere from directly over the last platform tile, up to 6 tiles out, works as long as you're close to the ceiling.

dragonsinth: Basically there's a 7-tile window that works, because ending the spark anywhere from -3 to +3 allows you to fall onto to the block.

region/brinstar/red/Red Brinstar Fireflea Room.json
```diff
@@ -416,6 +416,52 @@
...
     },
+    {
+      "link": [1, 2],
+      "name": "Low Energy Spark to Block (Come In Shinecharged)",
+      "entranceCondition": {
+        "comeInShinecharged": {}
+      },
+      "requires": [
+        {"notable": "Low Energy Spark to Block"},
+        {"shineChargeFrames": 42},
```
blkerby:
```suggestion
        {"shineChargeFrames": 30},
```

blkerby:
- The strats should end at node 3 rather than node 2.
- I would remove the part of the note about using the Waver for i-frames, as there are several possible options to progress from node 3 to node 2 which are covered in other strats.
- The comeInShinecharging should have length: 5 (door shell isn't included on entry since Samus spawns past it) and openEnd: 0 (since we don't want to include sliding off the runway). It looks like this was an error in the existing 1->2 strat (id: 10), which could also use a correction if you don't mind.

dragonsinth:
> * The strats should end at node 3 rather than node 2.

Whoops, fixed.

> * I would remove the part of the note about using the Waver for i-frames, as there are several possible options to progress from node 3 to node 2 which are covered in other strats.

Done.

> * The comeInShinecharging should have length: 5 (door shell isn't included on entry since Samus spawns past it) and openEnd: 0 (since we don't want to include sliding off the runway). It looks like this was an error in the existing 1->2 strat (id: 10), which could also use a correction if you don't mind.

Fixed, but I should note that in this room, all of the `comeInShinecharging` strats from node 2 are inconsitent with each other.  It looks like they should all be length 4, openEnd 0?  Although-- I don't know enough about spikesuit / x-mode to know, maybe the intent is to fall onto the spikes.

region/brinstar/red/Red Brinstar Fireflea Room.json
dragonsinth: How'd you get a number that low?  @kjbranch told me I should come in with 0 momentum.

https://videos.maprando.com/video/10995 I measured 174 - 131 - 1 = 42

dragonsinth: If it makes a difference, I had Spark Booster no Blue.

blkerby: You can spark from closer to the door. Updated the suggestion to 30 frames after messing with it a bit more. The randomizer has leniency on shinecharge frames so it won't end up being expected to be done this tight except on the highest difficulties.

https://videos.maprando.com/video/11003

blkerby:
> Fixed, but I should note that in this room, all of the `comeInShinecharging` strats from node 2 are inconsitent with each other. It looks like they should all be length 4, openEnd 0?

Right, I think the node 2 ones should all be length 4, openEnd 0. At best, sliding off would make you take an extra spike hit which isn't modeled.

The way it works in general:
- for a spikesuit you could slide off as long as you can be fully morphed before you reach the spikes. At node 2, I doubt there would be enough vertical space to make that work (though I haven't tested it), and even if somehow possible it would be need higher tech requirements, so making it openEnd: 0 should be right either way. At node 1, I think you could slide off into a spikesuit with no problem since there's lots of space.
- for an X-mode blue suit, you need to bounce before the X-mode, so sliding off would mean the bounce happens on the spikes, meaning you would need to take an extra spike hit. I would probably just shorten the runways to make them openEnd: 0 rather than trying to add an alternative with an extra spike hit.

But yeah, looks like node 1 also has other strats done inconsistently. We could probably use another pass through the whole game's logic to see if there are more inconsistencies like this (there probably are).

region/brinstar/red/Red Brinstar Fireflea Room.json
dragonsinth: Okay true. I was shooting my video to show the optimal positioning more than the minimum frames.  So you'd only have to get it that tight on insane; on "normal" settings the shinespark leniency will be fine.

Updated all `comeInShinecharging` strats from 1 to be `5, 0`, and all the ones from 2 to be `4, 0`.

dragonsinth: FWIW, I didn't realize openEnd: 1 on the end of the charge meant falling off the block, I assumed it just meant you had like half a block more because there wasn't an immediate wall.

dragonsinth:
I think this should be good to go now

